PR #5249, opened 2021-06-09
fix(linux): Fix packaging on ppa

ermshiperete:
When building packages on ppa the TERM environment variable is set to "unknown", which isn't a valid value but which we didn't account for previously.


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.64-alpha](https://keyman.com/downloads/releases/alpha/15.0.64)

PR #5247, opened 2021-06-09
fix(android): rotation is not updating keyboard

mcdurdin:
Fixes #5211.

PR #5262, opened 2021-06-10
fix(android): rotation is not updating keyboard 🍒

### LINKED ISSUE #5211: bug(android): Inconsistent OSK dimensions on rotation - system keyboard
During testing of #5201, I noticed the OSK height doesn't reliably resize to the correct height when toggling between portrait and landscape orientation. (landscape usually ends up "taller" than what the default dimension specifies)

Going into the globe button menu and existing seems to refresh OSK to the correct dimensions.
This affects both in-app and system keyboard; this issue is about the system keyboard. See #5252 for in-app.

Keyman 15

mcdurdin:
Fixes #5211.

Cherry-pick of #5247.


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.276](https://keyman.com/downloads/releases/stable/14.0.276)

PR #5281, opened 2021-06-14
fix(linux): Adjust version of dependency for ibus-keyman

ermshiperete:
This change updates the version number for the keyman-keyboardprocessor dependency of ibus-keyman to a version that doesn't have wrong path for i386.

### DIFF
--- a/linux/ibus-keyman/debian/control
+++ b/linux/ibus-keyman/debian/control
@@ -10,7 +10,7 @@ Build-Depends:
  libgtk-3-dev,
  libibus-1.0-dev (>= 1.2),
  libjson-glib-dev (>= 1.4.0),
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

I'm assuming the jenkins CI builds will pass once this gets merged

ermshiperete:
Eventually it will, yes. However, it needs to build 15.0.68  first - having the package for only Bionic isn't sufficient as I now realize. So we should wait with merging until that happened (seems I can't change a PR to draft once it's approved).

mcdurdin:
> seems I can't change a PR to draft once it's approved

The link to do this is up the top right, under `Reviewers`  :grin:

keyman-server:

PR #5298, opened 2021-06-16
fix(developer): kmconvert help match reality

mcdurdin:
kmconvert built-in help was missing `-targets`, misspelled `-full-copyright`, and the parameter checking was skipping the last parameter (which only mattered if the last parameter was `-nologo`, but still...).


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.70-alpha](https://keyman.com/downloads/releases/alpha/15.0.70)

PR #5299, opened 2021-06-16
fix(developer): kmconvert help match reality 🍒

mcdurdin:
Cherry-pick from #5298.

kmconvert built-in help was missing `-targets`, and the parameter checking was skipping the last parameter (which only mattered if the last parameter was `-nologo`, but still...).


darcywong00 (approved):
lgtm

keyman-server:

Issue #5316, opened 2021-06-19
bug: Keyman for Mac Breaks Installation of GFF Gurage


dyacob:
**Describe the bug**
The installation of the GFF Gurage keyboard appears to be broken.  Under the Keyman configuration dialog the keyboard filename is seen ("gff_gurage" in the screenshot) rather than the human-friendly name "ጉራጊና (Gurage)" and the keyboard icon is missing.  When Keyman is the selected input method, the keyboard also does not appear under the  "Keyboards" menu.

Another oddity is that clicking the "i" info button in the Configuration dialog, the information dialog that appears has empty data (see screenshot).  The "?" button for documentation does present the expected documentation.  However, the "i" and "?" button for any following keyboard will then work identically (empty info dialog, and the Gurage documentation appears again for the other keyboards).

**To Reproduce**
1. Go to Keyman > Configuration ... > Download ክeyboard...
2. Install GFF Amharic, then GFF Gurage and GFF Geez (the order may not be important)
3. The broken view of the GFF Gurage keyboard should not be visible in the Configuration dialog.
4. Click Info (i) and Help (?) buttons to review dialog contents.

**Expected behavior**
The keyboard should install fully like other keyboards.

**Screenshots**

_Configuration Dialog_
![Screen Shot 2021-06-18 at 8 16 34 PM (2)](https://user-images.githubusercontent.com/8098333/122625460-3db4af00-d073-11eb-9b32-7f0144afe26d.png)

_Empty Info Dialog_
![Screen Shot 2021-06-18 at 8 21 28 PM (2)](https://user-images.githubusercontent.com/8098333/122625467-43aa9000-d073-11eb-987d-b628cdecd024.png)

---

**Keyman for macOS**
 - OS: Big Sur Version 11.4
 - Keyman version: 14.0.276
 - Target Application: Keyman Configuration

---

**Keyboard**
 - Keyboard name: GFF Gurage
 - Keyboard version: 0.8
 - Language name: Gurage (sgw-Ethi)

**Additional context**
The issue has been observed in other versions of both Keyman for macOS and earlier versions of the GFF Gurage keyboard.  The Keyboard does install fine on Windows 10 and had been fine under Keyman 13 and the previous version macOS.

mcdurdin:
Version 14.0.277 is now available which should address this.

PR #5330, opened 2021-06-21
fix(mac): kmp.inf is windows-1252 🍒

mcdurdin:
Cherry-pick of #5328.

Relates to #5316.

Keyman for mac is still reading from kmp.inf, but was treating it as UTF-8. However, it is windows-1252 encoding. This was stopping Keyman from loading the file due to the presence of a (C) symbol.

Keyman for mac *should* be loading kmp.json. This needs to be addressed as a separate issue.


rc-swag (approved):
lgtm

keyman-server:

PR #5338, opened 2021-06-23
refactor(android): setKeyboard and setKeymanLanguage

mcdurdin:
This refactors the cross-boundary call to `setKeymanLanguage` to use JSON wrapping for the content. This is (a) safer, and (b) less mucking around anyway.

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -507,60 +507,37 @@ final class KMKeyboard extends WebView {
         KMManager.getLatestKeyboardFileVersion(getContext(), packageID, keyboardID) : null;
     }
 
+    if(kOskFont == null || kOskFont.isEmpty())
+      kOskFont = kFont;
+
+    JSONObject jDisplayFont = makeFontPaths(kFont);
+    JSONObject jOskFont = makeFontPaths(kOskFont);
+
+    txtFont = getFontFilename(jDisplayFont);
+    oskFont = getFontFilename(jOskFont);
+
     String kbKey = KMString.format("%s_%s", languageID, keyboardID);
 
     setKeyboardRoot(packageID);
     String keyboardPath = makeKeyboardPath(packageID, keyboardID, keyboardVersion);
 
-    String tFont = "''";
-    String oFont = null;
-    if (kFont == null) {
-      txtFont = "";
-    } else {
-      if (FileUtils.hasFontExtension(kFont)) {
-        txtFont = kFont;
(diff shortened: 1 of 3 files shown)


android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -1056,8 +1032,22 @@ private String getSubkeyText(String keyID, String keyText) {
    * @param font String font JSON object as a string
    * @return String of modified font information with full paths. If font is invalid, return "''"
```
darcywong00, line 1033: This now returns a JSONObject

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -1056,8 +1032,22 @@ private String getSubkeyText(String keyID, String keyText) {
...
+  private JSONObject makeFontPaths(String font) {
+
+    if(font == null) {
+      return null;
+    }
+
     try {
+      if (FileUtils.hasFontExtension(font)) {
+        JSONObject jfont = new JSONObject();
+        jfont.put(KMManager.KMKey_FontFamily, font.substring(0, font.length()-4));
```
darcywong00, line 1044: Do we get `IndexOutOfBoundsException` if font.length() < 4?

mcdurdin: No, because `hasFontExtension` cannot return `true` in that case (i.e. `.ttf` has length 4). It's a little bit assumed but it's safe.

darcywong00 (approved):
lgtm

keyman-server:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.80-alpha](https://keyman.com/downloads/releases/alpha/15.0.80)

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -507,60 +507,37 @@ public boolean setKeyboard(String packageID, String keyboardID, String languageI
         KMManager.getLatestKeyboardFileVersion(getContext(), packageID, keyboardID) : null;
     }
 
-    String kbKey = KMString.format("%s_%s", languageID, keyboardID);
+    if(kOskFont == null || kOskFont.isEmpty())
+      kOskFont = kFont;
 
-    setKeyboardRoot(packageID);
```
darcywong00, line 512: `setKeyboardRoot(packageID)` needs to remain before the font path munging. Otherwise, issue #5418 results.

PR #5341, opened 2021-06-23
fix(windows): osk scaling mismatch on horz axis

mcdurdin:
Font scaling on individual keys would ignore the vertical axis if the horizontal axis reached the key width threshold, rather than calculating the minimum of the two axes. This lead to weird effects, such as adding characters to a key could make the font larger rather than smaller!

Before:

image

After:

image

Reported by @jmdurdin.


darcywong00 (approved):
lgtm

keyman-server:

PR #5342, opened 2021-06-23
fix(windows): osk scaling mismatch on horz axis 🍒

mcdurdin:
Cherry-pick of #5341.

Font scaling on individual keys would ignore the vertical axis if the horizontal axis reached the key width threshold, rather than calculating the minimum of the two axes. This lead to weird effects, such as adding characters to a key could make the font larger rather than smaller!

Reported by @jmdurdin.


darcywong00 (approved):
lgtm

mcdurdin:
@jmdurdin, this pull request will be updated when the 14.0 release containing it is published. (You should get a notification)

keyman-server:

PR #5363, opened 2021-06-24
fix(linux): Don't add duplicate entries when reinstalling keyboard

### LINKED ISSUE #5308: bug(linux): km-package-install should not add an extra entry in ibus when updating an existing package
**Describe the bug**

```shell
km-package-install -f sample.kmp
km-package-install -f sample.kmp
```

This will result in two entries in ibus, for a single package. We should not be re-adding the package to ibus when it is already listed.

However:
```shell
km-package-uninstall sample
```
does remove both entries.

---

 - OS: Ubuntu 20.04 
 - Keyman version: 14.0.276

ermshiperete:
When we update/reinstall an existing keyboard we don't want to add a new entry to the list of keyboards.

Fixes #5308.


darcywong00 (approved):
lgtm

keyman-server:

PR #5371, opened 2021-06-25
 fix(linux): Fix restarting ibus when running with sudo 

### LINKED ISSUE #5309: bug(linux): sudo km-package-install -s to install shared packages does not restart ibus as needed
**Describe the bug**

```shell
$ sudo km-package-install -s -f sample.kmp
Neither sentry-sdk nor raven is available. Not enabling Sentry error reporting.
WARNING:could not find connected IBus.Bus
WARNING:Failed to restart IBus
WARNING:'NoneType' object has no attribute 'exit'
WARNING:could not find connected IBus.Bus
```

 - OS: Ubuntu 20.04 
 - Keyman version: 14.0.276

ermshiperete:
When installing in the shared area the user has to run `km-package-install` with sudo so that he has sufficient permissions.
In that case we have to do extra work to restart ibus for the logged-in user instead of root.

Fixes #5309.

Also fixes a crash in `km-package-install -p foo`.

### DIFF
--- a/linux/keyman-config/keyman_config/ibus_util.py
+++ b/linux/keyman-config/keyman_config/ibus_util.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 
+import os
 import time
 import gi
 import logging
@@ -68,15 +69,21 @@ def restart_ibus_subp():
 
 
 def restart_ibus(bus=None):
-    try:
-        if not bus:
-            bus = get_ibus_bus()
-        logging.info("restarting IBus")
-        bus.exit(True)
-        bus.destroy()
-    except Exception as e:
-        logging.warning("Failed to restart IBus")
-        logging.warning(e)
+    realuser = os.environ.get('SUDO_USER')
+    if realuser:
+        # we have been called with `sudo`. Restart ibus for the real user.
+        logging.info('restarting IBus by subprocess for user %s', realuser)
+        subprocess.run(['sudo', '-u', realuser, 'ibus', 'restart'])
+    else:
+        try:
+            if not bus:
+                bus = get_ibus_bus()
+            logging.info("restarting IBus")
+            bus.exit(True)
+            bus.destroy()
+        except Exception as e:
+            logging.warning("Failed to restart IBus")
+            logging.warning(e)
 
 
 def bus_has_engine(bus, name):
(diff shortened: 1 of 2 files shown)


linux/keyman-config/keyman_config/ibus_util.py
```diff
@@ -68,15 +69,21 @@ def restart_ibus_subp():
...
-        bus.exit(True)
-        bus.destroy()
-    except Exception as e:
-        logging.warning("Failed to restart IBus")
-        logging.warning(e)
+    realuser = os.environ.get('SUDO_USER')
+    if realuser:
+        # we have been called with `sudo`. Restart ibus for the real user.
+        logging.info('restarting IBus by subprocess for user %s', realuser)
+        subprocess.run(['sudo', '-u', realuser, 'ibus', 'restart'])
```
mcdurdin, line 76: LGTM but I wonder if it is more coherent to actually call ourselves rather than call out to `ibus`? It just feels a bit funny to be restarting ibus in two different methods, depending on whether or not we are sudo'd.

mcdurdin (approved):

linux/keyman-config/keyman_config/ibus_util.py
ermshiperete: Yes, that would be nicer. Unfortunately, when running as `sudo` we don't have access to the user's ibus interface. That would still leave the option of calling `ibus restart` for the regular user, but I'm hesitant because of potential side effects.

ermshiperete: Yes, that would be nicer. Unfortunately, when running as `sudo` we don't have access to the user's ibus interface. That would still leave the option of calling `ibus restart` for the regular user, but I'm hesitant because of potential side effects.

keyman-server:

linux/keyman-config/keyman_config/ibus_util.py
mcdurdin: I would expect us to call into a helper python script to do the restart (e.g. something like `ibus_util.py -restart-ibus`)

mcdurdin:
Are we going to cherry-pick this to stable-14.0?

mcdurdin:
Not sure if this is related to this PR or something new but getting this message:

```
mcdurdin@thark-ubuntu2004:~$ km-package-install --version
km-package-install version 15.0.76
mcdurdin@thark-ubuntu2004:~$ sudo km-package-install -s -p khmer_angkor
Can't connect to IBus.
```

PR #5372, opened 2021-06-25
refactor(linux): Centralize paths

ermshiperete:
Introduce methods to get paths instead of hard-coding the values. This also deprecates four methods (`user_keyman_dir`,
`user_keyman_font_dir`, `user_keyboard_dir`, and `get_isntall_area_path`).

This change also fixes a crash trying to install a package in the shared area.

### DIFF
--- a/linux/keyman-config/keyman_config/deprecated_decorator.py
+++ b/linux/keyman-config/keyman_config/deprecated_decorator.py
@@ -1,0 +1,25 @@
+#!/usr/bin/python3
+# based on https://stackoverflow.com/a/40301488
+import logging
+
+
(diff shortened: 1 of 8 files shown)


mcdurdin (approved):

keyman-server:

mcdurdin:
Will we cherry-pick this to stable-14.0?

ermshiperete:
> Will we cherry-pick this to stable-14.0?

No, I don't think we gain much on stable-14.0 from this PR.

mcdurdin:
I thought that:

> This change also fixes a crash trying to install a package in the shared area.

is worth cherry-picking

PR #5383, opened 2021-06-28
fix(windows): #5336 appcontext get to handle small buffer

rc-swag:
Get correctly handles caller buffers smaller than the current context buffer. Ensuring the last code point is not truncated
Fixes  #5336

PR #5390, opened 2021-06-29
fix(common): get handle caller buffer size small than internal Fixes …

### LINKED ISSUE #5336: bug(windows): AppContext::Get incorrectly handles buffersizes smaller then CurContext buffer
**Describe the bug**

Get should copy the current context into the supplied buffer managed by the caller of size n. If size n is smaller than the AppContext buffer (source buffer) it should ensure it is returning the charters closest to the caret position and as many characters next to this as size n allows.  Currently, it returns the characters furthest from the caret and is truncated when it reaches the size n of the caller's buffer. 
Part 2 of this bug is that it could write past the end of the caller's buffer.

**Unit Test Required**
A unit test should be written that fails with the current implementation. That will then be used to verify the fix to both these bugs

**Implementation Hints:**
A solution for Get would be to use AppContext::Buf passing in the buffer size as the argument to get a pointer to the position in the CurContext to start the copy.  (as buf already handles sentinel and unit16 pairs) copy from the returned pointer to the termination of the buffer. 

**Keyman for Windows:**
 - Keyman version:

rc-swag:
This is equivalent to the #5383 PR. Which was for the AppContext class this is KMX_context class in the core engine.
Fixes #5336
Also made the GetFullContext return a const for safety

### DIFF
--- a/common/core/desktop/src/kmx/kmx_context.cpp
+++ b/common/core/desktop/src/kmx/kmx_context.cpp
@@ -35,7 +35,8 @@ KMX_WCHAR *KMX_Context::BufMax(int n)
 	if(CurContext == p || n == 0) return p; // empty context or 0 characters requested, return pointer to end of context
 
   KMX_WCHAR *q = p;
-	for(; p != NULL && p > CurContext && (intptr_t)(q-p) < n; p = decxstr(p, CurContext));
+	for(; p != NULL && p > CurContext && (intptr_t)(q-p) < n; p = decxstr(p, CurContext))
+  ; // ; on new line to tell compiler empty loop is intentional
 
   if((intptr_t)(q-p) > n) p = incxstr(p); // Copes with deadkey or supplementary pair at start of returned buffer making it too long
 
--- a/common/core/desktop/src/kmx/kmx_context.cpp
+++ b/common/core/desktop/src/kmx/kmx_context.cpp
@@ -29,6 +29,18 @@ void KMX_Context::Add(KMX_WCHAR ch)
   //DebugLog("KMX_Context::Add(%x):  EXIT [%d]: %s", ch, pos, Debug_UnicodeString(CurContext));
 }
 
+KMX_WCHAR *KMX_Context::BufMax(int n)
+{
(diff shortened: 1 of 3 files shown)


mcdurdin (approved):

mcdurdin:
Note: wait for checks to pass before merging 😁

mcdurdin:
macos build failed:

```
[17:01:16][Step 1/1] c++ -Isrc/libkmnkbp0.0.dylib.p -Isrc -I../../../src -fcolor-diagnostics -Wall -Winvalid-pch -Wnon-virtual-dtor -Werror -std=c++14 -O3 -isystem../../../include -isysteminclude -DKMN_KBP_EXPORTING -Wall -Wctor-dtor-privacy -Wdouble-promotion -Wendif-labels -Wextra -Wno-unknown-pragmas -Wno-missing-field-initializers -Wnon-virtual-dtor -Wshadow -Wimplicit-fallthrough -Wno-double-promotion -Wshorten-64-to-32 -Wno-ctor-dtor-privacy -Wno-non-virtual-dtor -fvisibility=hidden -fvisibility-inlines-hidden -MD -MQ src/libkmnkbp0.0.dylib.p/kmx_kmx_context.cpp.o -MF src/libkmnkbp0.0.dylib.p/kmx_kmx_context.cpp.o.d -o src/libkmnkbp0.0.dylib.p/kmx_kmx_context.cpp.o -c ../../../src/kmx/kmx_context.cpp
[17:01:16][Step 1/1] ../../../src/kmx/kmx_context.cpp:38:87: error: for loop has empty body [-Werror,-Wempty-body]
[17:01:16][Step 1/1]         for(; p != NULL && p > CurContext && (intptr_t)(q-p) < n; p = decxstr(p, CurContext));
[17:01:16][Step 1/1]                                                                                              ^
[17:01:16][Step 1/1] ../../../src/kmx/kmx_context.cpp:38:87: note: put the semicolon on a separate line to silence this warning
```

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.77-alpha](https://keyman.com/downloads/releases/alpha/15.0.77)

PR #5391, opened 2021-06-29
fix(developer): builder and editor commands were ignored

mcdurdin:
As we changed the User Agent string to have 'Keyman' instead of 'TIKE', the builder and editor commands were being ignored. This meant that changes to files would not be saved, among other terrible things.

This change was introduced with the Chromium update.


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.79-alpha](https://keyman.com/downloads/releases/alpha/15.0.79)

PR #5399, opened 2021-06-30
fix(linux): Fix auto-generated help

ermshiperete:


mcdurdin (approved):
I am a little concerned about the maintainability of the script help2md but let's go with it.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.79-alpha](https://keyman.com/downloads/releases/alpha/15.0.79)

PR #5407, opened 2021-07-01
feat(android): Spacebar text controls 🍒

mcdurdin:
Cherry-pick of #5349, #5338.

Relates to #949.

* Refactors the cross-boundary call to `setKeymanLanguage` to use JSON wrapping for the content. This is (a) safer, and (b) less mucking around anyway.
* Adds setSpacebarText() API to Keyman Engine for Android
* Adds displayName API to Keyman Engine for Android
* Sets default spacebar text to keyboard name for FirstVoices Keyboards
* Does not include the preference changes to Keyman for Android -- these will be surfaced as a feature in 15.0.

### DIFF
--- a/android/KMEA/app/src/main/assets/keyboard.html
+++ b/android/KMEA/app/src/main/assets/keyboard.html
@@ -128,49 +128,10 @@
       return true;
     }
 
-    function setKeymanLanguage(keyboardName, internalName, languageName, langId, kbdFile, font, oskFont, package) {
-      var kmw=window['keyman'];
-      var kbdInterface=window.KeymanWeb;
-
-      // Defaults for missing arguments
-      switch (arguments.length) {
-        case 1:
-          internalName = keyboardName;
-          if (keyboardName == 'English') {
-              internalName = 'us';
-          }
-          break;
-        case 2:
-          languageName = keyboardName;
-          break;
-        case 3:
-          langId='';
-          break;
-        case 6:
-          oskFont = font;
-          break;
-      }
-
-      var k = {
-        KN:keyboardName,
-        KI:'Keyboard_'+internalName,
-        KLC:langId,
-        KL:languageName,
-        KF:kbdFile,
-        KP:package
-      };
-      if (font) {
-        k.KFont = font;
-        //window.console.log('keyboard.html assigning k.KFont = ' + font);
-      }
-      if (oskFont) {
-        k.KOskFont = oskFont;
-        //window.console.log('keyboard.html assigning k.KOskFont = ' + oskFont);
-      }
-      kbdInterface.registerStub(k);
-
-      kmw['setActiveKeyboard'](package + '::Keyboard_'+internalName,langId);
-      kmw['osk']['show'](true);
+    function setKeymanLanguage(k) {
+      KeymanWeb.registerStub(k);
+      keyman.setActiveKeyboard(k.KP + '::'+k.KI, k.KLC);
+      keyman.osk.show(true);
     }
 
     /**
--- a/android/KMEA/app/src/main/assets/keyboard.html
+++ b/android/KMEA/app/src/main/assets/keyboard.html
@@ -134,6 +134,11 @@
       keyman.osk.show(true);
     }
 
+    function setSpacebarText(mode) {
+      keyman.options['spacebarText'] = mode;
+      keyman.osk.show(true);
+    }
+
     /**
      * Inserts the selected string <i>s</i>
      * @param dn  Number of pre-caret code points (UTF+8 characters) to delete
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -1032,7 +1032,7 @@ final class KMKeyboard extends WebView {
    * 1. Replace "source" keys for "files" keys
    * 2. Create full font paths for .ttf or .svg
    * @param font String font JSON object as a string
-   * @return String of modified font information with full paths. If font is invalid, return "''"
+   * @return JSONObject of modified font information with full paths. If font is invalid, return `null`
    */
   private JSONObject makeFontPaths(String font) {
 
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -505,60 +505,37 @@ final class KMKeyboard extends WebView {
         KMManager.getLatestKeyboardFileVersion(getContext(), packageID, keyboardID) : null;
     }
 
+    if(kOskFont == null || kOskFont.isEmpty())
+      kOskFont = kFont;
+
+    JSONObject jDisplayFont = makeFontPaths(kFont);
+    JSONObject jOskFont = makeFontPaths(kOskFont);
+
+    txtFont = getFontFilename(jDisplayFont);
+    oskFont = getFontFilename(jOskFont);
+
     String kbKey = KMString.format("%s_%s", languageID, keyboardID);
 
     setKeyboardRoot(packageID);
     String keyboardPath = makeKeyboardPath(packageID, keyboardID, keyboardVersion);
 
-    String tFont = "''";
-    String oFont = null;
-    if (kFont == null) {
-      txtFont = "";
-    } else {
-      if (FileUtils.hasFontExtension(kFont)) {
-        txtFont = kFont;
-        tFont = KMString.format("{\"family\":\"font_family_%s\",\"files\":[\"%s%s\"]}", kFont.substring(0, kFont.length() - 4), keyboardRoot, kFont);
-      } else {
-        txtFont = getFontFilename(kFont);
-        if (!txtFont.isEmpty()) {
-          tFont = makeFontPaths(kFont);
-        }
-      }
+    JSONObject reg = new JSONObject();
+    try {
+      reg.put("KN", keyboardName);
+      reg.put("KI", "Keyboard_" + keyboardID);
+      reg.put("KLC", languageID);
+      reg.put("KL", languageName);
+      reg.put("KF", keyboardPath);
+      reg.put("KP", packageID);
+
+      if (jDisplayFont != null) reg.put("KFont", jDisplayFont);
+      if (jOskFont != null) reg.put("KOskFont", jOskFont);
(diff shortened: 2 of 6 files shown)


mcdurdin:
# User Testing

@keymanapp/testers 

1. Please perform a basic acceptance test on Keyman for Android in attached artifact. There should be no visible change from previous stable-14.0 release - no new crashes etc :)
2. FirstVoices Keyboards:
    * Please perform a basic acceptance test on FirstVoices Keyboards in attached artifact. There should be no new crashes or behaviour changes apart from the point below.
    * The keyboard should now show just the name of the keyboard, not the langtags.json language name. For example, St’át’imcets keyboard should show only 'St’át’imcets' and not 'Lillooet' (langtags.json language name).

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -505,60 +515,38 @@ public boolean setKeyboard(String packageID, String keyboardID, String languageI
...
     }
 
-    String kbKey = KMString.format("%s_%s", languageID, keyboardID);
+    if(kOskFont == null || kOskFont.isEmpty())
+      kOskFont = kFont;
 
-    setKeyboardRoot(packageID);
-    String keyboardPath = makeKeyboardPath(packageID, keyboardID, keyboardVersion);
+    JSONObject jDisplayFont = makeFontPaths(kFont);
+    JSONObject jOskFont = makeFontPaths(kOskFont);
```
jahorton, line 522: As noted by #5424, this would introduce a bug to `stable-14.0`: #5418.

jahorton (changes requested):

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -505,60 +515,38 @@ public boolean setKeyboard(String packageID, String keyboardID, String languageI
...
-          oskFont = null;
-        }
-      }
-    }
+    String kbKey = KMString.format("%s_%s", languageID, keyboardID);
 
-    if (oFont == null) {
-      oFont = tFont;
-    }
+    setKeyboardRoot(packageID);
```
jahorton, line 529: (In particular, the relocation of this specific line.)

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -505,60 +515,39 @@ public boolean setKeyboard(String packageID, String keyboardID, String languageI
         KMManager.getLatestKeyboardFileVersion(getContext(), packageID, keyboardID) : null;
     }
 
-    String kbKey = KMString.format("%s_%s", languageID, keyboardID);
-
     setKeyboardRoot(packageID);
```
mcdurdin, line 518: This should address #5418 for `stable-14.0`. 

@keymanapp/testers: Please test per comment: https://github.com/keymanapp/keyman/pull/5407#issuecomment-873907736.

MakaraSok:
> I'd like to add a test re: #5418
> 
> 1. Install sil_cameroon_qwerty keyboard for a language.
> 2. Verify the OSK (diacritics) of sil_cameroon_qwerty is correct (no tofu)
> 3. Restart Keyman
> 4. When sil_cameroon_qwerty loads, verify the OSK is still correct

@darcywong00 There is no tofu on the diacritics, but the default is loaded right after the keyboard finishes the installation, is this a new normal?

darcywong00 (commented):
Changes lgtm. Will approve after @keymanapp/testers tests

MakaraSok:
> 1. Please perform a basic acceptance test on Keyman for Android in attached artifact. There should be no visible change from previous stable-14.0 release - no new crashes etc :)

Similar behavior to the above comment, the below screenshot is shown right after St̓át̓imcets keyboard was installed. No obvious error is found, but usually EuroLatin keyboard comes with a predictive text on.

No predictive text on EuroLatin keyboard after St̓át̓imcets keyboard was installed, unless manually switch to this keyboard via the globe key.
Screenshot_1625534654

Now, when the globe key is pressed, the St̓át̓imcets keyboard is being selected. See the screenshot below. Notice EuroLatin (SIL) keyboard in the background.
Screenshot_1625535018

MakaraSok:
> * The keyboard should now show just the name of the keyboard, not the langtags.json language name. For example, St’át’imcets keyboard should show only 'St’át’imcets' and not 'Lillooet' (langtags.json language name).

This is shown as expected. See the screenshot below.
Screenshot_1625535208

android/KMEA/app/src/main/assets/keyboard.html
```diff
@@ -128,49 +128,15 @@
...
-        //window.console.log('keyboard.html assigning k.KFont = ' + font);
-      }
-      if (oskFont) {
-        k.KOskFont = oskFont;
-        //window.console.log('keyboard.html assigning k.KOskFont = ' + oskFont);
-      }
-      kbdInterface.registerStub(k);
+    function setKeymanLanguage(k) {
+      KeymanWeb.registerStub(k);
+      keyman.setActiveKeyboard(k.KP + '::'+k.KI, k.KLC);
```
darcywong00, line 133: While working on improved globe key button, I noticed some weirdness in switching keyboards.
I think cause `KeymanWeb.registerStub(k)` is internally changing k.KI to be namespaced

https://github.com/keymanapp/keyman/blob/651932f1bbbb6d7f724b2e602e0c042f227bd982/web/source/kmwembedded.ts#L165

Not sure why it's changed now...

darcywong00: @MakaraSok I think this line is related to the screenshots you posted yesterday.

```suggestion
      keyman.setActiveKeyboard(k.KI, k.KLC);
```

mcdurdin: Ouch, now I understand this. It would be far better for registerStub to not be mutating this parameter (deep copy first)! @jahorton, what do you think?

jahorton: I'm cool with that.  So, sounds like we'll throw #5438's fix (when ready) in with this PR once that one's ready?  Or will we keep those bits separate?  (Either way, sounds like we're rejecting the code suggestion above as-is.)

mcdurdin: > So, sounds like we'll throw #5438's fix (when ready) in with this PR once that one's ready?

Yes. I think that makes sense. The issue that it is fixing is introduced in stable-14.0 in this PR, so happy to hold off until that one is approved and tested and then cherry-pick it into this PR.

darcywong00: #5438 is approved and merged, so it can be 🍒 picked to this PR now

mcdurdin:
@keymanapp/testers please re-test with the latest version (currently building as I write this)

jahorton (approved):

mcdurdin:
@keymanapp/testers just making sure this is on your radar for re-testing

MakaraSok:
> @darcywong00 There is no tofu on the diacritics, but the default is loaded right after the keyboard finishes the installation, is this a new normal?

No tofu and the keyboard does appear right after the installation complete.
Screenshot_1626668200


@mcdurdin There is no noticeable crash during the test. 

Here are the results of the test for what should be shown on the spacebar. 

**The FV keyboard - St’át’imcets -- Lillooet is still shown**
Screenshot_1626668183

**EuroLatin (SIL)**
Screenshot_1626668486


**Khmer Angkor**
Screenshot_1626668341

mcdurdin:
> **The FV keyboard - St’át’imcets -- Lillooet is still shown**

This doesn't look like it was done in the FirstVoices Keyboards app?

Also, please make sure you take 'PASS' or 'FAIL' on each test result; I think I need to find a way to better label tests so we can all keep track of what is being tested.

MakaraSok:
> This doesn't look like it was done in the FirstVoices Keyboards app?

It is not on the FirstVoices Keyboards app. It is the FirstVoices keyboard on Keyman app for Android based off of the artifact on this PR build.

Which of the artifacts on this PR build should be used to test FirstVoices Keyboards app?

mcdurdin:
> Which of the artifacts on this PR build should be used to test FirstVoices Keyboards app?

You are right, this artifact was missing. It should be present as app-debug.apk in the next build on Test-14.0 (Keyman - Android).

mcdurdin:
> You are right, this artifact was missing. It should be present as app-debug.apk in the next build on Test-14.0 (Keyman - Android).

The FirstVoices Keyboards artifact is now present at https://build.palaso.org/viewLog.html?buildId=280804&buildTypeId=KeymanAndroid_TestPullRequests&tab=artifacts#%2Frelease, called app-debug.apk (this has been moved to a separate folder for future builds for clarity).

MakaraSok:
## Keyman for Android
**TEST-KM-ACCEPT: PASSED** - no crashes experienced, but the spacebar label looks different from that of the current stable. 
   - current stable: Language (KeyboardName)
   - this PR: Language - KeyboardName
 
**TEST-KM-NO-TOFU: PASSED** - no tofu as expected.

##  FirstVoices Keyboards for Android
**TEST-FV-ACCEPT: PASSED**
**TEST-FV-SPACEBAR: PASSED**

mcdurdin:
> but the spacebar label looks different from that of the current stable.

This is intentional, good pick 😁

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.278](https://keyman.com/downloads/releases/stable/14.0.278)

mcdurdin:
Thanks @MakaraSok for this testing. The format of the test results is excellent -- really easy to review and be sure that we've captured everything!

android/KMEA/app/src/main/java/com/tavultesoft/kmea/data/Keyboard.java
```diff
@@ -144,6 +150,7 @@ protected void fromJSON(JSONObject installedObj) {
       this.isNewKeyboard = installedObj.getBoolean(KB_NEW_KEYBOARD_KEY);
       this.font = installedObj.getString(KB_FONT_KEY);
       this.oskFont = installedObj.getString(KB_OSK_FONT_KEY);
+      this.displayName = installedObj.getString(KB_DISPLAY_NAME_KEY);
```
jahorton, line 153: Well, in retrospect... we did miss something.  The version in #5349 uses `optString` here.

Issue #5013, opened 2021-05-03
feat(developer): Keyman Core integration with debugger 🐞


mcdurdin:
This is a holding issue for the Keyman Core integration for Developer. We'll be filling this in as we complete planning and design.

**Objective:** 

1. replace existing keystroke debugger with: integration with a standalone version of Keyman Core (not part of Keyman for Windows), and add debugging hooks to Keyman Core to support this.
2. replace existing OSK keyboard import which currently relies on keyman32 with: integration with a standalone version of Keyman Core (not part of Keyman for Windows).

Possible stopgap (or stepping stone): copy existing keyman32 and strip it back to what is required to run standalone in Keyman Developer. 

---

**Keyman Developer:**
 - OS: Windows 
 - Keyman version: target 15.0

mcdurdin:
To do emerging from #5513:

- [ ] Keyboard options
- [ ] Regression testing
- [ ] User-selected base layout
- [ ] Test mode does not work -- do we even need to keep it?
- [ ] Refactor ForceKeyboard property in UfrmDebug.pas
- [ ] Ensure Action_Char and Action_Deadkey have same method of inserting text
- [ ] Ensure insertion point position is tracked correctly during single-step debugging
- [ ] Define behaviour around selection (in DebugEvent.pas)
- [ ] Hide Debug Events form
- [ ] `// TODO: remove +1 -1 messes for deadkey codes //GetDeadkeyMarker(i);` in UfrmDebug.pas
- [ ] `// TODO: this method of inserting chars differs from Action_Char;` in UfrmDebug.pas

mcdurdin:
# User Testing

This is a test on the latest alpha release, as there are multiple PRs that have all merged to master that relate to the Debugger.

* TEST_INSTALL: install and start the debugger

  1. Install Keyman Developer latest alpha
  2. Open any (desktop) keyboard project, load the .kmn, then press F5.
  3. The debugger should start with no errors.

* TEST_RULE_SIMPLE: with khmer_angkor, test the <kbd>x</kbd> key generates ខ

  1. Start the keyboard debugger, press <kbd>x</kbd>

* TEST_RULE_COMPLEX: with khmer_angkor, test that <kbd>x</kbd><kbd>SHIFT</kbd>+<kbd>E</kbd><kbd>j</kbd><kbd>m</kbd><kbd>r</kbd> generates ខ្មែរ.

[...]

PR #5425, opened 2021-07-06
feat(common/core/desktop): kmx debugger basic infrastructure 🐞

mcdurdin:
Part of #5013.

Adds all the debug infrastructure:

* public API headers
* public API implementation classes and functions
* basic unit tests
* helper classes for recording debug events
* minimal implementation of kmx-specific debug event signalling -- just BEGIN and END events.
* tangential: unit test assertion macros and coloured output

Follow-up PRs will add:

* all debug event signals in kmx processor
* unit tests for debug event signals
* console-based keyboard regression tests
* ? console-based keyboard debugger
* integration with Keyman Developer interactive debugger


common/core/desktop/src/kmx/kmx_debugger.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <kmx/kmx_processor.h>
+
+using namespace km::kbp;
+using namespace kmx;
+
+void KMX_DebugItems::push_item(uint8_t type, uint32_t flags, LPGROUP group, LPKEY key, PKMX_WCHAR context) {
+  _items->assert_push_entry();
+  km_kbp_state_debug_item item = {type, flags};
+  item.kmx_info.rule = key;
+  //item.processor.store = store;
```
darcywong00, line 18: Can this be removed?

common/core/desktop/src/km_kbp_debug_api.cpp
```diff
@@ -0,0 +1,63 @@
...
+  return KM_KBP_STATUS_OK;
+}
+
+uint8_t
+km_kbp_state_debug_get(
+  km_kbp_state const *state
+) {
+  assert(state);
+  if(!state) {
+    return -1;
```
darcywong00, line 37: Isn't this returning uint8_t?

common/core/desktop/tests/unit/kmnkbd/debug_api.cpp
```diff
@@ -0,0 +1,166 @@
...
+  setup();
+  test_debugging_disabled();
+
+  // Test 2: Debugging enabled, no rule match
+  setup();
+  test_debugging_no_rule_match();
+
+  // Test 3: Debugging enabled, function key pressed, no rule match
+  setup();
+  test_debugging_function_key();
```
darcywong00, line 158: How to tests 2 and 3 verify no rule match?
Would it involve `KM_KBP_DEBUG_FLAG_NOMATCH`?

mcdurdin: At present, they don't as we need to expand the debug logging -- then the tests will fail again and I will update them... but that's in the next PR.

common/core/desktop/src/debug.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+using debug_item = km_kbp_state_debug_item;
+
+class debug_items : public std::vector<debug_item>
+{
+private:
+  bool _is_enabled;
+public:
+  template<typename... Args>
+  debug_items(Args&&... args);
```
ermshiperete, line 27:
```suggestion
    debug_items(Args&&... args);
```

common/core/desktop/src/debug.hpp
```diff
@@ -0,0 +1,74 @@
...
+{
+
+using debug_item = km_kbp_state_debug_item;
+
+class debug_items : public std::vector<debug_item>
+{
+private:
+  bool _is_enabled;
+public:
+  template<typename... Args>
```
ermshiperete, line 26: Can we put this on one line, or indent the second line?

```suggestion
  template<typename... Args> debug_items(Args&&... args);
```

common/core/desktop/src/debug.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <vector>
+
+#include <keyman/keyboardprocessor.h>
+#include <keyman/keyboardprocessor_debug.h>
+
+namespace km {
+namespace kbp
+{
+
+using debug_item = km_kbp_state_debug_item;
```
ermshiperete, line 19: Why do we define this alias? I'm not sure it adds clarity.

common/core/desktop/src/kmx/kmx_debugger.h
```diff
@@ -0,0 +1,99 @@
...
+  push_item(KM_KBP_DEBUG_RULE_ENTER, 0, group, key, context);
+}
+
+inline
+void KMX_DebugItems::push_rule_exit(LPGROUP group, LPKEY key, PKMX_WCHAR context) {
+  push_item(KM_KBP_DEBUG_RULE_EXIT, 0, group, key, context);
+}
+
+inline
+void KMX_DebugItems::push_begin(km_kbp_state_debug_key_info *key_info, uint32_t flags) {
```
ermshiperete, line 85: Wouldn't it be better to have `push_begin` and `push_end` after the c'tor so that the order is the same as in the declaration above?

ermshiperete (commented):
lgtm

common/core/desktop/src/debug.hpp
mcdurdin: It just matches what we are doing with `action_item`; I agree with you that it is unnecessary. Removing it.

common/core/desktop/src/kmx/kmx_debugger.h
mcdurdin: Done.

common/core/desktop/src/km_kbp_debug_api.cpp
```diff
@@ -0,0 +1,63 @@
...
+  return state->debug_items().is_enabled() ? 1 : 0;
+}
+
+km_kbp_state_debug_item const *
+km_kbp_state_debug_items(
+  km_kbp_state const *state,
+  size_t *num_items
+) {
+  assert(state && state->debug_items().size() > 0);
+  if (!state || state->debug_items().empty()) {
```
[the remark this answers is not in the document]
mcdurdin: Yeah I'm of two minds about this. For shorter content, what I've done here is slightly less clear. But for longer content, e.g. long class names, the style in keyboardprocessor.h is hard to use because the open paren is so far right.

This matches the [style guide](https://github.com/keymanapp/keyman/wiki/Keyman-Code-Style-Guide#code-format) I think and I'm aiming to try and use that pattern. (Disclaimer I just added this format to the style guide 🙈🙉🙊).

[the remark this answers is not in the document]
rc-swag: Haha just in time style guide.   I still think there is nothing wrong with the "If its one line keep it one line rule" 
 ELSE then have the YES  examples on the linked [style guide](url) .   ( keeping the seconde NO)
 This is because of all the functions in xstring.h for example including   `u16ncpy` and all functions similar are just going to blow out and make a lot of noise in the file.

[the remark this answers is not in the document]
mcdurdin: See update I just made to style guide

rc-swag (approved):
lgtm

ermshiperete (approved):

keyman-server:

PR #5434, opened 2021-07-06
chore(windows): remove backup build step

mcdurdin:
We no longer need to backup debug files as they are saved to the symbol store during a release build. This reduces the amount of storage required for a release build and should speed up the build also.

Also fixes the broken build where backup had not been updated for the new consistent build paths.

I have updated the build configuration script to skip the backup step if the backup files are not present post-build.


rc-swag (approved):
lgtm

keyman-server:

Issue #1887, opened 2019-07-23
feat(android): The Globe key has a poor user experience


mcdurdin:
> Is there a reason why they can't use the Globe key to switch for quick entry?

I'd have to do some additional user testing, but I think they just don't know what the globe key _does_. Also, on my device (Android 9.0), pressing the globe key is a hinderance for a few reasons:

 1. Opening the modal dialog closes the keyboard (at least in Textra, my texting app, and a few other apps I've tested), thus meaning I have to reselect the text entry I was using whenever I switch the keyboard.
 2. The modal dialog it brings up occasionally locks up, and tries to download the keyboard index (I think), making it unresponsive to inputs for a second or two.
 3. It requires three touches to get to the Latin QWERTY keyboard: one to press the globe; one to select the alternate language. On my device, the languages are way above where my hands were resting to type in my language, so I have to readjust my grip to select either one of the two of my installed languages; finally, one more press to reselect the text area I was typing in. When I'm done writing my word or so in the Latin script, I must press the globe again, then select the language, having to readjust my grip in the process, and reselect the text input. Contrast this with one keypress to switch to Latin, I type my word, and one keypress to switch back.

A feature request to make this a quicker transition _without_ using the globe key and without adding another script to the existing keyboard is to make it so that a very deliberate swipe on the spacebar key to the left swaps to next installed keyboard. A deliberate swipe to the right swaps to the previous installed keyboard. That way, there is a very easy way to swap between installed languages to prevent breaking the flow of the typist.

_Originally posted by @eddieantonio in https://github.com/keymanapp/keyboards/pull/820#issuecomment-514381334_

darcywong00:
Adding feedback from another user via email:

>  ... And although I think the km Khmer keyboard is designed better than Google's, I don't use it because of the slow switch between gboard and km (not between 1 km keyboard to another).

mcdurdin:
> A feature request to make this a quicker transition _without_ using the globe key and without adding another script to the existing keyboard is to make it so that a very deliberate swipe on the spacebar key to the left swaps to next installed keyboard. A deliberate swipe to the right swaps to the previous installed keyboard. That way, there is a very easy way to swap between installed languages to prevent breaking the flow of the typist.

See also https://community.software.sil.org/t/keyboard-change-from-space-bar/3331

mcdurdin:
From https://community.software.sil.org/t/keyman-globe-key-is-it-nessecary/3467/2:

[...]

Issue #4753, opened 2021-03-23
bug(android/app): Backspace after switching system keyboard sometimes fails


darcywong00:
Found while generating screenshots of Keyman usage.

Intermittently, right after switching to a Keyman system keyboard, the backspace fails to delete a character. This feels similar to 14.0 bugs with the in-app context not resetting (#3809).

This may not consistently repro:
1. In an app (e.g. Google Keep or Messages), type with the gff_amharic keyboard (dictionary loaded)
2. Use the globe button to switch to another keyboard (khmer_angkor)
3. The keyboard is dismissed, so click the end of the line to display the keyboard
4. Attempt to backspace

![amharic context](https://user-images.githubusercontent.com/7358010/112080400-84018b80-8bb4-11eb-817d-a51189988fa8.png)

Expected result: the last character gets deleted
Actual result: no change on context
I've also seen this intermittently switching from Gboard context to Keyman keyboard.

Possible workarounds: 
1. Type another character, and then backspace is possible
2. Switch to other system keyboard, then switch back to Keyman keyboard. Backspace should work

### Configuration
Android Pixel 2 emulator (SDK 29)
Keyman 14.0.627-beta
Keyboards used: gff_amharic 2.0 w/ dictionary and khmer_angkor 1.07

darcywong00:
Preliminary investigation
the Keyman system keyboard isn't loaded by the time `SystemKeyboard.onStartInput()` is ready to check the context (and reset it)

darcywong00:
Backspace works fine in 13.0.6222 stable

darcywong00:
Binary search through the alpha builds

14.0.32 - pass
14.0.33 - sil_euro_latin keyboard fails to load
14.0.35 - fail

darcywong00:
The startup race condition is also impacting the numeric field override (`KMManager.setNumericLayer(KeyboardType.KEYBOARD_TYPE_SYSTEM);`)

darcywong00:
Testing numeric layer with suggestions disabled.
On the Contacts app, Phone number field, I encountered the KMW exception
```
keyman-sentry.js:208 Could not load model 'nrc.en.mtnt': Cannot read property 'postConfigure' of undefined
```

PR #5437, opened 2021-07-07
feat(android/engine): Improve globe key experience

darcywong00:
Addresses the immediate changes of #1887 

Description updated due to review comments...

## Keyman for Android changes
* Change shortpress globe actions from `GLOBE_KEY_ACTION_SHOW_MENU` to `GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD`
* Special  lock-screen handling of system keyboard preserved
* For keyboard.html, I also found `KeymanWeb.registerStub()` was handling the package namespacing for `keyman.setActiveKeyboard()`

Remaining items on the issue (external keyboard changes, flick, and menu) are deferred to future PRs

## User Testing
@keymanapp/testers 
Setup
* Install the PR build
* Install a second Keyman keyboard
* From the Keyman Settings menu, set Keyman as a system keyboard and set as default
* From the device Settings menu, assign a lock screen password "1234" (also strongly recommend a finger print unlock if using a physical device)
* For the following tests
    * short press globe is held less than 500ms
    * long press globe is longer than 500ms
    * When verifying a selected keyboard appears, also verify the suggestion banner shows if there's an associated lexical model

### Short press
1. In the Keyman app, short press the globe several times
2. Verify the in-app keyboard switches to the next Keyman keyboard each  time

### Long press
1. In the Keyman app, long press the globe
2. Verify the Keyboard Picker menu appears
3. Select another Keyman keyboard
4. Verify the selected keyboard appears

### Short press
1. Load a separate  app like Chrome and select a text field to bring up the  Keyman keyboard
2. Short press the globe several times
3. Verify the system keyboard switches to the next Keyman keyboard each time

### Long press 
1. Long press the globe 
2. Verify the Keyboard Picker menu appears
3. Select another Keyman keyboard
4. Verify the selected keyboard appears
5. Long press the globe
6. Select the "Next Input Method" button
7. Verify the next system IME is selected
8. From the system Settings, select Keyman as the system IME

### Lock the screen
1. Lock the device screen
2. For unlocking the screen, verify Keyman keyboard appears.
3. Short press the globe key
4. Verify the next Keyman keyboard appears
5. Long press the globe key
6. Verify the next Keyman keyboard appears (instead of the Keyboard Picker menu)
7. Short press the globe key several times
8. Verify eventually the next system IME appears
9. Use the system IME to unlock the device


mcdurdin (commented):
The concept looks good. There seem to be a few unrelated fixes scattered through, which I would like to see addressed separately. It will be important to coordinate the gesture management with @jahorton's concurrent OSK refactor work.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -1365,6 +1366,9 @@ public static boolean setKeyboard(Context context, int position) {
   public static void switchToNextKeyboard(Context context) {
     int index = KeyboardController.getInstance().getKeyboardIndex(KMKeyboard.currentKeyboard());
     index++;
+    if (index >= KeyboardController.getInstance().get().size()) {
+      index = 0;
+    }
```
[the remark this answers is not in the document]
darcywong00: I can split your other suggestions into separate PRs. This one is just adding robustness, since KeyboardController will automatically revert to index 0.

I'd like to keep this one on this PR though, since we don't assign `switchToNextKeyboard`

jahorton:
> Addresses the immediate changes of #1887
> ## KeymanWeb changes
> 
>     * Add timer to distinguish shortpress and longpress on globe key

This does not belong in KeymanWeb.

web/source/kmwembedded.ts
```diff
@@ -13,27 +13,67 @@
...
+  export class GlobeKeyState {
+    longpressGlobeKey: boolean = false;
+    globeKeyTimer: number;
+
+    setGlobeKeyTimer(): void {
+      this.longpressGlobeKey = false;
+
+      this.globeKeyTimer = window.setTimeout(function() {
+        this.longpressGlobeKey=true;
+      }.bind(this), 500)
```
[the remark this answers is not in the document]
jahorton: Probably so.  This is also the wrong spot to do this; the timer stuff belongs in the Android app... or, at worst, its keyboard.html.

jahorton (changes requested):
There should be no changes to kmwembedded.ts here.

mcdurdin:
> This does not belong in KeymanWeb.

Can you please explain your rationale?

jahorton:
> > This does not belong in KeymanWeb.
> 
> Can you please explain your rationale?

Sure.  Subkeys are also longpresses, yet we don't see a timer in kmwembedded.ts to handle those.  We should be consistent in our approach for the two longpress types... and the fact that we don't see a timer (within KMW) for subkeys implicitly means there's no _need_ for a timer for this case, either.  There's also an internal message thread between Darcy and I where I previously mentioned better spots to do this.

Secondly, there _will_ be unnecessary collateral effects for the iOS engine.

Thirdly, even if changes _were_ required on the WebView side... there's Android's keyboard.html for Android-specific changes.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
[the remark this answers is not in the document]
darcywong00: Though maybe I can bundle this with the registerAssociatedLexicalModel call onto the separate PR and temporarily assign the globe key to switchToNextKeyboard

darcywong00:
> Sure. Subkeys are also longpresses, yet we don't see a timer in kmwembedded.ts to handle those. We should be consistent in our approach for the two longpress types... and the fact that we don't see a timer (within KMW) for subkeys implicitly means there's no _need_ for a timer (within KMW) for this case, either.

Current, a longpress on a key triggers to regular press first (key preview), then later the subkeys appear.
Do you have a suggestion so a longpress on globe doesn't trigger a shortpress globe action first?

web/source/kmwembedded.ts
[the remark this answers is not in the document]
darcywong00: Won't we be able to detect longpress globe with #5387 ?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -1378,6 +1382,8 @@ public static void switchToNextKeyboard(Context context) {
     if (SystemKeyboard != null) {
       SystemKeyboard.setKeyboard(kbInfo);
     }
+
+    registerAssociatedLexicalModel(kbInfo.getLanguageID());
```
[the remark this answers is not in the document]
darcywong00: Split to #5439

android/KMEA/app/src/main/assets/keyboard.html
```diff
@@ -130,7 +131,8 @@
 
     function setKeymanLanguage(k) {
       KeymanWeb.registerStub(k);
-      keyman.setActiveKeyboard(k.KP + '::'+k.KI, k.KLC);
+      // KMW will handle the k.KP:: namespacing
+      keyman.setActiveKeyboard(k.KI, k.KLC);
```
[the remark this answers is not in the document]
darcywong00: Split to #5438

jahorton:
> > Sure. Subkeys are also longpresses, yet we don't see a timer in kmwembedded.ts to handle those. We should be consistent in our approach for the two longpress types... and the fact that we don't see a timer (within KMW) for subkeys implicitly means there's no _need_ for a timer (within KMW) for this case, either.
> 
> Current, a longpress on a key triggers to regular press first (key preview), then later the subkeys appear.
> Do you have a suggestion so a longpress on globe doesn't trigger a shortpress globe action first?

So, some sort of 'state' field that notes the gesture's base key is the language-menu key?

web/source/kmwembedded.ts
```diff
@@ -13,27 +13,35 @@
...
+          }
+        } else {
+          if(this.menuEvent) {
+            this.highlightKey(this.menuEvent, false);
+          }
+          if(typeof(window['menuKeyUp']) == 'function') {
+            window['menuKeyUp']();
+          }
+          this.menuEvent = null;
+        }
```
jahorton, line 40: Just going to point out at this point, `keyman['doGlobeKey'](true)` is called where `keyman['showKeyboardList']()` was, while `keyman['doGlobeKey'](false)` is called where `keyman['menuKeyUp']()` was.

You could very easily have Android's keyboard.html make the original two calls simple wrappers for your `doGlobeKey` without losing _anything_.

darcywong00: ah, that works out nice. thanks for the tip!

web/source/kmwembedded.ts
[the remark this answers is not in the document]
darcywong00: Per @jahorton , this PR will use GestureDetector in KMKeyboard to handle the globe longpress.

darcywong00:
I've reverted the KMW changes. This should be ready for @keymanapp/testers and review

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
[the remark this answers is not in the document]
mcdurdin: Can we rebase this on #5439 please?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -1967,15 +1995,36 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
+            // Handle shortpress globe
+            if (InAppKeyboard.keyboardPickerEnabled) {
+              if (inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SHOW_MENU) {
+                showKeyboardPicker(context, KeyboardType.KEYBOARD_TYPE_INAPP);
+              } else if (inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD) {
+                switchToNextKeyboard(context);
+              }
+            } else {
               switchToNextKeyboard(context);
             }
```
mcdurdin, line 2027: This would be easier to follow if we extracted out the actions from the tests, for example:


```java
          if (globeKeyState == GlobeKeyState.GLOBE_KEY_STATE_LONGPRESS) {
            // Longpress globe
            doGlobeKeyLongpressAction();
            globeKeyState = GlobeKeyState.GLOBE_KEY_STATE_UP;
          } else if (globeKeyState == GlobeKeyState.GLOBE_KEY_STATE_UP) {
            // Handle shortpress globe
            doGlobeKeyShortpressAction();
          }
```

These could be refactored slightly more (plus we have the whole in-app vs system keyboard mess -- one of these days this needs cleanup), but as a start:

```java
  void doGlobeKeyLongpressAction() {
    if (InAppKeyboard.keyboardPickerEnabled) {
      showKeyboardPicker(context, KeyboardType.KEYBOARD_TYPE_INAPP);
    }
  }

  void doGlobeKeyShortpressAction() {
    if (InAppKeyboard.keyboardPickerEnabled) {
      if (inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SHOW_MENU) {
        showKeyboardPicker(context, KeyboardType.KEYBOARD_TYPE_INAPP);
      } else if (inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD) {
        switchToNextKeyboard(context);
      }
    } else {
      switchToNextKeyboard(context);
    }
  }
```

web/source/kmwembedded.ts
```diff
@@ -13,7 +13,7 @@
 /*****************************************/
 
 namespace com.keyman.osk {
-  VisualKeyboard.prototype.optionKey = function(this: VisualKeyboard, e: KeyElement, keyName: string, keyDown: boolean) {
+   VisualKeyboard.prototype.optionKey = function(this: VisualKeyboard, e: KeyElement, keyName: string, keyDown: boolean) {
```
mcdurdin, line 16: I know this is whitespace only, but can we avoid any changes to kmwembedded.ts -- please revert it.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -2206,11 +2254,25 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
+        // Update globeKeyState
+        int start = url.indexOf("keydown=") + 8;
+        String value = url.substring(start);
+        boolean globeKeyDown = !value.isEmpty() && Boolean.valueOf(value);
+        if (globeKeyState != GlobeKeyState.GLOBE_KEY_STATE_LONGPRESS) {
+          if(globeKeyDown) {
+            globeKeyState = GlobeKeyState.GLOBE_KEY_STATE_DOWN;
+          } else {
+            globeKeyState = GlobeKeyState.GLOBE_KEY_STATE_UP;
+          }
```
mcdurdin, line 2266:
```suggestion
          globeKeyState = globeKeyDown ? GlobeKeyState.GLOBE_KEY_STATE_DOWN : GlobeKeyState.GLOBE_KEY_STATE_UP;
```

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -2206,11 +2254,25 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
         editor.putBoolean(KMManager.KMKey_ShouldShowHelpBubble, false);
         editor.commit();
 
+        // Update globeKeyState
+        int start = url.indexOf("keydown=") + 8;
+        String value = url.substring(start);
+        boolean globeKeyDown = !value.isEmpty() && Boolean.valueOf(value);
```
mcdurdin, line 2260: I'm seeing this url parsing pattern over and over again. I'd like to see us using a URL parser class instead and then having a helper function if necessary, e.g.

```java
  boolean globeKeyDown = getUrlBooleanValue("keydown");
```

This removes the unnecessary local variable definitions of `start` and `value`. 

In fact, even better would be to extract the entirety of the if test for url.indexOf("globeKeyAction") to a function:

```java
  } else if (url.indexOf("globeKeyAction") >= 0) {
    handleGlobeKeyAction(getUrlBooleanValue(url, "keydown"));
  }
```

mcdurdin (changes requested):
Would you be able to do some refactoring on the overly large `shouldOverrideUrlLoading` functions? I've also made some comments which apply to both the system keyboard and in-app keyboard sections. The less duplication we have here, the better...

MakaraSok:
Tested on Emulator Pixel_4_API_21 (Android 5.0.2):

- Short press is working as expected in app and as a system keyboard.
- Longpress does show the keyboard picker after the finger is lifted, but from the user point of view it should have appeared when it's hold long enough. Right now, if the globe key is not released, the keyboard picker won't appear. This is consistent on both in-app and system keyboard.

> **Long press**
> 
> 1. Long press the globe
> 2. Verify the Keyboard Picker menu appears
> 3. Select another Keyman keyboard
> 4. Verify the selected keyboard appears

The keyboard does appear when switching the keyboard while using the Google search bar, but not any other apps like Message, browser, contact...etc.

This is what is shown after switching the keyboard in a text field on Contact app. Only when you click on in the field, the keyboard would appear.

Screenshot_1625725146

> 5. Long press the globe
> 6. Select the "Close Keyman" button (will be renamed to "Next Input Method")

There is no "Close Keyman" button to select, but there is a button there now read "Next Input Method". Clicking on it does not seem to do anything. 

Screenshot_1625725284

Clicking in the field again, the Android build-in keyboard would appear.

Screenshot_1625725503


> **Lock the screen**
> 
> 1. Lock the device screen
> 2. For unlocking the screen, verify Keyman keyboard appears.
> 3. Short press the globe key
> 4. Verify the next Keyman keyboard appears
> 5. Long press the globe key
> 6. Verify the next Keyman keyboard appears (instead of the Keyboard Picker menu)

The next Keyman keyboard doesn't appear, but when short/long press on the globe key, it switches between the Android keyboard and Keyman keyboard. Note that only two Keyman keyboards are installed, i.e. EuroLatin (SIL) and Cameroon QWERTY. 

On lockscreen, when press the power button:

Screenshot_1625726294

On lockscreen, when press the globe key:

Screenshot_1625726299

On lockscreen, press the globe key again to see EuroLatin keyboard and again to see the Android keyboard.

MakaraSok:
On a separate note, the backspace key is not working. Pressing on it would not delete anything.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
darcywong00: I did the refactoring. For now, the doGlobeKey*() functions only handle IN_APP keyboard.
the System Keyboard is complicated by the lock-screen code.

darcywong00:
> There is no "Close Keyman" button to select, but there is a button there now read "Next Input Method". Clicking on it does not seem to do anything.

@MakaraSok Can you reproduce this and tell me the steps? I ran the test build and with the system keyboard in a Chrome field, I
1. longpress globe to bring up Keyboard Picker
2. Click the "Next Input Method". 
This closed Keyman and switched to the next IME (Gboard in my device)

MakaraSok:
Retest on Pixel_3a_API_30_x86 running Android 11
---


> ## In-app keyboard  globe action
> ### Short press
> 1. In the Keyman app, short press the globe several times
> 2. Verify the in-app keyboard switches to the next Keyman keyboard each  time
> 

**Passed**

> ### Long press
> 1. In the Keyman app, long press the globe
> 2. Verify the Keyboard Picker menu appears
> 3. Select another Keyman keyboard
> 4. Verify the selected keyboard appears
> 

**Passed** - But the Keyboard Picker menu does appear after lifting the finger right after releasing the globe key. Without release the globe key the menu won't appear no matter how long the globe key is pressed. I would suggest that the menu appear after the key is press for a certain length of time.


> ## System keyboard globe action
> ### Short press
> 1. Load a separate  app like Chrome and select a text field to bring up the  Keyman keyboard
> 2. Short press the globe several times
> 3. Verify the system keyboard switches to the next Keyman keyboard each time
>

**Failed?** - it looks like either Keyman or Chrome crashes when trying to trigger the keyboard to appear by pressing in the text input area on google.com of the Chrome browser. 

Pitch black screen appears:
Screenshot_1626429740

But then its working:
image

 
> ### Long press 
> 1. Long press the globe 
> 2. Verify the Keyboard Picker menu appears
> 3. Select another Keyman keyboard
> 4. Verify the selected keyboard appears
> 5. Long press the globe
> 6. Select the "Close Keyman" button (will be renamed to "Next Input Method")
> 7. Verify the next system IME is selected
> 8. From the system Settings, select Keyman as the system IME
> 

**Passed** - but the same behavior as when testing in-app.

> ### Lock the screen
> 1. Lock the device screen
> 2. For unlocking the screen, verify Keyman keyboard appears.
> 3. Short press the globe key
> 4. Verify the next Keyman keyboard appears
>

**Failed* - the keyboard does switch to the next Keyman keyboard, but a non-Keyman keyboard (Gboard) appears when pressing the globe key again. I don't think this is intended, isn't it?

> 5. Long press the globe key
> 6. Verify the next Keyman keyboard appears (instead of the Keyboard Picker menu)
>

**Failed** - not the next Keyman keyboard that appears, but the non-Keyman keyboard (Gboard) does appear.

> 7. Short press the globe key several times
> 8. Verify eventually the next system IME appears
> 9. Use the system IME to unlock the device
>  

From this point onwards, the keyboard that shows up after a short press is a non-Keyman keyboard.

Before a short press:
image

After a short press:
image

Let's have a remote session on this if you need further clarification on any result.

darcywong00:
@MakaraSok 

> *Failed - the keyboard does switch to the next Keyman keyboard, but a non-Keyman keyboard (Gboard) appears when pressing the globe key again. I don't think this is intended, isn't it?

5. Long press the globe key
6. Verify the next Keyman keyboard appears (instead of the Keyboard Picker menu)
> Failed - not the next Keyman keyboard that appears, but the non-Keyman keyboard (Gboard) does appear.

7. Short press the globe key several times
8. Verify eventually the next system IME appears
9. Use the system IME to unlock the device

---

On the Lock screen, the globe key will switch to the next Keyman keyboard until the end of the Keyman keyboard list.
Then it will switch to the next system IME.
If there's only 1 Keyman keyboard, the globe key would go to the next system IME.

MakaraSok:
Retested: The keyboard only switch between EuroLatin (SIL) and the Gboard after the first round of cycle through the Keyman keyboards.

Two Keyman keyboards are on the list: EuroLatin (SIL) and Bunong (SIL). Sometimes the switch is between EuroLatin (SIL) and the Gboard, some other time between Bunong (SIL) and Gboard. But never EuroLatin (SIL) > Bunong (SIL) > Gboard and repeat.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -440,6 +450,36 @@ public static boolean executeHardwareKeystroke(
...
+   */
+  private static void doGlobeKeyShortpressAction(Context context, KeyboardType keyboard) {
+    if (keyboard == KeyboardType.KEYBOARD_TYPE_INAPP && InAppKeyboard != null) {
+      if (InAppKeyboard.keyboardPickerEnabled && inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SHOW_MENU) {
+        showKeyboardPicker(context, KeyboardType.KEYBOARD_TYPE_INAPP);
+      } else if (inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD) {
+        switchToNextKeyboard(context);
+      } else {
+        switchToNextKeyboard(context);
+      }
```
mcdurdin, line 479:
```suggestion
      } else { /* inappKbGlobeKeyAction == GlobeKeyAction.GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD */
        switchToNextKeyboard(context);
      }
```

Is that really right?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -2112,6 +2153,26 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
       return false;
     }
+
+    private void handleGlobeKeyAction(boolean globeKeyDown) {
+      // Update globeKeyState
+      if (globeKeyState != GlobeKeyState.GLOBE_KEY_STATE_LONGPRESS) {
+        globeKeyState = globeKeyDown ? GlobeKeyState.GLOBE_KEY_STATE_DOWN : GlobeKeyState.GLOBE_KEY_STATE_UP;
+      }
+
+      if (KMManager.shouldAllowSetKeyboard()) {
```
mcdurdin, line 2163: So is there any situation where we might have `shouldAllowSetKeyboard()` returning `false` and leaving us with a `globeKeyState` which is not `GLOBE_KEY_STATE_UP`?

mcdurdin (approved):
I think this is looking good, assuming all tests pass. There is definitely more beneficial refactoring we can do in the Android code base but this PR is probably not the place for it!

mcdurdin:
> On a separate note, the backspace key is not working. Pressing on it would not delete anything.

Can you give us more detail on this @MakaraSok?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
darcywong00: yeah, that should be fine too.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
darcywong00: I don't think it's possible. But just in case I've added an else to clear `globeKeyState`

darcywong00:
> > On a separate note, the backspace key is not working. Pressing on it would not delete anything.
> 
> Can you give us more detail on this @MakaraSok?

I wonder if it's something similar to  #4753

MakaraSok:
## User Testing - Final Round ;)
---

@keymanapp/testers 
Setup
* **PASSED** - Install the PR build without any issue _on Android 5.1 and 11 simulators_
* **PASSED** - Install a second Keyman keyboard
* **PASSED** - From the Keyman Settings menu, set Keyman as a system keyboard and set as default
* **PASSED** - From the device Settings menu, assign a lock screen password "1234" (also strongly recommend a finger print unlock if using a physical device)
* For the following tests
    * short press globe is held less than 500ms
    * long press globe is longer than 500ms
    * When verifying a selected keyboard appears, also verify the suggestion banner shows if there's an associated lexical model

## In-app keyboard  globe action
### Short press
1. **PASSED** - In the Keyman app, short press the globe several times
2. **PASSED** - Verify the in-app keyboard switches to the next Keyman keyboard each  time -_the loading is not as smooth an experience for the users._

### Long press
1. **PASSED** - In the Keyman app, long press the globe
2. **PASSED** - Verify the Keyboard Picker menu appears
3. **PASSED** - Select another Keyman keyboard
4. **PASSED** - Verify the selected keyboard appears

## System keyboard globe action
### Short press
1. **PASSED** - Load a separate  app like Chrome and select a text field to bring up the  Keyman keyboard
2. **PASSED** - Short press the globe several times
3. **PASSED** - Verify the system keyboard switches to the next Keyman keyboard each time -_the loading is not as smooth an experience for the users._

### Long press 
1. **PASSED** - Long press the globe
2. **PASSED** - Verify the Keyboard Picker menu appears
3. **PASSED** - Select another Keyman keyboard
4. **FAILED/PASSED** - Verify the selected keyboard appears - Failed in Android 5.1 (the OSK disappears unless click in the text input area), but passed on Android 11
5. **PASSED** - Long press the globe
6. **FAILED** - Select the "Close Keyman" button (will be renamed to "Next Input Method") - _No "Close Keyman" button, but there is a button at bottom named "NEXT INPUT METHOD"_
7. **FAILED/PASSED** - Verify the next system IME is selected - _Failed in Android 5.1 (the OSK disappears unless click in the text input area), but passed on Android 11_
8. From the system Settings, select Keyman as the system IME

### Lock the screen
1. Lock the device screen
2. **PASSED** - For unlocking the screen, verify Keyman keyboard appears.
3. **PASSED** - Short press the globe key
4. **FAILED** - Verify the next Keyman keyboard appears - _press it three times the last keyboard would be the other system IME, not looping around the Keyman keyboards. Lock the screen again, set Keyman as the system keyboard. Turn the screen off and back on to see Keyman keyboard, but one short press on the globe key would switch to the other system IME. This is probably because of the end of the keyboard list has reached and it switch automatically to the next system IME._
5. Long press the globe key
6. **FAILED** - Verify the next Keyman keyboard appears (instead of the Keyboard Picker menu) - _either long or short press would switch to the next system IME instead of Keyman keyboard._
7. **FAILED** - Short press the globe key several times - _either long or short press would switch to the next system IME instead of Keyman keyboard._
8. **FAILED** - Verify eventually the next system IME appears - _one tap on the globe keys bring me to the next system IME_
9. Use the system IME to unlock the device

jahorton:
> ## User Testing - Final Round ;)

Ah, the optimism.

keyman-server:

PR #5439, opened 2021-07-07
fix(android/engine): register lexical model in switchToNextKeyboard

darcywong00:
Split out from #5437 and follow-on #5438

This fixes the following in switchToNextKeyboard
* Check for valid keyboard index
* register associated lexical model

Note: This contains temporary globe action `GlobeKeyAction.GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD` which  must be reverted before merging.

## User Testing
@keymanapp/testers 
1. Load the PR build
2. Install the following keyboards: khmer_angkor, gff_amharic, sil_cameroon_qwerty
3. Wait for the keyboards to install and the associated dictioanry for gff_amharic
4. In the Keyman app press the globe key several times (switches to next keyboard w/o KeyboardPicker)
* Verify sil_euro_latin shows suggestions
* Verify khmer_angkor does not show suggestions
* Verify gff_amharic shows suggestions
* Verify sil_cameroon_qwerty does not show suggestions

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
@@ -152,7 +152,7 @@ public final class KMManager {
 
   private static boolean didLogHardwareKeystrokeException = false;
 
-  private static GlobeKeyAction inappKbGlobeKeyAction = GlobeKeyAction.GLOBE_KEY_ACTION_SWITCH_TO_NEXT_KEYBOARD; // For testing only. Do not merge to production
+  private static GlobeKeyAction inappKbGlobeKeyAction = GlobeKeyAction.GLOBE_KEY_ACTION_SHOW_MENU;
   private static GlobeKeyAction sysKbGlobeKeyAction = GlobeKeyAction.GLOBE_KEY_ACTION_SHOW_MENU;
   // This is used to keep track of the starting system keyboard index while the screen is locked
   private static int sysKbStartingIndexOnLockScreen = -1;
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
@@ -152,7 +152,7 @@ public final class KMManager {
 
   private static boolean didLogHardwareKeystrokeException = false;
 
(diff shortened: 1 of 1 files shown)


mcdurdin:
Are there any situations where this could have been broken in stable-14.0?

darcywong00:
> Are there any situations where this could have been broken in stable-14.0?

I ran a test build of stable-14.0 and this needs to be 🍒 picked to it too

MakaraSok:
Tested on Emulator Pixel_2_API_29 (Android 10), all tests pass. The switch is not very smooth comparing with Gboard and Samsung keyboard, but it is not related to this PR.

One thing from the users point of view, is it possible to double tap or longpress to bring up the keyboard picker. It is a little painful to have to cycle through to the last keyboard on the list, i.e. while on EuroLatin (SIL), it takes at least 3 keypresses on the globe key to get to a Cameroon keyboard.

MakaraSok:
Also, "Afade" language was selected to install for "Cameroon QWERTY". After the installation was finished, "Baka" was shown before "Afade". When cycle through the list by repeatedly pressing on the globe key, you'll see "Baka - Cameroon QWERTY" and then "Afade - Cameroon QWERTY" ...etc.

Note that "Baka" was not selected during the keyboard installation.

Screenshot_1625717233
Screenshot_1625717219

darcywong00:
> One thing from the users point of view, is it possible to double tap or longpress to bring up the keyboard picker

Yes, that's what #5437 will do to change 
* shortpress globe to switch keyboard
* longpress globe to bring up keyboard picker

MakaraSok:
It seems like the "Baka" one has been preselected because that's what I see when searching for "Cameroon".

Screenshot_1625722510

Now in the installation process, I'm supposed to select a language, so "Afade" it is. Now by the end of the installation both "Baka" and "Afade" are installed.

This doesn't happen though when tested using the way method on 14.0.277.

darcywong00:
>This doesn't happen though when tested using the way method on 14.0.277.

Ah, the alpha builds search the keyboards from keyman-staging.com. Maybe it's providing a default language tag for the keyboard. Anyways, that's separate from this PR

mcdurdin (approved):
LGTM. Don't forget to remove the test code before merge.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.84-alpha](https://keyman.com/downloads/releases/alpha/15.0.84)

keyman-server:

PR #5468, opened 2021-07-19
chore(android): improve error reporting for kmw

mcdurdin:
Reports the actual error message instead of 'sendKMWError' as the primary message to Sentry.

A further improvement would be to capture the details of the error as data on the error rather than as a breadcrumb.

Yet further would be if we could report the error back from the web browser itself, giving us source context rather than Android context...


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.86-alpha](https://keyman.com/downloads/releases/alpha/15.0.86)

PR #5470, opened 2021-07-20
feat(common/core): debug action index 🐞

mcdurdin:
Adds a first_action index to each debug event (not all debug events need it). This allows the debugger to partially execute a rule including the possibility of multiple updates to context.

### DIFF
--- a/common/core/desktop/src/kmx/kmx_debugger.h
+++ b/common/core/desktop/src/kmx/kmx_debugger.h
@@ -27,7 +27,8 @@ private:
     LPGROUP group = nullptr,
     LPKEY key = nullptr,
     PKMX_WCHAR context = nullptr,
-    PKMX_WORD index_stack = nullptr
+    PKMX_WORD index_stack = nullptr,
+    int first_action = 0
   );
   void fill_store_offsets(
     km_kbp_state_debug_kmx_info *info,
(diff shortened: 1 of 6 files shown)


common/core/desktop/src/kmx/kmx_debugger.h
```diff
@@ -36,20 +37,22 @@ class KMX_DebugItems
...
-  void push_end(uint32_t flags);
-  void push_group_enter(LPGROUP group);
-  void push_group_exit(LPGROUP group, uint32_t flags);
-  void push_nomatch_enter(LPGROUP group);
-  void push_nomatch_exit(LPGROUP group);
-  void push_match_enter(LPGROUP group);
-  void push_match_exit(LPGROUP group);
+  void push_end(uint16_t first_action, uint32_t flags);
+  void push_group_enter(uint16_t first_action, LPGROUP group);
+  void push_group_exit(uint16_t first_action, uint32_t flags, LPGROUP group);
```
ermshiperete, line 42: Why `group` last here? I guess you could say it's always last, but it's always the second parameter except here.

ermshiperete (approved):

common/core/desktop/src/kmx/kmx_debugger.h
mcdurdin: I was trying to match the order of struct members.

keyman-server:

Issue #5052, opened 2021-05-10
chore(common/core/desktop): Caps-related system stores are not supported


mcdurdin:
There is currently a gap around support for the three [Caps Lock related system stores](https://help.keyman.com/developer/language/reference/caps):

* `&CapsAlwaysOff` store, `CAPS ALWAYS OFF` statement
: Manages CAPS LOCK behaviour - pressing Caps Lock key has no effect

* `&CapsOnOnly` store, `CAPS ON ONLY` statement
: Manages CAPS LOCK behaviour - pressing Caps Lock key only switches on Caps Lock state

* `&ShiftFreesCaps` store, `SHIFT FREES CAPS` statement
: Manages CAPS LOCK behaviour - pressing Shift key turns off Caps Lock

It is not clear yet whether work is required in common/core/desktop or in consumers of it.

ermshiperete:
**Note:** The implementation introduced a number changes in core significant for developers:

- **BREAKING** Removed `capsLock` from `KMX_Environment`; use modifier flag instead.
- **BREAKING** `km_kbp_process_event` now has an additional parameter, `is_key_down`.
- `KM_KBP_IT_CAPSLOCK` action type:  Added new union member `capsLock` to `km_kbp_action_item`. When the core needs to 
  enable or disable Caps Lock, it creates a `KM_KBP_IT_CAPSLOCK` action with `km_kbp_action_item.capsLock` set to `1` to enable 
  or `0` to disable Caps Lock.

MakaraSok:
@ermshiperete Is this ready for testing?

ermshiperete:
Almost. I discovered a bug (#5591) - testing should wait until the fix for that is merged (#5601) and packaged. I'll let you know.

MakaraSok:
OK. thank you. 🙏

ermshiperete:
@MakaraSok This is ready for testing now. Changes should be available in 15.0.98+

MakaraSok:
@ermshiperete Tried `15.0.100-alpha` and the app cannot be opened.

Here is what was run to install the version:
```
sudo add-apt-repository ppa:keymanapp/keyman
sudo apt install keyman onboard-keyman
```

MakaraSok:
Running this `km-config` opens the Configuration. This test will be done tomorrow.

MakaraSok:
# User Testing

| Notes |
| --- |
| Checked, **PASSED** -- works as expected/indicated; otherwise, **FAILED**. |
[...]

PR #5497, opened 2021-07-27
chore(common/core): Linux side of caps-lock stores 🚧

ermshiperete:
- implement Linux side of cap-lock action processing
- Turn off capslock on key down
    
   This change improves capslock handling with a capsAlwaysOff keyboard. Previously, the caps lock indicator was turned on while 
   holding down the Caps Lock key. With this change we turn it off immediately. This is done by forcing caps lock off even if the
   state already seems to match.
    
-  Also, for capsAlwaysOff keyboards skip further processing of caps lock key after we turned capslock off again.

Implements part of #5052.


linux/ibus-keyman/src/engine.c
```diff
@@ -27,6 +27,16 @@
...
 #include <gio/gio.h>
+
+#ifdef GDK_WINDOWING_X11
+#include <gdk/gdkx.h>
+#include <X11/XKBlib.h>
+#endif
+
+#ifdef GDK_WINDOWING_WAYLAND
+#include <gdk/gdkwayland.h>
+#endif
```
mcdurdin, line 38: This seems to be mixing wayland support into this PR? Is it possible to leave out all new wayland-related code and put that into a separate PR?

Do we see scenarios where both wayland and x11 need to be supported simultaneously?

linux/ibus-keyman/src/engine.c
```diff
@@ -238,32 +253,39 @@ static void reset_context(IBusEngine *engine)
...
+    keyman->xdisplay = GDK_DISPLAY_XDISPLAY(keyman->display);
+  } else
+#endif
+#ifdef GDK_WINDOWING_WAYLAND
+  if (GDK_IS_WAYLAND_DISPLAY(keyman->display)) {
+  } else
+#endif
+  {
+    g_error("unsupported GDK backend");
+  }
```
mcdurdin, line 280: I take it this somewhat awkward construct is quite deliberate -- to allow for any permutation of `#define GDK_WINDOWING_X11` and `GDK_WINDOWING_WAYLAND` (with X11 taking precedence over Wayland if both are defined)?

mcdurdin (approved):
LGTM. I am not sure about including the wayland `#ifdefs` but they are probably okay...?

linux/ibus-keyman/src/engine.c
ermshiperete: > Do we see scenarios where both wayland and x11 need to be supported simultaneously?

yes, the way current Ubuntu versions are configured is that it they build with Wayland, but also have the X11 support enabled for apps that don't support Wayland yet. And I think this will continue for several more years...

keyman-server:

PR #5498, opened 2021-07-27
chore(common/core): Allow shell script for kmcomp

ermshiperete:
This change allows to build the keyboards on Linux if a `kmcomp` script can be found in PATH. Otherwise we check for and use `wine` and `kmcomp.exe`.


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.93-alpha](https://keyman.com/downloads/releases/alpha/15.0.93)

PR #5500, opened 2021-07-27
fix(linux): isKeyDown flag was reversed

mcdurdin:


mcdurdin:
Already addressed in #5497.

PR #5512, opened 2021-07-29
chore(common/core): refactor backspace handling 🐞

mcdurdin:
Moves the responsibility for deletion of markers when the users presses the backspace key, from `kmx_processor::process_event` to `KMX_ProcessEvent::ProcessGroup`. This has no effect on usage of Keyman Core, but:

1. Simplifies the association between context and action further
2. Ensures that `m_actions` and `state->actions` will always be the same length, making future refactoring of these simpler
3. Allows us to use the index into the m_actions array for Debug events, which simplifies the interleaving of these for the upcoming Debugger.

When the user presses backspace at start of context (where there may be markers in the cached context, but no characters), the kmx processor will delete any markers prior to insertion point before requesting that the consumer emit the backspace virtual key back to the application, to allow the application to handle backspace at start of text.

### DIFF
--- a/common/core/desktop/src/kmx/kmx_actions.cpp
+++ b/common/core/desktop/src/kmx/kmx_actions.cpp
@@ -49,13 +49,7 @@ KMX_BOOL KMX_Actions::QueueAction(int ItemType, KMX_DWORD dwData)
     break;
 
   case QIT_BACK:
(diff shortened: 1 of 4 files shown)


ermshiperete (approved):
lgtm

keyman-server:

keyman-server:

keyman-server:

PR #5513, opened 2021-07-29
feat(developer): debugger uses Keyman Core 🐞

mcdurdin:
Start of the refactor of the debugger to use Keyman Core. Basic events work but there is a lot to tidy up yet.

Some items to complete:
- [x] Single-step mode
- [x] Breakpoints
- [x] Debug status windows
- [x] Refactor debugging code out of frmDebug so we can share it for the visual keyboard import tool and command line regression testing tools
- [x] Setup appropriate path for kmnkbp0.dll

For follow-up PRs:
- [ ] Keyboard options
- [ ] Regression testing
- [ ] User-selected base layout
- [ ] Test mode does not work -- do we even need to keep it?
- [ ] Refactor ForceKeyboard property in UfrmDebug.pas
- [ ] Ensure Action_Char and Action_Deadkey have same method of inserting text
- [ ] Ensure insertion point position is tracked correctly during single-step debugging
- [ ] Define behaviour around selection (in DebugEvent.pas)
- [ ] Hide Debug Events form
- [ ] `// TODO: remove +1 -1 messes for deadkey codes //GetDeadkeyMarker(i);` in UfrmDebug.pas
- [ ] `// TODO: this method of inserting chars differs from Action_Char;` in UfrmDebug.pas

This is a big PR. I am happy for it to be reviewed in stages. It is probably best to treat UfrmDebug.pas as an entirely new file because it is essentially a rewrite. There are a number of other files which need further refactoring (all the files without namespaced filenames, essentially) which I won't try and do in this PR -- it's huge enough as it is.

Can also do a peer code review if you prefer.


windows/src/developer/TIKE/debug/Keyman.System.Debug.DebugEvent.pas
```diff
@@ -0,0 +1,393 @@
...
+  end;
+end;
+
+procedure TDebugEventList.Action_EmitKeystroke(const key: Word);
+var
+  event: TDebugEvent;
+begin
+  case key of
+    VK_TAB:    Action_Char(9);  // Emit a tab character
+    VK_RETURN: Action_Char(13); // New line character
```
rc-swag, line 160: Magic numbers but at least the comment tells save you looking it up

rc-swag (commented):
Just a few comments otherwise looks ready to go.

windows/src/developer/TIKE/debug/Keyman.System.Debug.DebugEvent.pas
mcdurdin: Inclined to let these ones slide given they are chars and not virtual keys (Delphi has no `\t`, instead it uses `#9`, so here I could do `Ord(#9)` for Tab but it makes no difference).

rc-swag (approved):
lgtm

keyman-server:

PR #5519, opened 2021-07-29
feat(common/core): Implement shiftFreesCaps and capsOnOnly 🚧

ermshiperete:
Relates to #5052


common/core/desktop/src/kmx/kmx_capslock.cpp
```diff
@@ -60,8 +64,10 @@ void KMX_ProcessEvent::ResetCapsLock(KMX_DWORD &modifiers) {
  * @return TRUE to skip further processing, FALSE to continue with normal key handling
  */
 KMX_BOOL KMX_ProcessEvent::KeyCapsLockPress(KMX_DWORD &modifiers, KMX_BOOL isKeyDown) {
-  if (m_keyboard.Keyboard->dwFlags & KF_CAPSONONLY && !isKeyDown) {
-    SetCapsLock(modifiers, TRUE);
+  if (m_keyboard.Keyboard->dwFlags & KF_CAPSONONLY) {
+    if (isKeyDown)
+      SetCapsLock(modifiers, TRUE, TRUE);
```
mcdurdin, line 69:
```suggestion
    if (!isKeyDown) {
      SetCapsLock(modifiers, TRUE, TRUE);
    }
```

I think (dim dark past here) that on Windows we can only change caps lock state by emitting a caps lock key press (i.e. keydown+keyup pair), so we *have* to do it on key up, as if it is already down it will not change the caps lock state. This would also be the reason you saw holding caps + pressing 'a' would give a lower case 'a': caps lock cannot not be reset until the caps lock is released by the user.

common/core/desktop/tests/unit/kmx/032 - caps control.kmn
```diff
@@ -19,8 +19,8 @@ group(Main) using keys
 + [NCAPS K_2] > 'fail.'
 + [CAPS K_2] > 'pass.'
 
-+ [NCAPS K_3] > 'fail.'
-+ [CAPS K_3] > 'pass.'
++ [NCAPS SHIFT K_3] > 'fail.'
++ [CAPS SHIFT K_3] > 'pass.'
```
mcdurdin, line 23:
```suggestion
c Shift will free caps lock when it is released, so this test should still have Caps Lock on.
+ [NCAPS SHIFT K_3] > 'fail.'
+ [CAPS SHIFT K_3] > 'pass.'
```

mcdurdin (changes requested):

common/core/desktop/src/kmx/kmx_capslock.cpp
ermshiperete: I'm not convinced. On Windows, with a non-Keyman keyboard the capslock indicator turns lights up on keydown. With a CapsAlwaysOff Keyman keyboard, the capslock indicator stays off when holding down the caps lock key. If capslock could be deactivated only on keyup, I'd expect the capslock indicator to be on while pressing the caps lock key.

In this code here, calling `SetCapsLock` on `isKeyDown` makes CapsOnOnly behave similar to CapsAlwaysOff.

If you think it can only be done on keyup on Windows, then maybe we should call it on both keydown and keyup?

mcdurdin: Yeah, I am not sure to be honest. I would have to dig in and test, which I won't have time to look into just now. @rc-swag might be able to do this with you?

mcdurdin: Okay, let's see how it goes with what you have here -- we'll need to test on both platforms with the final result.

common/core/desktop/tests/unit/kmx/032 - caps control.kmn
```diff
@@ -31,3 +32,5 @@ group(Main) using keys
 + [NCAPS K_6] > 'fail.'
 + [CAPS K_6] > 'pass.'
 
++ [CAPS K_7] > 'fail.'
```
mcdurdin, line 35:
```suggestion
c Shift was pressed and released ([K_SHIFT]) so Caps should be off before this
+ [CAPS K_7] > 'fail.'
```

mcdurdin (approved):

common/core/desktop/src/kmx/kmx_capslock.cpp
ermshiperete: ok, further testing shows that @mcdurdin was right :smile:. If we do it in KeyDown, capslock still toggles. So I changed it to KeyUp again.

ermshiperete:
Merged into #5497

Issue #5017, opened 2021-05-03
feat(android): detect Chrome version on Keyman install and prompt for upgrade


mcdurdin:
**Is your feature request related to a problem? Please describe.**

While Android allows for an evergreen Chrome browser, we are observing that many users still have an older version installed. This breaks Keyman, even for version 14. For version 15, our minimum version will be bumped to 57 to support WASM, and this will increase the scope of this problem.

**Describe the solution you'd like**

We should be able to detect the version of Chrome that is installed, and prompt the user to update it and block use of the keyboard (or progressively reduce functionality? stretch goal) until that prerequisite is filled.

If possible, we should try and lead the user to the method of running the upgrade, e.g. Play Store, or manual install? Some research may be required here.

If possible, we should backport this to 14.0 and just set the Chrome minver to 37(?) for Keyman 14.

---

**Keyman for iPhone/iPad/Android:**
 - Device: older Android devices
 - OS: Android 
 - Keyman version: 15.0

darcywong00:
Play Sore link: https://play.google.com/store/apps/details?id=com.google.android.webview

darcywong00:
Spec question: Is this check performed every time the app starts? (vs trying to install ldml keyboard)

mcdurdin:
Yes, on app start makes sense.

darcywong00:
>  block use of the keyboard (or progressively reduce functionality? stretch goal) until that prerequisite is filled.

Can you give more clarification on the "degraded" mode if the minimum Chrome version isn't 57?
The Keyman app should still function with sil_euro_latin. Maybe just block kmp installation if the keyboard keyman version >= 15.0?

mcdurdin:
> Can you give more clarification on the "degraded" mode if the minimum Chrome version isn't 57?

* If chrome version `<57`, allow installation of keyboards but if the keyboard requires v15 then show a warning at keyboard install time, at a minimum, and then if possible, don't allow the keyboard to be selected.

* If chrome version `<37`, then we should block all usage of the touch keyboard entirely.

darcywong00:
So for the scope of things:
* Checking the Chrome version would be in the Keyman app
* keyboard selection / usage of touch keyboard is KMManager (Keyman Engine).

Then the Keyman engine change would impact 3rd party keyboards / FV Android / KAB?

mcdurdin:
> Then the Keyman engine change would impact 3rd party keyboards / FV Android / KAB?

[...]

PR #5520, opened 2021-07-29
feat(android): Add API for checking Chrome version

darcywong00:
Address #5017

* Adds an API `KMManager.getEngineMode()` that queries the device's Chrome version and returns an enum of the Keyman Engine for Android mode:
```
    public enum EngineModeType {
      ENGINE_MODE_TYPE_DISABLED, // WebView doesn't support touch keyboard features
      ENGINE_MODE_TYPE_DEGRADED, // WebView supports touch keyboards but not LDML keyboards
      ENGINE_MODE_TYPE_FULL;     // WebView supports touch keyboards and LDML keyboards
    }
```
Currently, the tiers are:
Chrome not installed or Chrome version <= 37.0 : `ENGINE_MODE_TYPE_DISABLED`
37.0 < Chrome version <= 57.0: `ENGINE_MODE_TYPE_DEGRADED`
57.0 < Chrome version: `ENGINE_MODE_TYPE_FULL`

* Rename WebViewUtil to WebViewUtils
* Add check in Keyman app so if the mode is not `ENGINE_MODE_TYPE_FULL`, the user is prompted to the Play Store to update Chrome (screenshot below from test build where `checkChromeVersion` was modified to go to Play Store)
playstore-prompt

Degraded mode of Keyman Engine (touch keyboard, switching, keyboard package install) will be handled on separate PR


android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
```diff
@@ -0,0 +1,98 @@
...
+      PackageInfo pInfo;
+      pInfo = context.getPackageManager().getPackageInfo("com.android.chrome", PackageManager.GET_ACTIVITIES);
+      if (pInfo != null) {
+        return pInfo.versionName;
+      }
+    } catch (PackageManager.NameNotFoundException e) {
+      KMLog.LogInfo(TAG, "Chrome not installed");
+    }
+    return "";
+  }
```
mcdurdin, line 97: Could we avoid needing to depend on chrome info by doing this?

```
mWebView.getSettings().getUserAgentString()
```

And parsing for Chrome/x.x.x.x?

https://stackoverflow.com/a/29218966/1836776

mcdurdin (changes requested):

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
mcdurdin: Possible alternative: https://stackoverflow.com/a/65005751/1836776

mcdurdin:
Per https://developer.android.com/about/versions/nougat/android-7.0.html#webview, I am concerned that checking Chrome package version may not be adequate on older versions:

>Starting with Chrome version 51 on Android 7.0 and above, the Chrome APK on your device is used to provide and render Android System WebViews. This approach improves memory usage on the device itself and also reduces the bandwidth required to keep WebView up to date (as the standalone WebView APK will no longer be updated as long as Chrome remains enabled).
>
>You can choose your WebView provider by enabling Developer Options and selecting WebView implementation. You can use any compatible Chrome version (Dev, Beta or Stable) that is installed on your device or the standalone Webview APK to act as the WebView implementation.

If that is true, then testing the webview version per user agent string may be the safest approach.

android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/MainActivity.java
```diff
@@ -764,6 +767,24 @@ private void checkSendCrashReport() {
     KMManager.setMaySendCrashReport(maySendCrashReport);
   }
 
+  private void checkChromeVersion() {
+    if (WebViewUtils.getEngineModeType(context,"") != EngineModeType.ENGINE_MODE_TYPE_FULL) {
+      // Notify minimum Chrome version needed
+      String message = "Minimum Chrome version 57 needed for Keyman.";
```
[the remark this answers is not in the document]
darcywong00: Moved to a localizable string on the banner

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
```diff
@@ -0,0 +1,98 @@
...
+
+public final class WebViewUtils {
+  public static final String TAG = "WebViewUtils";
+
+  // Keyman Engine functionality based on Chrome version
+  public enum EngineModeType {
+    ENGINE_MODE_TYPE_DISABLED, // WebView doesn't support touch keyboard features
+    ENGINE_MODE_TYPE_DEGRADED, // WebView supports touch keyboards but not LDML keyboards
+    ENGINE_MODE_TYPE_FULL;     // WebView supports touch keyboards and LDML keyboards
+  }
```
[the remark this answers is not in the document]
darcywong00: Changed to
```
  // Keyman Engine functionality based on Chrome version
  public enum EngineWebViewVersionStatus {
    DISABLED, // WebView doesn't support touch keyboard features
    DEGRADED, // WebView supports touch keyboards but not LDML keyboards
    FULL;     // WebView supports touch keyboards and LDML keyboards
  }
```

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
darcywong00: Changed to parse the user agent string for the Chrome version with a Regex

darcywong00: I believe the user agent string of a system WebView on older devices will also show up as a Chrome version

darcywong00:
Updated screenshot of the banner with a button that goes to the Play Store
Screenshot_1627699601

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
```diff
@@ -0,0 +1,104 @@
...
+  public static final String TAG = "WebViewUtils";
+
+  // Keyman Engine functionality based on Chrome version
+  public enum EngineWebViewVersionStatus {
+    DISABLED, // WebView doesn't support touch keyboard features
+    DEGRADED, // WebView supports touch keyboards but not LDML keyboards
+    FULL;     // WebView supports touch keyboards and LDML keyboards
+  }
+
+  private static final String CHROME_INSTALL_PATTERN_FORMATSTR = "^.*Chrome/([\\d\\.]+)\\s.*$";
```
mcdurdin, line 24: This will fail if the UA string has `Chrome(x.y)` as the last token because of the `\s`.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Get the Chrome version. Returns empty string if Chrome not installed.
+   * @param context - The context
+   * @return String
+   */
+  private static String getChromeVersion(Context context) {
+    if (context == null) {
+      KMLog.LogInfo(TAG, "Chrome not installed");
+      return "";
```
mcdurdin, line 90: Suggest returning `"0.0"` for not installed, so we can still do version comparisons?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    WebView mWebView = new WebView(context);
```
mcdurdin, line 93: Do you construct a new instance because we may not have an instance available? It just seems costly to do that, wonder if it is necessary :) (or do we just not worry?)

mcdurdin (approved):
We'll need to do some testing on down-level devices, 5.0, 7.0 in particular to ensure that this works correctly.

But LGTM barring minor nits.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
darcywong00: I considered passing the Keyboard WebView as a parameter. But thinking ahead, we might not have that instance available during kmp installation.

mcdurdin: I guess we could reuse that instance if available or make a new one if not?

android/KMEA/app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,10 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.VIBRATE" />
+    <queries>
+        <!-- Permission needed to query for Chrome version -->
+        <package android:name="com.android.chrome" />
+    </queries>
```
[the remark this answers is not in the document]
darcywong00: Good catch. Removed

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/WebViewUtils.java
```diff
@@ -0,0 +1,104 @@
...
+  public static final String TAG = "WebViewUtils";
+
+  // Keyman Engine functionality based on Chrome version
+  public enum EngineWebViewVersionStatus {
+    DISABLED, // WebView doesn't support touch keyboard features
+    DEGRADED, // WebView supports touch keyboards but not LDML keyboards
+    FULL;     // WebView supports touch keyboards and LDML keyboards
+  }
+
+  private static final String CHROME_INSTALL_PATTERN_FORMATSTR = "^.*Chrome/([\\d\\.]+).*$";
```
mcdurdin, line 24:
```suggestion
  private static final String CHROME_INSTALL_PATTERN_FORMATSTR = "^.*Chrome/([\\d.]+).*$";
```
?

Are Java regexes greedy by default?

darcywong00: I think I read Java regex is greedy by default. My previous experience is that it also has to match the entire line.

darcywong00:
> We'll need to do some testing on down-level devices, 5.0, 7.0 in particular to ensure that this works correctly.

## User Testing
@keymanapp/testers 

Load the test build on a device/emulator with Android 5.0 Lollipop (SDK 21).
Android 5.0 devices come with WebView 37.0 which is lower than the minimum version that Keyman wants.

1. Launch Keyman and close the "Get Started" panel
2. Verify the app displays a banner with a link to the Play Store for updating Chrome

3. Click the blue "UPDATE CHROME" button
4. Verify the Google Chrome profile on the Play Store is launched

MakaraSok:
## User Testing
@keymanapp/testers 

Load the test build on a device/emulator with Android 5.0 Lollipop (SDK 21).
Android 5.0 devices come with WebView 37.0 which is lower than the minimum version that Keyman wants.

- [x] **PASSED** 1. Launch Keyman and close the "Get Started" panel
- [x] **PASSED** 2. Verify the app displays a banner with a link to the Play Store for updating Chrome
- [x] **PASSED** 3. Click the blue "UPDATE CHROME" button
- [x] **PASSED** 4. Verify the Google Chrome profile on the Play Store is launched

keyman-server:

Issue #5011, opened 2021-05-03
spec(windows): Keyman Core integration 🥑


mcdurdin:
# Keyman Core Integration for Windows

## Table of Contents
1. [Introduction](#introduction)
2. [Architecture](#architecture)
1. [Changes Required to Keyman Engine for Windows](#changes-required-to-keyman-engine-for-windows)

# Introduction 

A keyboard processor implements keyboarding rules. The current Keyman engine on windows has a native processor implementing keyboarding rules for Keyman formatted keyboards (kmx) only. The Keyman core desktop is a Keyman processor that will allow the processing of different keyboard rule formats. Currently it is used by Keyman for Linux. This item of work is to integrate this Keyman core processor into Keyman for Windows. Allowing the Keyman core to cross-platform shared logic. This will enable future keyboard formats to be implemented including LDML Keyboards. 

# Architecture

See the [original design document for diagram](https://docs.google.com/document/d/12AAdSvHBQfCrs1F3zE1m0NDG5qiXABPLCf6ZrhN7Clk/edit?usp=sharing)

Diagram 1.0 gives an overview of how the Keyman Core processor interacts with the Keyman Engine for Windows. The Keyman Keyboard Processor API is consumed by the kmprocess interfacing with its Application Action Queue. The Keyman Core processor shall manage the keyboard rules, state, keyboard options and context. The core processor shall be included as a static library. The context and action queue shall be cached in the Keyman Engine for Windows. This will be used to update the core prior to processing each key event. After each key event, the context cache and action queue in the Keyman Engine for Windows shall be updated from the Keyman Core. 

#

# Changes Required to Keyman Engine for Windows

Include the core Keyman Keyboard Processor as a static library. 

## Use Registry Flag for Initial Integration.

 A registry flag for turning on and off the new core processor shall be provided. This will ease initial integration and regression testing. The flag shall be named `REGSZ_CoreIntegration` and be included with the current debug flags. The following changes in the rest of this specification are stated as if the core integration flag is set. If it is not set it will use the existing processor and data structures.

## Change to K32_load.cpp

There will be one instantiation of the processor per keyboard. 

[...]

PR #5522, opened 2021-07-29
chore(windows): split keyman64 header to keymanengine

rc-swag:
Ground work for #5011 . This PR breaks out the include files and definitions from keyman64.h into keymanengine.h This will allow for projects like kmanalyze to just include keyman64.h and not keymanengine.h and therefore the whole keyman core library.

### DIFF
--- a/windows/src/global/inc/keymanengine.h
+++ b/windows/src/global/inc/keymanengine.h
@@ -1,0 +1,254 @@
+/*
+ * Keyman is copyright (C) SIL International. MIT License.
+ *
+ * Name:             KeymanEngine
+ * Description:      Main header file for windows engine
+ *
+*/
+/***************************************************************************/   // I4006   // I4169
+
+#ifndef _KEYMANENGINE_H
+#define _KEYMANENGINE_H
+
+#ifndef _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES
+#define _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES 1
+#endif
+
+#ifndef _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES_COUNT
+#define _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES_COUNT 1
+#endif
+
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600
+#endif
+
+#ifndef STRICT
+#define STRICT
+#endif
+
+#include <windows.h>
+#include <assert.h>
+#include <msctf.h>
+#include "compiler.h"
+
+/***************************************************************************/
+
+typedef BOOL(WINAPI* IMDLLHOOKProc)(HWND hwndFocus, WORD KeyStroke, WCHAR KeyChar, DWORD shiftFlags);
+
+typedef struct tagIMDLLHOOK
+{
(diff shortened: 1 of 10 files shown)


windows/src/global/inc/keymanengine.h
```diff
@@ -0,0 +1,254 @@
+/*
+ * Keyman is copyright (C) SIL International. MIT License.
+ *
+ * Name:             KeymanEngine
+ * Description:      Main header file for windows engine
+ *
+*/
+/***************************************************************************/   // I4006   // I4169
```
rc-swag, line 8: Not sure how this version history works for e.g. `// I4006` but I have brought them all over throughout this file.

windows/src/global/inc/keymanengine.h
```diff
@@ -0,0 +1,254 @@
...
+DWORD EthnologueStringCodeToDWord(PWSTR EthCode);
+
+PWSTR  GetSystemStore(LPKEYBOARD kb, DWORD SystemID);
+
+DWORD ExceptionMessage(LPSTR Proc, LPEXCEPTION_POINTERS ep);
+
+void keybd_shift(LPINPUT pInputs, int* n, BOOL isReset, LPBYTE const kbd);
+
+//#define KEYEVENT_EXTRAINFO_KEYMAN 0xF00F0000   // I4370
+#ifndef _KEYMAN64_LIGHT
```
rc-swag, line 217: Should we update this define (and therefore all the places it is defined in projects) to KEYMANENGINE_LIGHT?

windows/src/global/inc/keymanengine.h
```diff
@@ -0,0 +1,254 @@
...
+#include "k32_tsf.h"
+
+void ReportActiveKeyboard(WORD wCommand);   // I3933   // I3949
+void SelectKeyboardHKL(DWORD hkl, BOOL foreground);  // I3933   // I3949   // I4271
+BOOL SelectKeyboardTSF(DWORD KeymanID, BOOL foreground);   // I3933   // I3949   // I4271
+BOOL ReportKeyboardChanged(WORD wCommand, DWORD dwProfileType, UINT langid, HKL hkl, GUID clsid, GUID guidProfile);
+void ProcessModifierChange(UINT key, BOOL isUp, BOOL isExtended);   // I4793
+
+#endif  // _KEYMAN64_LIGHT
+#endif	// _KEYMANENGINE_H
```
rc-swag, line 254: fix indents

windows/src/global/inc/keymanengine.h
mcdurdin: Yeah, these are legacy issue references. We have the legacy issues in a private repo, e.g. https://github.com/sillsdev/keyman-issues/issues, so you can search for the tag `I4006` which returns https://github.com/sillsdev/keyman-issues/issues/3997.

windows/src/global/inc/keymanengine.h
mcdurdin: In theory we should be able to remove it once we finish all the refactoring as I reckon its use is another example of this patching.

mcdurdin (approved):
LGTM.

keyman-server:

keyman-server:

keyman-server:

PR #5529, opened 2021-07-30
refactor(linux): Cleanup import of os.path

ermshiperete:
This change imports `os` instead of `os.path`. This is the recommended way. It also reduces confusion whether to use `os.path.whatever()` or just `path.whatever()`.


darcywong00 (approved):
lgtm

I'm assuming the Jenkins build fail is unrelated

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.93-alpha](https://keyman.com/downloads/releases/alpha/15.0.93)

PR #5534, opened 2021-08-01
fix(android): avoid error with empty font data

### LINKED ISSUE #5509: bug(android): JSONException: End of input at character 0 of
This seems to be arising when receiving some keyboard font data which is in an unexpected format. However it is not clear what the data is. 

Step 1: collect more data, I think (report on the string that is causing an exception).

This appears to have come out of changes in #5338.

Sentry Issue: [KEYMAN-ANDROID-17P](https://sentry.keyman.com/organizations/keyman/issues/9250/?referrer=github_integration)

```
JSONException: End of input at character 0 of 
    at org.json.JSONTokener.syntaxError(JSONTokener.java:460)
    at org.json.JSONTokener.nextValue(JSONTokener.java:101)
    at org.json.JSONObject.<init>(JSONObject.java:164)
    at org.json.JSONObject.<init>(JSONObject.java:181)
    at com.tavultesoft.kmea.KMKeyboard.makeFontPaths(KMKeyboard.java:1065)
...
(13 additional frame(s) were not displayed)
```

mcdurdin:
Fixes #5509.

If a keyboard has empty font data string, then we should avoid attempting to parse it.

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -1048,7 +1048,7 @@ final class KMKeyboard extends WebView {
    */
   private JSONObject makeFontPaths(String font) {
 
-    if(font == null) {
+    if(font == null || font.equals("")) {
       return null;
     }
 


darcywong00 (approved):
lgtm

PR #5536, opened 2021-08-02
chore(linux): Consolidate Debian source packages 📦

### LINKED ISSUE #5022: chore(linux): consolidate Debian source packages
Note: as a part of #3869, consolidate the different Debian source packages into a single one that builds all binary packages (suggestion/wish from Debian maintainers).

ermshiperete:
This change consolidates the Debian packages into a single source package, with the exception of the legacy KMFL packages. We put the `Debian` directory in the top-level directory because the Debian tools work better if `Debian` is a top-level directory.

Fixes #5022.


mcdurdin:
> We put the `Debian` directory in the top-level directory because the Debian tools work better if `Debian` is a top-level directory.

I really am not keen on this 😡 -- imagine if we had to do this for every distro and supported environment... it's already bad enough having .flake8, .editorconfig, package.json and package-lock.json, linux-code-workspace, lerna.json, crowdin.yml, etc, etc in our repo root. The slow creeping in of additional files and directories is something to avoid as far as we possibly can.

mcdurdin (changes requested):
If absolutely forced to, could you do a `ln -s debian linux/debian` for a top-level `debian` folder during a build? So that we can avoid having platform-specific junk cluttering the top-level in-repo. Yeah, I'm rabbiting on here but it just causes problems down the line if we don't keep the platforms as isolated as possible.

.gitignore
```diff
@@ -171,3 +171,13 @@ lerna-debug.log
...
+
+# Linux packaging related
+results/
+*.deb
+*.ddeb
+*.dsc
+*.build
+*.buildinfo
+*.changes
+*.tar.?z
```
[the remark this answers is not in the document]
ermshiperete: ok, I reverted that change.

ermshiperete:
Moved the `debian` subdirectory `linux`.

linux/debian/man/km-config.1
```diff
@@ -0,0 +1,21 @@
+.\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.48.4.
```
mcdurdin, line 1: Is it possible to generate these at build time rather than storing in repo?

linux/scripts/dist.sh
```diff
@@ -69,54 +69,50 @@ for proj in ${autotool_projects}; do
...
+        --tar-ignore=linux/keyboardprocessor --tar-ignore=linux/kmflcomp \
+        --tar-ignore=linux/libkmfl --tar-ignore=mac --tar-ignore=node_modules --tar-ignore=oem \
+        --tar-ignore=linux/build* --tar-ignore=common/core/desktop/build \
+        --tar-ignore=resources/devbox --tar-ignore=resources/git-hooks \
+        --tar-ignore=resources/scopes --tar-ignore=resources/web-environment \
+        --tar-ignore=resources/build/*.lua --tar-ignore=resources/build/jq* \
+        --tar-ignore=resources/build/vswhere* --tar-ignore=results \
+        --tar-ignore=web --tar-ignore=windows --tar-ignore=keyman_1* \
+        --tar-ignore=dist --tar-ignore=.pbuilderrc --tar-ignore=VERSION \
+        --tar-ignore=scripts -Zgzip -b .
```
mcdurdin, line 94: This is pretty fragile, because as we add other files and folders in the root we have to remember to update this, buried away in /linux. Is there a way we can improve this? For example, making a temp folder which has a copy of specific files and folders to _include_, and working from that? That's perhaps not entirely wonderful either but I think this as it stands is going to need constant maintenance for unrelated work.

docs/linux-packaging.md
```diff
@@ -89,8 +89,9 @@ are scattered over several source repos:
     - [jenkins.sh](https://github.com/keymanapp/keyman/blob/master/linux/scripts/jenkins.sh)
       gets called from `lsdev-pipeline-library` to create a source package.
 
-  - [linux/*/debian](https://github.com/keymanapp/keyman/tree/master/linux/ibus-kmfl/debian) -
-    each package has a separate `debian` subdirectory with the meta data for the Linux package.
+  - [debian](https://github.com/keymanapp/keyman/tree/master/debian) - this is the `debian`
```
mcdurdin, line 92:
```suggestion
  - [linux/debian](https://github.com/keymanapp/keyman/tree/master/linux/debian) - this is the `debian`
```

mcdurdin (changes requested):
Looking good - minor suggestions at this point, but one maintenance thing I'd like to try and avoid if we can

linux/scripts/dist.sh
ermshiperete: It doesn't really matter if there are additional files and folders in the source package - just makes it a little bigger. But if we copy specific files and put those in the source package we might miss added necessary files which would cause build failures. This needs maintenance in both cases, but in the former we can do the maintenance when we get to it. I think I'd prefer to leave it as is.

linux/debian/man/km-config.1
ermshiperete: Oops - yes! How did these files end up here? They do get generated...

linux/scripts/dist.sh
mcdurdin: ok fair enough

linux/scripts/dist.sh
```diff
@@ -16,36 +16,36 @@ THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BA
...
-            autotool_projects=""
+        if [ "$1" == "keyman" ]; then
+            legacy_projects=""
             extra_projects="$1"
         else
-            autotool_projects="$1"
+            legacy_projects="$1"
             extra_projects=""
         fi
     fi
```
mcdurdin, line 51: This looks like the if/else conditions are essentially identical except for the parameter position. Could you consolidate?

mcdurdin (approved):

linux/scripts/dist.sh
ermshiperete: Consolidate the consolidation? Done.

ermshiperete:
The following PRs were merged into this one: #5542 and #5543

keyman-server:

PR #5541, opened 2021-08-03
fix(android/app): Fix cleanup when progress dialog cancelled

### LINKED ISSUE #5242: bug(android): going back while `progressDialog` is visible does not destroy the activity
**Describe the bug**

If `progressDialog` has been created (e.g. from `downloadKmp`), if you touch the back button, it dismisses the progress dialog but does not destroy it, which makes it impossible to re-enter and start another download.

**To Reproduce**

1. Start to download a keyboard.
2. As soon as the down progress dialog is shown, click Back `<`.
3. Try again.
4. See that nothing happens.

---

 - Device: Android 
 - OS: Android 11 
 - Keyman version: 14.0.275-stable

darcywong00:
Fixes #5242

When the progress dialog for downloading a kmp is cancelled, sometimes it gets left in a state preventing further kmp's from being downloaded.

* This sets the cancel handler on progressDialog to do cleanup.
* Also cleaned up imports on LanguagesSettingsActivity (when I was searching for references of progressDialog)

Initially putting the PR in draft since I haven't been able to repro the reported issue.

### DIFF
--- a/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/DownloadResultReceiver.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/DownloadResultReceiver.java
@@ -47,6 +47,12 @@ public class DownloadResultReceiver extends ResultReceiver {
         packageIntent.putExtras(bundle);
         context.startActivity(packageIntent);
         break;
+      case FileUtils.DOWNLOAD_CANCELLED :
+        downloadedFilename = resultData.getString("filename");
+        String message = String.format(context.getString(R.string.cancelled_downloading_package), downloadedFilename);
+        Toast.makeText(context, message,
+          Toast.LENGTH_SHORT).show();
+        break;
     }
     super.onReceiveResult(resultCode, resultData);
   }
--- a/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/DownloadResultReceiver.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/DownloadResultReceiver.java
@@ -48,10 +48,7 @@ public class DownloadResultReceiver extends ResultReceiver {
         context.startActivity(packageIntent);
         break;
       case FileUtils.DOWNLOAD_CANCELLED :
-        downloadedFilename = resultData.getString("filename");
-        String message = String.format(context.getString(R.string.cancelled_downloading_package), downloadedFilename);
-        Toast.makeText(context, message,
-          Toast.LENGTH_SHORT).show();
+        // NO-OP
         break;
     }
     super.onReceiveResult(resultCode, resultData);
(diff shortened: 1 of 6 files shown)


darcywong00:
Though I can't repro the reported issue, adding `cleanupPackageInstall()` should fix things because it cleans up progressDialog to null. It's already being called if exceptions occur during the kmp download.

jahorton:
I'm not clear on how that method helps with cleanup.  I mean, it seems to help with the _dialog_, sure, but what about partial download artifacts?

darcywong00:
@mcdurdin said in a PM
> maybe that the progressdialog never showed again? it's been a while...

This PR only cleans up progressDialog. 

The kmp download is handled in a separate background service and I haven't found a way to cancel that.

jahorton:
> @mcdurdin said in a PM
> 
> > maybe that the progressdialog never showed again? it's been a while...
> 
> This PR only cleans up progressDialog.
> 
> The kmp download is handled in a separate background service and I haven't found a way to cancel that.

Ah.  `cleanupPackageInstall` isn't a very fitting name for what it does, then - `cleanupProgressDialog` or something similar would fit better.  (Granted, it wasn't introduced in this PR.)

mcdurdin:
I think we can let the download continue in the background if the user clicks Back, so long as there is a way to signal that the user is no longer interested in the results of the download, and throw it away rather than opening UI for it?

Remembering that the user may start a second download after the current one which may run concurrently and which they *are* interested in!

darcywong00:
> I think we can let the download continue in the background if the user clicks Back, so long as there is a way to signal that the user is no longer interested in the results of the download, and throw it away rather than opening UI for it?
> 
> Remembering that the user may start a second download after the current one which may run concurrently and which they _are_ interested in!

ok, while we can't stop the download from finishing, I've added hooks to notify the user about the download being cancelled. And it prevents attempting to install a cancelled download.

Screenshot_1628136428

android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/DownloadResultReceiver.java
```diff
@@ -47,6 +47,12 @@ protected void onReceiveResult(int resultCode, Bundle resultData) {
         packageIntent.putExtras(bundle);
         context.startActivity(packageIntent);
         break;
+      case FileUtils.DOWNLOAD_CANCELLED :
+        downloadedFilename = resultData.getString("filename");
+        String message = String.format(context.getString(R.string.cancelled_downloading_package), downloadedFilename);
+        Toast.makeText(context, message,
+          Toast.LENGTH_SHORT).show();
+        break;
```
mcdurdin, line 55: I wonder if we really need the toast to tell the user that the action they just cancelled was cancelled 😬

mcdurdin (approved):
LGTM

android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/DownloadResultReceiver.java
darcywong00: > so long as there is a way to signal that the user is no longer interested in the results of the download

Oh, I thought you meant signal to the user. 🤷‍♂️ OK, I can make that a no-op

keyman-server:

PR #5556, opened 2021-08-05
fix(windows): tsysinfo format grid correctly

mcdurdin:
Minor fix to make the module list format correctly with chromium browser instead of mshtml, plus a tweak to build to copy xsl files as well to bin.


darcywong00 (approved):
lgtm

keyman-server:

PR #5559, opened 2021-08-05
fix(windows): wrap text in keyboard installation dialog

### LINKED ISSUE #4883: bug(windows): keyboard installation prompt lacks wordwrap / scrolling
**Describe the bug**
When installing a keyboard, the Message doesn't wrap, so the end is clipped off.

**To Reproduce**
1. Install the keyboard from https://github.com/keymanapp/keyboards/pull/1519 (I had the keyboard loaded in Keyman Developer, built it, then used the Install button to install it.)
2. Observe that the Message is clipped. It doesn't wrap to display the entire text. (I can select the Message line, copy and paste it into Notepad to see that all that's missing is the end of "Nigeria." So, in this case, I'm not missing much.
3. Click Cancel avoid installing the keyboard.

**Expected behavior**
The Message should wrap to the following line(s). Alternatively, the Message area could be scrollable.

Has this always been the behaviour? Or has something changed recently. 

**Screenshots**
image

---

Keyman for Windows:
    OS: Windows 10 20H2 x64
    Keyman version: 14.0.271

Keyman Developer: (since I launched the install from Developer)
    OS: Windows 10 20H2 x64
    Keyman Developer version: 14.0.271.0

---

**Keyboard**
 - Keyboard name: Obolo Chwerty
 - Keyboard version: v1.1.2 (PR https://github.com/keymanapp/keyboards/pull/1519 as of 2021-04-15)

SabineSIL:
Fixes #4883.

keyboard installation prompt wordwrap

Before:
image
image

After:
image
image

### DIFF
--- a/windows/src/desktop/kmshell/xml/installkeyboard.css
+++ b/windows/src/desktop/kmshell/xml/installkeyboard.css
@@ -123,5 +123,6 @@ div {
 }
 .otherdetails {
   font-size: 13.3px;
+  overflow-wrap: anywhere;
   vertical-align: top;
 }


keyman-server:

mcdurdin:
Thank you for your contribution!

Yes, it's one line but I recognise that there were a lot of steps to get to that one line fix!

---

I've made a couple of edits to your pull request description and title to fit with our normal workflow pattern:

* You had 4833 instead of 4883 for your bug reference
* I updated the title to describe the fix rather than the bug reference
* I've added the `Fixed #4883` text to allow the issue to auto-close when we merge this PR
* I've added the sample images from your google doc into the description for future reference

You might like to take a look at https://github.com/keymanapp/keyman/wiki/Pull-Request-and-Commit-workflow-notes for more detail on how we format and prepare pull requests. But no stress, this is still awesome!

I won't worry about doing a build because the change is minimal and certainly will not break the build. 😁

mcdurdin (approved):
Looks Good To Me (LGTM)

mcdurdin:
A note for future work: you may want to create a branch on your local repo so you have the ability to update `master` on your local repo rather than doing all work on your `master` branch.

mcdurdin:
@megahirt, @ermshiperete tagging for your reference.

keyman-server:

PR #5562, opened 2021-08-06
fix(common/core/desktop): split smp for context chars

### LINKED ISSUE #5465: bug(common/core): unicode32 char is cast to 16 bit char losing surrogate info
**Describe the bug**

KMX_ProcessEvent::PostString  converts(incorrectly) a surrogate pair unicode-16 to a UTF-32. Which then gets cast away again.

**To Reproduce**
kmx_processevent.cpp
`KMX_ProcessEvent::PostString  line  m_actions.QueueAction(QIT_CHAR, Uni_SurrogateToUTF32(*p, *(p+1)));`

kmx_actions.cpp 
`KMX_Actions::QueueAction
case QIT_CHAR:
    m_context->Add((KMX_WORD) dwData);`

**Expected behavior**

I think just consective calls to QueueAction should fix this as long as *(p+1) is not null
`m_actions.QueueAction(QIT_CHAR, *p);`
`m_actions.QueueAction(QIT_CHAR, *(p+1));`

### LINKED ISSUE #5561: bug(core): smp chars in subgroups do not appear to process correctly
**Describe the bug**

Given the following keyboard:

```
store(&version) '10.0'
begin Unicode > use(Main)
group(Main) using keys

+ [K_1] > U+10333 use(second)

group(second)

U+10333 > 'x'
```

I would expect typing <kbd>1</kbd> to result in `x`, but instead it results in `U+10333` (`𐌳`). I tested this keyboard on Keyman for Windows with legacy core and it worked correctly, so this bug is constrained to Keyman Core.

---

**Keyman Core:**
 - OS: Windows 10 
 - Keyman version: 15.0.94-alpha

mcdurdin:
Fixes #5465.
Fixes #5561.

If a non-BMP Unicode char was emitted into the internal context, it was being truncated to a WORD, which caused data loss for multi-group keyboards.

Includes an update to a unit test to validate this fix.


common/core/desktop/src/kmx/kmx_actions.cpp
```diff
@@ -45,7 +45,12 @@ KMX_BOOL KMX_Actions::QueueAction(int ItemType, KMX_DWORD dwData)
     break;
 
   case QIT_CHAR:
-    m_context->Add((KMX_WORD) dwData);
+    if(Uni_IsSMP(dwData)) {
+      m_context->Add(Uni_UTF32ToSurrogate1(dwData));
+      m_context->Add(Uni_UTF32ToSurrogate2(dwData));
+    } else {
+      m_context->Add((KMX_WORD) dwData);
+    }
```
darcywong00, line 53: Does l.44 also need this check?

mcdurdin: No, because that's a deadkey marker that can only be in range 1-fffd.

darcywong00 (approved):
lgtm

Should this get 🍒 -picked to stable.14.0?

mcdurdin:
> Should this get 🍒 -picked to stable.14.0?

I'm not in a rush to cherry-pick this because we are not receiving reports of users running into problems with it...  We could do it, but it's just more busy work, so inclined to just let it wait for 15.0.

mcdurdin:
@rc-swag, @ermshiperete FYI

keyman-server:

PR #5567, opened 2021-08-07
chore(linux): Add TC publish script

ermshiperete:
The `publish.sh` script runs as a build step on TC to upload to downloads.keyman.com. Previously we had it directly on TC, but since we consolidated the source packages the script has to differ between alpha and stable branch.


mcdurdin (approved):

mcdurdin (changes requested):
Problems with this, can we discuss.

PR #5586, opened 2021-08-16
fix(windows): handle restoring modal dialogs consistently

### LINKED ISSUE #5574: bug(windows): When help/hint dialog is minimized, it cannot be restored
This was reported by a user via email. Video describing the symptom below
Unfortunately, this was an isolated case as he was unable to reproduce the issue in later attempts.

### Configuration
Windows 10 Version 20H2, OS Build 19042.1110
Keyman for Windows  14.0.279

> I tried exiting the program just now, and it shut down without even prompting me to confirm. Reloaded the program, exited it again, and this time it asked me to confirm. But no problem getting to the window this time. Tried it a third time, same result as the second. So I haven't been able to reproduce the problem yet. 

https://user-images.githubusercontent.com/7358010/128788173-b2277c7a-9e62-4f00-8575-6d1810aa17fb.mp4

mcdurdin:
Fixes #5574.

When a user presses Win+M, Win+D, or clicks the desktop button on the taskbar, all windows get minimized, even those that do not have the `WS_MINIMIZE` flag. Delphi modal dialog forms are not expecting this, and do not restore correctly later.

This scenario arises only when a form is created with a `nil` owner and then `form.ShowModal` is called.

The fix was already implemented on Splash and HTML forms, but differently; I have chosen to use the implementation from the Splash form as it is simpler and seems to work correctly.

### DIFF
--- a/windows/src/desktop/kmshell/main/UfrmHTML.dfm
+++ b/windows/src/desktop/kmshell/main/UfrmHTML.dfm
@@ -70,9 +70,4 @@ object frmHTML: TfrmHTML
     TabOrder = 4
     OnClick = cmdForwardClick
   end
-  object ApplicationEvents1: TApplicationEvents
-    OnMessage = ApplicationEvents1Message
-    Left = 232
-    Top = 160
-  end
 end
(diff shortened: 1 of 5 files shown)


darcywong00 (approved):
lgtm

keyman-server:

PR #5594, opened 2021-08-17
fix(developer): get Test Mode working again 🐞

mcdurdin:
Relates to #5013.

This resolves the broken Test Mode with Keyman Core. Previously, Test Mode just dumped the keyboard into Keyman for Windows, and let it do its thing. Given that we no longer have the dependency on Keyman for Windows, the Test window must use Keyman Core to process keystrokes, and that means there are some differences in how Test Mode operates in 15.0:

* Behind the scenes, the Test mode is just the Debug mode with breakpoints and single step disabled, and the Debug Status window hidden
* Test mode requires debug information in the keyboard (see note below)
* You can always switch between Test and Debug mode
* Deadkeys are always shown in the debug window

Test Mode now requires debug information to be included in the keyboard in order for deadkeys to be tracked correctly in the editor. While it would be possible to remove the need for this, it does not add any significant benefit to do so -- doing so would mean that the tester would have less visibility into deadkeys, and complicate the Core marker tracking in the debug window.

I considered removing Test Mode altogether, but it still feels like a useful view for playing with a keyboard without all the debugging paraphernalia cluttering the display. It's a little slower than full native Keyman for Windows (aka the 14.0 Test Mode), because the debug hooks are still running in the background, but it is quite usable.

Note the removal of `Application.ProcessMessages`. This resolves the issue of keystrokes being swallowed by the debugger during non-interactive debug runs, making rapid input work properly in both Debug Mode and Test Mode.

### DIFF
--- a/windows/src/developer/TIKE/dialogs/UfrmMustIncludeDebug.dfm
+++ b/windows/src/developer/TIKE/dialogs/UfrmMustIncludeDebug.dfm
@@ -6,14 +6,10 @@ inherited frmMustIncludeDebug: TfrmMustIncludeDebug
   Caption = 'Keyman Developer Debugger'
   ClientHeight = 133
   ClientWidth = 393
-  Color = clBtnFace
-  Font.Charset = DEFAULT_CHARSET
-  Font.Color = clWindowText
-  Font.Height = -11
   Font.Name = 'MS Sans Serif'
-  Font.Style = []
-  OldCreateOrder = False
   Position = poScreenCenter
+  ExplicitWidth = 399
+  ExplicitHeight = 162
   PixelsPerInch = 96
   TextHeight = 13
   object Label1: TLabel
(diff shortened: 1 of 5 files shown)


windows/src/developer/TIKE/dialogs/UfrmMustIncludeDebug.dfm
```diff
@@ -23,19 +19,22 @@ inherited frmMustIncludeDebug: TfrmMustIncludeDebug
     Height = 39
```
darcywong00, line 19: Just checking the longer caption still fits in the height.

darcywong00 (approved):
lgtm

windows/src/developer/TIKE/dialogs/UfrmMustIncludeDebug.dfm
mcdurdin: Yes, it does 😁

PR #5597, opened 2021-08-18
fix(developer): F6 in debugger and cef focus 🐞

mcdurdin:
Relates to #5013.

If user presses F6 in the debugger (while not in run mode), will now switch to editor view, and vice-versa.

Also fixes the inversion of the focus test in CEF so that focus controls work correctly for web hosts.


darcywong00 (approved):
lgtm

PR #5607, opened 2021-08-19
fix(linux): Fix uploading to launchpad 📦

ermshiperete:
Follow-up of #5536.


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.97-alpha](https://keyman.com/downloads/releases/alpha/15.0.97)

PR #5621, opened 2021-08-23
chore(android/app): Add in-app help for adjusting keyboard height

darcywong00:
This adds in-app help for the feature in #5606 of a new Settings menu for adjusting the on-screen keyboard height.

* Also reorganized "Settings" menu help to match the order from #5606.
* Split some "Settings" menus to a separate screen
* Removed unused `settings-screen-at.png`  image

PR #5622, opened 2021-08-23
chore(android): Allow Info Activity to navigate back

darcywong00:
Follow-on to #5621

I noticed when navigating the InfoActivity help, the back button abruptly kills the activity, which is a nuisance when navigating between help screens.

This updates `onBackPressed()` so the help menus can go back and forth.


jahorton:
Is there a way to provide an "exit" button as well that performs the original behavior, or will users have to drill all the way back to the top in order to exit the help section?  I don't see anything obvious for this at present.

darcywong00:
> Is there a way to provide an "exit" button as well that performs the original behavior, or will users have to drill all the way back to the top in order to exit the help section? I don't see anything obvious for this at present.

OK, I've added a close button that will always be on the top right of the Info Activity
Screenshot_1629856305

PR #5625, opened 2021-08-23
chore(linux): Fix uploading to Sentry 🍒

ermshiperete:
On local builds we shouldn't upload errors to Sentry. The `build-utils.sh` script sets the `UPLOAD_SENTRY` environment variable
for that purpose. Also, Sentry doesn't like having an empty string as pkgversion (which should no longer happen with the rest of this change, but...)

🍒-pick of #5623


darcywong00 (approved):
lgtm

ermshiperete:
@keymanapp-test-bot skip

keyman-server:

PR #5647, opened 2021-08-27
feat(developer): Debugger keyboard options 🐞

### LINKED ISSUE #5642: feat(developer): support keyboard options in debugging
Split out from #5013.

Feature: Keyboard options

- [x] enumerate options in use within the keyboard
- [x] provide interface to view current option values
- [x] provide interface to update option values
- [x] #5695
- [ ] ~~`save()` support~~ (not required for 15.0, I think)
- [x] ensure option cannot be changed by user while in the middle of a debug single-step batch

---

**Keyman Developer:**
 - Keyman Developer version: 15.0

mcdurdin:
Fixes #5642.
Relates to #5013.

This adds support for viewing and setting [keyboard options](https://help.keyman.com/developer/language/guide/variable-stores) in the debugger. 

* The debugger will scan the keyboard for all possible option stores and their known values, and make those available in an Options tab in the Debug Status window, both to show what the current options are and also to allow the developer to change them.

* It is possible to set arbitrary values if you prefer, as a `if(store != 'value')` rule in the keyboard may need that (this is possible in real use as options.htm allows you to set arbitrary values that are never referenced in the .kmx).

This initial implementation does not have 100% coverage of keyboard options; specifically:

1. keyboard options are not saved between debug sessions.
2. in single-step debugging, the keyboard options will be shown as updated at the start of a keystroke batch, not when the rule which calls `set()` is fired. This level of granularity is not currently visible from the engine. It does not impact the correctness of the rule processing, only the view of the current state of the keyboard option.

I do not anticipate supporting either of those functions in 15.0.

This commit renames the existing `TfrmDebugStatus_Options` to `TfrmDebugStatus_Platform`. I had originally planned to show both the platform and keyboard options in the same debug status window, but for space reasons I have split them. (I'm sorry if this makes reading the PR a little more challenging.)

There is a significant amount of rewriting of kmxfileutils.pas and debugkeyboard.pas. This extends support for working with kmx files and modernises the code (e.g. using generics for list types).

Note in particular the `ExpandSentinel` function in kmxfileutils.pas. This is copied from CompileKeymanWeb.pas but has different internal references, so cannot be used as-is -- CompileKeymanWeb works with intermediate object 'files' from the kmx compiler, whereas the debugger works with final binary .kmx files which have a different in-memory layout. It may be possible to unify this in future but that would involve additional risk. (Seriously, I would probably prefer to rewrite CompileKeymanWeb.pas in C++, Rust or TypeScript than to invest in merging these two use cases.)

image

# User Testing

With the following keyboard:

```
store(&VERSION) '10.0'
store(&NAME) 'Keyboard Options '
store(&TARGETS) 'any'

begin Unicode > use(main)
 
group(Main) using keys

store(optkey) 'ralt'
store(optord) 'before'

if(optkey = 'ralt')  + [RALT K_QUOTE] > dk(quote)
if(optkey = 'rctrl') + [RCTRL K_QUOTE] > dk(quote)
if(optkey = '')      + [K_QUOTE] > dk(quote)

if(optkey = 'ralt') + [SHIFT K_QUOTE] > set(optkey = 'rctrl')
if(optkey = 'rctrl') + [SHIFT K_QUOTE] > set(optkey = '')
if(optkey = '') + [SHIFT K_QUOTE] > set(optkey = 'ralt')

if(optord = 'before') + '<' > set(optord = 'after')
if(optord = 'after') + '<' > set(optord = 'before')

+ 'a' > 'a'

match > use(a)

group(a)

if(optord = 'before') dk(quote) 'a' > 'á'
if(optord = 'after') 'a' dk(quote) > 'á'
```

* TEST_OPTION_INITIAL: Confirm that the options shown when first starting debugging are 'ralt' and 'before'
* TEST_OPTION_TRACKED: Confirm that pressing <kbd>&lt;</kbd> swaps `optord` between 'after' and 'before'
* TEST_OPTION_TRACKED_2: Confirm that pressing <kbd>"</kbd> swaps `optkey` between 'ralt', 'rctrl', and '' (empty string).
* TEST_OPTION_MANUALLY_CHANGED: Confirm that changing `optkey` to '' in the Options debug status window does actually cause the debugger to use that option when typing (see TEST_WORKING below on how to test this).
* TEST_WORKING: Confirm that when `optord` is 'after', that you have to type <kbd>a</kbd><kbd>'</kbd>, and when `optord` is 'before', you have to type <kbd>'</kbd><kbd>a</kbd>.
  - note: <kbd>Right Alt</kbd>+<kbd>'</kbd> and <kbd>Right Ctrl</kbd>+<kbd>'</kbd> are not currently being captured by the debugger. This is a separate issue that I am working on resolving. Do not fail the test because of this :wink: (This means you should set `optkey` to '' (empty string) before testing.


MakaraSok:
- **TEST_OPTION_INITIAL: FAILED** System Error. Code 126.
An error occurs soon after pressing <kbd>F5</kbd>.
   image

   Keyman Developer crashes soon after pressing the OK button.
   ```
   Details for error report for Keyman app:
     Application Identifier: tike-15.0.104-alpha-test-5647
     Error Identifier: 00EA75925A8546D9781FB7B75DE46FC6
   ```

   The following tests are blocked.
- **TEST_OPTION_TRACKED: BLOCKED**
- **TEST_OPTION_TRACKED_2: BLOCKED**
- **TEST_OPTION_MANUALLY_CHANGED: BLOCKED**
- **TEST_WORKING: BLOCKED**

mcdurdin:
@keymanapp-test-bot retest

Based on the fixes in #5649, the new build should include kmnkbp0-0.dll so the debugger should start.

MakaraSok:
@mcdurdin It looks like the build fell over again.

MakaraSok:
Should this be tested after this PR #5649 is merged? Even now that the build went through, the error persists.

MakaraSok:
- **TEST_OPTION_INITIAL: PASSED** 'ralt' and 'before' do show up when starting the debugger
- **TEST_OPTION_TRACKED: PASSED** "before" and "after" on `optord` do swap around when pressing <kbd>Shift</kbd> <kbd><</kbd>
- **TEST_OPTION_TRACKED_2: PASSED**  'ralt', 'rctrl', and '' (empty string) on `optkey` do swap around when pressing <kbd>Shift</kbd> <kbd>"</kbd>
- **TEST_OPTION_MANUALLY_CHANGED: PASSED** choosing '' on `optkey` and 'after' and type <kbd>a</kbd> <kbd>'</kbd>, the output is the expected `U+00E1`.
- **TEST_WORKING: PASSED** <kbd>a</kbd> <kbd>'</kbd> does output the expected character, see above.

darcywong00 (approved):
lgtm

keyman-server:

PR #5664, opened 2021-09-01
fix(web,android/app): Select numeric layer when entering a numeric field

### LINKED ISSUE #4850: bug(android/engine): System Keyboard startup doesn't consistently override to numeric layer
The startup race condition is also impacting the numeric field override (`KMManager.setNumericLayer(KeyboardType.KEYBOARD_TYPE_SYSTEM);`)

_Originally posted by @darcywong00 in https://github.com/keymanapp/keyman/issues/4753#issuecomment-804706136_

Steps to repro
1. Install app and set Keyman as a system keyboard
2. Exit the Keyman app
3. Start another app that uses Numeric intents (e.g "Contacts" --> Add a new contect --> Set a phone number)
4. Expected Result: When Keyman starts on sil_euro_latin, it should select numeric layer

Actual result: default layer is shown

darcywong00:
Fixes #4850

This fixes the timing in `SystemKeyboard.onStartInput()` so that `KMManager.resetContext()` can happen and later switch to numeric layer as needed.

* Also includes numeric text fields for temporarily disabling predictions
* KeymanWeb also adds a check for `this.activeKeyboard`

## User Testing
Setup: Install the test build. Enable Keyman as a system keyboard.
The tests assume English locale where the default sil_euro_latin keyboard also has English suggestions.
    
* **TEST_CONTACTS**: Tests that Keyman will toggle between default layer and numeric layer depending on text field type
1. After Keyman is enabled as a system keyboard, close Keyman
2. On the device, open the Contacts app (one that lets you create a new contact and enter names (text) and phone numbers (numbers)
3. With Keyman as the system keyboard, select a name field (First or Last)
4. Verify the Keyman keyboard is in the default layer
5. Verify the Keyman keyboard has suggestions enabled
6. Type a letter or select a suggestion
7. Verify the Name is updated accordingly
5. Select a phone number field
6. Verify the Keyman keyboard switches to the numeric layer (if the keyboard includes a numeric layer)
7. Verify the suggesitons are disabled
8. Type a number
9. Verify the phone number is updated accordingly
10. Select another Name field or email
11. Verify the Keyman keyboard switches to the default layer
12. Verify suggestions are re-enabled

Screenshot of entering a phone number
Screenshot_1630469337


common/core/web/keyboard-processor/src/text/keyboardProcessor.ts
```diff
@@ -714,9 +714,11 @@ namespace com.keyman.text {
     };
 
     setNumericLayer(device: utils.DeviceSpec) {
-      let layout = this.activeKeyboard.layout(device.formFactor);
-      if(layout.getLayer('numeric')) {
-        this.layerId = 'numeric';
+      if (this.activeKeyboard) {
```
darcywong00, line 717: This check guards against console exceptions

MakaraSok:
- **TEST_CONTACTS: BLOCKED** When click in text field, the suggestion is not always on.
   It is off and the banner is blank at times. Tested on Android 11, API 30.
  
   Things are working well on Android 5.0, API 21. _But one issue I observe is that when a suggestion is picked from the banner, a space would be added after it automatically -- which is not necessary. This happens across Android versions._

MakaraSok:
The result is now updated to PASSED because, after rubber ducky with @Darcy, this issue is mild. The suggestion would start working as soon as one starts typing in the field when the suggestion banner is blank.

- **TEST_CONTACTS: PASSED**

MakaraSok:
It is worth to note that the suggestion banner is blank only when switch from a blank number field, i.e. phone, to a text field, i.e. email.

common/core/web/keyboard-processor/src/text/keyboardProcessor.ts
mcdurdin: This check feels like it belongs higher up -- in caller within Keyman for Android -- because it must be happening when Keyman is not fully initialised. 

Not that it's entirely a bad thing to have a sanity check here but feels like it is side-stepping a more fundamental problem.

mcdurdin (approved):
LGTM apart from that one query on this.activeKeyboard in setNumericLayer

common/core/web/keyboard-processor/src/text/keyboardProcessor.ts
darcywong00: Should I have keyboard.html do something like this then?
```
    // Tell KMW to switch to "numeric" layer
    function setNumericLayer() {
      var kmw=window['keyman'];
      if (kmw.core.activeKeyboard) {  // <--- add this
        kmw.setNumericLayer();
      }
    }
```

mcdurdin: Yes. Although I'd test that `kmw` and `kmw.core` are also initialised (`if (kmw && kmw.core && kmw.core.activeKeyboard)`)

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.111-alpha](https://keyman.com/downloads/releases/alpha/15.0.111)

PR #5669, opened 2021-09-03
fix(common/core): insure list pntr is incremented

rc-swag:
There was a bug in `km_kbp_options_list_size` which meant it would never exit if there was at least 1 item in the list. 
The option_api unit tests are incomplete so this was not picked up.

### DIFF
--- a/common/core/desktop/src/km_kbp_options_api.cpp
+++ b/common/core/desktop/src/km_kbp_options_api.cpp
@@ -25,7 +25,7 @@ km_kbp_options_list_size(km_kbp_option_item const *opts)
   if (!opts)  return 0;
 
   auto n = 0;
-    for (; opts->key; ++opts){
+  for (; opts->key; ++opts) {
     ++n;
   }
 
--- a/common/core/desktop/src/km_kbp_options_api.cpp
+++ b/common/core/desktop/src/km_kbp_options_api.cpp
@@ -25,7 +25,9 @@ km_kbp_options_list_size(km_kbp_option_item const *opts)
   if (!opts)  return 0;
 
   auto n = 0;
-  while (opts->key) ++n;
+    for (; opts->key; ++opts){
+    ++n;
+  }
 
   return n;
 }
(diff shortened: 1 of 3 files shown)


rc-swag:
@keymanapp-test-bot skip

common/core/desktop/src/km_kbp_options_api.cpp
```diff
@@ -25,7 +25,9 @@ km_kbp_options_list_size(km_kbp_option_item const *opts)
   if (!opts)  return 0;
 
   auto n = 0;
-  while (opts->key) ++n;
+    for (; opts->key; ++opts){
```
ermshiperete, line 28: Shouldn't this check for `opts != NULL` as well?

And while you're at it, the indentation is wrong for this line.

ermshiperete (dismissed):

common/core/desktop/src/km_kbp_options_api.cpp
darcywong00: Isn't that checked in l. 24 `assert(ops)`?

darcywong00:
Does `km_kbp_options_list_size` get exercised in any of the unit tests?

common/core/desktop/src/km_kbp_options_api.cpp
ermshiperete: Too much C#! I guess I meant `*opts != NULL`

rc-swag:
> Ah, can we make a unit test to exercise this?
Yes I can do that.  I have it tested it on the windows side through a google test but I could make a test for this in the common core. I did actually try getting the option_api working but it was a bit more then the quick fix I wanted to get out :)

common/core/desktop/src/km_kbp_options_api.cpp
rc-swag: The options list should be terminated #define KM_KBP_OPTIONS_END { 0, 0, 0 }
The for loop second statement is checking if opts->key is defined and therefore the same as opts->key!=NULL.
See line 66 in the same file.

mcdurdin:
```suggestion
  for (; opts->key; ++opts) {
```

rc-swag:
> Ah, can we make a unit test to exercise this?

Done - commit #42c13856f84453144c67cd6d85d67acd06f63829

mcdurdin (approved):

mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.116-alpha](https://keyman.com/downloads/releases/alpha/15.0.116)

PR #5685, opened 2021-09-13
fix(windows): fallback to filename if `&name` not set 🍒

### LINKED ISSUE #5683: bug(windows): keyboard with `&name` does not appear in keyboard picker
**Describe the bug**
A .kmx without &name does not install properly in Windows (should fallback to filename). This did work in earlier versions of Keyman; think this regression arose in 14.0.

**To Reproduce**

Create a keyboard without a `&name` system store and install it. It will appear in Keyman Configuration but not in the keyboard picker.

---

 - OS: Windows
 - Keyman version: 14.0

mcdurdin:
Fixes #5683.

Cherry-pick of #5684.

If `store(&name)` is missing from a keyboard, then fallback to the filename of the keyboard (sans extension). This fixes a regression in 14.0 keyboard registration.

I opted not to make this change in kmxfile but rather in places which use it in the Keyman Engine COM API, because I wanted kmxfile to remain truthful about all details of what it is reading from the file.

Given that the COM API will now never return an empty string for the name of the keyboard, there are  mitigations for this in Keyman Configuration which are unnecessary, but it is not harmful for them to remain there.

# User Testing

(Same test as with Keyman 15.0 in #5684)

TEST_NO_NAME: Verify that keyboards with no `&name` store work correctly.

1. Create a keyboard that has no `store(&name)` header. Install it.
2. The keyboard should be visible in Keyman Configuration, and also in the keyboard picker.

### DIFF
--- a/windows/src/engine/kmcomapi/com/keyboards/keymankeyboardfile.pas
+++ b/windows/src/engine/kmcomapi/com/keyboards/keymankeyboardfile.pas
@@ -195,7 +195,9 @@ end;
 
 function TKeymanKeyboardFile.Get_Name: WideString;
 begin
-  Result := FKeyboardInfo.KeyboardName;
+  if FKeyboardInfo.KeyboardName = ''
+    then Result := Get_ID
+    else Result := FKeyboardInfo.KeyboardName;
 end;
 
 function TKeymanKeyboardFile.Get_Version: WideString;
--- a/windows/src/engine/kmcomapi/processes/keyboard/kpinstallkeyboard.pas
+++ b/windows/src/engine/kmcomapi/processes/keyboard/kpinstallkeyboard.pas
@@ -316,7 +316,7 @@ var
   FLanguages: TArray<Integer>;
   FLanguageInstalled: Boolean;
   ki: TKeyboardInfo;
-  kbdname: string;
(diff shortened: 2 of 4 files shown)


darcywong00 (approved):
lgtm

MakaraSok:
* **TEST_NO_NAME: FAILED** The installed keyboards without header `store(&NAME)` don't show up in the keyboard picker, yet they are all seen in the Configuration. Not only that, even the ones with the header line won't be found in the picker, i.e. sil_ipa keyboard.

mcdurdin:
> * Not only that, even the ones with the header line won't be found in the picker, i.e. sil_ipa keyboard.

This sounds like there was something else wrong with the installation. Can you make sure you've uninstalled any possible Keyman apps (Keyman and Keyman Developer) from the test machine, rebooted, and re-test?

@keymanapp-test-bot retest

MakaraSok:
* **TEST_NO_NAME: PASSED** but there is an issue icon

     The Naro keyboard is shown wit a blue N icon in the keyboard picker, but it's blank in the Configuration.
     

     The `sił_ipa` keyboard does show the icon consistently in both places.

keyman-server:

PR #5689, opened 2021-09-14
fix(web): Add checks for empty rows in OSK

darcywong00:
Addresses KeymanWeb portion of #5327

If a touch keyboard has an empty row, several of the `keys` calculations generate exceptions.

## User Testing
Please wait for review comments before testing
* Use emulator for Android 5.0 (SDK 21) which exercises the polyfill code.

* **TEST_EMPTY_ROW** Verify that an OSK with an empty row displays
1. Install the PR build
2. Download and install [afghan_turkmen.kmp](https://darcywong00.github.io/examples/afghan/afghan_turkmen.kmp) which has an empty top row in the OSK
3. When the keyboard is installed, verify the OSK displays fine with an empty top row. There's no associated dictionary, so the appearance will look like a broken suggestion banner.

Screenshot_1631609760

### DIFF
--- a/common/core/web/keyboard-processor/src/keyboards/activeLayout.ts
+++ b/common/core/web/keyboard-processor/src/keyboards/activeLayout.ts
@@ -49,7 +49,7 @@ namespace com.keyman.keyboards {
      * Matches the key code as set within Keyman Developer for the layout.
      * For example, K_R or U_0020.  Denotes either physical keys or virtual keys with custom output,
      * with no additional metadata like layer or active modifiers.
-     * 
+     *
      * Is used to determine the keycode for input events, rule-matching, and keystroke processing.
      */
     public get baseKeyID(): string {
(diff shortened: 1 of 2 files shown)


jahorton:
To be clear - the top row is _intentionally_ empty, no keys?

jahorton:
If not intentionally empty, that should probably be addressed in this block:

https://github.com/keymanapp/keyman/blob/85aa4190c0c176f095fbab58336e703c38b9714b/common/core/web/keyboard-processor/src/keyboards/activeLayout.ts#L638-L665

mcdurdin:
> This block currently assumes any erroneous empty rows will be appended, rather than prepended as seen here.

It would be possible for empty rows to be anywhere in the rows array, technically.

darcywong00:
> To be clear - the top row is _intentionally_ empty, no keys?

This keyboard came from the community site. Likely, the top row was **not** intentionally empty, since the desktop keyboard looks fine:

MakaraSok:
Is this ready for testing now?

darcywong00:
> Is this ready for testing now?

No, I'll put this back to "draft" while I'm still reworking it.
Please wait for @jahorton 's review and approval before testing...

darcywong00:
Abandoning for #5699

PR #5692, opened 2021-09-14
fix(common/core): incxstr checks all character while incrementing

### LINKED ISSUE #4604: bug(windows): incxstr needs extra robustness for malformed strings
**Describe the bug**
In the incxstr function, there are a number of places where we don't check a character but jump over it, assuming it is valid (`p += 2` kinda thing). We need to check every character as we increment so we don't miss a U+0000 end of string with malformed data. 

Discovered while working on #4596.

---

 - Keyman version: 14.0.252-beta

SabineSIL:
Fixes #4604.

In a string some characters after UC_SENTINEL (FFFF) had not been checked while incrementing due to a p+=2 movement of the pointer. If a \0 had been located between UC_SENTINEL (FFFF)  and another control character the \0 han not been detected in some cases which then resulted in a buffer overrun,

### DIFF
--- a/common/core/desktop/src/kmx/kmx_xstring.cpp
+++ b/common/core/desktop/src/kmx/kmx_xstring.cpp
@@ -119,33 +119,55 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
 
 PKMX_WCHAR km::kbp::kmx::incxstr(PKMX_WCHAR p)
 {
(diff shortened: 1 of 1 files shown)


keyman-server:

mcdurdin:
Thank you for this PR! Would you be able to include the unit test that you wrote as well please?

mcdurdin:
@keymanapp-test-bot skip 

I don't think we need user testing for this -- unit tests more appropriate.

mcdurdin:
Replaced by #5712

PR #5696, opened 2021-09-15
feat(developer): track changes to option values better in debugger 🐞

### LINKED ISSUE #5695: feat(developer): update the option value at time that rule is processed when in single step debugging
Currently, the debugger updates the displayed values of option stores at the start of a batch instead of actually when the value is set by a `set()` statement in a rule. We should refactor this so that the displayed value is updated at the correct time

mcdurdin:
Fixes #5695.

When the debugger is in single-step mode, reflects the value of option stores at the time of the change rather than at the beginning of the batch.

# User Testing

* TEST_SET_OPTION: verify that option is set at time of rule match

With the following keyboard or similar, start debugging, set the debugger into single step mode (`>>` button),
and then watch the Options tab in the Debug Status pane to verify that the option value for `opt` is changed
from `1` to `0` when the single step (`>` button) steps over the `+ 'a'` rule (there will be two steps on the
rule: the first is arriving at the rule, the second is processing the rule -- it should happen on the second).

```
store(&VERSION) '10.0'
begin Unicode > use(main)
store(opt) '1'
group(main) using keys

+ 'a' > set(opt = '0')
```


windows/src/developer/TIKE/Tike.dpr
```diff
@@ -15,7 +15,6 @@ uses
   VersionInfo in '..\..\global\delphi\general\VersionInfo.pas',
   compile in '..\..\global\delphi\general\compile.pas',
   KeymanDeveloperOptions in 'main\KeymanDeveloperOptions.pas',
-  int_kmdebug in '..\..\global\delphi\general\int_kmdebug.pas',
```
mcdurdin, line 18: Discovered that this file was completely unused so removed it to reduce confusion...

MakaraSok:
* **TEST_SET_OPTION: PASSED** The `opt` does change from `1` to `0` on the second step where the rule was being processed.

darcywong00 (approved):
lgtm

keyman-server:

PR #5699, opened 2021-09-15
fix(common/core/web): Remove empty rows in OSK

darcywong00:
Replaces #5689 in addressing the KeymanWeb portion of #5327

If a touch keyboard had empty rows (compiled with older versions of Keyman Developer), the OSK would generate exceptions.
This cleans up how `ActiveLayout.correctLayerEmptyRowBug()` removes the empty rows.

For testing, I checked in the afghan_turkmen keyboard as included on the issue. I also created a custom `empty_row` keyboard in Developer 13 and edited the js file so there's an empty row in each of the 3 layers (different row ID).

## User Testing

* **TEST_ANDROID** Verifies OSK handles empty rows in the embedded app
Use an Android 5.0 emulator so the polyfill code is exercised.
1. Load the PR build of Keyman for Android
2. Test several of the default sil_euro_latin keys and verify they function correctly
3. Download and install [afghan_turkmen.kmp](https://darcywong00.github.io/examples/afghan/afghan_turkmen.kmp) which has an empty top row defined on the default layer
4. After the keyboard finished installing and switches, verify the afghan_turkmen OSK is fine (no empty rows)
5. Verify the default keys work. Note: since the attached keyboard came from the community site, it has an issue where the shift button doesn't switch layers

* **TEST_WEB** Verifies OSK 
Use the CI web test page artifacts using the new page `testing/empty-row/` in mobile viewport
1. Load the test page
2. With the empty_row keyboard, verify the default layer on the OSK is fine and keys function correctly
3. Use the shift to switch to the shift layer
4. Verify the shift layer OSK is fine and keys function correctly
5. Use the numeric key to switch to the numeric layer
6. Verify the numeric layer OSK is fine and keys function correctly
7. Use the globe key to switch to the afghan_turkmen keyboard
8. Verify the default layer on the afghan_turkmen OSK is fine (no empty rows)
9. Verify the keys work (except for the shift button, see the mentioned limitation in TEST_ANDROID)


web/testing/empty-row/empty_row-1.0.js
```diff
@@ -0,0 +1,521 @@
...
+    this.KMBM = 0x0000;
+    this.KVKL = {
+        "phone": {
+            "font": "Tahoma",
+            "displayUnderlying": false,
+            "layer": [{
+                "id": "default",
+                "row": [{
+                    "id": "1"
+                }, {
```
darcywong00, line 21: row 1 is a blank row on the default layer.
Also ll. 214, (row 2 on shift layer) and
ll. 423 (row 3 on numeric layer)

common/core/web/keyboard-processor/src/keyboards/activeLayout.ts
```diff
@@ -675,28 +671,11 @@ namespace com.keyman.keyboards {
...
-        for(i=rows.length; i>0; i--) {
-          if(rows[i-1]['key'].length > 0) {
-            break;
-          }
-        }
-
-        if(i < rows.length) {
-          rows.splice(i-rows.length,rows.length-i);
-        }
-      }
```
darcywong00, line 698: This section should already be handled in correctLayerEmptyRowBug

common/core/web/keyboard-processor/src/keyboards/activeLayout.ts
```diff
@@ -652,15 +652,11 @@ namespace com.keyman.keyboards {
         let layer=layers[n];
         let rows=layer['row'];
         let i: number;
-        for(i=rows.length; i>0; i--) {
-          if(rows[i-1]['key'].length > 0) {
-            break;
+        for(i=rows.length-1; i>=0; i--) {
+          if(!Array.isArray(rows[i]['key'])) {
+            rows.splice(i, 1)
```
darcywong00, line 657: The previous implementation wasn't removing an empty row if it was the top one (rows[0])

jahorton: Should also include the `.length` check, as `Array.isArray([]) == true`.  What you put here does catch `undefined` cases (which was missing before), but not empty-array cases (which were caught before).

web/testing/empty-row/empty_row-1.0.js
```diff
@@ -0,0 +1,521 @@
+KeymanWeb.KR(new Keyboard_empty_row());
+
+function Keyboard_empty_row() {
+    this.KI = "Keyboard_empty_row";
+    this.KN = "empty row";
```
jahorton, line 5: This case is more of a `Keyboard_undefined_row` - the keys array is undefined.

It may be wise to have both this case and an `Keyboard_empty_row` - where the top row's `.keys` entry is defined, but just as `[]`.

jahorton (changes requested):

web/testing/empty-row/empty_row-1.0.js
darcywong00: Per @jahorton , I've changed row 1 to to test array length == 0
```
"row": [{
  "id": "1",
  "key": []
}, {
...
```

common/core/web/keyboard-processor/src/keyboards/activeLayout.ts
darcywong00: Gotcha. I've updated to include both checks

web/testing/empty-row/empty_row-1.0.js
darcywong00: Tweaked row 1 (see above)

jahorton: Right, I suppose the `afghan` test-resource does test the `undefined` case well enough.  :+1:

jahorton (approved):
LGTM pending user testing.

darcywong00:
If we 🍒 -pick this to stable-14.0, should also get ed490fcbeffa12ecbd9dc7fdd7a42e35e5d6c720

MakaraSok:
* **TEST_ANDROID: PASSED** The default keyboard works as expected and the `afghan_turkmen` works as indicated.
    


* **TEST_WEB: PASSED** Tested on Chrome emulator (android) on Windows and everything works as indicated in the instructions.

keyman-server:

PR #5649, opened 2021-08-27
fix(developer): include Keyman Core in Keyman Developer 🐞

mcdurdin:
Fixes #5648.

Includes kmnkbp0-0.dll in the installer for Keyman Developer.

# User Testing

* TEST_BASIC: Load a keyboard in Keyman Developer, and press F5 to start debugging. It should start!

PR #5711, opened 2021-09-17
fix(developer): add version info to Core 🐞

### LINKED ISSUE #5672: bug(windows): kmnkbp0-0.dll does not include version information
**Describe the bug**

kmnkbp0-0.dll needs to include a version information resource. This could be added by the Developer build rather than by the Keyman Core build, so that we can use the version information generation infrastructure included in the windows/ Makefiles.

sigcheck verify.dpr (i3633) is temporarily skipping this file in #5649 so we can continue testing and validation. Once we have version information, we should remove this patch from verify.dpr.

---

**Keyman for Windows:**
 - Keyman version: 15.0.106

mcdurdin:
Fixes #5672.

* Adds version information to meson build for Keyman Core DLL
* Removes verify.dpr patch added in #5649 so it will be tested at build

@keymanapp-test-bot skip

### DIFF
--- a/common/core/desktop/src/version.rc
+++ b/common/core/desktop/src/version.rc
@@ -1,0 +1,36 @@
+#include <windows.h>
+#include "include/keyman/keyboardprocessor_version.h"
+
+// Encoding: UTF8 (compile with rc /c65001)
+// Use rc /n to append null to all strings
+
+1 VERSIONINFO
+  FILEVERSION		VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, 0
+  PRODUCTVERSION	VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, 0
+  FILEFLAGSMASK		0x3fL
+  FILEFLAGS		0x0L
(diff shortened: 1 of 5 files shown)


darcywong00:
Shouldn't the PR scope be common/core/desktop and possible windows (test) ?

common/core/desktop/src/version.rc
```diff
@@ -0,0 +1,36 @@
+#include <windows.h>
```
darcywong00, line 1: I'm take it Linux doesn't build this since the Jenkins build passed.

mcdurdin:
> Shouldn't the PR scope be common/core/desktop and possible windows (test) ?

The Core build artifact that this impacts is only used in Developer (kmnkbp0-0.dll) -- Windows will static link this.

User test not really required; the verify.dpr change means that the version resource will be tested as a unit test.

> I'm take it Linux doesn't build this since the Jenkins build passed.

Correct.

darcywong00 (approved):
lgtm

keyman-server:

PR #5712, opened 2021-09-17
fix(common/core): incxstr checks all character while incrementing 

### LINKED ISSUE #4604: bug(windows): incxstr needs extra robustness for malformed strings

SabineSIL:
Fixes #4604.

In a string some characters after UC_SENTINEL (FFFF) had not been checked while incrementing due to a p+=2 movement of the pointer. If a \0 had been located between UC_SENTINEL (FFFF) and another control character the \0 han not been detected in some cases which then resulted in a buffer overrun.


keyman-server:

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
-    case CODE_IFSYSTEMSTORE: return p+3;
-    case CODE_SETOPT:   return p+2;
-    case CODE_SETSYSTEMSTORE: return p+2;
-    case CODE_RESETOPT: return p+1;
-    case CODE_SAVEOPT:  return p+1;
-    default:        return p;
+    // UC_SENTINEL(FFFF) with UC_SENTINEL_EXTENDEDEND(0x10) == variable length
+    if (*(p + 1) == CODE_EXTENDED) {
+      p += 2;
+      while (*(p - 1) && *p && *p != UC_SENTINEL_EXTENDEDEND)
```
mcdurdin, line 135:
```suggestion
      while (*p && *p != UC_SENTINEL_EXTENDEDEND)
```

We've just tested p-1 (was p+1) on line 133. We don't need to test it again.

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
+        p++;
+
+      //if (!(*(p))) return p;
+      if (*p == 0) return p;
+      if (*p == UC_SENTINEL_EXTENDEDEND)		return p + 1;
```
mcdurdin, line 140: Once we get here, *p MUST be UC_SENTINEL_EXTENDEDEND, so:
```suggestion
      return p + 1;
```

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
-    case CODE_RESETOPT: return p+1;
-    case CODE_SAVEOPT:  return p+1;
-    default:        return p;
+    // UC_SENTINEL(FFFF) with UC_SENTINEL_EXTENDEDEND(0x10) == variable length
+    if (*(p + 1) == CODE_EXTENDED) {
+      p += 2;
+      while (*(p - 1) && *p && *p != UC_SENTINEL_EXTENDEDEND)
+        p++;
+
+      //if (!(*(p))) return p;
```
mcdurdin, line 138:
```suggestion
```

Don't leave commented code in!

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
+    case CODE_CLEARCONTEXT:		deltaptr = 3; break;
+    case CODE_CALL:						deltaptr = 3; break;
+    case CODE_CONTEXTEX:			deltaptr = 3; break;
+    case CODE_IFOPT:          deltaptr = 5; break;
+    case CODE_IFSYSTEMSTORE:	deltaptr = 5; break;
+    case CODE_SETOPT:					deltaptr = 4; break;
+    case CODE_SETSYSTEMSTORE:	deltaptr = 4; break;
+    case CODE_RESETOPT:				deltaptr = 3; break;
+    case CODE_SAVEOPT:				deltaptr = 3; break;
+    default:									deltaptr = 2;
```
mcdurdin, line 160: For safety, I think we should check *every* CODE_ that is defined, and default should then have deltaptr = 1.

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
+    }
+
+    // check for \0 between FFFF and next printable character
+    for (int i = 0; i < (deltaptr); i++) {
+      if (*p==0)
+        return p;
+      p += 1;    }
```
mcdurdin, line 167:
```suggestion
    for (int i = 0; i < deltaptr; i++) {
      if (*p==0) {
        return p;
      }
      p++;
    }
```

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
   {
-    if(*p >= 0xD800 && *p <= 0xDBFF && *(p+1) >= 0xDC00 && *(p+1) <= 0xDFFF) return p+2;
-    return p+1;
+    if (*p >= 0xD800 && *p <= 0xDBFF && *(p + 1) >= 0xDC00 && *(p + 1) <= 0xDFFF) return p + 2;
+    return p + 1;
   }
-
-  p+=2;
-  switch(*(p-1))
+  else
```
mcdurdin, line 130: The `else {}` is not needed here because the previous `if` block always `return`s. So we can move the code contained in the else block out to the parent level.

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -119,33 +119,57 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
+    case CODE_DEADKEY:				deltaptr = 3; break;
+    case CODE_CLEARCONTEXT:		deltaptr = 3; break;
+    case CODE_CALL:						deltaptr = 3; break;
+    case CODE_CONTEXTEX:			deltaptr = 3; break;
+    case CODE_IFOPT:          deltaptr = 5; break;
+    case CODE_IFSYSTEMSTORE:	deltaptr = 5; break;
+    case CODE_SETOPT:					deltaptr = 4; break;
+    case CODE_SETSYSTEMSTORE:	deltaptr = 4; break;
+    case CODE_RESETOPT:				deltaptr = 3; break;
+    case CODE_SAVEOPT:				deltaptr = 3; break;
```
mcdurdin, line 159: I think we should define the delta size for each of those `CODE_` values in the header file, e.g.

```
#define DELTA_CODE_ANY 3
```

and then use those delta values here.

Even better: we could have a const array for the delta sizes and avoid the switch statement altogether (just test bounds).

mcdurdin (changes requested):
This is looking good so far -- thank you for your contribution. I've got a lot of suggestions but don't be scared 😱 by those; it's all part of the process 😄!

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: HI Marc I changed incxstr. default for deltaptr must is set to 2 since we didn`t move before checking *(p+1)

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -116,37 +116,43 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
-    case CODE_RESETOPT: return p+1;
-    case CODE_SAVEOPT:  return p+1;
-    default:        return p;
+  // CODE_PTR defined in kmx_processevent.h
+  for (int i = 0; i < size(CODE_PTR[0]); i++) {
+    if (*(p + 1) == CODE_PTR[0][i]) {
+      deltaptr = CODE_PTR[1][i];
+      break;
+    }
   }
```
mcdurdin, line 144: I think we should be able to do this without needing a loop. If we define a `CODE__SIZE` array in kmx_file.h, e.g.:

```
/// number of parameters for each CODE_### control code
extern const int CODE__SIZE[];
```

and then in kmx_file.cpp (copying from the #defines for CODE_### in kmx_file.h):

```
#include <assert.h>
...

/// number of parameters for each CODE_### control code
const int CODE__SIZE[] = {
  -1, // undefined       0x00
  1,  // CODE_ANY        0x01
  2,  // CODE_INDEX      0x02
  0,  // CODE_CONTEXT    0x03
...
  // ... use 0 for undefined CODE_ values, e.g. 0x09
};

// Ensure that all CODE_### sizes are defined
static_assert(sizeof(CODE__SIZE)/sizeof(CODE_SIZE_[0]) == CODE_LASTCODE + 1);
```

then we can do, in `incxstr`: 

```
if(*p > CODE_LASTCODE || CODE__SIZE[*p] == -1) {
  // Invalid CODE_ value
  return *p;
}

int deltaptr = CODE__SIZE[*p];
p++;
//...
```

common/core/desktop/src/kmx/kmx_processevent.h
```diff
@@ -138,7 +138,15 @@ const char *Debug_VirtualKey(KMX_WORD vk);
...
-
+const int CODE_PTR[2][14] = {
+      {CODE_ANY, CODE_NOTANY, CODE_INDEX, CODE_USE, CODE_DEADKEY, CODE_CLEARCONTEXT, CODE_CALL, CODE_CONTEXTEX, CODE_IFOPT,
+       CODE_IFSYSTEMSTORE, CODE_SETOPT, CODE_SETSYSTEMSTORE, CODE_RESETOPT, CODE_SAVEOPT},
+      {3, 3, 4, 3, 3, 3, 3, 3, 5, 5, 4, 4, 3, 3}};
+
+const int CODE_PTR[2][14] = {
+      {CODE_ANY, CODE_NOTANY, CODE_INDEX, CODE_USE, CODE_DEADKEY, CODE_CLEARCONTEXT, CODE_CALL, CODE_CONTEXTEX, CODE_IFOPT,
+       CODE_IFSYSTEMSTORE, CODE_SETOPT, CODE_SETSYSTEMSTORE, CODE_RESETOPT, CODE_SAVEOPT},
+      {3, 3, 4, 3, 3, 3, 3, 3, 5, 5, 4, 4, 3, 3}};
```
mcdurdin, line 149: See my commentary on your change in `incxstr`; this data belongs in kmx_file.h / kmx_file.cpp 😁 -- but probably in the format I suggested?

mcdurdin (changes requested):
This is looking good; I see that the process is coming together. I have made an additional suggestion on how we can eliminate one of the loops.

(For performance reasons we may decide to eliminate all the loops in a future pass, but let's go for correctness first!)

mcdurdin:
Build failed:

```
[15:49:49][Step 1/1] In file included from ../../../src/kmx/kmx_processor.hpp:13:0,
[15:49:49][Step 1/1]                  from ../../../src/km_kbp_keyboard_api.cpp:16:
[15:49:49][Step 1/1] ../../../src/kmx/kmx_processevent.h:146:25: error: redefinition of ‘const int km::kbp::kmx::CODE_PTR [2][14]’
[15:49:49][Step 1/1]  const int CODE_PTR[2][14] = {
[15:49:49][Step 1/1]                          ^
[15:49:49][Step 1/1] ../../../src/kmx/kmx_processevent.h:141:11: note: ‘const int km::kbp::kmx::CODE_PTR [2][14]’ previously defined here
[15:49:49][Step 1/1]  const int CODE_PTR[2][14] = {
[15:49:49][Step 1/1]            ^~~~~~~~
```

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,255 @@
...
+  //std::string path(argv[1]);
+
+  if(argc < 2) {
+   return error_args();
+ }
+
+  auto arg_color = std::string(argv[1]) == "--color";
+  if(arg_color && argc < 3) {
+    return error_args();
+  }
```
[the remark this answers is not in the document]
SabineSIL: this was changed last week also

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: this is included in todays commit

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: this was changed in the version I committed today

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: ... removed :-)

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: changed :-)

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: this change is made

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: OK

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,255 @@
...
+  console_color::enabled = console_color::isaterminal() || arg_color;
+
+  //arg_path = argv[arg_color ? 2 : 1];
```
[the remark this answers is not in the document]
SabineSIL: ...removed

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,255 @@
...
+constexpr const auto help_str = "\
+debug_api [--color] <SOURCE_PATH>|--print-sizeof\n\
+\n\
+  --color         Force color output\n\
+  --print-sizeof  Emit structure sizes for interop debug\n\
+  SOURCE_PATH     Path where debug_api.cpp is found; kmx files are\n\
+                  located relative to this path.\n";
+
+int error_args() {
+  std::cerr << "debug_api: Invalid arguments." << std::endl;
```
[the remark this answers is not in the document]
SabineSIL: ... is removed

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+}
+
+constexpr const auto help_str = "\
+debug_api [--color] <SOURCE_PATH>|--print-sizeof\n\
+\n\
+  --color         Force color output\n\
+  --print-sizeof  Emit structure sizes for interop debug\n\
+  SOURCE_PATH     Path where debug_api.cpp is found; kmx files are\n\
+                  located relative to this path.\n";
```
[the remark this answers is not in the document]
SabineSIL: done

common/core/desktop/tests/unit/kmnkbd/meson.build
```diff
@@ -2,18 +2,13 @@
...
-# History:      19  Oct 2018 - TSE - Added test for context API functions.
 #
 
 defns=['-DKMN_KBP_STATIC']
 tests = [
-  ['context-api', 'context_api.cpp'],
-  ['keyboard-api', 'keyboard_api.cpp'],
-#  ['options-api', 'options_api.cpp'],
-  ['state-api', 'state_api.cpp'],
-  ['debug-api', 'debug_api.cpp'],
```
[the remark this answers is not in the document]
SabineSIL: ... put them back in

common/core/desktop/src/kmx/kmx_processevent.h
```diff
@@ -138,7 +138,6 @@ const char *Debug_VirtualKey(KMX_WORD vk);
 const char *Debug_UnicodeString(PKMX_WCHAR s, int x = 0);
 const char *Debug_UnicodeString(std::u16string s, int x = 0);
 const char *Debug_ModifierName(KMX_UINT modifiers);
-
```
mcdurdin, line 141: Please restore this file as it is only whitespace change.

common/core/desktop/src/kmx/kmx_file.h
```diff
@@ -219,6 +219,8 @@ namespace kmx {
 #define K_MODIFIERFLAG  0x007F
 #define K_NOTMODIFIERFLAG 0xFF00   // I4548
 
+
+
```
mcdurdin, line 223:
```suggestion
```

Remove these additional lines.

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -3,14 +3,15 @@
   Authors:          mcdurdin
 */
 #include <vector>
+#include <array>
 #include <iterator>
 #include <codecvt>
 #include <locale>
 #include "kmx_processevent.h"
 #include "utfcodec.hpp"
+//#include <assert.h>
```
mcdurdin, line 12: Don't leave in commented code!

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -3,14 +3,15 @@
...
 
-
-
+using namespace std;
```
mcdurdin, line 14: We shouldn't need to use `std`.

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -116,37 +117,37 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
+PKMX_WCHAR km::kbp::kmx::incxstr(PKMX_WCHAR p) {
 
-PKMX_WCHAR km::kbp::kmx::incxstr(PKMX_WCHAR p)
-{
-  if(*p == 0) return p;
-  if(*p != UC_SENTINEL)
-  {
-    if(*p >= 0xD800 && *p <= 0xDBFF && *(p+1) >= 0xDC00 && *(p+1) <= 0xDFFF) return p+2;
-    return p+1;
+PKMX_WCHAR km::kbp::kmx::incxstr(PKMX_WCHAR p) {
```
mcdurdin, line 122: This code doesn't compile -- can you make sure it compiles and then I will review?

mcdurdin (changes requested):
I'll have more to review later -- but note that the code doesn't currently compile.

Please can you restore the .kmx files you removed (e.g.
`git checkout master "common/core/desktop/tests/unit/kmx/000 - null keyboard.kmx"`)

common/core/desktop/src/kmx/kmx_processevent.h
SabineSIL: removed lines

common/core/desktop/src/kmx/kmx_processevent.h
SabineSIL: OK

common/core/desktop/src/kmx/kmx_file.h
SabineSIL: ...removed...

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: changed this file, included your proposals

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,230 @@
+/*
+ * Keyman is copyright (C) SIL International. MIT License.
+ *
+ * Keyman Core - Debugger API unit tests
```
mcdurdin, line 4:
```suggestion
 * Keyman Core - KMX Extended String unit tests
```

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    -1,  // undefined                0x00
+    1,   // CODE_ANY                 0x01
+    2,   // CODE_INDEX               0x02
+    0,   // CODE_CONTEXT             0x03
+    0,   // CODE_NUL                 0x04
+    1,   // CODE_USE                 0x05
+    0,   // CODE_RETURN              0x06
+    0,   // CODE_BEEP                0x07
+    1,   // CODE_DEADKEY             0x08
+    0,   // CODE_BACKSPACE           0x09
```
mcdurdin, line 32:
```suggestion
    -1,  // unused                   0x09
```

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    2,   // CODE_EXTENDED            0x0A
+    0,   // CODE_EXTENDEDEND         0x0B
```
mcdurdin, line 34:
```suggestion
    -1,  // CODE_EXTENDEDEND         0x0B (unused)
```

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    0,   // CODE_SWITCH              0x0C
```
mcdurdin, line 35:
```suggestion
    1,   // CODE_SWITCH              0x0C
```

Per https://github.com/keymanapp/keyman/blob/d1dd8b022e7323dc44339d52aabb5380299429e6/windows/src/developer/kmcmpdll/Compiler.cpp#L2383-L2386

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    0,   // CODE_KEY                 0x0D
+    1,   // CODE_CLEARCONTEXT        0x0E
```
mcdurdin, line 37:
```suggestion
    0,   // CODE_CLEARCONTEXT        0x0E
```

Per https://github.com/keymanapp/keyman/blob/d1dd8b022e7323dc44339d52aabb5380299429e6/windows/src/developer/kmcmpdll/Compiler.cpp#L2100-L2102

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    0,   // CODE_NUL                 0x04
+    1,   // CODE_USE                 0x05
+    0,   // CODE_RETURN              0x06
+    0,   // CODE_BEEP                0x07
+    1,   // CODE_DEADKEY             0x08
+    0,   // CODE_BACKSPACE           0x09
+    2,   // CODE_EXTENDED            0x0A
+    0,   // CODE_EXTENDEDEND         0x0B
+    0,   // CODE_SWITCH              0x0C
+    0,   // CODE_KEY                 0x0D
```
mcdurdin, line 36:
```suggestion
    -1,  // CODE_KEY                 0x0D (never used)
```

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    1,   // CODE_CLEARCONTEXT        0x0E
+    1,   // CODE_CALL                0x0F
+    0,   // UC_SENTINEL_EXTENDEDEND  0x10
```
mcdurdin, line 39:
```suggestion
    -1,  // UC_SENTINEL_EXTENDEDEND  0x10 (not valid with UC_SENTINEL)
```

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+    1,   // CODE_CONTEXTEX           0x11
+    1,   // CODE_NOTANY              0x12
+    2,   // CODE_KEYMAN70_LASTCODE   0x13
+    3,   // CODE_IFOPT               0x14
+    1,   // CODE_SAVEOPT             0x15
+    1,   // CODE_RESETOPT            0x16
+    3,   // CODE_IFSYSTEMSTORE       0x17
+    2    // CODE_LASTCODE            0x18
```
mcdurdin, line 47:
```suggestion
    2,   // CODE_SETOPT              0x13
    3,   // CODE_IFOPT               0x14
    1,   // CODE_SAVEOPT             0x15
    1,   // CODE_RESETOPT            0x16
    3,   // CODE_IFSYSTEMSTORE       0x17
    2    // CODE_SETSYSTEMSTORE      0x18
```

common/core/desktop/src/kmx/kmx_file.cpp
```diff
@@ -18,6 +19,37 @@ KMX_BOOL KMX_ProcessEvent::Load(km_kbp_path_name KeyboardName)
...
+};
+
+// Ensure that all CODE_### sizes are defined
+static_assert(sizeof(CODE__SIZE) / sizeof(CODE__SIZE[0]) == (CODE_LASTCODE + 1), "Size of array CODE_SIZE not correct");
```
mcdurdin, line 51:
```suggestion
static_assert(sizeof(CODE__SIZE) / sizeof(CODE__SIZE[0]) == (CODE_LASTCODE + 1), "Size of array CODE__SIZE not correct");
```

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,230 @@
...
+  // --- Test for FFFF +CODE_SAVEOPT -----------------------------------------------------------------------------------------------------------
+  p = (PKMX_WCHAR)u"\U0000FFFF\U00000015\U00000001";
+  q = incxstr(p);
+  assert(q == p + 3 );
+
+  // --- Test for FFFF +default ----------------------------------------------------------------------------------------------------------------
+  p = (PKMX_WCHAR)u"\U0000FFFF\U00000004";
+  q = incxstr(p);
+  assert(q == p + 2);
+
```
mcdurdin, line 126: These tests are all good. It is important that we add some more normal case tests for longer strings, so we can be certain that `incxstr` is not simply terminating at the final nul, e.g.

```
p = (KMX_WCHAR)u"\U0000FFFF\U00000004\U00001234\U00002468";
q = incxstr(p);
assert(q == p + 2);
```

It might be good to duplicate all the tests to do this. Or you may want to create a helper function to simplify your work:

```
  test_incxstr_offset(u"\U0000FFFF\U00000004", 2);
```

```
void test_incxstr_offset(PKMX_WCHAR p, int offset) {
  PKMX_WCHAR q = incxstr(p);
  assert(q == p+ offset);
  //TODO: allocate buffer to copy p into a new string with extra chars \U00001234\U00002468 
  //on the end and then test that string also, then free buffer.
  //If you do go with a helper function, then it would be good to add some other small tests
  //here, e.g. do the same thing with a UC_SENTINEL or a surrogate pair,
}
```

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -116,36 +115,37 @@ km_kbp_cp *km::kbp::kmx::u16tok(km_kbp_cp *p, km_kbp_cp ch, km_kbp_cp **ctx) {
...
+
+  if (*(p + 1) > CODE_LASTCODE || CODE__SIZE[*(p + 1)] == -1) {
+    return p + 1;
   }
+
+  int deltaptr = 2 + CODE__SIZE[*(p + 1)];
+
+  // check for \0 between FFFF and next printable character
+  for (int i = 0; i < deltaptr; i++) {
+    if (*p == 0)       return p;
```
mcdurdin, line 145:
```suggestion
    if (*p == 0)
      return p;
```

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -3,14 +3,13 @@
   Authors:          mcdurdin
 */
 #include <vector>
+#include <array>
```
mcdurdin, line 6: Not sure if we need this include now?

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -0,0 +1,230 @@
...
+int error_args() {
+  std::cerr << "test_kmx_xstring: Invalid arguments." << std::endl;
+  std::cout << help_str;
+  return 1;
+}
+
+int main(int argc, char *argv []) {
+  // TODO: choose/create a keyboard which has some rules
+  // Global setup
+  //std::string path(argv[1]);
```
mcdurdin, line 220:
```suggestion
```

mcdurdin (changes requested):
I think this is getting pretty close. I went through and verified the delta sizes by comparing to compiler.cpp. You'll see in a couple of places I came up with different values than you found. Can you verify each of the values once more, as this is pretty critical!

I also denoted unused values as -1. The unused values cannot be generated by the compiler, at least for the last 15 years or so!

The tests look really good. I realised we have one gap: we test that `incxstr` doesn't skip past `\0` but we never actually test that it finishes at the right spot in a longer string! I mean, from my reading of the code it does work correctly, but we should be testing that.

A few very minor suggestions otherwise!

Good work 😀

SabineSIL:
Thank you Marc for your reply. I will address those as soon as possible.

 

Unfortunately I am not able to make the meeting today as I came down with a cold.

I hope to make it on Monday.

 

Have a good meeting

 

Sabine

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
SabineSIL: I wrote a helper function but then realized that special care would have been necessary to ensure that the concatenated sequences would be stored in the buffer in a way incxstr()could be tested. The helper function worked OK for most cases but with \0 it wasn`t behaving correctly:
When a \0 is copied to the buffer with wcscpy()  and then other characters like "\U00001234\U00002468" are concatenated using  wcscat() the \0 in the buffer would be removed and only those 2 characters would be placed in the buffer. This makes incxstr() move the pointer forward instead of stopping at (the lost) \0. Therefore, I decided to add more tests without using a helper function.

common/core/desktop/src/kmx/kmx_file.cpp
SabineSIL: compared it with CODE_SETSYSTEMSTORE+1 instead of CODE_LASTCODE +1

common/core/desktop/src/kmx/kmx_file.cpp
SabineSIL: also changed

common/core/desktop/src/kmx/kmx_file.cpp
SabineSIL: changed

mcdurdin:
@ermshiperete @rc-swag do you want to take a read through the final code changes also before we merge?

PR #5719, opened 2021-09-19
chore(common/core): cleanup version headers

### LINKED ISSUE #5450: chore(common/core/desktop): keyboardprocessor.h should not need to be transformed
Currently keyboardprocessor.h is a large file with a .in source that is transformed in order to inject version numbers. It would be better for that version information to be included in a separate header with only those three lines, so that the transformed file can be small:

```c++
#define KM_KBP_LIB_CURRENT  @lib_curr@
#define KM_KBP_LIB_AGE      @lib_age@
#define KM_KBP_LIB_REVISION @lib_rev@
```

Then, include that transformed file in keyboardprocessor.h

mcdurdin:
Fixes #5450.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #5727, opened 2021-09-20
fix(developer): run kmlmc from Keyman source path

mcdurdin:
Makes it possible to run the kmlmc.cmd compiler wrapper script from Keyman Developer while debugging from source.

Adds a new class `TKeymanDeveloperPaths` which is similar to `TKeymanPaths`, to provide a future single location for all Keyman Developer path-related functions. There are a number scattered through the source at present, e.g. in `RedistFiles.pas`, so this is just a starting point.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.117-alpha](https://keyman.com/downloads/releases/alpha/15.0.117)

PR #5717, opened 2021-09-18
fix(android): Mirror arrow icons to support RTL

darcywong00:
This PR creates 4 drawable layouts so that the forward/back navigation arrows and text size up/down arrows will autoMirror for RTL layouts.
* ic_action_back
* ic_action_forward
* ic_action_textsize_up
* ic_action_textsize_down

Base branch set to `chore/common/crowdin-ar` to facilitate testing with the Arabic locale (#5706)

### Sample screenshots when Display Language in Arabic (RTL layout)
language settings (forward arrow pointing left)
Screenshot_1631928418

Text Size menu (increase text size pointing left, decrease text size pointing right)
Screenshot_1631928435

Info activity (order of navigation icons: close, forward, back)
Screenshot_1631928469

## User Testing
Run the PR build on an Android 5.0 emulator

* **TEST_LTR ** - Tests no regressions in English UI layout
1. Load Keyman for Android and use the default English UI(left to right)
2. From the main menu, go to the Text Size menu
3. Use the menu and verify:
    * clicking left arrow decreases the text size
    * clicking right arrow increases the text size
4. From the  main menu, select Info Activity 
5. Browse through the help and verify
    * clicking left arrow will navigate back
    * clicking right arrow will navigate forward (only enabled if there's a forward page to go to)
6. Click "x" to close the Info activity
7. From the Settings menu, install sil_cameroon_qwerty keyboard from keyman.com for any language
8. Verify keyboard installation wizard behaves as expected
9. Click through other "Settings" menus and verify
    * settings toolbar back arrow points to the left (behaves as normal)
    * forward arrows point to the right (behaves as normal)

* **TEST_RTL ** - Test layout of RTL icons
1. Load Keyman for Android
2. From the "Settings" menu, change the Display language to "Arabic" (an RTL layout). Note that forward arrows will point left and back arrows will point right.
3. From the main menu, go to the Text Size menu
4. Use the menu and verify:
    * clicking the left arrow increases the text size 
    * clicking the right arrow decreases the text size
5. Browse through the help and verify
    * clicking left arrow will navigate forward (only enabled if there's a forward page to go to)
    * clicking right arrow will navigate back
6. Click "x" to close the Info activity
7. From the Settings menu, install sil_cameroon_qwerty keyboard from keyman.com for any language
8. Verify keyboard installation wizard behaves as expected. Note the bottom left buttons go "forward" in the wizard, and the "OK" buttons will be on the bottom left.
9. Click through other "Settings" menu and verify
[...]

PR #5706, opened 2021-09-16
chore(common): Check in crowdin strings for Arabic

darcywong00:
This PR checks in the crowdin strings for Arabic, and will exercise how well the Keyman apps handle RTL languages.
* Also update the root crowdin.yml file so Android uses locale `ar` instead of `ar-SA`.
 
TODO:
- [ ] Windows hotkeys that use the `&amp;` strings within RTL strings
- [ ] Run xcode to include iOS strings

Screenshot of Keyman for Android settings menu
Screenshot_1631784571

## User Testing
* **TEST_ARABIC** - Check Keyman apps for Arabic strings
1. Load the PR build of the Keyman app for the various platforms
2. Set the app UI language to "Arabic" (this will be different across the apps)
3. Verify the UI changes to Arabic
4. Make note of any right-to-left issues in rendering / layout

PR #5732, opened 2021-09-21
fix(android): Mirror arrow icons for RTL layouts 🍒 

darcywong00:
Because it will take a while for #5706 to resolve (checking in Arabic strings), this is a 🍒 pick of #5717 which fixes Android arrow icons to mirror for RTL layouts.

## User Testing
We won't have an Arabic local available for testing the Keyman app, so use a [pseudolocale](https://developer.android.com/guide/topics/resources/pseudolocales) AR(XB). This will display English strings (RTL) and allow us to test the RTL layouts

<details><summary>Pseudolocale setup</summary>

1. Start an Android 10 emulator (but don't launch the Keyman app yet)
2. From the device's "Settings" --> About Phone --> Build number --> [enable developer settings](https://developer.android.com/studio/debug/dev-options). This usually involves clicking on the Android Build number multiple times.
3. After developer settings are enabled, reboot the Android emulator
4. When the emulator is back, go to device's "Settings" --> Languages --> cibarA (XB)
Screenshot_1632186847
    * If you can't find the (XB) pseudolocale, troubleshoot with the pseudolocale link at the top of this setup summary
</details>

* **TEST_RTL_ARROWS** - Tests that  arrows are correct for RTL
1. With the device in the AR(XB) locale, start the PR build of Keyman for Android
2. Verify the app shows English strings in an RTL layout
Screenshot_1632187618
3. In the app, go to "Settings"
4. On the "Installed languages" selection, verify the forward arrow points left. (Screenshot shows 2 installed languages, but you'll only have 1 off a fresh install)
Screenshot_1632186959
5. Click on "Installed languages" --> English
6. Verify the Forward arrows point left, and the back arrow points right
Screenshot_1632187808
7. Back out of Settings
8. From the top menu, select "Text Size" menu
9. Click on the forward arrow (pointing left)
10. Verify the text size increases
11. Click on the back arrow (pointing right)
12. Verify the text size decreases
Screenshot_1632187865


mcdurdin:
Do we need this given we have no RTL translations in master? Shouldn't it just wait for the Arabic translation to land?

darcywong00:
> Do we need this given we have no RTL translations in master? Shouldn't it just wait for the Arabic translation to land?

I was hoping to keep it separate from the Arabic translations PR so we'd have some flexibility. (e.g. another RTL language ends up finishing before Arabic. Crowdin has a Shuwa Arabic translation about ready to pull in)

mcdurdin:
I think the blocker is not the Arabic translation but the RTL work we need to do generally -- so all the RTL translations will be held up in the same way, won't they?

darcywong00:
Per discussion in standup, I'll abandon this PR and we'll have #5706 include any changes needed for RTL layouts

PR #5738, opened 2021-09-22
chore(common): support forked repos when triggering test builds

mcdurdin:
This change allows branches from forked repos to trigger test builds.

Note: this will not be run automatically, but only when we manually tell it to run in TeamCity.

@keymanapp-test-bot skip


jahorton (commented):
> Note: this will not be run automatically, but only when we manually tell it to run in TeamCity.

Where's the mechanism that handles this logic?

mcdurdin:
> > Note: this will not be run automatically, but only when we manually tell it to run in TeamCity.
> 
> Where's the mechanism that handles this logic?

In the TeamCity build configuration (can't be in-repo!)

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.119-alpha](https://keyman.com/downloads/releases/alpha/15.0.119)

PR #5750, opened 2021-09-27
fix: help.keyman.com script file cleanup

mcdurdin:
Ensures that the destination folder is cleaned before deployment.


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.282](https://keyman.com/downloads/releases/stable/14.0.282)

Issue #5396, opened 2021-06-29
feat(android): move all network and sd file access out of Keyman Engine into Keyman


mcdurdin:
**Is your feature request related to a problem? Please describe.**

Keyman Engine should not make network connections, nor should it access user files. The host app should provide access to resources as required.

There are two current points where Keyman Engine accesses network:

* Sentry error reporting
* Checking for resource updates / downloading resources

**Describe the solution you'd like**

1. Error reporting should be delegated to the host app (if necessary, provide a callback).
2. Checking for resource updates and downloading resources should be done in the host app also.

---

**Keyman for iPhone/iPad/Android:**
 - Device: Android 
 - Keyman version: 14.0 

---

Emerging from discussions at https://community.software.sil.org/t/kab-support-for-keyman-14-or-15/4857

darcywong00:
This also ties to some Keyboard App Builder users who would like the network permission removed from the Keyman Engine AndroidManifest.xml file.

Network access is also used to display keyboard help page (help.keyman) on the Keyboard Info pages.

mcdurdin:
> Network access is also used to display keyboard help page (help.keyman) on the Keyboard Info pages.

This should definitely be in Keyman, not Keyman Engine.

chrisvire:
I believe this covers our needs in Keyboard App Builder.  Thanks!

mcdurdin:
All PRs relating to this issue should be run past the KAB team before merge.

mcdurdin:
Per @chrisvire:

Main issues:
* sentry.io
* AndroidManifest.xml : READ_EXTERNAL_STORAGE, INTERNET, ACCCESS_NETWORK_STATE permissions
* `getUpdateTool().checkForResourceUpdates()`

darcywong00:
This will deprecate some of the KMManager [APIs](https://help.keyman.com/developer/engine/android/15.0/KMManager/):

* [addKeyboardDownloadEventListener](https://help.keyman.com/developer/engine/android/15.0/KMManager/addKeyboardDownloadEventListener)
* [getMaySendCrashReport](https://help.keyman.com/developer/engine/android/15.0/KMManager/getMaySendCrashReport)
* [hasConnection](https://help.keyman.com/developer/engine/android/15.0/KMManager/hasConnection.php)?
* [setMaySendCrashReport](https://help.keyman.com/developer/engine/android/15.0/KMManager/setMaySendCrashReport)
[...]

PR #5752, opened 2021-09-27
fix(android/engine): Remove unnecessary permissions from Manifest

darcywong00:
Starts to address #5396 

The Keyman Engine for Android classes for `cloud/`, `data/`, and `logic/` are too entangled to move from Keyman Engine to the Keyman app.
As a stopgap, this PR removes the unnecessary permissions from the engine's Android_Manifest.xml file:
*  INTERNET
* READ_EXTERNAL_STORAGE
* ACCESS_NETWORK_STATE (per review comment)
 
Some internal calls also updated to handle when those permissions aren't available.

Apps using Keyman Engine for Android can then add those permissions back in their Manifest file (e.g. Keyman and FirstVoices).

## User Testing

* **TEST_KEYMAN_SDK_21** - Tests Keyman on Android 5.0
1. Start an Android 5.0 emulator (SDK 21) and disable network access
2. Load the PR build of Keyman
3. On the "Get Started" menu -->Add a keyboard for your language
4. On the next menu, select "Install from keyman.com"
5. Verify a "Cannot connect to Keyman server!" dialog appears
6. Click "OK" to dismiss
7. On the device, enable network access
8. On the same Keyman menu, select "Install from keyman.com"
9. Verify a keyboard search page appears
10. Search for "khmer_angkor" keyboard and click through to Install the keyboard
11. Verify the keyboard package downloads and the keyboard installation wizard appears
12. Click "Install" and "OK"
13. Verify the khmer_angkor keyboard is installed
14. Open the "Chrome" app and save [sil_cameroon_qwertry.kmp](https://downloads.keyman.com/keyboards/sil_cameroon_qwerty/6.0.5/sil_cameroon_qwerty.kmp) to the device
15. Return to the Keyman app to the yellow text area
16. Go to "Settings" --> Install Keyboard or Dictionary --> Install from local file
17. Browse to Downloads --> select the downloaded sil_cameroon_qwerty.kmp file
18. Verify the keyboard installation wizard appears
19. Click "Next" --> select a language --> Install --> OK
20. Verify the sil_cameroon_qwerty keyboard is installed

* **TEST_KEYMAN_SDK_30** - Tests Keyman on Android 11.0

1. Repeat the steps for TEST_KEYMAN_SDK_21 using an Android 11.0 emulator (SDK 30). Remember the test starts with disabled network access.

* **TEST_KMSAMPLE1_SDK_30** - Tests app without INTERNET and ACCESS_NETWORK_STATE permissions
1. Start an Android 11 emulator (SDK 30) and disable network access
2. Load the PR build of KMSample1
3. Verify the app launches without crashing
4. Press various keys and suggestions and verify they work

### DIFF
--- a/android/KMEA/app/src/main/AndroidManifest.xml
+++ b/android/KMEA/app/src/main/AndroidManifest.xml
@@ -2,7 +2,6 @@
   xmlns:tools="http://schemas.android.com/tools"
   package="com.tavultesoft.kmea" >
 
-    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.VIBRATE" />
 
     <!-- Ensure these permissions aren't added by dependencies unless calling app intentionally adds them -->
--- a/android/KMEA/app/src/main/AndroidManifest.xml
+++ b/android/KMEA/app/src/main/AndroidManifest.xml
@@ -1,9 +1,12 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.tavultesoft.kmea" >
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
+  package="com.tavultesoft.kmea" >
 
-    <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.VIBRATE" />
+
+    <!-- Ensure these permissions aren't added by dependencies unless calling app intentionally adds them -->
+    <uses-permission android:name="android.permission.INTERNET" tools:node="remove" />
 
     <application
         android:theme="@style/AppTheme" >
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -127,8 +127,8 @@ final class KMKeyboard extends WebView {
 
     // Normally, this would be true to prevent the WebView from accessing the network.
     // But this needs to false for sending embedded KMW crash reports to Sentry (keymanapp/keyman#3825)
-    // Throws SecurityException if INTERNET permission not granted
     if (KMManager.hasInternetPermission(context)) {
+      // Throws SecurityException if INTERNET permission not granted
       getSettings().setBlockNetworkLoads(!KMManager.getMaySendCrashReport());
     }
 
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -23,9 +23,11 @@ import com.tavultesoft.kmea.util.FileUtils;
 import com.tavultesoft.kmea.util.KMLog;
 import com.tavultesoft.kmea.util.KMString;
 
+import android.Manifest;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.graphics.Color;
 import android.graphics.Rect;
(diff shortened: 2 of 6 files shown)


darcywong00:
tagging @chrisvire for running this by KAB

android/KMEA/app/src/main/AndroidManifest.xml
```diff
@@ -1,10 +1,13 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.tavultesoft.kmea" >
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
+  package="com.tavultesoft.kmea" >
 
-    <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
mcdurdin, line 5: Do we need this perm in Engine?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
```diff
@@ -125,7 +127,10 @@ public void initKMKeyboard(final Context context) {
 
     // Normally, this would be true to prevent the WebView from accessing the network.
     // But this needs to false for sending embedded KMW crash reports to Sentry (keymanapp/keyman#3825)
-    getSettings().setBlockNetworkLoads(!KMManager.getMaySendCrashReport());
+    // Throws SecurityException if INTERNET permission not granted
```
mcdurdin, line 130: I'm not sure I understand this comment -- does this mean `hasInternetPermission` will crash?

mcdurdin (approved):
These changes look good. We don't actually need to move everything out of Engine if the permissions are controlled by the app -- so long as Engine never tries to do things it is not permitted to do.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
darcywong00: `setBlockNetworkLoads()` throws SecurityException if INTERNET permission is not granted. That's why I added a `KMManager.hasInternetPermission()` check.

android/KMEA/app/src/main/AndroidManifest.xml
mcdurdin: So we should be able to remove this in Engine and add it in apps instead?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
mcdurdin: Okay, the comment should be moved inside the if() block then :grin:

android/KMEA/app/src/main/AndroidManifest.xml
darcywong00: Right. l.9 `tools:node="remove"` removes the INTERNET permission from the manifest (even if a dependency like sentry tries to add it back).

Then the KMAPro Manifest l.7 uses `tools:node="replace"` to ensure the INTERNET permission is "added".

mcdurdin: But I'm looking at `    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />`, not `INTERNET`?

darcywong00: I'll put this PR back in draft to also remove ACCESS_NETWORK_STATE permission

darcywong00: @keyman-test-bot retest TEST_KMSAMPLE1_SDK_30

mcdurdin (approved):
LGTM. This seems to me to be enough for KAB -- good work.

MakaraSok:
* **TEST_KEYMAN_SDK_21: PASSED**
   * when network disabled, the message reads `Cannot connect to Keyman server!` does appear.
   * when network enabled, 
      * the search page does appear in full, i.e. both the search bar and instructions to search.
      * the keyboard package does download and the keyboard installation wizard does appear
      * the keyboard installation completed successfully
      * Cameroon QWERTY keyboard installation from file completed successfully.

MakaraSok:
* **TEST_KEYMAN_SDK_30: PASSED**
   * when network disabled, the message reads `Cannot connect to Keyman server!` does appear.
   * when network enabled, 
      * the search page does appear in full, i.e. both the search bar and instructions to search.
      * the keyboard package does download and the keyboard installation wizard does appear
      * the keyboard installation completed successfully
      * Cameroon QWERTY keyboard installation from file completed successfully.
* **TEST_KMSAMPLE1_SDK_30: BLOCKED** Can't locate the build for this test. Shouldn't it be here: https://build.palaso.org/buildConfiguration/KeymanAndroid_TestSamplesAndTestProjects/291506?buildTab=artifacts.

darcywong00:
>  Can't locate the build for this test. Shouldn't it be here:

I've updated the CI config to generate Sample artifacts. You should be able to get KMSample1 and KMSample2 apks now.

MakaraSok:
* **TEST_KMSAMPLE1_SDK_30: PASSED** No crash encountered and the keyboard works as expected.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.125-alpha](https://keyman.com/downloads/releases/alpha/15.0.125)

PR #5769, opened 2021-09-29
chore(windows): Keyman Core UI and Settings 🥑

mcdurdin:
* Adds a taskbar icon to highlight to user when Keyman Core library is in use, with a crummy core.ico. This visual feedback will be removed in late beta. Note that this means there will be 2 icons visible when Core is active:

  image

* Renames the 'use common core' registry setting to match other flags
* Adds a System Setting to make the Keyman Core setting visible in System Settings and kmconfig.

# User Testing

* TEST_ICON_WORKING: verify that the new core icon is visible when Keyman Core is active
* TEST_CORE_SETTING: Check that you can set `engine.compatibility.use_keyman_core` to `0` in kmconfig (Keyman Configuration / Options / System Settings), restart Keyman and verify that the new core icon disappears -- this should use the legacy code path.

If you want to verify that Keyman Core is in use or not internally, try the chinese IMX keyboard: it won't work with Core but should work with legacy Windows.


windows/src/engine/keyman/UfrmKeyman7Main.pas
```diff
@@ -502,6 +504,40 @@ procedure TfrmKeyman7Main.GetTrayIconHandle;
   end;
 end;
 
+procedure TfrmKeyman7Main.CreateCoreTrayIcon;
+var
+  cust: IKeymanCustomisation;
+  istrm: IStream;
+  olestrm: TOleStream;
+const
+  REGSZ_CoreIntegation = 'use common core';
```
rc-swag, line 513:
```suggestion
  REGSZ_Flag_UseKeymanCore = 'Flag_UseKeymanCore';
```

windows/src/engine/keyman/UfrmKeyman7Main.pas
```diff
@@ -502,6 +504,40 @@ procedure TfrmKeyman7Main.GetTrayIconHandle;
...
+begin
+  if not Reg_GetDebugFlag(REGSZ_CoreIntegation, True) then
```
rc-swag, line 515:
```suggestion
  if not Reg_GetDebugFlag(REGSZ_Flag_UseKeymanCore , True) then
```

rc-swag (commented):
Just need to update to dephi to match registry flag name change.

rc-swag (approved):
lgtm

MakaraSok:
* **TEST_ICON_WORKING: PASSED** After installing and running Keyman, the icon does appear in the taskbar near the usual Keyman icon. 

   image

* **TEST_CORE_SETTING: PASSED** The icon does go away after changing the value of `engine.compatibility.use_keyman_core` to `0` and restarting Keyman.

   image

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.122-alpha](https://keyman.com/downloads/releases/alpha/15.0.122)

Issue #4608, opened 2021-03-05
chore(common): update js dependencies


mcdurdin:
Some NPM JS dependencies have security updates we should apply.

darcywong00:
https://github.com/keymanapp/keyman/security/dependabot

PR #5770, opened 2021-09-29
chore(developer): Update jszip version

darcywong00:
Addresses #4608 by updating jszip dependency to from `2.5.0` to `3.7.0`.

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM. I assume you verified compat between versions of jszip?

darcywong00:
> LGTM. I assume you verified compat between versions of jszip?

Oh, I only checked that developer/js/build.sh builds. I didn't look where jszip gets used

mcdurdin:
It'll be used in the lexical model compiler

darcywong00:
> LGTM. I assume you verified compat between versions of jszip?

Ah it wasn't as straightforward as I thought. Migration involved replacing the deprecated call `generate()` with `generateAsync()`.
I was also able to remove the previous jszip.d.ts type.

Reference:
https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html#type-option

developer/js/source/package-compiler/kmp-compiler.ts
```diff
@@ -145,8 +145,8 @@ export default class KmpCompiler {
     zip.file(kmpJsonFileName, JSON.stringify(kmpJsonData, null, 2));
 
     // Generate kmp file
-    var data = zip.generate({base64:false,compression:'DEFLATE'});
-    fs.writeFileSync(kmpFileName, data, 'binary');
-
+    zip.generateAsync({type: 'binarystring', compression:'DEFLATE'}).then(function (data) {
+      fs.writeFileSync(kmpFileName, data, 'binary');
+    });
```
mcdurdin, line 150: This is no longer synchronous, so I am concerned about side-effects. I'd like to investigate this further before we went ahead and merged.

mcdurdin: At the very least, `buildKmpFile` needs to return the `Promise` from `zip.generateAsync.then`, and the caller needs to handle it, to ensure that the zip file is actually ready for use before the next step in the compile. This potentially breaks a lot of the design, which has assumed synchronous use.

darcywong00: I've updated `buildKmpFile` to return the Promise, which also involved moving the call to `writeFileSync`.

mcdurdin (commented):
This looks good. But could you please update the test in developer\js\tests\test-package-compiler.ts to do the end-to-end test as well for completeness? 😁

darcywong00:
> This looks good. But could you please update the test in developer\js\tests\test-package-compiler.ts to do the end-to-end test as well for completeness? 😁

Per our discussion, if I can recompile the lexical-model repo with this change, that will suffice instead of the E2E test...

Will document that when I can repro

darcywong00:
> Will document that when I can repro

Here's the steps I did:
```
cd [path to lexical-models repo]/release/nrc/nrc.en.mtnt/source
node [path to keyman repo]/developer/js/dist/kmlmp.js nrc.en.mtnt.model.kps -o ../build/me.kmp
```

Output package me.kmp matches the built package nrc.en.mtnt.model.kmp

PR #5776, opened 2021-09-29
fix(developer): handle shortcut keys in debugger

### LINKED ISSUE #5718: bug(developer): debugger should capture all key events, even shortcuts such as Ctrl+A
Key events that have shortcuts associated are not making it into the keyboard debugger. We need to be capturing these key events before the shortcuts get a look-in, as this is what Keyman does within apps. If the key combo is not handled by the keyboard under debugging, then it can fall through to the shortcut.

Per @makarasok: 

```
+ [RCTRL K_A] > 'RightCtrl+A works!'  c not working
+ [LCTRL K_A] > 'LeftCtrl+A works!'   c not working

+ [RCTRL K_L] > 'RightCtrl+L works!'
+ [LCTRL K_L] > 'LeftCtrl+L works!'
```

Probably it is conflicting with the common shortcut "Ctrl+A". 

_Originally posted by @MakaraSok in https://github.com/keymanapp/keyman/issues/5697#issuecomment-920743502_

mcdurdin:
Fixes #5718.

The debugger memo control will now filter out action shortcut keys such as Ctrl+A, Ctrl+O, so that the keyboard being debugged can get the first chance to look at them and handle them if it wishes.

If the keyboard does not handle the keystroke, then a set of editor shortcut keys will be handled by the debugger, namely:

* Ctrl+A - select all
* Ctrl+C - copy
* Ctrl+V - paste
* Ctrl+X - cut
* Ctrl+Z - undo

Note that undo semantics are a little weird with the debugger, but that is not related to this fix. By a little weird, I mean that characters emitted by the keyboard being debugged do not flow into the undo buffer, so undo often has no effect. I am not planning to fix this idiosyncrasy at this time.

# User Testing

All these tests should be performed in the debugger.

* **TEST_BASELINE:** Make sure that general debugging works in a Keyman keyboard, and that text selection keys work.
* **TEST_SHORTCUT:** When debugging a keyboard that does not define Ctrl+A, press Ctrl+A and verify that it selects all text in the debug window.
* **TEST_HANDLED:** Create a keyboard that defines `+ [CTRL K_A]`. Test it and verify that Ctrl+A is handled by the keyboard and *doesn'* select all.

### DIFF
--- a/windows/src/developer/TIKE/child/UfrmDebug.pas
+++ b/windows/src/developer/TIKE/child/UfrmDebug.pas
@@ -825,6 +825,26 @@ procedure TfrmDebug.ExecuteEventAction(n: Integer);
     end;
   end;
 
+  procedure DoHandleShortcut(vk: UINT);
+  begin
+    // Because we disable shortcuts in the debug memo, there are a small set of
+    // editor Ctrl+Key shortcuts we need to rehandle here. Note that Ctrl+Ins,
+    // Shift+Ins, Shift+Del appear to be handled natively by the debug memo
+    // control (Windows code?), as are all cursor movement / selection keys,
+    // apart from Ctrl+A
+
+    if GetKeyState(VK_CONTROL) >= 0 then
+      Exit;
+
+    case vk of
+      Ord('A'): memo.SelectAll;
+      Ord('C'): memo.CopyToClipboard;
+      Ord('V'): memo.PasteFromClipboard;
+      Ord('X'): memo.CutToClipboard;
+      Ord('Y'): ; // redo not supported in edit
+      Ord('Z'): memo.Undo;
+    end;
+  end;
 
   procedure DoEmitKeystroke(dwData: DWord);
   var
@@ -850,6 +870,8 @@ procedure TfrmDebug.ExecuteEventAction(n: Integer);
         then msg.Msg := WM_SYSKEYUP
         else msg.Msg := WM_KEYUP;
       memo.Dispatch(msg);
+
+      DoHandleShortcut(LOBYTE(dwData));
 
       RealignMemoSelectionState(state);
     end;
@@ -1116,6 +1138,7 @@ begin
         begin
           StatusText := 'Simple Test';
           memo.ReadOnly := False;
+          memo.IsDebugging := True;
         end;
       duiDebugging:
         begin
@@ -1124,6 +1147,7 @@ begin
           StartBatch;
           StatusText := 'Debugging';
           memo.ReadOnly := True;
+          memo.IsDebugging := False;
         end;
       duiFocusedForInput:
         begin
@@ -1131,6 +1155,7 @@ begin
           FinishBatch;
           StatusText := 'Focused for input';
           memo.ReadOnly := False;
+          memo.IsDebugging := True;
           //SelectSystemLayout(True);
         end;
       duiReadyForInput:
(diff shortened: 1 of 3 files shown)


windows/src/developer/TIKE/main/RedistFiles.pas
```diff
@@ -124,8 +124,15 @@ function GetWixPath: string;
...
+var
+  root: string;
+const
+  DevProjectTemplatePath = 'windows\src\developer\kmconvert\data\';
 begin
-  Result := GetDebugPath('Debug_RedistTemplatePath', ExtractFilePath(ParamStr(0))+'projects\templates\');
+  if TKeymanPaths.RunningFromSource(root)
+    then Result := root + DevProjectTemplatePath
+    else Result := ExtractFilePath(ParamStr(0))+'projects\templates\';
+  Result := GetDebugPath('Debug_RedistTemplatePath', Result);
```
mcdurdin, line 135: This allows us to create new projects while debugging Keyman Developer from source, without fiddling with registry settings. I just threw it in here as a rider; it matches similar patterns in RedistFiles.pas and KeymanPaths.pas.

windows/src/developer/TIKE/child/UfrmDebug.pas
```diff
@@ -1124,13 +1147,15 @@ procedure TfrmDebug.SetUIStatus(const Value: TDebugUIStatus);
           StartBatch;
           StatusText := 'Debugging';
           memo.ReadOnly := True;
+          memo.IsDebugging := False;
```
darcywong00, line 1150: Why is this `False` in the duiDebugging case?

mcdurdin: Yeah, this one is a little tricky. The thing is, while single-step debugging, the control temporarily switches modes so characters can be inserted, etc, if these are a part of the action being debugged. But we don't want the user to make changes to the content of the control while single-step debugging -- if they are on a breakpoint and press Ctrl+X to delete content, then that will mess things up quite badly!

darcywong00 (approved):
lgtm

MakaraSok:
* **TEST_BASELINE: BLOCKED** Please clarify and illustrate what `text selection keys` are. Are they all listed right above the instructions?
* **TEST_SHORTCUT: PASSED** `Ctrl A` does select all texts in the debugger.

   

* **TEST_HANDLED: PASSED** The key defined does output the expected character and not selecting all texts in the debugger.

mcdurdin:
> * Please clarify and illustrate what `text selection keys` are. Are they all listed right above the instructions?

There would be two categories: text movement and text selection:
* Text movement keys would be all arrow keys <kbd>←</kbd>, <kbd>→</kbd>, <kbd>↑</kbd>, <kbd>↓</kbd>, <kbd>home</kbd>, <kbd>end</kbd>, <kbd>pgup</kbd>, <kbd>pgdn</kbd>, <kbd>ctrl</kbd>+<kbd>home</kbd>, <kbd>ctrl</kbd>+<kbd>end</kbd>.
* Then text selection keys are <kbd>shift</kbd>+any text movement combination from above, as well as <kbd>Ctrl</kbd>+<kbd>A</kbd>.

There may be other text movement keys but they would be less critical, and would probably work in any case :grin:

@keymanapp-test-bot retest TEST_BASELINE

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.143-alpha](https://keyman.com/downloads/releases/alpha/15.0.143)

PR #5782, opened 2021-09-30
fix(common): Fix cherry-pick labeling

ermshiperete:
My previous commit escaped the parens which should be part of the regex.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.123-alpha](https://keyman.com/downloads/releases/alpha/15.0.123)

PR #5787, opened 2021-10-04
chore: sentry.io dsn 👮‍♂️

mcdurdin:
Updates all the DSNs for sentry.io for all Keyman apps. Will cherry-pick

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/__init__.py
+++ b/linux/keyman-config/keyman_config/__init__.py
@@ -57,7 +57,7 @@ else:
             level=logging.INFO,           # Capture info and above as breadcrumbs
             event_level=logging.CRITICAL  # Send critical errors as events
         )
-        SentryUrl = "https://1d0edbf2d0dc411b87119b6e92e2c357@sentry.keyman.com/12"
+        SentryUrl = "https://1d0edbf2d0dc411b87119b6e92e2c357@o1005580.ingest.sentry.io/5983525"
         sentry_sdk.init(
             dsn=SentryUrl,
             environment=__environment__,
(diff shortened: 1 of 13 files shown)


linux/keyman-config/keyman_config/__init__.py
```diff
@@ -78,7 +78,7 @@ def _(txt):
             from raven import Client
             HaveSentryNewSdk = False
 
-            SentryUrl = "https://1d0edbf2d0dc411b87119b6e92e2c357:e6d5a81ee6944fc79bd9f0cbb1f2c2a4@sentry.keyman.com/12"
```
mcdurdin, line 81: note: secret part of DSN is deprecated

ermshiperete: Yes, but unfortunately the python-raven client still needs it :disappointed: - that's the cause for the build failure on Jenkins.

We either have to add it back in, or drop support for Ubuntu 18.04 Bionic (EOL April 2023).

ermshiperete: Another option for the packages we publish on llso, pso and ppa would be to update python3-raven to the latest version. Available in Ubuntu is 6.3.0; 6.7.0 made the DSN secret optional. We would then require at least version 6.7.0 of python3-raven. That might actually be an acceptable solution, because Ubuntu won't ship a newer Keyman version for Bionic.

But it's still more work than just adding the secret part back in...

mcdurdin: I've added the secret back in with a note.

ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.126-alpha](https://keyman.com/downloads/releases/alpha/15.0.126)

PR #5797, opened 2021-10-05
feat(developer): kmcompx: cross-platform compiler 🍃

mcdurdin:
Initial seed for kmcompx, builds under msvc in meson. Probably breaks Windows compiler build?

This is the starting point for the cross-platform build of kmcompx before we start stripping out Windowsisms.


developer/kmcompx/src/meson.build
```diff
@@ -0,0 +1,102 @@
...
+version_res = []
+
+# TODO: move these shared defines to a resources/meson folder?
+if compiler.get_id() == 'gcc' or compiler.get_id() == 'clang'
+  warns = [
+     '-Wall',
+     '-Wctor-dtor-privacy',
+     '-Wdouble-promotion',
+     '-Wendif-labels',
+     '-Wextra',
```
eli-schwartz, line 17: Meson's default options: `warning_level=1` will already set `-Wall`, level 2 adds `-Wextra`, and it is preferred to set it to `warning_level=2` in your project() options if you want `-Wextra` rather than hardcoding it in the build file (it will handle MSVC for free for you)

developer/kmcompx/src/meson.build
```diff
@@ -0,0 +1,102 @@
...
+     '-Wno-unknown-pragmas',
+     '-Wno-missing-field-initializers',
+     '-Wnon-virtual-dtor',
+     '-Wshadow'
+  ]
+  flags = [
+    '-fvisibility=hidden',
+    '-fvisibility-inlines-hidden'
```
eli-schwartz, line 25: You can instead add the kwarg `gnu_symbol_visibility: 'inlineshidden'` to the target. Meson will check whether the current compiler supports both these flags, then add both of them if supported.

mcdurdin:
@eli-schwartz thank you for your review!

I will feed your suggestions to @ermshiperete who is maintaining most of our meson build infrastructure. I think we are still stuck on 0.45 due to some packaging infrastructure but am not 100% sure; @ermshiperete will know. Note: this draft PR will not be merged as it is a base for #6026 which is being prepared by @SabineSIL.

eli-schwartz:
No problem :) I like helping people use meson more effectively.

Relying on meson 0.45 implies that you need to be able to build on Ubuntu 18.04 LTS (and 20.04 LTS isn't a suitable minimum) and need to rely on the Ubuntu version of meson rather than one locally upgraded. Which is a fair concern to have.

You can use the `meson_version: '>=0.45.0'` kwarg to [`project()`](https://mesonbuild.com/Reference-manual_functions.html#project) to specify the minimum version of meson you support -- meson will then warn you if you use features which are too new, which is handy for e.g. local development using newer versions of meson and not realizing that you are trying to do something that won't work in CI / buildbots.

mcdurdin:
@eli-schwartz Awesome, again thank you! 👍 

Up for a challenge? I don't suppose you'd be interested in looking into one of our more troublesome areas with meson: conflicting requirements between mseon, rust, emscripten, and MSVC for building common/core/desktop. We have a fairly complex set of requirements -- which you can probably tell just by the painful list of tools above!

1. We are building a library "kmkbp", both static and dynamic versions. For example, on Windows, the library is statically linked into the Keyman product, but dynamically linked into the Keyman Developer product (for reasons).

2. Rust code is currently linked into a separate library, but that's probably not our long-term preference. 

2. The library has C++ and Rust sources.

3. We need to compile native on each platform:
   * Windows (i386 and x86-64), 
   * mac (umm... both intel and m1, although this is a bridge we haven't crossed yet)
   * Linux (we package i386 and x86-64 versions mostly but other archs should also 'just work')

4. On Windows, we are launching from git bash or cmd (with MSVC env loaded via vcvarsall.bat). Both need to work, and preferably also just plain cmd without any MSVC env. We do prefer git bash. This is the most painful part of the build -- vcvarsall.bat is ridiculously slow and we call it multiple times during the build. Running this way we also lose a lot of the benefits of meson's build performance.

5. Oh yeah, we also need to build a WASM target of the library.

We were playing whack-a-mole trying to get everything to just work together -- particularly rust (cargo), but also emscripten. We ended up with a bash build script common/core/desktop/build.sh which works on our three target dev platforms, but this script is certainly not my happy place!

ermshiperete:
> Relying on meson 0.45 implies that you need to be able to build on Ubuntu 18.04 LTS (and 20.04 LTS isn't a suitable minimum) and need to rely on the Ubuntu version of meson rather than one locally upgraded. Which is a fair concern to have.

Unfortunately at least for now we still have to support Ubuntu 18.04 and use the meson version that comes with that. We might be able to drop support for Ubuntu 18.04 for the next Keyman version, depending on what we decide when we will release Keyman 16.

> You can use the `meson_version: '>=0.45.0'` kwarg to [`project()`](https://mesonbuild.com/Reference-manual_functions.html#project) to specify the minimum version of meson you support

Thanks for the tip! I'll add it!

developer/kmcompx/.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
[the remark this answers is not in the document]
ermshiperete: @eli-schwartz In what meson version did that get introduced? I don't see a `build/.gitignore` file on my machine.

developer/kmcompx/src/meson.build
ermshiperete: Unfortunately that's only available in 0.48

mcdurdin:
Replaced by #6026.

developer/kmcompx/meson.build
```diff
@@ -0,0 +1,34 @@
...
+                           'cpp_std=c++14',
+                           'b_vscrt=static_from_buildtype'])
+
+message('meson.project_version(): ' + meson.project_version() + '\n')
+compiler = meson.get_compiler('cpp')
+
+# lib_version = '0.0.0'
+
+if compiler.get_id() == 'msvc'
+  add_global_arguments('/source-charset:utf-8', language: ['c', 'cpp'])
```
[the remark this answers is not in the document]
ermshiperete: Unfortunately this is only true with Meson >= 0.60

PR #5803, opened 2021-10-06
feat(windows): Change caps to work with common core processor

### LINKED ISSUE #5652: feat(windows): Keyman Core Integration - Full support for Caps Lock system stores 🥑
The common desktop core has been upgraded with Caps Lock support the Windows engine needs to be updated to send correct key up/down state with the keyevent to core and then also correctly process the new actions with regards to Caps.

rc-swag:
The Keyman for Windows engine now  sends through the Modifier key presses to the core processor so it could evaluate the key press against the keyboard rules for Caps. The Windows engine now to process the KM_KBP_IT_CAPSLOCK action and synthesizes key presses accordingly.  
Fixes  #5652


rc-swag:
> This looks really clean. There's one functional change I think we need to make (`!Updateable`) which may have side-effects so will need testing. Apart from that LGTM.

I have done some testing this morning. Processing only on the Updatable pass doesn't work for `CapsOnly` && `ShiftFressCaps`. It also makes CapsAlwaysOff a bit flakey it will always produce lower case letters but with mashing the caps lock key you can get the caps lock indicator to stay on until the next key press.
I am looking into it more to see if I can get it to work.

MakaraSok:
> The test keyboard layouts are found in the keyman repo at `app/windows/src/test/caps_lock_stores`.

The path to the keyboard packages for testing this PR doesn't exist, but there is something like this `keyman\windows\src\test\manual-tests\caps-lock-stores`. Please confirm. Thanks.

rc-swag:
> > The test keyboard layouts are found in the keyman repo at `app/windows/src/test/caps_lock_stores`.
> 
> The path to the keyboard packages for testing this PR doesn't exist, but there is something like this `keyman\windows\src\test\manual-tests\caps-lock-stores`. Please confirm. Thanks.

Yes this is correct and I have updated the test instructions

MakaraSok:
## SUITE_SHIFT: CapsOnOnly/ShiftFreesCaps
- **TEST_CAPSONLY-1: PASSED** `pass.` does get output when <kbd>1</kbd> is pressed and released on the VM's soft keyboard with the CapsLock off.
- **TEST_CAPSONLY-2: PASSED** `pass.` does get output and `CapsLock` key indicator turned on when <kbd>CapsLock</kbd> is pressed and released and then <kbd>2</kbd> is pressed and released on the VM's soft keyboard with the CapsLock initially off.
- **TEST_CAPSONLY-3: PASSED** `pass.` does get output and `CapsLock` key indicator turned on when <kbd>CapsLock</kbd> is pressed and released twice and then <kbd>6</kbd> is pressed and released on the VM's soft keyboard with the CapsLock initially off.
- **TEST_CAPSONLY-4: PASSED** `pass.` does get output and `CapsLock` key indicator turned off when <kbd>CapsLock</kbd> is pressed and released and <kbd>shift</kbd> is pressed and held and then <kbd>3</kbd> is pressed and released on the VM's soft keyboard with the CapsLock initially off.
- **TEST_CAPSONLY-5: FAILED** The `CapsLock` and `Shift` key indicators are all stay on.

rc-swag:
Talking with @ermshiperete using the VM soft keyboard it does NOT actually send the <kbd>Shift</kbd> Shift Key Stroke through but rather will change the keys pressed for example if an <kbd>a</kbd> is pressed the soft keyboard itself will change that key to a <kbd>A</kbd>.  This means we can't do Test   TEST_CAPONLY-5 on a soft keyboard. I will still merge this PR as it passes when an actual shift stroke is sent through.    Thoughts?

PR #5806, opened 2021-10-07
fix(ios): move sentry settings responsibility to build agent 👮‍♂️ 🍒

mcdurdin:
Cherry-pick of #5805.

The build agent already has the relevant `SENTRY_` environment variables set in order to complete builds for all projects. Thus, we should exclude these values from the xcode build. However, we need to ensure that the `SENTRY_URL` and `SENTRY_ORG` variables are imported into the xcode environment from the build agent environment.

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.283](https://keyman.com/downloads/releases/stable/14.0.283)

PR #5808, opened 2021-10-07
feat(windows): manual keyboard caps tests

rc-swag:
Add manual test layouts to assist in testing the Caps Lock behaviour based on the Caps Lock Stores.  Also added a TestReadme.md with all the steps required for user and developer testing. 

@keymanapp-test-bot skip


mcdurdin:
Can we add the full source folder for each keyboard as found on @ermshiperete's repo, rather than just the .kmn?

rc-swag:
> Can we add the full source folder for each keyboard as found on @ermshiperete's repo, rather than just the .kmn?

Sure we can.  Like I said following the structure of what was done before there was just `.kmn` some had `.kmx` and some had the `build` folder as well.

mcdurdin:
> @mcdurdin Do you want just the source folder or do you want the next level up also with the .kpj ?

The next level up with the .kpj as well -- everything except the build folder and .kpj.user file.

mcdurdin:
> While I'm at these changes what do you think of renaming the capscontrol to shift_frees_caps and obviously updating all references in the TestReadme.md?

Sounds good to me.

PR #5819, opened 2021-10-07
chore: Create CODE_OF_CONDUCT.md

mcdurdin:
Based on the GitHub template; I have reviewed this and don't think I see any issues. At present, @mcdurdin and @darcywong00 will be responsible for maintenance of community standards through the email address listed in this document.

See: https://github.com/keymanapp/keyman/community/code-of-conduct/new?template=contributor-covenant -- this has no changes from the template so would prefer not to make changes to it ...

@keymanapp-test-bot skip

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.129-alpha](https://keyman.com/downloads/releases/alpha/15.0.129)

PR #5570, opened 2021-08-09
fix(android/engine): UX improvement when 1 keyboard installed

darcywong00:
Continues globe key refactoring from #5437 and fixes #5545

A short-press on globe key to switch Keyman keyboards when only 1 Keyman keyboard installed leads to a confusing UX.
This PR makes the following improvement:
* When only 1 Keyman keyboard is installed, short press on globe key becomes SHOW_MENU

This PR also refactors both inapp and system keyboard globe actions to use the same methods for handling lock screen overrides, short press and long press keys.

A recap of expected inapp and system keyboard globe functionality:
* When 1 keyboard installed, shortpress shows keyboard picker menu
* When 2+ keyboards installed, short press switches to the next Keyman keyboard
* On the lock screen, both short press and longpress switches to the next keyboard (Keyman or system IME)

## User Testing
@keymanapp/testers 
The tests are split into two phases:
1. behavior of globe key when 1 keyboard installed (off default Keyman installation)
2. behavior of globe key when 2+ keyboards installed

### Globe Key when 1 Keyboard installed
Setup
* Install PR build and set Keyman as a System Keyboard
* On the device, set a simple lock screen password "1234". (If using a physical device, recommend a fingerprint unlock as backup)

#### In app keyboard globe key (1 keyboard installed)
*  In the Keyman app, short press the globe key and release
*  Verify the Keyboard Picker menu appears (for SIL_Euro_Latin)
*  Click the back button to exit
* In the Keyman app, long press the globe key and release
* Verify the Keyboard Picker menu appears (for SIL_Euro_Latin)
* Click the back button to exit

#### System keyboard globe key (1 keyboard installed)
* Launch a separate app like Chrome
* Select the omnibar and display the Keyman OSK
* Short press the globe key and release
* Verify the Keyboard Picker menu appears and that it includes the "Next Input Method" button
* Click the back button to exit
* Display the Keyman OSK
* Long press the globe key and release
* Verify the Keyboard Picker menu appears and that it includes the "Next Input Method" button

#### System keyboard lock screen (1 keyboard installed)
* With Keyman selected as the system keyboard, lock the device
* Hit the power button again to enter the password
* Short press the globe key and release
* Verify the keyboard switches to the next IME (e.g. Google Gboard)
* While still at the password field, switch to Keyman 
* Longpress the globe key and release
* Verify the keyboard switches to the next IME
* Unlock the device

----
### Globe Key when 2+ Keyboard installed
Continued Setup
[...]

PR #5836, opened 2021-10-18
chore(android/engine): Address globe key TODO

darcywong00:
Addresses TODO from https://github.com/keymanapp/keyman/pull/5570#discussion_r689151866

In `handleGlobeKeyAction`, if `KMManager.shouldAllowSetKeyboard()` is false, then `switchToNextKEyboard()` should not be allowed.

At the time this PR is created, the branch contains test code https://github.com/keymanapp/keyman/commit/140e45d93736a102e045815e56bae71fb998938a which disables `shouldAllowSetKeyboard` to enable user testing. 

Because this test code needs to be reverted before merging to production, reviewers should wait until user testing is complete.

## User Testing
* **TEST_DISABLED_GLOBE_ACTION** - This test can only be run before the test code (mentioned above) gets reverted.
1. Load the PR build
2. From the "Settings" menu, search for another keyboard and install (e.g. khmer_angkor)
3. When the keyboard package finishes installing, verify the keyboard remains the default sil_euro_latin keyboard
4. Short press the globe key
5. Verify the keyboard does **not** switch
6. Long press the globe key
7. Verify the keyboard does **not** switch

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
@@ -159,7 +159,7 @@ public final class KMManager {
 
   private static InputMethodService IMService;
   private static boolean debugMode = false;
-  private static boolean shouldAllowSetKeyboard = false; // TEST CODE to remove after user testing: Do not merge to production
+  private static boolean shouldAllowSetKeyboard = true;
   private static boolean didCopyAssets = false;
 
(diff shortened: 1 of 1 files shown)


MakaraSok:
* **TEST_DISABLED_GLOBE_ACTION: PASSED** There is one error though after the installation is finishing. Please see the screenshot.

   

   Note: This was marked as PASSED because all expected behaviors are confirmed: (1) the keyboard stays in sil_euro_latin after the installation and (2) short/long press on the globe key does nothing.

PR #5842, opened 2021-10-20
fix(common/core):  Update decxstr to check all characters while decrementing

### LINKED ISSUE #5812: bug(common): decxstr needs extra robustness for malformed strings
@ermshiperete writes: It seems we need similar checks in `decxstr`.

@mcdurdin adds:
Re `decxstr`, you are probably correct, particularly for the `*(p-1)` case where we don't test the `CODE_xxx` type. `decxstr` doesn't have a failure mode, but we should define what it does for malformed xstrings. However, as far as I can tell, we can't ever jump out of the buffer, unlike with `incxstr` -- in the worst case we'll return `UC_SENTINEL` as a standalone char (which will cause trouble elsewhere, probably).

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/issues/4604#issuecomment-901465686_

SabineSIL:
Fixes #5812 (related to #4604).

While working on the incxstr-function (#5712) we established an array named CODE__SIZE[] filled with data used for pointer movements. The decxstr-function is now adapted to use this array also. 
Since changes are made to some CODE_xxx values ( CODE_EXTENDED (not checked -> 2); CODE_SWITCH (not checked -> 1); CODE_CLEARCONTEXT ( 1->0) ) decxstr may behave differently for those CODE_xxx. 

@keymanapp-test-bot skip


common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -173,43 +173,19 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
...
+  else
   {
-    switch(*(p-2))
+    for (int i = 1; i < 5; i++)
     {
-      case CODE_INDEX:
-      case CODE_SETOPT:
-      case CODE_SETSYSTEMSTORE:
-        return p-3;
+      //   *(p - i) == UC_SENTINEL  &&  next CODE_ right to UC_SENTINEL ( looked up in CODE__SIZE+1) has value i
```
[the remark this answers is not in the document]
SabineSIL: ...done :-)

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -173,43 +173,18 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
...
-      case CODE_RESETOPT:
-      case CODE_SAVEOPT:
-        return p-2;
-    }
-  }
-  else if(p > pStart+2 && *(p-3) == UC_SENTINEL)
+  else
   {
-    switch(*(p-2))
+    for (int i = 1; i < 5; i++)
```
ermshiperete, line 180: I think we could start at `i = 2` since we checked `*(p-1)` above.

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -173,43 +173,18 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
...
     {
-      case CODE_INDEX:
-      case CODE_SETOPT:
-      case CODE_SETSYSTEMSTORE:
-        return p-3;
+      //   p right of pstart &&  *(p - i) == UC_SENTINEL  &&  next CODE_ right of UC_SENTINEL ( looked up in CODE__SIZE+1) has value i
+      if ((p >= pStart + i)  && (*(p - i) == UC_SENTINEL) && (CODE__SIZE[*(p - i + 1)] + 1 == i))
+        return (p - (CODE__SIZE[*(p - i + 1)] + 1));
```
ermshiperete, line 184: This would be easier to humanly digest as:

```suggestion
        return (p - i);
```

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -16,7 +16,587 @@
 using namespace km::kbp::kmx;
 using namespace std;
 
-void test_incxstr() {
+PKMX_WCHAR find_ptr_to_last_character(PKMX_WCHAR p_first) {
+  int length = wcslen((const wchar_t *)p_first); 
+ if (length > 0)
```
ermshiperete, line 21:
```suggestion
  if (length > 0)
```

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -16,7 +16,587 @@
...
+    PKMX_WCHAR p;       // pointer end of input
+    PKMX_WCHAR q;       // pointer output
+
+
+    // -------------------------------------------------------------------------------------------------------------------------------------------------------
+    // more tests: check if we might end up left of pstart
+    // -------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    p_start = (PKMX_WCHAR)u"\u0001";
+    p       = find_ptr_to_last_character(p_start);
```
ermshiperete, line 40: Since this is a unit test where we know the content of the pointer it might be better (aka easier to read) to explicitly state where `p` points to, i.e. in this case `p = p_start`. On the other hand this looses the reminder that `p` points to the last character.
Another option would be to assert `p`:
```
p       = find_ptr_to_last_character(p_start);
assert(p == p_start);
```
but I don't particularly like that option either...

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -16,7 +16,587 @@
...
+     // ---- character
+     // --------------------------------------------------------------------------------------- 
+   
+     p_start = (PKMX_WCHAR)u"abcd";     
+     p       = find_ptr_to_last_character(p_start);
+     q       = decxstr(p, p_start);
+       
+     assert(q == (p - 1) );
+
+         // ---------------------------------------------------------------------------------------
```
ermshiperete, line 169:
```suggestion
    // ---------------------------------------------------------------------------------------
```

ermshiperete (approved):
A few suggestions, but otherwise lgtm

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
mcdurdin: I am happy with the consistency of the test setup as it is for this.

mcdurdin (changes requested):

common/core/desktop/src/kmx/kmx_xstring.cpp
mcdurdin: One other thing: can we avoid the use of the magic number `5` here and instead define a constant `#define CODE__SIZE_MAX 5` alongside the `CODE__SIZE` definition?

SabineSIL: OK

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: ...changed it

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
SabineSIL: OK

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
SabineSIL: ...did no changes...

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
SabineSIL: changed it

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -173,43 +173,18 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
...
   {
-    switch(*(p-2))
+    for (int i = 1; i < 5; i++)
     {
-      case CODE_INDEX:
-      case CODE_SETOPT:
-      case CODE_SETSYSTEMSTORE:
-        return p-3;
+      //   p right of pstart &&  *(p - i) == UC_SENTINEL  &&  next CODE_ right of UC_SENTINEL ( looked up in CODE__SIZE+1) has value i
+      if ((p >= pStart + i)  && (*(p - i) == UC_SENTINEL) && (CODE__SIZE[*(p - i + 1)] + 1 == i))
```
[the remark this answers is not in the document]
SabineSIL: included several tests

mcdurdin (approved):
LGTM.

Can you tidy up the style of the function to match the [style guide](https://github.com/keymanapp/keyman/wiki/Keyman-Code-Style-Guide)?

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -173,43 +175,19 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
...
-      case CODE_CLEARCONTEXT:
-      case CODE_CALL:
-      case CODE_CONTEXTEX:
-      case CODE_RESETOPT:
-      case CODE_SAVEOPT:
-        return p-2;
-    }
-  }
-  else if(p > pStart+2 && *(p-3) == UC_SENTINEL)
+  else
```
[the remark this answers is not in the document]
SabineSIL: removed them

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -173,43 +175,19 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
   {
     return p-1;
   }
+
   else if(*(p-1) == UC_SENTINEL) return p-1;
```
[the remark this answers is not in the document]
SabineSIL: I wonder if that ever will be a problem since we want to end up at the next UC_SENTINEL to the left if possible?
- For "normal" characters (u"\uFFFFa") or Surrogate Pairs we move 1 to the left to UC_SENTINEL.
- For all CODE__xxx where (CODE__SIZE[*(q + 1)] + 1 == i) we jump directly to next UC_SENTINEL to the left.
Are there any cases where we would run into problems when we are one position right to a UC_SENTINEL and we are then moving 1 step to the left to that UC_SENTINEL?

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -150,15 +150,17 @@ PKMX_WCHAR km::kbp::kmx::incxstr(PKMX_WCHAR p) {
...
   }
 
   p--;
-  if(*p == UC_SENTINEL_EXTENDEDEND)
-  {
+  if(*p == UC_SENTINEL_EXTENDEDEND) {
     int n = 0;
-    while(*p != UC_SENTINEL && n < 10) { p--; n++; }
+    while(*p != UC_SENTINEL && n < 10) {
+      p--; n++; }
```
mcdurdin, line 163:
```suggestion
    while(p >= pStart && *p != UC_SENTINEL && n < 10) {
      p--; n++;
    }
```

common/core/desktop/src/kmx/kmx_xstring.cpp
```diff
@@ -167,49 +169,28 @@ PKMX_WCHAR km::kbp::kmx::decxstr(PKMX_WCHAR p, PKMX_WCHAR pStart)
...
-      case CODE_IFOPT:
-      case CODE_IFSYSTEMSTORE:  // I3432
-        return p-4;
-    }
+  
+  q  = p - 2;
+  for (int i = 2; i < CODE__SIZE_MAX && q >= pStart; i++, q--) {
+    //  *q == UC_SENTINEL &&  *(q + 1) is within CODE__SIZE && next CODE_ right of UC_SENTINEL ( looked up in CODE__SIZE+1) has value i
+    if (*q == UC_SENTINEL &&  *(q + 1) <= CODE_LASTCODE     && CODE__SIZE[*(q + 1)] + 1 == i)
+      return (p - i);
```
mcdurdin, line 190:
```suggestion
    if (*q == UC_SENTINEL &&  *(q + 1) <= CODE_LASTCODE     && CODE__SIZE[*(q + 1)] + 1 == i) {
      return q;
    }
```

mcdurdin (changes requested):
I don't think this is an easy one to get to 100%! I just spotted something else in the virtual key pattern...

But it's close now, very close!

mcdurdin:
Tests failed:

```
[10:33:01][Step 1/1] Test failed with 0 at ../../../tests/unit/kmnkbd/test_kmx_xstring.cpp:46:
```

Which is:

https://github.com/keymanapp/keyman/blob/ba0092973e8ccb28951220e965e9472a80a45406/common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp#L43-L46

And given that the code for that UC_SENTINEL is out of range, I think the test is wrong; it should be:

```c++
 p_start = (PKMX_WCHAR)u"abc\uFFFF\uF000d"; 
 p       = find_ptr_to_last_character(p_start); 
 q       = decxstr(p, p_start); 
 assert(q == (p - 1)); 
```

mcdurdin:
I am also concerned that the Windows Core build passed where the other platforms didn't. We need to find out why.

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: OK

common/core/desktop/src/kmx/kmx_xstring.cpp
SabineSIL: **### _I am also concerned that the Windows Core build passed where the other platforms didn't. We need to find out why._**

I have pushed again, in case there were problems with my files. Also I have 2 more ideas:

Incxstr/decxstr are only changed in **\common\core\desktop\src\kmx\kmx_xstring.cpp** so far. They are not changed in **\windows\src\global\vc\xstring.cpp** yet.

There are changes in pointer movement in the new version of decxstr so some tests will not give the same results as CODE__SIZE[] for new decxstr uses different values of  pointer movement compared to the old version of decxstr. This applies to 3 cases : 

- o	0x0A ( not specified in old decxstr <-> CODE__SIZE[0x0A]= 2 in new decxstr) 
- o	0x0C (not specified in old decxstr  <-> CODE__SIZE[0x0C]= 1 in new decxstr) 
- o	0x0E (specified  as 1 in old decxstr <-> CODE__SIZE[0x0E]=0 in new decxstr) 

__________________________________________________________________________________

_**And given that the code for that UC_SENTINEL is out of range, I think the test is wrong; it should be…**_

I am puzzled by the behavior using 0xF000. Here are the results of the original decxstr-function: 


- p_start = (PKMX_WCHAR)u"abc\uFFFF\U0001F609d";	 ------> assert(q == (p - 2));
- p_start = (PKMX_WCHAR)u"abc\uFFFFzd"; ------> assert(q == (p - 2)); 
- p_start = (PKMX_WCHAR)u"abc\uFFFF\uF000d";	 ------> assert(q == (p - 1));


If we decxstr a surrogate pair in the original decxstr-function, the pointer moves 2 to the left. When using characters like u"abc\uFFFFzd"  in the original function the pointer will be moved 2 to the left. If we take 0xF0000 instead of z the pointer is supposed to move one to the left?  Is there a difference in the use of characters and characters of the private use area and how would one discern them? Are they always used after UC_SENTINEL?  Or do we simply move all characters >0xF000 just one step to the left?

To not produce errors in tests I removed the tests for 0xF000 until I am really sure how to process those characters. 
__________________________________________________________________________________

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -16,7 +16,635 @@
 using namespace km::kbp::kmx;
 using namespace std;
 
-void test_incxstr() {
+PKMX_WCHAR find_ptr_to_last_character(PKMX_WCHAR p_first) {
+  int length = wcslen((const wchar_t *)p_first);
```
mcdurdin, line 20: Turns out a hard cast hides all manner of evil! On macOS and Linux, `wchar_t` is a UTF-8 type. Which causes great havoc because we have a UTF-16 string 😢 

```suggestion
  int length = std::u16string(p_first).length();
```

Using `u16string` is the appropriate fix.

common/core/desktop/src/kmx/kmx_xstring.cpp
mcdurdin: > To not produce errors in tests I removed the tests for 0xF000 until I am really sure how to process those characters.

See my comment here: https://github.com/keymanapp/keyman/pull/5842#discussion_r749770382. With that in place, we should be able to re-enable tests.

SabineSIL: p_start = (PKMX_WCHAR)u"abc\uFFFF\U0001F609d"; ------> assert(q == (p - 2));
p_start = (PKMX_WCHAR)u"abc\uFFFFzd"; ------> assert(q == (p - 2));
p_start = (PKMX_WCHAR)u"abc\uFFFF\uF000d"; ------> assert(q == (p - 1));

What I don`t understand is:
-  in the second line we start out at "d", move 2 to the left and end up at "FFFF" (and skip "z" ?) 
-  in the third line we start out at "d", move 1 to the left and end up at "F000"

Why do we treat these similar cases differently? 
Why do we want to end **AT** FFFF in one case and **ONE RIGHT TO** FFF in the other?
Shouldn`t we stop  **ONE RIGHT TO** FFF in both cases? Was there a reason to set it up like that in the original code?

mcdurdin: You are right. The behaviour was poorly defined for invalid sequences.

I have gone ahead and defined the behaviour when we have an invalid UC_SENTINEL sequence:

1. UC_SENTINEL followed by an invalid code: treat as individual characters
2. UC_SENTINEL followed by a valid code but insufficient number of parameters: treat as individual characters

Note that scenario 2 can arise when we call `decxstr` on a pointer to any parameter to the code sequence. In a correctly formed string, this should never arise., so for example, in this scenario, we are pointing to the parameter for CODE_ANY:

```c++
  p_start = (PKMX_WCHAR)u"\uFFFF\u0001\u0002";
  p = p_start + 2; // \u0002 in the middle of the otherwise valid UC_SENTINEL CODE_ANY sequence
  q = decxstr(p, p_start);
  assert(q == p - 1);
```

However, note the outcome of the following:

```c++
  p_start = (PKMX_WCHAR)u"\uFFFF\u0001\u0001";
  p = p_start + 3; // nul, i.e. at the end of the valid UC_SENTINEL CODE_ANY sequence
  q = decxstr(p, p_start);
  assert(q == p - 3);
```

In this situation, we are starting at the end of a valid UC_SENTINEL sequence, so `decxstr` correctly jumps back to the start of the sequence.

I pushed an update that applies that fix and updates the tests, so you'll want to pull that locally and verify that it makes sense to you :grin:

mcdurdin (approved):
LGTM!

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.152-alpha](https://keyman.com/downloads/releases/alpha/15.0.152)

PR #5849, opened 2021-10-22
fix(android/app): Check KMP file exists before attempting to extract

### LINKED ISSUE #5848: bug(android): FileNotFoundException when extracting kmp file
Sentry Issue: [KEYMAN-ANDROID-92](https://sentry.io/organizations/keyman/issues/2735300110/?referrer=github_integration)

```
ErrnoException: open failed: ENOENT (No such file or directory)
    at com.tavultesoft.kmapro.PackageActivity.onCreate(PackageActivity.java:80)
...
(21 additional frame(s) were not displayed)

FileNotFoundException: /data/user/0/com.tavultesoft.kmapro/cache/basic_kbdsora.kmp: open failed: ENOENT (No such file or directory)
    at com.tavultesoft.kmapro.PackageActivity.onCreate(PackageActivity.java:80)
...
(16 additional frame(s) were not displayed)
```

darcywong00:
Fixes #5848 and will need to 🍒 pick to stable-14.0

This checks that the downloaded kmp file exists before attempting to extract it.

---

# DO NOT MERGE
**Note for reviewers, this contains test code in https://github.com/keymanapp/keyman/commit/6fb6238ea9f772bf6f5742a8bc8a1ac51f1472d6 
After user testing, I will revert the test code before merging to production**

---

## User Testing
Note: This will only pass while the test code above is in.
* **TEST_KMP_EXISTS** - Checks kmp file exists before PackageActivity() attempts to extract it
1. Load the PR build
2. From the "Settings" menu, search for the "basic_kbdsora" keyboard and download to install
3. Verify a Toast notification briefly appears saying "Failed to extract" (The kmp file no longer exists, so the PackageActivity finishes)

### DIFF
--- a/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/PackageActivity.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/PackageActivity.java
@@ -65,6 +65,11 @@ public class PackageActivity extends AppCompatActivity implements
     Bundle bundle = getIntent().getExtras();
     if (bundle != null) {
       kmpFile = new File(bundle.getString("kmpFile"));
+      if (!kmpFile.exists()) {
+        KMLog.LogError(TAG, kmpFile.getAbsolutePath() + " not found. Unable to extract");
+        showErrorToast(context, getString(R.string.failed_to_extract));
+        return;
+      }
       installMode = (KmpInstallMode) bundle.getSerializable("installMode");
       if(installMode == null) installMode = KmpInstallMode.Full;
       lastInstallMode = installMode;
@@ -230,6 +235,7 @@ public class PackageActivity extends AppCompatActivity implements
     setResult(1);
     cleanup();
     finish();
+    MainActivity.cleanupPackageInstall();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


mcdurdin:
This addresses the symptom but not the cause. Why is the .kmp file not found?

darcywong00:
> This addresses the symptom but not the cause. Why is the .kmp file not found?

It's hard for me to tell from the [Sentry breadcrumbs](https://sentry.io/organizations/keyman/issues/2735300110/?project=5983520&referrer=github_integration). There's quite a few app stop and restarts. Maybe PackageActivity is happening multiple times for the same kmp?

mcdurdin:
OK. It's hard to know if this will do much? It seems to be rare given there is only one report. I don't have a real problem with merging this but do wish we could isolate root cause, but given its rareness, don't spend any more time on it I think.

mcdurdin:
* TEST_KMP_EXISTS: PASSED

mcdurdin:
@darcywong00 looks like you merged this with test code still in it?

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.144-alpha](https://keyman.com/downloads/releases/alpha/15.0.144)

Issue #5777, opened 2021-09-30
feat(windows): Implement an API call in common core so that a platform engine can register preserved keys.🥑


rc-swag:
In Keyman for Windows a number of key combinations involving keys such ALT, ALTGR are not sent through for processing unless they are preserved. Preserved in this case meaning we register certain combinations to be sent through to for processing. 
To achieve this a new API call will need to be added to the common/core processor that will provide the ability to get an array of all the keys defined in the group rules of a keyboard. Then the Windows engine will be updated to use this list to register the preserved keys for the keyboard.

PR #5850, opened 2021-10-22
feat(common/core/desktop): Allow preserved key support

rc-swag:
Fixes: #5777

Add `km_kbp_keyboard_get_key_rules` to get a list of virtual keys and the shift modifiers that are used in the keyboard rules. 
The data structure for each item in the list is  
``` 
typedef struct {
  km_kbp_virtual_key key;   // The key the belongs
  uint32_t modifier_flag;   // modifier flag for the rule
} km_kbp_keyboard_key_rules;
```

There is more information in the `kmx` processor implmentation for a key. `Line`; `dpOutput`; `dpContext`;. These seemed to be processor specific, and they are coupled to `Group` again which is a `kmx` specific rule layout. I wasn't sure what would be useful to have in a list of "keys". So if I am after guidance if this data structure should contain more information in this public api call to make it a more useful API call.  The other option would be to make this an encode/decode pair with the structured data type which would then allow more info to be added to `km_kbp_keyboard_key_rules` without breaking previous implementations. That is it would be backward compatible.

# User Testing
All tests are for keyman for windows. 

**TEST_PRESERVEDKEYS** 
This is to test the key rules which contain <kbd>AltGr</kbd> or <kbd>Shift</kbd> + <kbd>AltGr</kbd> work correctly.

* Open LibreOffice or Notepad 
* Observere output of  ឱ​

* Observere output of ᧨


common/core/desktop/include/keyman/keyboardprocessor.h
```diff
@@ -659,6 +659,12 @@ typedef struct {
   km_kbp_option_item const * default_options;
 } km_kbp_keyboard_attrs;
 
+typedef struct {
+  km_kbp_virtual_key key;   // The key the belongs
+  uint32_t modifier_flag;   // modifier flag for the rule
```
mcdurdin, line 664:
```suggestion
  km_kbp_virtual_key key;
  uint32_t modifier_flag;
```

I'm not sure that the line comments add relevant info? 😁

common/core/desktop/include/keyman/keyboardprocessor.h
```diff
@@ -725,6 +731,34 @@ km_kbp_status
 km_kbp_keyboard_get_attrs(km_kbp_keyboard const *keyboard,
                           km_kbp_keyboard_attrs const **out);
 
+/*
+```
+### `km_kbp_keyboard_get_key_list`
+##### Description:
+Returns the list of keys that belong to the keyboard rules. The matching dispose
+call needs to be called to free the memory.
```
mcdurdin, line 739:
```suggestion
Returns the unordered full set of modifier+virtual keys that are handled by the
keyboard. The matching dispose call needs to be called to free the memory.
```

common/core/desktop/include/keyman/keyboardprocessor.h
```diff
@@ -725,6 +731,34 @@ km_kbp_status
...
+- `KM_KBP_STATUS_INVALID_ARGUMENT`: If non-optional parameters are null.
+##### Parameters:
+- __keyboard__: A pointer to the opaque keyboard object to be queried.
+- __out__: A pointer to the result:
+    A pointer to a `km_kbp_keyboard_get_key_list` structure.
+
+```c
+*/
+
+/* Get keyboard rules */
```
mcdurdin, line 751:
```suggestion
```

common/core/desktop/include/keyman/keyboardprocessor.h
```diff
@@ -725,6 +731,34 @@ km_kbp_status
...
+KMN_API
+km_kbp_status
+km_kbp_keyboard_get_key_list(km_kbp_keyboard const *keyboard,
+                          km_kbp_keyboard_key **out);
+
+
+
+void km_kbp_keyboard_key_list_dispose(km_kbp_keyboard_key *key_list);
```
mcdurdin, line 759: Need to document this function too 😉

common/core/desktop/include/keyman/keyboardprocessor.h
```diff
@@ -725,6 +731,34 @@ km_kbp_status
...
+    A pointer to a `km_kbp_keyboard_get_key_list` structure.
+
+```c
+*/
+
+/* Get keyboard rules */
+KMN_API
+km_kbp_status
+km_kbp_keyboard_get_key_list(km_kbp_keyboard const *keyboard,
+                          km_kbp_keyboard_key **out);
```
mcdurdin, line 755:
```suggestion
km_kbp_keyboard_get_key_list(km_kbp_keyboard const *keyboard,
                            km_kbp_keyboard_key **out);
```

common/core/desktop/src/km_kbp_keyboard_api.cpp
```diff
@@ -81,3 +81,20 @@ km_kbp_keyboard_get_attrs(km_kbp_keyboard const *keyboard,
   *out = &keyboard->keyboard();
   return KM_KBP_STATUS_OK;
 }
+
+km_kbp_status
+km_kbp_keyboard_get_key_list(km_kbp_keyboard const *keyboard,
+                          km_kbp_keyboard_key **out)
```
mcdurdin, line 87:
```suggestion
km_kbp_keyboard_get_key_list(km_kbp_keyboard const *keyboard,
                             km_kbp_keyboard_key **out)
```

common/core/desktop/src/kmx/kmx_processor.cpp
```diff
@@ -194,3 +194,40 @@ constexpr km_kbp_attr const engine_attrs = {
...
+    {
+      for(auto j = decltype(p_group->cxKeyArray){0}; j < p_group->cxKeyArray; j++)
+      {
+        // If we have a key rule for the key add it to the list
+        rules[n].key = p_group->dpKeyArray[j].Key;
+        rules[n].modifier_flag = p_group->dpKeyArray[j].ShiftFlags;
+        n++;
+      }
+    }
+  }
```
mcdurdin, line 229: This may return the same key combination multiple times from multiple groups. I think that may be a bit ugly -- we probably need to make sure that they are not repeated. Might be best to build that with a hash table (scanning for repeats naively would be `O(n*n)`)?

We do have a limited set of possible virtual key codes -- at least until we support touch layouts in the future which can define arbitrary virtual keys (codes > 256).

common/core/desktop/include/keyman/keyboardprocessor.h
```diff
@@ -725,6 +731,34 @@ km_kbp_status
...
+##### Description:
+Returns the list of keys that belong to the keyboard rules. The matching dispose
+call needs to be called to free the memory.
+##### Return status:
+- `KM_KBP_STATUS_OK`: On success.
+- `KM_KBP_STATUS_INVALID_ARGUMENT`: If non-optional parameters are null.
+##### Parameters:
+- __keyboard__: A pointer to the opaque keyboard object to be queried.
+- __out__: A pointer to the result:
+    A pointer to a `km_kbp_keyboard_get_key_list` structure.
```
mcdurdin, line 746:
```suggestion
- __out__: A pointer to an array of `km_kbp_keyboard_key` structures,
           terminated by `KM_KBP_KEYBOARD_KEY_LIST_END`.
```

common/core/desktop/src/mock/mock_processor.cpp
```diff
@@ -182,6 +182,9 @@ namespace km {
       return engine_attrs;
     }
 
+    km_kbp_keyboard_key  * mock_processor::get_key_list() const {
+      return nullptr;
```
mcdurdin, line 186: This returns nullptr but we don't define that as a possible output for the api function `km_kbp_get_key_list()` -- shouldn't we return an empty array instead?

common/core/desktop/src/rust/rust_mock_processor.cpp
```diff
@@ -74,6 +74,9 @@ namespace km {
       return engine_attrs;
     }
 
+    km_kbp_keyboard_key  * rust_mock_processor::get_key_list() const  {
+      return nullptr;
```
mcdurdin, line 78: This returns nullptr but we don't define that as a possible output for the api function `km_kbp_get_key_list()` -- shouldn't we return an empty array instead? (Same comment as mock_processor)

common/core/desktop/tests/unit/kmx/kmx_key_list.cpp
```diff
@@ -0,0 +1,77 @@
...
+  // Setup state, environment
+  try_status(km_kbp_state_create(test_kb, test_env_opts, &test_state));
+
+  try_status(km_kbp_keyboard_get_key_list(test_kb,&kb_key_list));
+
+  km_kbp_keyboard_key *key_rule_it = kb_key_list;
+  auto n = 0;
+  for (; key_rule_it->key; ++key_rule_it) {
+    ++n;
+  }
```
mcdurdin, line 49: Can we assert the values returned as well? I would also like to see a test for a more complex keyboard with multiple groups and with the same keys referred in multiple groups (to validate that we don't list keys more than once). We cannot assume the list order so will need to be careful on testing results!

windows/src/engine/keyman32/preservedkeymap.cpp
```diff
@@ -38,11 +38,26 @@ class PreservedKeyMap
 {
 public:
   BOOL MapKeyboard(KEYBOARD *pKeyboard, PreservedKey **pPreservedKeys, size_t *cPreservedKeys);
+  /**
+   * Updates a map of preserved keys (with GUID) that are used in keyboard rules.
+   * Passing in NULL for pPreservedKeys will cause cPreservedKeys to be set to the number
+   * PreserveKeys needed for the supplied pKeyboard, this should be used in creating pPreservedKeys
+   * List to sufficent size. When pPreseredKeys list is passed the cPreservedKeys will be the actual
+   * count of the number of unique pPreseredKeys
```
mcdurdin, line 46:
```suggestion
   * Updates a map of preserved keys (with GUID) that are used in keyboard rules.
   * Passing in NULL for pPreservedKeys will cause cPreservedKeys to be set to the number
   * of preserved keys needed for the supplied pKeyboard; this should be used in creating 
   * pPreservedKeys list to sufficient size. When pPreservedKeys list is passed the 
   * cPreservedKeys will be the actual count of the number of unique pPreservedKeys
```

windows/src/engine/keyman32/preservedkeymap.cpp
```diff
@@ -38,11 +38,26 @@ class PreservedKeyMap
...
+   *
+   * @param   pKeyboard  the keyboard for which the rules will be extracted from
+   * @param   pPreservedKeys  list of or PreservedKeys
```
mcdurdin, line 49:
```suggestion
   * @param   pPreservedKeys  preallocated array for preserved keys to be filled in, or nullptr to retrieve required size
```

windows/src/engine/keyman32/preservedkeymap.cpp
```diff
@@ -38,11 +38,26 @@ class PreservedKeyMap
...
+   * @param   cPreservedKeys  number of preserved keys in pPreservedKeys - or the size pPreservedKeys needs to be
+   * @return  BOOL  return TRUE if pPreveredKeys was updated
```
mcdurdin, line 51:
```suggestion
   * @return  BOOL  return TRUE on success
```

windows/src/engine/keyman32/preservedkeymap.cpp
```diff
@@ -318,13 +370,114 @@ BOOL PreservedKeyMap::MapKeyboard(KEYBOARD *pKeyboard, PreservedKey **pPreserved
...
+    }
+    // use api to get key rules
+    return pkm.MapKeyboardCore(_td->lpActiveKeyboard->lpCoreKeyboard, pPreservedKeys, cPreservedKeys);
+
+  } else {
+    if (!_td->lpActiveKeyboard->Keyboard) {
+      return FALSE;
+    }
+    return pkm.MapKeyboard(_td->lpActiveKeyboard->Keyboard, pPreservedKeys, cPreservedKeys);
+  }
```
mcdurdin, line 481: Do we want to add the issue number for future removal of this code and related functions?

mcdurdin (changes requested):

common/core/desktop/tests/unit/kmx/kmx_key_list.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+  KM_KBP_OPTIONS_END
+};
+
+int error_args() {
+    std::cerr << "kmx: Not enough arguments." << std::endl;
+    return 1;
+}
+
+km_kbp_keyboard_key kb_key_expected_list[] = {{49,16384},{50,16384},{66,16416},{67,16384},{67,16448},{97,0},{98,0}};
```
rc-swag, line 29: Should I do this test better and add a header comment that kmx.cpp test file can use, with all the other tests?
c keys: [LCTRL SHIFT K_A][SHIFT K_B][RCTRL SHIFT K_C][K_SPACE][SHIFT K_D][SHIFT K_E][SHIFT K_F]
c expected: \u0E01\u0E02\u0E03\u0020\u0E04\u0E05\u0E06
c context: 
c key_list:

mcdurdin: > Should I do this test better and add a header comment that kmx.cpp test file can use, with all the other tests?

I think it's probably overkill for now.

> My keyboard is totally bogus. I guess I should make a real keyboard.

The keyboard doesn't need to be sensible, just needs to exercise the functionality :grin:

mcdurdin:
Jenkins failure:

```
[2021-11-09T09:18:44.341Z] 55/57 key_list                                FAIL     0.01 s
[2021-11-09T09:18:44.341Z] 
[2021-11-09T09:18:44.341Z] --- command ---
[2021-11-09T09:18:44.341Z] /<<PKGBUILDDIR>>/common/core/desktop/build/arch/release/tests/unit/kmx/key_list /<<PKGBUILDDIR>>/common/core/desktop/build/arch/release/tests/unit/kmx
[2021-11-09T09:18:44.341Z] --- stderr ---
[2021-11-09T09:18:44.341Z] Test failed with 7 at ../../../tests/unit/kmx/kmx_key_list.cpp:46:
[2021-11-09T09:18:44.341Z]   km_kbp_keyboard_load(full_path.native().c_str(), &test_kb)
```

common/core/desktop/tests/unit/kmx/049 - groups ctrlalt.kmn
```diff
@@ -0,0 +1,28 @@
+c Description: Tests multiple groups. Purpose here is to interleave output and backspaces from kmx processor to consumer
```
ermshiperete, line 1: Looks like this test doesn't run (because it's not listed in `meson.build`). I see that it's used in `kmx_key_list.cpp`. If it's intended to be used only from that test than IMO it would be better and less confusing to rename it to `kmx_key_list.kmn` or something like that.

mcdurdin: Agreed

rc-swag: Yes, when I was first setting it up I as going to create a new `expected key list` comment to go in the header of the kmn so that it would test with `kmx.cpp`. I ended up with having it all testested from `kmx_key_list.cpp` I will rename this to the suggested name.

common/core/desktop/tests/unit/kmx/kmx_key_list.cpp
```diff
@@ -0,0 +1,86 @@
+/*
+  Copyright:    © 2018 SIL International.
+  Description:  Tests for key rules list generation in the kmx processor
+  Create Date:  30 Oct 2018
```
ermshiperete, line 4:
```suggestion
  Copyright:    © 2021 SIL International.
  Description:  Tests for key rules list generation in the kmx processor
  Create Date:  15 Nov 2021
```

common/core/desktop/tests/unit/kmx/kmx_key_list.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+  KM_KBP_OPTIONS_END
+};
+
+int error_args() {
+    std::cerr << "kmx: Not enough arguments." << std::endl;
+    return 1;
+}
+
+km_kbp_keyboard_key kb_key_expected_list[] = {{49,16384},{50,16384},{66,16416},{67,16384},{67,16448},{97,0},{98,0}};
```
ermshiperete, line 29: The test might be easier to understand if we'd use the `KM_KBP_VKEY_*` and `KM_KBP_MODIFIER_*` consts here, especially when looking at the .kmn file and trying to understand the expected list here.

windows/src/engine/keyman32/preservedkeymap.cpp
```diff
@@ -318,13 +371,114 @@ BOOL PreservedKeyMap::MapKeyboard(KEYBOARD *pKeyboard, PreservedKey **pPreserved
...
+    cKeys *= 2;
+  }
+
+  if (pPreservedKeys == NULL) {
+    *cPreservedKeys = cKeys;
+    km_kbp_keyboard_key_list_dispose(kb_key_list);
+    return TRUE;
+  }
+
+  if (*cPreservedKeys < cKeys) {
```
ermshiperete, line 420: Shouldn't we set `*cPreservedKeys = cKeys` here as well?

windows/src/engine/keyman32/preservedkeymap.cpp
```diff
@@ -318,13 +371,114 @@ BOOL PreservedKeyMap::MapKeyboard(KEYBOARD *pKeyboard, PreservedKey **pPreserved
...
+  if ((err_status != KM_KBP_STATUS_OK) || (kb_key_list ==nullptr)) {
+    return FALSE;
+  }
+
+  km_kbp_keyboard_key *key_rule_it = kb_key_list;
+  for (; key_rule_it->key; ++key_rule_it) {
+    ++cRules;
+  }
+  cKeys = cRules;
+  if (cKeys == 0) {
```
ermshiperete, line 404: Shouldn't we set `*cPreservedKeys = cKeys` here as well?

ermshiperete (commented):
Looks like `049 - groups ctrlalt.kmx` is still present!?

windows/src/engine/keyman32/preservedkeymap.cpp
rc-swag: If we get to this if statement it means the size of the array of PreservedKeys `pPreservedKeys` is might going to be big enough to store all the keys.  So it returns False.

When `::MapKeyboardCore` is called if `pPreservedKeys` is NULL, we set *cPreservedKeys to the size pPreservedKeys  needs to be. 

So in the other cases we don't set `cPreservedKeys`

windows/src/engine/keyman32/preservedkeymap.cpp
rc-swag: I guess we could check if `pPreservedKeys == NULL` and set `*cPreservedKeys = cKeys` basically setting it to zero. but in this case we are returning FALSE anyway.  So if the function is returning FALSE for failure, I don't think the `*cPreservedKeys` should be trusted.
This also mimics the behaviour of the current `::MapCore`

mcdurdin (approved):
Changes LGTM

MakaraSok:
* **TEST_PRESERVEDKEYS: PASSED** The characters are output per instructions given, see the screenshot below.
image

rc-swag:
@mcdurdin  The comment @ermshiperete  made about using constants instead of magic numbers in the `kmx_list.cpp` raised an issue. I now realise what I am putting in the `km_kbp_keyboard_key` is not a "virtual key" but kmx `KEY` that can either be a "Virtual Key" or a "Character Code" determined by the modifier flag  `KM_KBP_MODIFIER_VIRTUALKEY`.  So we have a kmx engine detail in the keyboard_processor API. This is not good.  I then spent some time thinking how I could change this but I wasn't sure what we should do?

mcdurdin:
> I now realise what I am putting in the `km_kbp_keyboard_key` is not a "virtual key" but kmx `KEY` that can either be a "Virtual Key" or a "Character Code" determined by the modifier flag `KM_KBP_MODIFIER_VIRTUALKEY`. So we have a kmx ~~engine~~ _processor_ detail in the ~~keyboard_processor~~ _Keyman Core_ API. 

Two possible solutions. I lean towards the first.

## Translate the characters to US vkeys before returning from kmx processor

This is probably the cleanest way to solve the issue. `MapUSCharToVK` is defined in preservedkeymap.cpp and CharToKeyConversion.cpp (which we should consolidate...), so that'll need to be ported over to kmx processor.

```c++
        v_key = p_group->dpKeyArray[j].Key;
        modifier_flag = p_group->dpKeyArray[j].ShiftFlags;
        if(modifier_flag == 0) {
          if(!MapUSCharToVK(&v_key, &modifier_flag)) continue;
        }
  
        map_rules[std::make_pair(v_key,modifier_flag)] = modifier_flag;
```

## Live with it

This alternative is also okay, although not great.

We have many kmx processor details that inform the design of the Keyman Core API. The fact that the kmx processor can report chars as well as vkeys should be documented, not ignored :-).

However, let's split out the char pattern as a separate bit property rather than encoding it into the `modifier_flag`:

```
typedef struct {
  
  km_kbp_virtual_key key;   /// a virtual key, or an ASCII character if `is_character` is non-zero.
  uint16_t modifier_state;  /// bitmask from `km_kbp_modifier_state` enum 
  uint8_t is_character;     /// if non-zero, `key` is an ascii character corresponding to
                            /// the US English key cap (shifted or unshifted), and
                            /// `modifier_state` is ignored.
} km_kbp_keyboard_key;
```

(Noted also, use `modifier_state` not `modifier_flag`, and it should be `uint16_t` not `uint32_t`!)

That hides the worst of the implementation details (a `modifier_state` that included the `KM_KBP_MODIFIER_VIRTUALKEY` flag would differs from all other uses of modifiers in the API -- ignoring debug api which is a bit special anyway). If other engines don't need to list chars in this API, then that's fine; the `is_character` field will be `0` in those cases.

mcdurdin:
Per [discussion](https://github.com/keymanapp/keyman/pull/5850#issuecomment-969854135), moving back to draft to eliminate use of characters in place of virtual keys outside the boundaries of kmx processor.

rc-swag:
> Per [discussion](https://github.com/keymanapp/keyman/pull/5850#issuecomment-969854135), moving back to draft to eliminate use of characters in place of virtual keys outside the boundaries of kmx processor.

This work has now been done and ready for re-review.

common/core/desktop/tests/unit/kmx/049 - groups ctrlalt.kmn
rc-swag: @ermshiperete  I fixed the what I thought where the issues with kmx_key_list and the meson.build, but the jenkins build is still failing. Can you also add my user to the build so I can view it in Jenkins?

mcdurdin:
Jenkins error:

```
[2021-11-23T04:36:16.697Z] ../../../common/core/desktop/src/kmx/ibus_keyman_tests-kmx_processor.o: In function `km::kbp::kmx_processor::get_key_list() const':

[2021-11-23T04:36:16.697Z] ./linux/ibus-keyman/tests/../../../common/core/desktop/src/kmx/kmx_processor.cpp:221: undefined reference to `km::kbp::kmx::MapUSCharToVK(unsigned short, unsigned short*, unsigned int*)'

[2021-11-23T04:36:16.697Z] collect2: error: ld returned 1 exit status
```

ermshiperete:
> It's a shame we are not using meson here!

Indeed. I have it on my to-do list though :smile: - probably should create an issue for it: #5981

keyman-server:

PR #5854, opened 2021-10-26
fix(web): Allow touch-alias elements to clear text selection

### LINKED ISSUE #5853: bug(web): Highlight and replace/delete is not working
@katelem24 reports on [community site](https://community.software.sil.org/t/feedback-on-keyman-15-0-alpha/5405/24)

> Highlight and replace/delete is not working.
When text is highlighted, whatever action is applied to the highlighted text affects only the last letter in the highlight. So, to delete a paragraph, you have to delete it one letter at a time or switch to another keyboard that can highlight and delete all at once.

For Keyman web, any action taken only affects the character at where the insertion point was before the selection/highlight.
>
>I think it also affects the current released version on Android (I can’t tell, I’ve since been using the 15-alpha versions).

darcywong00:
Fixes #5853 

Original issue was reported on Keyman for Android (hence the android scope on the branch name), but the underlying behavior seems to be from KeymanWeb engine.

When the user selects text and presses a key, the expectation is for the selected text to be replaced by the processed keystroke.
Currently when selected text is passed to KeymanWeb, the output includes both selected text and keystroke. Effectively, the "selected text" remains.

If the user selects text beyond the working text area, I don't think we want to do any deletion/replacement.

## User Testing
For this test, load the web/testing/prediction-mtnt/ test page (includes English suggestion banner with English keyboard). Use Chrome browser in mobile viewport with the Developer console open (F12)
Ignore console exceptions about `Uncaught TypeError` since they're resolved by #5873

* **TEST_WEB_MOBILE** - Test text selection and replacement
1. In the default text area, type a few sentences
2. Click on the test page outside of all the text areas/ text fields
3. Hold down <kbd>shift</kbd> and click the text area to select enough text so the entire text selection includes both test page text and text from the text area
4. Type a key
5. Verify the selected text is **not** replaced
6. Verify the typed key is inserted into the text area
7. Click in the original text area
7. Hold down <kbd>shift</kbd> and click the text area to select enough text so the entire text selection is in the text area
8. Type a key
9. Verify the selected text is replaced with the typed key

* **TEST_WEB_SUGGESTION** - Test text selection and suggestions
1. In the default text area, type a few sentences
2. Click on the test page outside of all the text areas/text fields
3. Hold down <kbd>shift</kbd> and click the text area to select enough text so the entire text selection includes both test page text and text from the text area
4. Click on a suggestion
5. Verify the word at the cursor is replaced with the suggested word
6. Click in the original text area
7. Hold down <kbd>shift</kbd> and click the text area to select enough text so the entire text selection is in the text area
8. Click on a suggestion
9. Verify the selected text is replaced with the suggestion


web/source/dom/targets/touchAlias.ts
```diff
@@ -25,15 +31,29 @@ namespace com.keyman.dom.targets {
...
 
     getDeadkeyCaret(): number {
       return this.root.getTextCaret();
     }
 
     getTextBeforeCaret(): string {
+      if (this.hasSelection()) {
+        // If there's selected text, use nul for context
+        return '';
+      }
```
mcdurdin, line 56: This change needs to be elsewhere I think -- there are multiple `getTextBeforeCaret()` functions and they all need to be handling this situation.

Note that contentEditable.ts has the exact opposite of this test.

I think `hasSelection()` is returning true if the caret is in the text area even if there is no actual selection. But I would have to dig deeper to be sure.

mcdurdin (changes requested):
I think there is sadly a fair bit more work we need to do on this for web -- the `hasSelection` function is a bit funny -- it seems to be conflating `hasFocus()` with `hasSelection()`, for example see contentEditable.ts?

mcdurdin:
FWIW, we don't currently support selection on touch alias elements in KeymanWeb ... the caret does not track the document selection. So I think we can abandon this until such time as we have a full selection model on web touch alias.

darcywong00:
> So I think we can abandon this until such time as we have a full selection model on web touch alias.

Too bad. So from the original issue reported on the Android app:
> Keyman Engine for Android should pass selected text as null context to KeymanWeb engine

Should that still be implemented on a separate PR?

mcdurdin:
> > Keyman Engine for Android should pass selected text as null context to KeymanWeb engine
> 
> Should that still be implemented on a separate PR?

Yes, but let's review what the status is for all apps on this:

* Keyman for Android
* Keyman for iOS
* KeymanWeb (non-touch mode)
* Keyman for Windows
* Keyman for Linux
* Keyman for macOS

We may need to check multiple modes in Windows, macOS as well.

darcywong00:
I'll abandon this PR. Further investigations to be added to #5853

PR #5859, opened 2021-10-28
chore(linux): Set correct version number on ibus component

ermshiperete:
@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.138-alpha](https://keyman.com/downloads/releases/alpha/15.0.138)

PR #5860, opened 2021-10-28
chore(common/core): generate kmp.json

ermshiperete:
This change generates kmp.json based on the test files specified in meson.build. It will also generate .kmx files in the build directory and copies the .kmn files to it as well.

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.138-alpha](https://keyman.com/downloads/releases/alpha/15.0.138)

PR #5879, opened 2021-11-03
feat(developer): kmcompx - remove Windows dependencies 🍃

SabineSIL:
[Kmcompx_FirstIdeas.docx](https://github.com/keymanapp/keyman/files/7466818/Kmcompx_FirstIdeas.docx)

this is what I am about to do for the kmcompx-project:

My idea was to create a new .h-file (kmx_kmcomp_inc.h) where I include everything that is necessary for changing variable types and functions. This file then would be included in Compiler.cpp and all the other files where changes need to be made in VarTypes and functions. 

image

image

I am right at the beginning and have just included a typedef for KMX_DWORD in my new kmx_kmcomp_inc.h. In Compiler.cpp I started to change some return-variables of a few functions. 
Before going further I just want to make sure that I understood right and am on the right track. Therefore I have some questions:

* You wrote:The list of source files that you will need to work with will be found in the meson.build files in that pull request, under developer/kmcompx. (kmcompx is a 'working title' for this project!). 

  As I understand **the files found in developer/kmcompx/src/meson.build are the ones to change?** 

  image

* **Is it OK to create a seperate .h-file for kmcompx?** 
* **Is there any naming convention for a .h-file in keyman?** (Is there a more appropriate name for kmx_kmcomp_inc.h ?) 
* At the moment kmx_kmcomp_inc.h is located at the same location as kmx_base.h: \common\core\desktop\src\kmx. **Is there a better location?**

### DIFF
--- a/common/core/desktop/src/kmx/kmx_kmcomp_inc.h
+++ b/common/core/desktop/src/kmx/kmx_kmcomp_inc.h
@@ -1,0 +1,8 @@
+#pragma once
+#include <stdint.h>
+
+
+typedef uint32_t            KMX_DWORD;
+//typedef int32_t             KMX_BOOL;
+//typedef uint8_t             KMX_BYTE;
+//typedef uint16_t            KMX_WORD;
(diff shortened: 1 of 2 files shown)


keyman-server:

SabineSIL:
replaced by new one #6026

PR #5881, opened 2021-11-03
chore(linux): Add integration tests for ibus-keyman 🎫

ermshiperete:
This change adds integration/regression tests for ibus-keyman. It re-uses the unit tests that we have for core.

Also modifies one test so that the context doesn't use characters used in the keyboard. This makes it easier to setup in ibus-keyman tests.

Note that not all tests are yet passing.

@keymanapp-test-bot skip


mcdurdin (approved):
I think LGTM -- there's a lot of new code there but it should be okay I hope. Will these tests run as part of our CI?

ermshiperete:
> I think LGTM -- there's a lot of new code there but it should be okay I hope. Will these tests run as part of our CI?

Eventually yes, that's the plan.

darcywong00:
What's the distinction between `/linux/ibus-keyman/test` and `/linux/ibus-keyman/tests`?

ermshiperete:
> What's the distinction between `/linux/ibus-keyman/test` and `/linux/ibus-keyman/tests`?

`/linux/ibus-keyman/test` is an older attempt that doesn't work reliably. I'll delete this directory in a follow-up PR.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.141-alpha](https://keyman.com/downloads/releases/alpha/15.0.141)

PR #5883, opened 2021-11-05
chore(linux): Update method comments

ermshiperete:
@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/keymanutil.c
+++ b/linux/ibus-keyman/src/keymanutil.c
@@ -400,7 +400,7 @@ keyman_put_options_todconf(gchar *package_id,
     gchar *needle = g_strdup_printf("%s=", option_key);
     gchar *kvp = g_strdup_printf("%s=%s", option_key, option_value);
 
-    if (*options != NULL)
+    if (options != NULL)
     {
         int index = 0;
         gboolean option_updated = FALSE;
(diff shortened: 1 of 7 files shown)


linux/ibus-keyman/src/keymanutil.c
```diff
@@ -369,6 +368,7 @@ keyman_get_options_queue_fromdconf(gchar *package_id,
             }
             index++;
         }
+        g_strfreev(options);
```
mcdurdin, line 371: I note on line 352 you have: 

```
    if (*options != NULL)
```

But `keyman_get_options_fromdconf` can return `NULL` so we should be testing for that, not `*options` surely.

mcdurdin (changes requested):

linux/ibus-keyman/src/keymanutil.c
ermshiperete: Thanks for noticing. Done.

mcdurdin (commented):
Yep, that looks better. Saw you fixed another location.

Okay, so now I am reading `keyman_put_options_todconf` and it looks like the reallocs in that are conflating an array-of-pointers with the array-containing-strings (which does not really exist in C++). See for example lines 422, 430. It's probably going to "work" much of the time, assuming `kvp` is longer than the existing options array size... but it's certainly not right and would cause grief sometimes...

ermshiperete:
:+1: You're right, this doesn't work - it looked suspicious but since it's always been that way... :smile: I added unit tests and that showed that it doesn't work if it has to reallocate memory. Should be fixed now :smile:.

mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.148-alpha](https://keyman.com/downloads/releases/alpha/15.0.148)

PR #5889, opened 2021-11-08
chore(linux): Run ibus-keyman tests as part of the build 🎫

ermshiperete:
`glib-tap.mk`, `tap-driver.sh` and `tap-test` are copied from (an older version of) glib. (The current version of glib uses meson instead of automake and so deleted these files, although they are still mentioned in the glib docs...)

@keymanapp-test-bot skip


linux/ibus-keyman/.gitignore
```diff
@@ -6,6 +6,7 @@ src/ibus-engine-keyman
 src/print_kmp
 src/print_kmpdetails
 src/test/keymanutil_tests
+src/test/keymanutil-tests
```
mcdurdin, line 9: Do we have both of these folders? That seems a bit confusing!

mcdurdin (approved):
LGTM. I am a little hesitant about running interactive-style tests on a build agent... (we don't do that for other platforms except web, and that's been a hassle for a while!)

linux/ibus-keyman/.gitignore
ermshiperete: Note that these are the test executables. I thought that I'd create both, depending on how we build, but it seems that was a left-over.

ermshiperete (commented):
> I am a little hesitant about running interactive-style tests on a build agent... (we don't do that for other platforms except web, and that's been a hassle for a while!)

I don't particularly like it either. It's fragile, and it significantly increases build time.

What do you suggest? Not run it during a build? Or leave it on for now, see how it goes and disable it if it causes to much failures?

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.154-alpha](https://keyman.com/downloads/releases/alpha/15.0.154)

mcdurdin:
> I don't particularly like it either. It's fragile, and it significantly increases build time.
> 
> What do you suggest? Not run it during a build? Or leave it on for now, see how it goes and disable it if it causes to much failures?

I wonder. Can run them in the Test build configuration but not the Release build configuration? That way we don't accidentally break the release builds; test builds are less of a problem.

ermshiperete:
> I wonder. Can run them in the Test build configuration but not the Release build configuration? That way we don't accidentally break the release builds; test builds are less of a problem.

Good idea. I changed the TC config so that for builds on `master` or `stable*` we keep what we did previously, for PRs we run the tests. We'll see how this goes. 

#5975 adds the missing Makefile targets to the `stable-14.0` branch so that PR builds will work there as well (although they won't run any ibus-keyman tests, but it simplifies TC config if we can treat all PRs the same).

mcdurdin:
> Good idea. I changed the TC config so that for builds on `master` or `stable*` we keep what we did previously, for PRs we run the tests. We'll see how this goes.

I don't think this is necessary? We have two TC configurations: Release-14.0 and Test-14.0. I would like all the tests to run on all branches in Test-14.0, but only basic tests in Release-14.0.

The reason is that we run Test-14.0 on `master` and `stable-*` every time we merge a branch into these, as an additional verification that merged changes don't cause a problem. So when we come to a nightly release build in `Release-14.0`, we can be reasonably confident that the changes have already been through a full unit test run on that branch, without having to worry about fragility breaking the release build.

ermshiperete:
OK, I adjusted the TC configs: Release-14.0 won't run the new tests, Test-14.0 will always run them.

PR #5894, opened 2021-11-11
feat(developer): support for U_xxxx_yyyy

### LINKED ISSUE #2858: feat(developer): allow `U_xxxx_yyyy` to output two Unicode characters
If a key on a touch layout needs to output two Unicode characters, the designer must assign a T_xxxx label, then use the .kmn file to cause that T_xxxx code to output the two Unicode characters.

Just as using U_zzzz on a touch layout produces U+zzzz, using U_xxxx_yyyy would produce U+xxxx and U+yyyy.

The current use of T_ codes and code in the .kmn file.

**Additional context**
Would need to consider how to handle non-BMP characters.
Once you allow two characters, someone is sure to want three, etc.

mcdurdin:
Fixes #2858.

This feature requires KeymanWeb 15.0.

Adds support for `U_xxxx_yyyy_....` identifiers so that you can emit more than a single Unicode character with touch keys without needing kmn code. For example, to emit U+1000 U+1001 with a single keystroke, you'd need to call the key, e.g. `T_1000_1001` and include a rule `+ [T_1000_1001] > U+1000 U+1001`. Now you can do the same thing without a rule for it.

See https://github.com/keymanapp/help.keyman.com/pull/487 for documentation.

# User Testing

Test in Keyman Developer, KeymanWeb test, for iPhone or iPad.

* **TEST_U_xxxx_yyyy:** The pull request includes a keyboard which you can load into Keyman Developer and test with KeymanWeb to validate that multi-character outputs for the touch keyboard are generated correctly.


rc-swag (approved):
lgtm

rc-swag:
**TEST_U_xxxx_yyyy**: **PASSED**

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.146-alpha](https://keyman.com/downloads/releases/alpha/15.0.146)

PR #5895, opened 2021-11-11
feat(developer): validate .kps files during compile

### LINKED ISSUE #5464: feat(developer): kmcomp should validate .kps file format

If a .kps file includes a typo, such as the example below, the compiler should complain about it: 
 https://github.com/keymanapp/keyboards/blob/a208c7376054ebb9b8ef53c46142f7a6af57e02a/release/b/balochi_phonetic/source/balochi_phonetic.kps#L60-L71 

This caused the font to be ignored during the compile, which is not desired.

There does not appear to be an existing DTD for .kps files.

---

**Keyman Developer:**
 - OS: any 
 - Keyman Developer version: 15.0

mcdurdin:
Fixes #5464.

The keyboards repository already has code to validate .xsd files, using xmllint. This adds the same functionality directly into kmcomp and TIKE.

# User Testing

* TEST_SCHEMA: Verify that existing .kps files compile correctly -- choose at random!
* TEST_DODGY_SCHEMA: Manually edit a .kps file in a text editor to introduce an unsupported XML element, for example, add a `<Foo />` element somewhere. The compile of this file should fail with a "helpful" error message.

### DIFF
--- a/common/schemas/kps/README.md
+++ b/common/schemas/kps/README.md
@@ -1,0 +1,6 @@
+# kps.xsd
+
+Master version: https://github.com/keymanapp/api.keyman.com/blob/master/schemas/kps/7.0/kps.xsd
+
+## 2021-07-19 7.0
+* Initial version 7.0
--- a/common/schemas/kps/kps.xsd
+++ b/common/schemas/kps/kps.xsd
@@ -1,0 +1,213 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  .kps file schema, version 7.0
+
+  Copyright (C) SIL International
+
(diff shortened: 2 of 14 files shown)


rc-swag:
**TEST_SCHEMA**: **PASSED**
**TEST_DODGY_SCHEMA **: **PASSED**
Works well on the command line. For Developer, an unknown schema tag just silently gets removed and compiling continues. 
Developer fails to load if an invalid kps file is loaded at startup.

rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.145-alpha](https://keyman.com/downloads/releases/alpha/15.0.145)

Issue #5904, opened 2021-11-14
bug(web): 15.0 alpha not loading on keymanweb.com


mcdurdin:
**Describe the bug**

Loading https://keymanweb.com/?tier=alpha# fails.

```
kmwkeyboards.ts:1227 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'resolve')
    at b.register (kmwkeyboards.ts:1227)
    at e.register (kmwbase.ts:375)
    at addKeyboards (keyboards.php?tier=alpha&version=15.0:3)
    at ?tier=alpha:159
b.register @ kmwkeyboards.ts:1227
e.register @ kmwbase.ts:375
addKeyboards @ keyboards.php?tier=alpha&version=15.0:3
(anonymous) @ ?tier=alpha:159
Promise.then (async)
(anonymous) @ ?tier=alpha:157
```

---

**KeymanWeb:**
 - OS: Windows 10 
 - Browser: Chrome 95 
 - KeymanWeb Version: 15.0.146 
 - URL: https://keymanweb.com/?tier=alpha#

PR #5908, opened 2021-11-15
fix(web): avoid error if timerid not valid

mcdurdin:
Relates to #5904.

If a call to api.keyman.com takes longer than 10 seconds, the timerid would no longer be valid, even if the call returns successfully. This threw a cascade of errors.

There was a secondary issue where keymanweb.com was passing in an invalid timerid, causing the same error cascade. This is fixed in keymanapp/keymanweb.com#38.

@keymanapp-test-bot skip

We'll do user testing on keymanapp/keymanweb.com#38 when it merges. This particular PR is not particularly easy to user-test, although you could force it with a breakpoint and delay in KeymanWeb unminified test page if you were really keen.

### DIFF
--- a/web/source/keyboards/kmwkeyboards.ts
+++ b/web/source/keyboards/kmwkeyboards.ts
@@ -73,7 +73,7 @@ namespace com.keyman.keyboards {
 
     /**
      * Utility to convert stubs to KeyboardStub[]
-     * @param arg 
+     * @param arg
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.147-alpha](https://keyman.com/downloads/releases/alpha/15.0.147)

PR #5922, opened 2021-11-17
fix(android/engine): Don't pass selected text to KeymanWeb

darcywong00:
Fixes Android portion of #5853 and replaces #5854
Because KeymanWeb currently doesn't support selection on touch alias elements, this updates Keyman Engine for Android to not pass the selected context to KeymanWeb engine. This way, Keyman Engine for Android handles the deletion of selected text and inserts the typed string per KeymanWeb.

### User Testing
Setup: Load test build of Keyman for Android and set Keyman as a system keyboard
Can also install another keyboard (e.g. khmer_angkor)
Note: if you click an available suggestion to replace selected text, this is still buggy and beyond the scope of this PR.

* **TEST_INAPP_KEYBOARD** - Tests how selected text is handled in the Keyman app
1. Start Keyman
2. Type some text
3. Drag and select some of the typed text
4. Click a key/longpress key
5. Verify the selected text is replaced with the pressed key.

* **TEST_SYSTEM_KEYBOARD** - Tests how selected text is handled in the Keyman system keyboard
1. Launch a separate app where you can type in a text field / text area (e.g. Contacts). Don't use Chrome omnibar as it may apply its own auto-corrections
2. With the Keyman keyboard, type some text
3. Drag and select some of the typed text
4. Click a key/longpress key
5. Verify the selected text is replaced with the pressed key.

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
@@ -1889,9 +1889,12 @@ public final class KMManager {
 
   public static boolean updateSelectionRange(KeyboardType kbType, int selStart, int selEnd) {
     boolean result = false;
+    // Currently, KMW doesn't support selection on touch alias elements
+    // so pass null context to KMW. See issue #5853
+    int selectionEnd = Math.max(selStart, selEnd);
     if (kbType == KeyboardType.KEYBOARD_TYPE_INAPP) {
       if (InAppKeyboard != null && InAppKeyboardLoaded && !InAppKeyboardShouldIgnoreSelectionChange) {
-        InAppKeyboard.loadJavascript(KMString.format("updateKMSelectionRange(%d,%d)", selStart, selEnd));
+        InAppKeyboard.loadJavascript(KMString.format("updateKMSelectionRange(%d,%d)", selectionEnd, selectionEnd));
         result = true;
       }
 
@@ -1906,7 +1909,7 @@ public final class KMManager {
           }
         }
 
-        SystemKeyboard.loadJavascript(KMString.format("updateKMSelectionRange(%d,%d)", selStart, selEnd));
+        SystemKeyboard.loadJavascript(KMString.format("updateKMSelectionRange(%d,%d)", selectionEnd, selectionEnd));
         result = true;
       }
 


android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -1888,9 +1888,12 @@ public static boolean updateText(KeyboardType kbType, String text) {
 
   public static boolean updateSelectionRange(KeyboardType kbType, int selStart, int selEnd) {
     boolean result = false;
+    // Currently, KMW doesn't support selection on touch alias elements
```
mcdurdin, line 1891: Touch alias elements are not relevant to Keyman for Android -- they are only used in KeymanWeb in a browser.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
```diff
@@ -1888,9 +1888,12 @@ public static boolean updateText(KeyboardType kbType, String text) {
...
+    // so pass null context to KMW. See issue #5853
```
mcdurdin, line 1892: Does this actually pass in null context, or just no selection? To test this, we need a keyboard with a context rule, for example:

```
'a' + 'b' > 'x'
```

Then, have the text `123a` selected, and type <kbd>b</kbd>. The selection should be deleted, and `b` emitted -- not `x`.

mcdurdin (changes requested):

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMManager.java
darcywong00: Ah, it's just passing no selection, so KMW still has the context.

Testing with xinaliq keyboard and the selection isn't correct...

How should a null context be passed to KMW then?

mcdurdin: Yes, that should do it, I think.

darcywong00: Maybe it's moot calling `updateKMSelectionRange()` afterwards.

The app seems to behave with only `updateKMText()`

darcywong00:
Going back to Draft to characterize the following

https://github.com/keymanapp/keyman/issues/5853#issuecomment-972529787
> More notes. We'll want to check all Keyman platforms to verify these 3 scenarios:
> 
> 1. When text is selected, an inserted character should replace the selection
> 2. When text is selected, backspace should only delete the selection
> 3. When text is selected, treat context as empty for keyboard rules (intentionally ignore text before selection too, since that might be pretty far and confusing to consider)

darcywong00:
Superseded by #6274

Issue #5944, opened 2021-11-18
bug(web): sil_euro_latin osk disappears on 2nd click


mcdurdin:
**Describe the bug**

On a desktop browser, clicking on any "key" in the sil_euro_latin OSK causes the OSK to disappear -- but only on the 2nd click. The first click works fine.

![image](https://user-images.githubusercontent.com/4498365/142512083-fb98b9d3-93cc-442b-aff4-44953902731f.png)

This issue is present in 13.0-stable through 15.0-alpha. I have not tested earlier versions. It does not seem to repro on keymanweb.com but does repro on http://localhost/keymanweb/testing/unminified.html. This suggests there may be interactions between the OSK and the keyboard selector UI?

I thought at first that it was related to an issue within the `setFocusTimer()` function that is referenced in #5909, but the issue arises whether or not the `setFocusTimer()` functionality is present.

I resolved the error arising in #5939 locally and that did not address this problem.

I note that the `keymanweb.KSF()` (aka `saveFocus()`) function has no effect. This is probably the root cause, and I am guessing it is probably related to earlier refactors.

**To Reproduce**

1. Go to http://localhost/keymanweb/testing/unminified.html
2. Select Swedish
3. Click on a letter "key" on the sil_euro_latin OSK.
4. Click on the same letter key or any other letter key again.
5. Observe the OSK disappearing (as focus is lost)

**Expected behavior**

Well, the OSK shouldn't disappear...

---

**KeymanWeb:**
 - KeymanWeb Version: observed in 13.0-stable - 15.0-alpha

PR #5946, opened 2021-11-19
fix(web): shorten setFocusTimer()

### LINKED ISSUE #5909: bug(web): clicking a control and then outside a control within 1 second does not hide the touch OSK
**Describe the bug**

If you focus a control and then rapidly click outside any edit control (within 1 second) to blur, the blur does not happen and the touch OSK does not get dismissed.

This appears to be an interaction inside the inter-control switching code, where there is a focus timer that expires 1 second after a focus event:

https://github.com/keymanapp/keyman/blob/09f4aed77e3e9076e7da10e1f041c9e4724d3f1b/web/source/dom/domEventHandlers.ts#L30-L37

---

**KeymanWeb:**
 - OS: any 
 - Browser: mobile browsers at least? 
 - KeymanWeb Version: 15.0.146-alpha

mcdurdin:
Fixes #5909.

The `setFocusTimer()` delay was 1000ms. However, it seems that a much shorter delay accomplishes the same outcome for switching between controls, as the important factor is to block asynchronously-arriving events from cancelling the OSK / focus controls for touch alias elements.

I have set the delay to 50ms. This is fast enough that a click in a control followed by a click on the page area will always cancel the OSK, but gives the space needed for queued focus/blur events have time to be processed before cancelling the timer.

# User Testing

This is really a 'regression' test to make sure that the OSK continues to function predictably.

* TEST_DESKTOP: load the unminified test page on a desktop browser, and using both mouse and keyboard, switch between input fields, interact with the OSK and make sure that the OSK stays visible when it should (when a KMW-enabled text field is focused), and is hidden when no text field is focused. (Take note of #5944 if you are using sil_euro_latin for any tests and don't fail on that basis.)

* TEST_MOBILE: load the unminified test page on a mobile device, and using touch, switch between input fields, interact with the OSK and make sure that the OSK stays visible when it should (when a KMW-enabled text field is focused), and is hidden when no text field is focused. Make sure that using a hardware keyboard also works correctly.

* TEST_BLUR: On a mobile device, try focusing a text field by touching it, and then rapidly touching a blank area. Observe the OSK disappearing. Do this a few times to make sure that the OSK always disappears as expected.


web/source/dom/domEventHandlers.ts
```diff
@@ -32,7 +32,7 @@ namespace com.keyman.dom {
 
       this.focusTimer = window.setTimeout(function() {
         this.focusing=false;
-      }.bind(this), 1000)
+      }.bind(this), 50)
```
darcywong00, line 35: Would it be better to define a constant?

darcywong00 (approved):
lgtm

web/source/dom/domEventHandlers.ts
mcdurdin: In theory yes but not sure that in practice it makes much difference here 😁

MakaraSok:
* **TEST_DESKTOP: PASSED** Nothing strange is experienced.
* **TEST_MOBILE: PASSED** The OSK displays and works correctly. The hardware keyboard can also output text in the text area.
* **TEST_BLUR: PASSED** There seems to a very tiny brief delay between the click and the disappearance of the OSK. Very tiny, so this is a PASSED.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.155-alpha](https://keyman.com/downloads/releases/alpha/15.0.155)

PR #5958, opened 2021-11-21
fix(web): improve keyboard switch performance

mcdurdin:
Summary of these changes:

Two targeted improvements; more optimization is possible but IMO this is a significant enough improvement to go through test:

1. In oskView.ts, avoid calling `VisualKeyboard.refreshLayout` twice from `OSKView.refreshLayout`. This roughly halves the time spent in `VisualKeyboard.refreshLayout` for a keyboard switch.
2. In modelManager.ts, we avoid a very expensive `unloadModel` / `loadModel` sequence from `registerModel` by checking to see if the model spec we are passed is already registered. This avoids an expensive callback to Keyman for Android which causes the banner to 'bounce'.

Longer notes:

So `VisualKeyboard.refreshLayout` is called 20+ times during a keyboard switch. It is a very expensive call. It is the bulk of the time spent when you press the globe key in Keyman for Android. This seems ripe for optimization; see the following functions in keyboard.html:

* `setKeymanLanguage` -> `setActiveKeyboard` -> 6 calls
* `enableSuggestions` -> `registerModel` -> 8 calls
* `stateChange` -> 6+ calls

`OSKView.refreshLayout` always seems to call `VisualKeyboard.refreshLayout` twice: first with `setSize`, and then again itself.

Keyman for Android and KeymanWeb seem to fight over who is responsible for selecting models (KMW first calls `loadModel`, then Keyman for Android calls `unloadModel`, `loadModel`, through `enableSuggestions`).

The fix for rotation (`correctOSKTextSize()`), as called from `stateChange()`, is expensive as well, because it rebuilds the entire keyboard from scratch.

Along with that, we have the size of the webview changing as the model is unloaded and reloaded and lots of back-and-forth between the KeymanWeb and KMEA which seems unnecessary.

Shouldn't KeymanWeb be responsible for loading and unloading models, once they are all registered? We should only need to register the models once at page load time (when config changes, do we just reload the page?).

# User Testing

* **TEST_ANDROID:** Test keyboard switching in Keyman for Android, along with rotation and basic usage of the keyboard. Confirm that no visual glitches arise. Make sure you use keyboards with and without models. Emulator is okay.

* **TEST_IOS:** Test keyboard switching in Keyman for iPhone and iPad, along with rotation and basic usage of the keyboard. Confirm that no visual glitches arise. Make sure you use keyboards with and without models. Simulator is okay.

* **TEST_WEB:** Test keyboard switchin in KeymanWeb prediction-ui test page, in mobile view. Use keyboards with and without models. Confirm that no visual glitches arise. Make sure you use keyboards with and without models. Chrome mobile device emulation should be okay.

### DIFF
--- a/android/KMEA/app/src/main/assets/keyboard.html
+++ b/android/KMEA/app/src/main/assets/keyboard.html
@@ -118,8 +118,7 @@
       //window.console.log('onStateChange change: ' + change);
 
       // Refresh KMW OSK
-      var kmw=window['keyman'];
-      kmw['correctOSKTextSize']();
+      keyman.correctOSKTextSize();
 
       fragmentToggle = (fragmentToggle + 1) % 100;
       if(change != 'configured') { // doesn't change the display; only initiates suggestions.
--- a/web/source/osk/oskView.ts
+++ b/web/source/osk/oskView.ts
@@ -22,7 +22,7 @@ namespace com.keyman.osk {
     manual      = "manual",
     automatic = "automatic"
   }
-  
+
   export abstract class OSKView {
     _Box: HTMLDivElement;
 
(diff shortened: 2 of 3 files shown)


MakaraSok:
* **TEST_ANDROID: PASSED** The basic usage of the keyboard is OK and no visual glitch experienced, but one [issue](https://github.com/keymanapp/keyman/issues/5965) found when testing.
* **TEST_IOS: PASSED** The basic usage of the keyboard is OK and no visual glitch experienced, but one [issue](https://github.com/keymanapp/keyman/issues/5965#issuecomment-977712569) found when testing.
* **TEST_WEB: PASSED** No obvious issue experienced when testing using this [test page](https://build.palaso.org/repository/download/Keymanweb_TestPullRequests/297860:id/testing/prediction-ui/) in Chrome on macOS.

darcywong00 (approved):
lgtm

keyman-server:

PR #5959, opened 2021-11-22
fix(windows): fix loadkeyboardoptions core memory error

### LINKED ISSUE #5951: bug(windows): SIL IPA Keyboard causes catastrophic crashing on 15 alpha
When using Keyman 15 alpha (which unfortunately for me is the only version that works on this machine), the SIL IPA keyboard, whenever it is used, causes the following applications to crash:
- My test text editor (Notepad++)
- My Discord client
- My web browser (Vivaldi 4.3.2439.56 64 bit)
- `explorer.exe`

**To Reproduce**
1. Select the IPA keyboard
2. Select an empty file in Notepad++
3. Press any key
4. Watch as your system implodes

**Expected behavior**
The IPA keyboard is usable, or at the very least not actively destructive.

 - OS: Windows 10 Pro x64 20H2 (build 19042.1288)
 - Keyman version: 15.0.151-alpha
 - Target Application: Notepad++, Vivaldi, explorer.exe

---

**Keyboard**
 - Keyboard name: IPA (SIL)
 - Keyboard version: 1.8.6
 - Language name: und-Latn

**Additional context**
I have *attempted* to turn off the Keyman Core engine, however it seems as though that setting is not being respected.
image
image

I have also not actually tried any other Keyman keyboards, but standard Windows keyboards work fine even when Keyman is running.

rc-swag:
Fixes #5951 

LoadKeyboardOptionsCore no longer frees the memory that has not been allocated.

# User Testing

**TEST_KEYBOARDOPTIONS_CHANGE**:  make sure changing the default keyboard options don't cause a crash.

This test requires IPA (SIL) keyboard to be installed (or a keyboard that has modifiable keyboard options).
Step 1. Go to `Keyman Configuration`, select the IPA (SIL) keyboard, click on `Keyboard options` and choose `Before`. You may need to exit Keyman configuration before the default behaviour changes.  
Step 2. Open Notepad, start typing anything and notepad should NOT crash.

Note: `Before` `After` setting. This will change the sequence to produce the IPA character for example `>b` => `ɓ`  for before and if the setting is after it is `b>` => `ɓ`

### DIFF
--- a/windows/src/engine/keyman32/keyboardoptions.cpp
+++ b/windows/src/engine/keyman32/keyboardoptions.cpp
@@ -1,18 +1,18 @@
 /*
   Name:             keyboardoptions
   Copyright:        Copyright (C) SIL International.
-  Documentation:    
-  Description:      
+  Documentation:
+  Description:
   Create Date:      25 May 2010
 
   Modified Date:    28 Nov 2012
   Authors:          mcdurdin
-  Related Files:    
-  Dependencies:     
+  Related Files:
+  Dependencies:
 
-  Bugs:             
-  Todo:             
-  Notes:            
+  Bugs:
+  Todo:
+  Notes:
   History:          25 May 2010 - mcdurdin - I1632 - Keyboard Options
                     31 May 2010 - mcdurdin - I2397 - Assertion failure - keyboard options loading twice
                     28 Nov 2012 - mcdurdin - I3594 - V9.0 - Remove old SelectKeyboard code and related messages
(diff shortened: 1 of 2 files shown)


windows/src/engine/keyman32/keyboardoptions.cpp
```diff
@@ -230,63 +230,57 @@ void LoadKeyboardOptionsREGCore(LPINTKEYBOARDINFO kp, km_kbp_state* const state)
...
+  int n = 0;
+  for (auto kpc = keyboardAttrs->default_options; kpc->key; kpc++) {
+    keyboardOpts[n].scope = KM_KBP_OPT_KEYBOARD;
+    keyboardOpts[n].key   = kpc->key;
+    LPCWSTR coreKey = reinterpret_cast<LPCWSTR>(kpc->key);
+    WCHAR val[256];
+    if (r.ReadString(coreKey, val, sizeof(val) / sizeof(val[0])) && val[0]) {
+      km_kbp_cp* cp = new km_kbp_cp[wcslen(val) + 1];
+      wcscpy_s(reinterpret_cast<LPWSTR>(cp), wcslen(val) + 1, val);
+      keyboardOpts[n].value = cp;
```
mcdurdin, line 264: I feel like we should add `CloneKMKBPCPFromWSTR()`

windows/src/engine/keyman32/keyboardoptions.cpp
```diff
@@ -230,63 +230,57 @@ void LoadKeyboardOptionsREGCore(LPINTKEYBOARDINFO kp, km_kbp_state* const state)
...
 
-      if (r.ReadString(buf, val, sizeof(val) / sizeof(val[0])) && val[0])
-      {
-        val[255] = 0;
-        keyboardOpts[n].scope = KM_KBP_OPT_KEYBOARD;
-
-        km_kbp_cp* cp = new km_kbp_cp[wcslen(buf) + 1];
-        wcscpy_s(reinterpret_cast<LPWSTR>(cp), wcslen(buf) + 1, buf);
-        keyboardOpts[n].key = cp;
+  if (!(r.OpenKeyReadOnly(REGSZ_KeymanActiveKeyboards) && r.OpenKeyReadOnly(kp->Name) && r.OpenKeyReadOnly(key))) {
```
mcdurdin, line 240:
```suggestion
  BOOL hasData = r.OpenKeyReadOnly(REGSZ_KeymanActiveKeyboards) && r.OpenKeyReadOnly(kp->Name) && r.OpenKeyReadOnly(key);
```

We need to be setting to default if no data available, for consistency, right?

windows/src/engine/keyman32/keyboardoptions.cpp
```diff
@@ -230,63 +230,57 @@ void LoadKeyboardOptionsREGCore(LPINTKEYBOARDINFO kp, km_kbp_state* const state)
...
+  size_t listSize = km_kbp_options_list_size(keyboardAttrs->default_options);
+  km_kbp_option_item* keyboardOpts = new km_kbp_option_item[listSize + 1];
+
+  int n = 0;
+  for (auto kpc = keyboardAttrs->default_options; kpc->key; kpc++) {
+    keyboardOpts[n].scope = KM_KBP_OPT_KEYBOARD;
+    keyboardOpts[n].key   = kpc->key;
+    LPCWSTR coreKey = reinterpret_cast<LPCWSTR>(kpc->key);
+    WCHAR val[256];
+    if (r.ReadString(coreKey, val, sizeof(val) / sizeof(val[0])) && val[0]) {
```
mcdurdin, line 261:
```suggestion
    if (hasData && r.ReadString(coreKey, val, sizeof(val) / sizeof(val[0])) && val[0]) {
```

windows/src/engine/keyman32/keyboardoptions.cpp
```diff
@@ -230,63 +230,57 @@ void LoadKeyboardOptionsREGCore(LPINTKEYBOARDINFO kp, km_kbp_state* const state)
...
+  if (!(r.OpenKeyReadOnly(REGSZ_KeymanActiveKeyboards) && r.OpenKeyReadOnly(kp->Name) && r.OpenKeyReadOnly(key))) {
+    return FALSE;
+  }
 
-        cp = new km_kbp_cp[wcslen(val) + 1];
-        wcscpy_s(reinterpret_cast<LPWSTR>(cp), wcslen(val) + 1, val);
-        keyboardOpts[n].value = cp;
-      }
-      n++;
+  // Get the list of default options to determine size of list
```
mcdurdin, line 244: We probably should move lines 244-252 up to near the top of the function e.g. to line 237, and then if listSize == 0, return TRUE. Avoids all memory allocation and registry access where there are no keyboard options.

mcdurdin (changes requested):
This actually looks pretty good to me. Just a couple of minor changes requested.

windows/src/engine/keyman32/keyboardoptions.cpp
rc-swag: Ha yes I was tempted to do that last night..   should have gone with that thought.

windows/src/engine/keyman32/keyboardoptions.cpp
```diff
@@ -237,7 +237,7 @@ BOOL IntLoadKeyboardOptionsCore(LPCSTR key, LPINTKEYBOARDINFO kp, km_kbp_state*
 
   RegistryReadOnly r(HKEY_CURRENT_USER);
 
-  if (!(r.OpenKeyReadOnly(REGSZ_KeymanActiveKeyboards) && r.OpenKeyReadOnly(kp->Name) && r.OpenKeyReadOnly(key))) {
+  BOOL hasData = r.OpenKeyReadOnly(REGSZ_KeymanActiveKeyboards) && r.OpenKeyReadOnly(kp->Name) && r.OpenKeyReadOnly(key);
     return FALSE;
   }
```
mcdurdin: oops...

windows/src/engine/keyman32/keyboardoptions.cpp
rc-swag: I have added it but I also remember why I didn't do it was because this was the only place in keyboardoptions.cpp that would take advantage of it.

mcdurdin (approved):
LGTM

MakaraSok:
**TEST_KEYBOARDOPTIONS_CHANGE: FAILED** No crash experience along the way, but:
   * Keyman Configuration doesn't have to be closed when changing the keyboard option to "Before" for it to take effect.
   * it is impossible to switch the keyboard option back to "After". No error experience when switching it back, but it doesn't take affect at all even after closing Keyman Configuration.

Note that Keyman Configuration lives by this motto "Changes are applied immediately"; if it doesn't cover this case, we might have to notify the user, otherwise they would assume that any change in the Configuration would be effective immediately.

If the issue mentioned above is intended, feel free to PASS this test.

rc-swag:
I did further testing and I found that you had to eitherː
- restart notepad for it to take effect
- change to a different keyboard and back ( without restarting notepad )
- Also opening another app like LibreOffice will see the change working there but not in notepad. 

I found this is the exact same behaviour without the keyman common core running. 
As a user cannot actually work with the keyman due to the crashing I am going to merge this and raise the configuration issue as it is really a  separate issue.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.156-alpha](https://keyman.com/downloads/releases/alpha/15.0.156)

Issue #5613, opened 2021-08-20
chore(linux): Enhance unit tests for ibus-keyman


ermshiperete:
- [x] automatically run unit tests for ibus-keyman during a build
- [ ] allow to test with both `IBUS_CAP_SURROUNDING_TEXT`-aware and non-aware modes

ermshiperete:
Not complete yet...

PR #5966, opened 2021-11-24
fix(linux): Fix ibus-keyman integration tests 🎫

ermshiperete:
Previously we overwrote our test data, so we basically ran the same test repeatedly under different names.

Properly running them showed some failures which this change also fixes or skips.

Part of #5613.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #5968, opened 2021-11-24
chore(linux): add non-surrounding-text feature to tests 🎫

### LINKED ISSUE #5613: chore(linux): Enhance unit tests for ibus-keyman

ermshiperete:
This also fixes a failing test (`048 - modifier keys keep context`) if surrounding text is not supported. If the user presses the
capslock key we shouldn't reset the context if no rules match.

This PR also refactors some tests:
- long context tests according to discussion on Slack
- tests that set option to use a test specific option to reduce interference between tests

I had to change the line number in `debug_api.cpp` and I'm not quite sure why that was necessary, but without it the tests failed.

This change also adds more core unit tests and fixes a problem in `KMX_Context::Add()`.

Look at the individual commits if there seem to be to many changes in this PR.

Fixes #5613.

@keymanapp-test-bot skip


common/core/desktop/tests/unit/kmx/021 - options.kmn
```diff
@@ -1,17 +1,17 @@
 c Description: Tests basic option rules
 c keys: [K_A][K_1][K_A][K_0][K_A]
 c expected: no foo.foo.no foo.
-c context: 
+c context:
 
 store(&version) '10.0'
 
-store(foo) '0'
+store(foo21) '0'
```
mcdurdin, line 8: > * tests that set option to use a test specific option to reduce interference between tests

This should not be necessary: the store names must be specific to a keyboard. Keyboards should be able to reuse store names without problem. 

Is this caused by something in the test harness or by an issue in Keyman Core?

common/core/desktop/tests/unit/kmx/041 - long context and deadkeys.kmn
```diff
@@ -22,4 +25,7 @@ c Test 3: Test that we haven't split a deadkey code: UC_SENTINEL CODE_DEADKEY <d
 c         because: CODE_DEADKEY = U+0008; deadkey_number = U+0001
 U+0008 U+0001 'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh12' + '3' > 'FAIL TEST 3'
 U+0001 'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh12' + '3' > 'FAIL TEST 3'
-'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh12' + '3' > context '3'
+'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh12' + '3' > 'abcdefghijklmnopqrstuvwxyz123'
+
+c Convert to a visible result:
+dk(1) 'abcdefghijklmnopqrstuvwxyz123' + '4' > 'PASS'
```
mcdurdin, line 31: This changes the meaning of this test. If we get to MAXCONTEXT, that means that additional content at the start of the cached context should be chopped off. This means that the `dk(1)` should be permanently lost, so if it isn't it probably means that we are not testing the context boundary as we think it should be.

In fact, this final test of `[K_4]` should be inverted: if `dk(1)` is still there, then we haven't tested what we want so it should fail... 😆:

```
c deadkey should have been trimmed from cached context;
c if it is not then we have not reached the cached context
c boundary!
dk(1) 'abcdefghijklmnopqrstuvwxyz123' + '4' > 'FAIL'
c This should match if the above rule doesn't.
'abcdefghijklmnopqrstuvwxyz123' + '4' > 'PASS'
```

common/core/desktop/tests/unit/kmx/042 - long context and split deadkeys.kmn
```diff
@@ -1,14 +1,23 @@
 c Description: Tests that we don't split a dk in context. Note that we start with a context with length MAXCONTEXT.
```
mcdurdin, line 1: This has the same issue as test 041 I think?

linux/ibus-keyman/tests/ibusimcontext.c
```diff
@@ -967,6 +954,27 @@ _ibus_context_forward_key_event_cb(
   }
   GdkEventKey *event = _create_gdk_event(ibusimcontext, keyval, keycode, state);
   gdk_event_put((GdkEvent *)event);
+
+  if (!surrounding_text_supported && keyval == IBUS_KEY_BackSpace && ibusimcontext->text->len > 0) {
+    int index = ibusimcontext->text->len - 1;
+    int len   = 1;
+    do {
+      guchar c = (guchar)ibusimcontext->text->str[index];
+      if (c <= 0x7F) { // ASCII character
```
mcdurdin, line 963:
```suggestion
      // delete a single UTF-8 codepoint, code unit (byte) by code unit
      if (c <= 0x7F) { // ASCII character
```

linux/ibus-keyman/tests/run-tests.sh
```diff
@@ -30,34 +30,24 @@ function cleanup() {
...
+  case $1 in
+    -k) ARG_K=$1 ;;
+    --tap) ARG_TAP=$1 ;;
+    --surrounding-text) ARG_SURROUNDING_TEXT=$1 ;;
+    --no-surrounding-text) ARG_NO_SURROUNDING_TEXT=$1 ;;
+    --) shift && break ;;
+    *) echo "Error: Unexpected argument \"$1\". Exiting." ; exit 4 ;;
+  esac
+  shift || (echo "Error: The last argument is missing a value. Exiting."; false) || exit 5
+done
```
mcdurdin, line 43: It'd be good to document these flags with `--help|-h|-?` 😁

linux/ibus-keyman/tests/testfixture.cpp
```diff
@@ -220,12 +240,16 @@ get_involved_keys(IBusKeymanTestsFixture *fixture, km::tests::key_event test_eve
   // process key
   guint keyval = (test_event.modifier_state & KM_KBP_MODIFIER_SHIFT) ? gdk_keyval_to_upper(test_event.vk)
                                                                      : gdk_keyval_to_lower(test_event.vk);
+
+  // REVIEW: the keyval we use here are not correct for < 0x20 and >= 0x7f
+  // See /usr/include/X11/keysymdef.h and /usr/include/ibus-1.0/ibuskeysyms.h
```
mcdurdin, line 245: Can you explain which circumstances this is a problem?

linux/ibus-keyman/tests/run-tests.sh
```diff
@@ -106,5 +118,5 @@ echo "kill -9 $!" >> $PID_FILE
 sleep 1s
 
 echo "Starting tests..."
-./ibus-keyman-tests ${ARG_K-} ${ARG_TAP-} --directory $TESTDIR "${TESTFILES[@]}"
+./ibus-keyman-tests ${ARG_K-} ${ARG_TAP-} ${ARG_SURROUNDING_TEXT-} ${ARG_NO_SURROUNDING_TEXT-} --directory $TESTDIR "${TESTFILES[@]}"
```
mcdurdin, line 121: You pass -k and --tap in here but I can't see how those are used in testfixture.cpp?

mcdurdin (changes requested):
A few changes requested here -- two issues with the kmx unit tests in particular. Cheers :)

common/core/desktop/tests/unit/kmx/021 - options.kmn
ermshiperete: That was a problem with the test harness. 

You're right that it is not necessary to have different store names. However, for a test I think it's a bad idea to use the same name since it can create interdependencies between the tests. If we want to test that multiple keyboards can use the same store we should create an explicit test for that.

common/core/desktop/tests/unit/kmx/041 - long context and deadkeys.kmn
ermshiperete: And we do fail...

common/core/desktop/tests/unit/kmx/042 - long context and split deadkeys.kmn
ermshiperete: yep

linux/ibus-keyman/tests/run-tests.sh
ermshiperete: Done

linux/ibus-keyman/tests/run-tests.sh
ermshiperete: They are consumed by the GLib testing framework.

linux/ibus-keyman/tests/testfixture.cpp
ermshiperete: That's not a problem, at least not at the moment, because we don't use keyval. But if that ever changes then it becomes a problem because our tests will behave different than the real client.

linux/ibus-keyman/tests/ibusimcontext.c
ermshiperete: Done

common/core/desktop/tests/unit/kmnkbd/test_kmx_context.cpp
```diff
@@ -0,0 +1,414 @@
...
+  context.Set(u"\uFFFF\u0008\u0001abc");
+  context.Set(u"def");
+  context.Get((KMX_WCHAR *)&buf, 2 * MAXCONTEXT);
+  assert_string_equal(buf, u"def");
+
+  context.Reset();
+  context.Set(u"\uFFFF\u0008\u0001abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh");
+  context.Set(u"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh1");
+  context.Get((KMX_WCHAR *)&buf, 2 * MAXCONTEXT);
+  assert_string_equal(buf, u"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh1");
```
rc-swag, line 133: What is this test - testing?

common/core/desktop/tests/unit/kmnkbd/test_kmx_context.cpp
```diff
@@ -0,0 +1,414 @@
...
+  assert_equal(p, last - 3);
+
+  p = context.Buf(2);
+  assert_equal(p, last - 4);
+
+  p = context.Buf(3);
+  assert_equal(p, last - 5);
+
+  p = context.Buf(4);
+  assert(p == NULL);
```
rc-swag, line 267: I can see why this in the `Buf` code why this returns NULL. However, is this a bug - is this the intent of `Buf` when the caller asks for more valid xstring units than is present in the `CurContext`?

common/core/desktop/tests/unit/kmnkbd/test_kmx_context.cpp
```diff
@@ -0,0 +1,414 @@
...
+  test_CharIsDeadkey();
+  test_CharIsSurrogatePair();
+  test_Set();
+  test_Add();
+  test_Delete();
+  test_Buf();
+  test_BufMax();
+
+  return 0;
+}
```
rc-swag, line 414: I like this unit test module testing all the context methods!!

rc-swag (approved):
LGTM

Just in this review highlighted that I'm not 100% sure why `KMX_Context::Buf`  is required to return null when the number of xstring units it requests is greater than the number available in the `CurContext`

ermshiperete (commented):
Added some explanations to the tests, plus some additional tests. And I realized that I miscounted the string length in the tests, so I fixed that as well and made it more explicit how long the long strings really are.

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -1208,6 +1208,27 @@ void test_decxstr() {
   assert(q == p + 2);
 }
 
+void test_xstrlen() {
+  assert(xstrlen((PKMX_WCHAR)u"") == 0);
+  assert(xstrlen((PKMX_WCHAR)u"1") == 1);
+  assert(xstrlen((PKMX_WCHAR)u"1234567890") == 10);
+  assert(xstrlen((PKMX_WCHAR)u"\uFFFF\u0008\u0001a") == 2);
+  assert(xstrlen((PKMX_WCHAR)u"\uFFFF\u0014\u0001\u0002\u0003a") == 2);
+  assert(xstrlen((PKMX_WCHAR)u"\uFFFF\u0017\u0001\u0002\u0003a") == 2);
```
mcdurdin, line 1217: I wonder if we can make these clearer with some #defines? e.g.

```c++
#define U_UC_SENTINEL u"\uFFFF"
#define U_CODE_ANY u"\u0001"
...
#define U_CODE_DEADKEY u"\u0008"

#define C_CODE_ANY(store) U_UC_SENTINEL U_CODE_ANY store
...
#define C_CODE_DEADKEY(deadkey) U_UC_SENTINEL U_CODE_DEADKEY deadkey

assert(xstrlen((PKMX_WCHAR) C_CODE_DEADKEY("\u0001") u"a") == 2);
```

Then we just need to verify those definitions once and all the tests should become (a) much easier to read, and (b) much less likely to have typos. (Of course, we still need to write the intentionally malformed test strings very carefully!)

common/core/desktop/tests/unit/test_assert.h
```diff
@@ -43,3 +43,33 @@
...
+#define assert_equal(actual, expected) { \
+  if (actual != expected) { \
+    std::wcerr << console_color::fg(console_color::BRIGHT_RED) \
+             << "Test failed at " << __FILE__ << ":" << __LINE__ << ":" \
+             << console_color::reset() \
+             << std::endl \
+             << "expected: " << expected << std::endl \
+             << "actual:   " << actual << std::endl; \
+    std::exit(100*__LINE__); \
+  } \
```
mcdurdin, line 59:
```suggestion
  if ((actual) != (expected)) { \
    std::wcerr << console_color::fg(console_color::BRIGHT_RED) \
             << "Test failed at " << __FILE__ << ":" << __LINE__ << ":" \
             << console_color::reset() \
             << std::endl \
             << "expected: " << (expected) << std::endl \
             << "actual:   " << (actual) << std::endl; \
    std::exit(100*__LINE__); \
  } \
```

common/core/desktop/tests/unit/test_assert.h
```diff
@@ -43,3 +43,33 @@
...
+#undef assert_string_equal
+#endif
+#define assert_string_equal(actual, expected) { \
+  if (u16cmp(actual, expected) != 0) { \
+    std::wcerr << console_color::fg(console_color::BRIGHT_RED) \
+             << "Test failed at " << __FILE__ << ":" << __LINE__ << ":" \
+             << console_color::reset() \
+             << std::endl \
+             << "expected: " << Debug_UnicodeString((PKMX_WCHAR)expected) << std::endl \
+             << "actual:   " << Debug_UnicodeString((PKMX_WCHAR)actual) << std::endl; \
```
mcdurdin, line 72:
```suggestion
  if (u16cmp((actual), (expected)) != 0) { \
    std::wcerr << console_color::fg(console_color::BRIGHT_RED) \
             << "Test failed at " << __FILE__ << ":" << __LINE__ << ":" \
             << console_color::reset() \
             << std::endl \
             << "expected: " << Debug_UnicodeString((PKMX_WCHAR)(expected)) << std::endl \
             << "actual:   " << Debug_UnicodeString((PKMX_WCHAR)(actual)) << std::endl; \
```

mcdurdin (changes requested):
I like most of this but there are a few things that I think we still need to work on :grin:

common/core/desktop/tests/unit/kmnkbd/test_kmx_context.cpp
mcdurdin: `Buf` is a weird one ... for historical reasons, that's what it does -- it was for performance during rule evaluation, saying, well if I can't even get enough characters to compare, I don't need to do the comparison. Function name is obviously awful!

rc-swag: Can I suggest we update the doxy explanation in `kmx_context.h` to include this explanation.  I can do it as a separate LHF PR. as all the `@brief`  tags need to be removed from that header fine as well.

mcdurdin: > Can I suggest we update the doxy explanation in kmx_context.h to include this explanation. I can do it as a separate LHF PR. as all the @brief tags need to be removed from that header fine as well.

Sounds good to me!

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
ermshiperete: I changed it for the new methods. For the existing tests in this file I'll do it in a separate PR. I also added a define for the surrogate pair, but I'm not sure if that makes things clearer or hides the fact that it's two bytes.

common/core/desktop/tests/unit/kmx/021 - options.kmn
ermshiperete: You're right. They are independent. Reverted the changes.

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
ermshiperete: Other tests are done in #6002

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -17,6 +17,19 @@
 using namespace km::kbp::kmx;
 using namespace std;
 
+#define U_UC_SENTINEL u"\uFFFF"
+#define U_UC_SURROGATE u"\U0001F609"
```
mcdurdin, line 21: I don't think I'd use this name, given it's just a random character!

And ... now I see:

> I also added a define for the surrogate pair, but I'm not sure if that makes things clearer or hides the fact that it's two bytes.

Yeah, I think it's better to give it a different name:

```suggestion
#define U_1F609_WINKING_FACE u"\U0001F609"
```

That may be overkill but it's clear and self-documenting!

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -17,6 +17,19 @@
...
+
+#define U_CODE_ANY u"\u0001"
+#define U_CODE_DEADKEY u"\u0008"
+#define U_CODE_IFOPT u"\u0014"
+#define U_CODE_IFSYSTEMSTORE u"\u0017"
+
+#define C_CODE_ANY(store) U_UC_SENTINEL U_CODE_ANY store
+#define C_CODE_IFOPT(opt, val1, val2) U_UC_SENTINEL U_CODE_IFOPT opt val1 val2
+#define C_CODE_IFSYSTEMSTORE(store, val1, val2) U_UC_SENTINEL U_CODE_IFSYSTEMSTORE store val1 val2
+#define C_CODE_DEADKEY(deadkey) U_UC_SENTINEL U_CODE_DEADKEY deadkey
```
mcdurdin, line 31: These look good. I think we should move them to a separate header file, perhaps even into kmx_file.h, below the corresponding integer #defines?

mcdurdin (approved):
LGTM 😉 [U+1F609]

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
ermshiperete: That's an idea to put the code point in the name. Done!

common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
ermshiperete: I'll do that in the other PR.

keyman-server:

PR #5963, opened 2021-11-23
feat(web): Start of Sentence support - part 1 🕊

mcdurdin:
Relates to #3621.

* Adds basic support for `begin newContext` and `begin postKeystroke` to compiler, including `gn` and `gpk` entry points.
* Adds basic support for `begin newContext` to KeymanWeb, calling into `gn` entry point, for mobile, touch and embedded contexts.

TODO in this PR:

* Compiler:
   - [x] `readonly` semantics
   - [x] usage constraints
   - [x] `&layerChanged` system store tests
* KeymanWeb:
   - [x] `begin postKeystroke` support'
   - [x] verify that `postKeystroke` fires after model events (#6004)
   - [ ] banner should be updated after newcontext, not before. (note: this may be an unrelated issue, as the banner does not currently respect layer selection for its hints.)

For follow-up pull requests:

* Developer IDE
   - [x] Support `newContext`, `postKeystroke` and `&layerChanged` tokens (#5984)
* Help
   - [x] Documentation for `newContext`, `postKeystroke` and `&layerChanged` (keymanapp/help.keyman.com#492)
   - [x] Full guide for Caps Lock support (keymanapp/help.keyman.com#492)
* Caps Lock Layer support (#3620)
   - [x] Caps Lock Layer (#5988, #5989)

* Keyman Core
   - [ ] `&layerChanged` always `0`

Optional for v15:

* KeymanWeb
   - [ ] Automated tests for start-of-sentence
* Keyman Core
   - [ ] `begin newContext`, `begin postKeystroke` support

# User Testing

### **SUITE_BASELINE:** Ensure nothing is broken!

* **TEST_BASELINE:** Run through a set of basic keyboard usage tests with SIL Eurolatin, Khmer Angkor, and other keyboards of your choice, to ensure that existing functionality has not changed. Focus on the touch keyboard interactions. Do this on Android and iOS (simulators okay), and KeymanWeb (unminified test page).

### **SUITE_START_OF_SENTENCE:** Test start-of-sentence detection

This PR includes a simple English keyboard that should detect start of sentence, called start_of_sentence_3621.kmn. 

[start_of_sentence_3621.kmp](https://drive.google.com/uc?id=1WbiesUOgI5Nby8xJM_vSy-Ip_-98nLCf&export=download)

  - It should set the keyboard to the Shift layer at the start of a text field, after full stop (`.`), exclamation (`!`) or question mark (`?`) and one or spaces (e.g. `. `).
[...]

PR #5972, opened 2021-11-25
fix(developer): building kmw keyboard had wrong message callback signature

mcdurdin:
Picked up during #5963 development.

If you ran `kmcomp <file.kmn> <out.js>`, compiler messages were passed to a function with an incorrect signature, resulting in an exception. This was a mode that was not widely used, which is why we haven't picked it up earlier.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/developer/kmcomp/main.pas
+++ b/windows/src/developer/kmcomp/main.pas
@@ -293,7 +293,7 @@ begin
   begin
     with TCompileKeymanWeb.Create do
     try
-      Result := Compile(nil, FInFile, FOutFile, FDebug, @CompilerMessage);   // I3681   // I4865   // I4866
+      Result := Compile(nil, FInFile, FOutFile, FDebug, @CompilerMessageW);   // I3681   // I4865   // I4866
     finally
       Free;
     end;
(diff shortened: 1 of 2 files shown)


windows/src/developer/kmcomp/main.pas
```diff
@@ -293,7 +293,7 @@ function CompileKeyboard(FInFile, FOutFile: string; FDebug, FWarnAsError: Boolea
   begin
     with TCompileKeymanWeb.Create do
     try
-      Result := Compile(nil, FInFile, FOutFile, FDebug, @CompilerMessage);   // I3681   // I4865   // I4866
+      Result := Compile(nil, FInFile, FOutFile, FDebug, @CompilerMessageW);   // I3681   // I4865   // I4866
```
darcywong00, line 296: Does l. 304 keep `@CompilerMessage`?

mcdurdin: Yes -- kmcmpdll still uses non-Unicode error messages, whereas the Delphi code uses Unicode messages.

darcywong00 (approved):
lgtm

keyman-server:

PR #5975, opened 2021-11-26
fix(linux): Add test targets to Makefile 🎫

ermshiperete:
The test targets got added on the `master` branch and the TC configuration adjusted so that PR builds run the tests. This change
adds these targets (but without running ibus-keyman tests) to the stable branch, so that TC builds for PRs will work.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #5979, opened 2021-11-29
chore(android): Specify build tools version on other projects

darcywong00:
While Android Studio Gradle plugin should automatically use a default version of build tools, we can specify it:
https://developer.android.com/studio/releases/build-tools

Attempting to resolve lint exceptions on the build agent.

@keymanapp-test-bot skip

### DIFF
--- a/android/README.md
+++ b/android/README.md
@@ -111,6 +111,7 @@ Keyman Engine for Android library (**keyman-engine.aar**) is now ready to be imp
 ```gradle
 android {
     compileSdkVersion 30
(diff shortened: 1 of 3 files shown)


darcywong00:
I'm convinced this will solve our build agent issues. But worth a try

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.158-alpha](https://keyman.com/downloads/releases/alpha/15.0.158)

PR #5987, opened 2021-11-30
fix(android/engine): Fix font paths

darcywong00:
Fixes #5914 

During the 14.0 refactoring of keyboard info (keyboard controller, Keyboard data object), the font info already contains the full paths to the .ttf files.

This fixes the font paths in KMKeyboard.

## User Testing
Setup: Install on Android emulator with SDK 21

* **TEST_FONT_PATHS** - Verifies font is applied for OSK keys and subkeys
1. Load the PR build of Keyman
2. Launch Keyman and exit the "Get Started" menu
3. Verify the keys and subkeys for the default sil_euro_latin keyboard are correct
3. From the "Settings" menu, install keyboards that use a TTF font. For example
    a. Test [U_xxxx_yyyy](https://darcywong00.github.io/examples/u2858/u_xxxx_yyyy_2858.kmp) keyboard 
    b. khmer_angkor (this verifies font is applied in reported issue #5985)
4. After the keyboard installs, verify the OSK is correct (no tofu boxes)
5. Select the U_xxxx_yyyy keyboard
6. Press the first Amharic key (under the `q` key and verify the output (two chars) is correct
7. Long-press the Old Italic key (3 characters under the `w` and `e` keys)
8. Verify the popup key is correct (no tofu boxes)
9. Release the popup key and verify the output is correct
10. Select khmer_angkor keyboard
11. Long-press a key and verify the font rendering is consistent with base keys.

### Screenshots of U_xxxx_yyyy keys
Screenshot_1638264938

subkey

Issue #5985, opened 2021-11-30
bug(android): the base keys and the subkeys are in different font face


MakaraSok:
**Describe the bug**

<img width="278" alt="image" src="https://user-images.githubusercontent.com/28331388/144007078-995bacfc-99a5-42c3-a2d7-f44af31a828a.png">

When testing https://github.com/keymanapp/keyman/pull/5906, it's noticed that the font face of the base keys and the subkeys are not consistent. The base keys are in Mondulkiri font (chosen for this keyboard), but the subkeys are in the Android default font (Noto Sans Khmer).

The font should have been consistent. If not, they looks kind of odd.

---

**Keyman for Android:**
 - Device: Pixel 4
 - OS: Android 10 - API 29
 - Keyman version 15.0.157-alpha-test-5906
 - Target application: Keyman

---

**Keyboard**
 - Keyboard name: khmer_angkor
 - Keyboard version: 1.1
 - Language name: Khmer

darcywong00:
Likely related to #5914

darcywong00:
Will resolve with that issue

PR #5990, opened 2021-12-01
fix(android/engine): Fix font paths 🍒 

darcywong00:
🍒 pick of #5987 to stable-14.0

This fixes the font paths since they already contain the full path.

## User Testing
Setup: Install on Android emulator with SDK 21

* **TEST_FONT_PATHS** - Verifies font is applied for OSK keys and subkeys
1. Load the PR build of Keyman
2. Launch Keyman and exit the "Get Started" menu
3. Verify the keys and subkeys for the default sil_euro_latin keyboard are correct
4. Note how the sil_euro_latin OSK is rendered (for comparison later)
5. From the "Settings" menu, install keyboards that use a TTF font. For example
    a. sil_cameroon_qwerty (Not using test U_xxxx_yyyy keyboard because 14.0 will crash trying to render those subkeys)
    b. khmer_angkor (this verifies font is applied in reported issue bug(android): the base keys and the subkeys are in different font face #5985)
6. After the keyboard installs, verify the OSK is correct (no tofu boxes)
7. Select the sil_cameroon_qwerty keyboard
8. Verify the OSK uses a different font from sil_euro_latin.
9. Long-press a key
10. Verify the popup key is correct (no tofu boxes)
11. Release the popup key and verify the output is correct
12. Select khmer_angkor keyboard
13. Long-press a key and verify the font rendering is consistent with base keys.

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -363,9 +363,7 @@ final class KMKeyboard extends WebView {
    * @return String
    */
   public static String textFontFilename() {
-    String fontPath = (txtFont.contains(KMManager.KMDefault_KeyboardFont)) ?
-      KMManager.getResourceRoot() : keyboardRoot;
-    return txtFont.isEmpty() ? "" : fontPath + txtFont;
+    return txtFont;
   }
 
   /**
@@ -373,9 +371,7 @@ final class KMKeyboard extends WebView {
    * @return String
    */
   public static String oskFontFilename() {
-    String fontPath = (oskFont.contains(KMManager.KMDefault_KeyboardFont)) ?
-      KMManager.getResourceRoot() : keyboardRoot;
-    return fontPath + oskFont;
+    return oskFont;
   }
 
   /**


mcdurdin (approved):

MakaraSok:
* **TEST_FONT_PATHS: FAILED** The EuroLatin (SIL) keyboard layouts look OK, but the key under <kbd>e</kbd> is empty on the default layer of U_xxxx_yyyy keyboard.

    

   More over the three keys do not work as expected. They output only the first character of the string instead of all of them.

   Long press on the key under <kbd>w</kbd> and <kbd>e</kbd> causes Keyman to crash (for both API 21 and 29). It doesn't crash though on other long press, i.e. long press of <kbd>.</kbd>

MakaraSok:
Also for the font face of the base keys and subkeys on Khmer Angkor. They are shown consistently.

image

darcywong00:
@keymanapp-test-bot retest TEST_FONT_PATHS

The test should use sil_cameroon_qwerty instead of U_xxxx_yyyy keyboard. Step 5a:
> Not using test U_xxxx_yyyy keyboard because 14.0 will crash trying to render those subkeys)

14.0 code can't render the U_xxxx_yyyy key so that's why it's blank.

MakaraSok:
Ah, I did't careful read the instructions thinking that it's all familiar. I took a glance at the instructions and went off -- which was why. 

I'll test this again thoroughly. Thank you for pointing it out, Darcy.

MakaraSok:
* **TEST_FONT_PATHS: PASSED** 
   * No tofu on both base keys and subkeys in `sil_cameroon_qwerty`. 
   * The font face is indeed different from that of the EuroLatin (SIL) keyboard.
   * Khmer Angkor's subkeys do have the same font as that of the base keys.

Now I feel a little less robust on the long presses. Long press a key and then move over quickly to long press another and see the preview stays instead of showing the subkeys.

keyman-server:

PR #5993, opened 2021-12-01
chore(linux): Remove lintian warning

ermshiperete:
Lintian complained "missing-build-dependency-for-dh-addon" which I'm not entirely sure is correct. However, there is an easy
workaround by removing an unnecessary parameter.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #5995, opened 2021-12-01
fix(developer): debugger crash with empty group

mcdurdin:
As the array length properties are unsigned integers, taking -1 from them resulted in an integer overflow when the array was empty. A gotcha with Delphi-style for loops!

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.159-alpha](https://keyman.com/downloads/releases/alpha/15.0.159)

PR #6000, opened 2021-12-02
chore(linux): Remove lintian warning :package:  :cherries: 

ermshiperete:
Lintian complained "missing-build-dependency-for-dh-addon" which I'm not entirely sure is correct. However, there is an easy
workaround by removing an unnecessary parameter.

:cherries:-pick of #5993.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.161-alpha](https://keyman.com/downloads/releases/alpha/15.0.161)

PR #6002, opened 2021-12-02
refactor(common/core): Use defines in all xstring tests

ermshiperete:
@keymanapp-test-bot skip

### DIFF
--- a/common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
+++ b/common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
@@ -17,18 +17,7 @@
 using namespace km::kbp::kmx;
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


common/core/desktop/tests/unit/kmnkbd/test_kmx_xstring.cpp
```diff
@@ -21,14 +21,47 @@ using namespace std;
 #define U_UC_SURROGATE u"\U0001F609"
```
darcywong00, line 21:
```suggestion
#define U_UC_SURROGATE u"\u0001F609"
```

To be consistent with all the other constants using `\u`.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.164-alpha](https://keyman.com/downloads/releases/alpha/15.0.164)

PR #6003, opened 2021-12-03
chore(common): add more description to Buf method

rc-swag:
Add more description to Buf method. As its behaviour was not correctly documented 
Also updated the doxy to remove the brief tag.
Made mirror change to appint.h
@keymanapp-test-bot skip

### DIFF
--- a/common/core/desktop/src/kmx/kmx_context.h
+++ b/common/core/desktop/src/kmx/kmx_context.h
@@ -23,85 +23,90 @@ public:
   KMX_Context();
 
   /**
-   * @brief Copy "source" AppContext to this AppContext
+   * Copy "source" AppContext to this AppContext
    *
-   * @param source KMX_Context to copy
+   * @param  source  KMX_Context to copy
    */
   void CopyFrom(KMX_Context *source);
 
   /**
-   * @brief  Add a single code unit to the Current Context. Not necessarily a complete code point
+   * Add a single code unit to the Current Context. Not necessarily a complete code point
    *
-   * @param Code unit to add
+   * @param  ch  Code unit to add
    */
   void Add(KMX_WCHAR ch);
 
   /**
-   * @brief  Removes a single code point from the end of the CurContext closest to the caret;
-   *         i.e. it will be both code units if a surrogate pair. If it is a deadkey it will
-   *         remove three code points: UC_SENTINEL, CODE_DEADKEY and deadkey value.
-   */
+  * Removes a single code point from the end of the CurContext closest to the caret;
+  * i.e. it will be both code units if a surrogate pair. If it is a deadkey it will
+  * remove three code points: UC_SENTINEL, CODE_DEADKEY and deadkey value.
+  */
   void Delete();
 
    /**
-   * @brief Clears the CurContext and resets the position - pos - index
+   * Clears the CurContext and resets the position - pos - index
    */
   void Reset();
 
    /**
-   * @brief  Copies the characters in CurContext to supplied buffer.
-   *         If bufsize is reached before the entire context was copied, the buf
-   *         will be truncated to number of valid characters possible with null character
-   *         termination. e.g. it will be one code unit less than bufsize if that would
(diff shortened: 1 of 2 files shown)


mcdurdin (approved):
LGTM. I think the failing linux test is unrelated.

common/core/desktop/src/kmx/kmx_context.h
```diff
@@ -23,85 +23,90 @@ class KMX_Context
...
   /**
-   * @brief  Returns a pointer to the character in the current context buffer which
-   *         will have at most n code units remaining until the the null terminating character.
-   *         Note: Unlike BufMax there is no checking for truncation of code points.
-   *         e.g. the pointer may point to a code unit that is half of a surrogate pair.
+   * Returns a pointer to the character in the current context buffer which
+   * will have n valid xstring units remaining until the the null terminating character.
+   * OR
+   * Returns NULL if there are less than n valid xstring units in the current context.
+   * Background this was historically for performance during rule evaluation, if I there
```
rc-swag, line 85:
```suggestion
   * Background this was historically for performance during rule evaluation, if there
```

windows/src/engine/keyman32/appint/appint.h
```diff
@@ -88,80 +88,84 @@ class AppContext
...
-   * @brief  Returns a pointer to the character in the current context buffer which
-   *         will have at most n code units remaining until the the null terminating character.
-   *         Note: Unlike BufMax there is no checking for truncation of code points.
-   *         e.g. the pointer may point to a code unit that is half of a surrogate pair.
+    /**
+   * Returns a pointer to the character in the current context buffer which
+   * will have n valid xstring units remaining until the the null terminating character.
+   * OR
+   * Returns NULL if there are less than n valid xstring units in the current context.
+   * Background this was historically for performance during rule evaluation, if I there
```
rc-swag, line 150:
```suggestion
   * Background this was historically for performance during rule evaluation, if there
```

common/core/desktop/src/kmx/kmx_context.h
```diff
@@ -23,85 +23,90 @@ class KMX_Context
...
-   *         will have at most n code units remaining until the the null terminating character.
-   *         Note: Unlike BufMax there is no checking for truncation of code points.
-   *         e.g. the pointer may point to a code unit that is half of a surrogate pair.
+   * Returns a pointer to the character in the current context buffer which
+   * will have n valid xstring units remaining until the the null terminating character.
+   * OR
+   * Returns NULL if there are less than n valid xstring units in the current context.
+   * Background this was historically for performance during rule evaluation, if there
+   * are not enough characters to compare, don't event attempt the comparison.
+   * Note: Unlike BufMax there is no checking for truncation of code points.
```
ermshiperete, line 87: `Buf` as well as `BufMax` use `decxstr` to move the pointer, so it does check for truncation of code points.

The difference is that `Buf` will return `NULL` if  n xstring units won't fit, whereas `BufMax` will return a pointer for (n-1) xstring units in that case.

At least that's how it's implemented in core, didn't check the Windows implementation.

rc-swag: > Buf as well as BufMax use decxstr to move the pointer, so it does check for truncation of code points.

Of the windows `::Buf` uses `decxstr` now as well. There is the old implementation there where it didn't use `decxstr` the line below commented out. So I can update the comment to remove line 87 

>The difference is that Buf will return NULL if n xstring units won't fit, whereas BufMax will return a pointer for (n-1) xstring units in that case.

No the difference is it returns NULL if there are less than n valid xstring units.  Not if n xstring units won't fit. 

As for the line in `::BufMax` 
`f((intptr_t)(q-p) > n) p = incxstr(p); // Copes with deadkey or supplementary pair at start of returned buffer making it too long`

I can't think that this would happen when using `decxstr` to move the pointer.

keyman-server:

Issue #3621, opened 2020-09-25
spec: Start of text/sentence selects `shift` layer


mcdurdin:
This is part of a broader specification for Caps Lock support on touch devices. Development of this touches multiple components; each component will be implemented in a separate PR, referencing this issue.

## Related features

* #3620: spec: Caps Lock layer for touch layouts
* #3720: spec: Suggestions respect capitalization
* #5029: spec: multitap and flick for touch keyboards

## Introduction

At the start of text, it would be helpful for the keyboard to select an appropriate layer. For the purposes of this discussion, this feature should only apply to a standard text control, and not be activated for email, numeric or other types of input controls.

For most Latin script languages, an appropriate layer for start of text is the `shift` layer. This is certainly not true for other scripts or for a subset of Latin script languages.

Similarly, at the start of a sentence, for most Latin script languages, the `shift` layer should be selected.

There is some ambiguity as to whether this functionality should be specified in a keyboard or in a model. An advantage of specifying this with a keyboard is that no model is required for reasonable behaviour, which is very helpful for keyboards such as Cameroon QWERTY which cover hundreds of languages. However, this analytical behaviour may belong more logically within a syntactic model.

As I believe a keyboard-based implementation will be necessary in any case, a model implementation could therefore be implemented later if time constrained.

## Virtual Key Modifiers

The `caps` layer will set the `CAPS` virtual key modifier for keystroke rules. This is something for keyboard developers to be aware of.

## Keyboard Implementation

The keyboard will have a separate entry point called `newContext` which references a group that does work whenever a new context is entered.

```
begin [keystroke] > use(main)
begin newContext > use(newContext)
begin postKeystroke > use(postKeystroke)

group(main) using keys
  ...

group(newContext) readonly
  nomatch > use(selectNextLayer)

group(postKeystroke) readonly
  if(&layerChanged = "0") > use(selectNextLayer)

group(selectNextLayer) readonly
  store(sentencePunctuation) '.?!'
  nul > layer(shift)
  any(sentencePunctuation) ' ' optany(' ') > layer(shift) 
```

The `newContext` entry point will be called whenever the user enters a new context, for example, by focusing an edit control, or by moving the insertion point, or tabbing to a new cell in a table. The `newContext` entry point will _not_ be called after a normal keystroke input event.

[...]

PR #6004, opened 2021-12-03
fix(web): call postKeystroke on banner touch 🕊

mcdurdin:
Relates to #3621.

Merges into #5963.

Ensures that the `begin postKeystroke` entry point is called when the user accepts a suggestion.

@keymanapp-test-bot skip

We'll do testing separately later on this.

### DIFF
--- a/web/source/osk/banner.ts
+++ b/web/source/osk/banner.ts
@@ -404,10 +404,6 @@ namespace com.keyman.osk {
       // Tell the keyboard that the current layer has not changed
       keyman.core.keyboardProcessor.layerChangedStore.set('0');
       // Call the keyboard's entry point.
-      /*const postRuleBehavior = keyman.core.keyboardProcessor.processPostKeystroke(this.hostDevice, outputTarget);
-      if(postRuleBehavior) {
-        postRuleBehavior.finalize(keyman.core.keyboardProcessor, outputTarget);
-      }*/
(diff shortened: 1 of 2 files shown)


web/source/osk/banner.ts
```diff
@@ -390,8 +390,28 @@ namespace com.keyman.osk {
       keyman.core.languageProcessor.addListener('suggestionsready', manager.updateSuggestions);
       keyman.core.languageProcessor.addListener('tryaccept', manager.tryAccept);
       keyman.core.languageProcessor.addListener('tryrevert', manager.tryRevert);
+      keyman.core.languageProcessor.addListener('suggestionapplied', this.suggestionApplied);
     }
 
+    /**
+     * Handler for post-porcessing once a suggestion has been applied: calls
```
rc-swag, line 397:
```suggestion
     * Handler for post-processing once a suggestion has been applied: calls
```

rc-swag (commented):
just noticed a typo. I think Darcy is probably best to give the final tick of approval.

darcywong00 (approved):
lgtm

web/source/osk/banner.ts
mcdurdin: I left this one there to see if anyone would pick it up.

Just kidding. Good catch 🐟

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.179-alpha](https://keyman.com/downloads/releases/alpha/15.0.179)

PR #6019, opened 2021-12-07
chore(windows): version.rc use macros instead of mkver ✈

mcdurdin:
This cleanup moves to using `#include "keymanversion.h"` in the various version.rc files instead of rewriting the version fields from a version.in template file with mkver, thus reducing our dependence on mkver -- and eliminating the version.in files altogether. There should be no material difference between the functionality in the two build processes.

Note: follow-up PR #6022 centralizes a few more of the values in the version.rc resources.

@keymanapp-test-bot skip

PR #6020, opened 2021-12-07
chore(windows): move from Borland make to nmake ✈

mcdurdin:
Merges into #6019.

While it may be smarter to move to a different build system altogether, it is also a *lot* more work. This was a fairly minor syntax refresh to get nmake working with our existing Makefile system. This is now no longer compatible with Borland make, so you will need to use `nmake` to build.

~~The presence of NMAKE.md informs builds systems that nmake should be used instead of make.~~ this is now done.

I have yet to update the build agents to use nmake based on the NMAKE.md signal, so initial builds will fail.

@keymanapp-test-bot skip


windows/src/Defines.mak
```diff
@@ -28,50 +28,50 @@ INSTALLPATH_KEYMANDEVELOPER=%ProgramFiles(X86)%\Keyman\Keyman Developer
 INSTALLPATH_KEYMANENGINE=%CommonProgramFiles(X86)%\Keyman\Keyman Engine
 
 !IFDEF DEBUG
-  MAKEFLAG_DEBUG=-DDEBUG
-  DELPHI_MSBUILD_FLAG_DEBUG=/p:Config=Debug
+MAKEFLAG_DEBUG="DEBUG=$(DEBUG)"
+DELPHI_MSBUILD_FLAG_DEBUG="/p:Config=Debug"
```
mcdurdin, line 32: nmake does not permit whitespace in the same way as Borland make. You'll see these changes throughout.

mcdurdin, line 32: Quotation marks are required around parameters that include an `=` sign -- otherwise nmake loses it.

windows/src/Defines.mak
```diff
@@ -28,50 +28,50 @@ INSTALLPATH_KEYMANDEVELOPER=%ProgramFiles(X86)%\Keyman\Keyman Developer
...
 !ELSE
-  DELPHI_MSBUILD_FLAG_DEBUG=/p:Config=Release
+DELPHI_MSBUILD_FLAG_DEBUG="/p:Config=Release"
 !ENDIF
 
 !IFDEF USERDEFINES
-  MAKEFLAG_USERDEFINES=-DUSERDEFINES
+MAKEFLAG_USERDEFINES="USERDEFINES=$(USERDEFINES)"
```
mcdurdin, line 38: `-D` command line parameter to define a macro is replaced with equivalent macro definition

windows/src/Defines.mak
```diff
@@ -118,13 +118,13 @@ DELPHIWARNINGS=-W+MESSAGE_DIRECTIVE -W+IMPLICIT_STRING_CAST -W+IMPLICIT_STRING_C
...
-DELPHIDPRPARAMS64=-Q -B -GD -H -VT -$C+ -$D+ -$J+ -$L+ -$O+ -$Q- -$R- -$W+ -$Y+ -E. $(DELPHIWARNINGS) -I$(DELPHIINCLUDES) -U$(DELPHIINCLUDES) -R$(DELPHIINCLUDES) -NSVcl;Vcl.Imaging;Vcl.Touch;Vcl.Samples;Vcl.Shell;System;Xml;Web;Soap;Winapi;System.Win -NU.\obj\Win64\$(TARGET_PATH) -E.\bin\Win64\$(TARGET_PATH)
-DELPHIDPKPARAMS=-Q -B -GD -VT -$C+ -$D+ -$J+ -$L+ -$O+ -$Q- -$R- -$W+ -$Y+ -E. $(DELPHIWARNINGS) -I$(DELPHIINCLUDES) -U$(DELPHIINCLUDES) -R$(DELPHIINCLUDES) -NSVcl;Vcl.Imaging;Vcl.Touch;Vcl.Samples;Vcl.Shell;System;Xml;Web;Soap;Winapi;System.Win -LE$(OUTLIB) -LN$(OUTLIB) -NSData -NUobj\Win32\$(TARGET_PATH)
+DELPHIDPRPARAMS=-Q -B -GD -H -VT -^$C+ -^$D+ -^$J+ -^$L+ -^$O+ -^$Q- -^$R- -^$W+ -^$Y+ -E. $(DELPHIWARNINGS) -I$(DELPHIINCLUDES) -U$(DELPHIINCLUDES) -R$(DELPHIINCLUDES) -NSVcl;Vcl.Imaging;Vcl.Touch;Vcl.Samples;Vcl.Shell;System;Xml;Web;Soap;Winapi;System.Win -NU.\obj\Win32\$(TARGET_PATH) -E.\bin\Win32\$(TARGET_PATH)
```
mcdurdin, line 121: The `$` sign must be escaped with `^` to avoid being treated as a macro identifier.

windows/src/Defines.mak
```diff
@@ -118,13 +118,13 @@ DELPHIWARNINGS=-W+MESSAGE_DIRECTIVE -W+IMPLICIT_STRING_CAST -W+IMPLICIT_STRING_C
...
+DELPHIDPRPARAMS64=-Q -B -GD -H -VT -^$C+ -^$D+ -^$J+ -^$L+ -^$O+ -^$Q- -^$R- -^$W+ -^$Y+ -E. $(DELPHIWARNINGS) -I$(DELPHIINCLUDES) -U$(DELPHIINCLUDES) -R$(DELPHIINCLUDES) -NSVcl;Vcl.Imaging;Vcl.Touch;Vcl.Samples;Vcl.Shell;System;Xml;Web;Soap;Winapi;System.Win -NU.\obj\Win64\$(TARGET_PATH) -E.\bin\Win64\$(TARGET_PATH)
+DELPHIDPKPARAMS=-Q -B -GD -VT -^$C+ -^$D+ -^$J+ -^$L+ -^$O+ -^$Q- -^$R- -^$W+ -^$Y+ -E. $(DELPHIWARNINGS) -I$(DELPHIINCLUDES) -U$(DELPHIINCLUDES) -R$(DELPHIINCLUDES) -NSVcl;Vcl.Imaging;Vcl.Touch;Vcl.Samples;Vcl.Shell;System;Xml;Web;Soap;Winapi;System.Win -LE$(OUTLIB) -LN$(OUTLIB) -NSData -NUobj\Win32\$(TARGET_PATH)
 
 !IFDEF NOUI
-DCC32="$(DCC32PATH)\dcc32.exe" $(DELPHIDPRPARAMS)
-DCC32DPK="$(DCC32PATH)\dcc32.exe" $(DELPHIDPKPARAMS)
+DCC32=cmd /c "$(DCC32PATH)\dcc32.exe" $(DELPHIDPRPARAMS)
+DCC32DPK=cmd /c "$(DCC32PATH)\dcc32.exe" $(DELPHIDPKPARAMS)
```
mcdurdin, line 127: nmake seems to crash dcc32.exe if it calls it directly. Unsure why. This makes me somewhat uncomfortable, but I have not yet taken the time to dig into the root cause. The `cmd /c` workaround works.

windows/src/Defines.mak
```diff
@@ -141,8 +141,7 @@ DCC64="$(DCC32PATH)\dcc64.exe" $(DELPHIDPRPARAMS64) -N0x64\ -Ex64\
 # Delphi MSBuild related commands and macros
 #
 
-# Warning: whitespace is horribly significant in the following macro -- particularly lack of space before &&
-DELPHI_MSBUILD=set DCC32PATH=$(DCC32PATH)&& $(ROOT)\src\buildtools\msbuild-wrapper.bat $(DELPHI_MSBUILD_FLAG_DEBUG)
+DELPHI_MSBUILD=$(ROOT)\src\buildtools\msbuild-wrapper.bat "$(DCC32PATH)" $(DELPHI_MSBUILD_FLAG_DEBUG)
```
mcdurdin, line 144: Embedding the `set` into the batch file works better than the `&&` trick (which doesn't work).

windows/src/desktop/kmshell/Makefile
```diff
@@ -17,7 +17,7 @@ icons:
 xml: sentry-init.js
     cd $(ROOT)\src\desktop\kmshell\xml
     -del /q $(ROOT)\bin\desktop\xml\*
-    copy * $(ROOT)\bin\desktop\xml\   #
```
mcdurdin, line 20: This `\` on the end of the copy target with a `#` to prevent make from treating it as a line extension trick doesn't work in nmake (can use `/I` to force same xcopy semantics, not necessary here with copy).

ermshiperete (approved):
Does `README.md` need to be adjusted?

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.166-alpha](https://keyman.com/downloads/releases/alpha/15.0.166)

PR #6021, opened 2021-12-07
chore(windows): use findstr instead of Borland grep ✈

mcdurdin:
Merges into #6020.

Removes one more dependency on Delphi tools

@keymanapp-test-bot skip


mcdurdin:
> I take it `findstr` doesn't have the equivalent of `grep -c`, but that's fine?

Correct. We only care about the exit code from `findstr` 😁

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.166-alpha](https://keyman.com/downloads/releases/alpha/15.0.166)

PR #6026, opened 2021-12-08
refactor(developer): kmcompx compiler cross platform 🍃

SabineSIL:
moved to new laptop - and branched on keymanapp/keyman instead of on SabineSIL:keyman

Replaces #5879.


mcdurdin:
Replaced by #6980

PR #6033, opened 2021-12-10
feat(developer): remember test page preferences on reload 🛒

mcdurdin:
Emerges from ELTW demos.

Remembers the user's selected model and device when they refresh the page.

# User Testing

* **TEST_MODEL_MEMORY:** Load a model in the web keyboard debugger in Developer. Switch the model on, and press F5 to refresh the page. Verify that the model is still selected. (Do this a couple of times with model selected or off.)

* **TEST_DEVICE_MEMORY:** Switch between various devices in the device picker. Verify that the device stays selected after you refresh the page.

PR #6034, opened 2021-12-10
feat(developer): tidy up package list in web test view 🛒

mcdurdin:
Merges into #6033.

Tidies up the package download page in the web test view, making it clearer and less verbose, while offering more detail.

## New Design

image

## Old Design

image

@keymanapp-test-bot skip

I think the screenshot is sufficient for testing :grin:

### DIFF
--- a/windows/src/developer/TIKE/project/Keyman.Developer.UI.Project.kpsProjectFileUI.pas
+++ b/windows/src/developer/TIKE/project/Keyman.Developer.UI.Project.kpsProjectFileUI.pas
@@ -137,7 +137,7 @@ begin
   if not FileExists(FCompiledName) then
     Exit(False);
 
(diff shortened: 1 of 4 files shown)


ermshiperete (approved):
Looks great. However, I don't agree with the `keymanapp-test-bot skip` labeling because that'll cause it to not be on the radar for getting tested.

mcdurdin:
> However, I don't agree with the `keymanapp-test-bot skip` labeling because that'll cause it to not be on the radar for getting tested.

@ermshiperete and I discussed this offline -- basically I just don't want to overload our tester :grin: and this is low risk and hopefully cosmetic.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.171-alpha](https://keyman.com/downloads/releases/alpha/15.0.171)

PR #6040, opened 2021-12-13
chore(common): Update crowdin strings for Amharic 📜

darcywong00:
This adds Amharic localization for macOS.

Also updates Amharic strings for Android and iOS.

## User Testing
Only testing macOS platform, since other platforms are just updating strings

* **TEST_MAC** - Tests Amharic localization on Keyman for macOS
1. Load the macOS PR build
2. Set the system language to Amharic
3. Launch Keyman for macOS
4. Verify strings are in Amharic.

### DIFF
--- a/android/KMEA/app/src/main/res/values-am-rET/strings.xml
+++ b/android/KMEA/app/src/main/res/values-am-rET/strings.xml
@@ -8,6 +8,11 @@
         <item quantity="other">የፊደል ገበታዎች</item>
     </plurals>
     <!-- Context: Title for list -->
+    <plurals name="title_other_input_methods">
+        <item quantity="one">የሚቀጥለውን የመቀበያ ዘዴ</item>
+        <item quantity="other">የሚቀጥለውን የመቀበያ ዘዴዎች</item>
+    </plurals>
+    <!-- Context: Title for list -->
     <string name="title_add_keyboard" comment="Add a new keyboard">አዲስ ኪቦርድ ጨምር</string>
     <!-- Context: Title for list -->
     <string name="title_languages_settings" comment="List of installed languages">የተጫኑ ቋንዎች</string>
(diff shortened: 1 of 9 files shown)


mcdurdin (approved):

MakaraSok:
* **TEST_MAC: FAILED** It can't be switch to Amharic, but Khmer does work though.

Catalina: the language doesn't switch to Amharic, even after a restart.
image

When the language is set to "Khmer," the UI language for Keyman does switch to Khmer though.
image


Big Sur: the language doesn't switch to Amharic, even after a restart.
image

Monterey: the same behaviors.

darcywong00:
@sgschantz - Does the macOS project also need to be run through XCode to add the Amharic locale?

sgschantz:
Yes, I need to specify it as a supported localization in the project, just like we need to do for iOS.

darcywong00:
@keymanapp-test-bot retest TEST_MAC

sgschantz:
>

MakaraSok:
- **TEST_MAC: PASSED** but some strings are still in English. See the screenshot below.

The UI does switch to the language, but all strings in the keyboard info page stay in English.

Catalina
image

BigSur
image

sgschantz:
@MakaraSok When I did the internationalization for the Mac, I missed all the strings in this view because it is a webview, and I didn't notice the labels coded in HTML. So these must still appear in English for the Khmer localization, too. 

I'm not sure how I missed that -- I might have assumed they had a different source like the help pages. I'll need to externalize the keyboard info strings so they can be localized.

MakaraSok:
@sgschantz Let me know when/if this needs to be retested.

sgschantz:
I think we need a new issue to track this because it is not related to language. You should see the same results for Khmer. After adding i18n support for this window, then we can go back and get the corresponding Amharic strings from crowdin.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.173-alpha](https://keyman.com/downloads/releases/alpha/15.0.173)

PR #6044, opened 2021-12-14
fix(linux): Improve ibus-keyman tests

ermshiperete:
This change will hopefully help some tests that occasionally fail on TC. After setting the options during the test setup we sync the
settings before running the test scenario.

@keymanapp-test-bot skip because this change affects tests only.


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.169-alpha](https://keyman.com/downloads/releases/alpha/15.0.169)

PR #6059, opened 2021-12-16
docs(windows): Add note on split user/admin accounts

mcdurdin:
If the computer has split user/admin accounts, then language association may not be completed automatically, and the user may need to add a language association themselves.

Ref https://community.software.sil.org/t/what-security-permissions-are-needed-to-install-a-keyboard-configuration-in-keyman/5695/10

Note: the topic `configure-computer` has a terrible url but I won't fix that here. Suggested future fix -- turn `configure-computer` into a redirect to `link-keyboard-to-language` or `associate-keyboard-with-language`?

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/desktop/help/start/download-and-install-keyboard.md
+++ b/windows/src/desktop/help/start/download-and-install-keyboard.md
@@ -71,6 +71,11 @@ Here's how to download and install a Keyman keyboard within Keyman:
 
     ![](../desktop_images/download-keyboard-step6.png)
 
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.171-alpha](https://keyman.com/downloads/releases/alpha/15.0.171)

PR #6070, opened 2021-12-17
feat(linux): Don't create Sentry events for errors

ermshiperete:
We don't want to send events to Sentry for errors that we handled.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/convertico.py
+++ b/linux/keyman-config/keyman_config/convertico.py
@@ -51,7 +51,7 @@ def checkandsaveico(icofile):
         # Using .bmp.png file extension so it won't conflict if the package already contains .png
         im4.save(bmpfile + '.png', 'png')
     except (IOError, OSError):
-        logging.critical("Cannot convert %s to png", icofile)
+        logging.error("Cannot convert %s to png", icofile)
         pass
     finally:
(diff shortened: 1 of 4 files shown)


ermshiperete:
Another option to get rid of the handled errors on Sentry would be to use `logging.error()` instead of `logging.critical()`.

linux/keyman-config/keyman_config/__init__.py
```diff
@@ -55,7 +55,7 @@ def _(txt):
 
         sentry_logging = LoggingIntegration(
             level=logging.INFO,           # Capture info and above as breadcrumbs
-            event_level=logging.CRITICAL  # Send critical errors as events
+            event_level=None              # Don't create events for errors
```
rc-swag, line 58: Won't this mean all no events are sent to sentry even unhandled events?

ermshiperete: What was I thinking??? Of course you're right.

ermshiperete: done.

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.175-alpha](https://keyman.com/downloads/releases/alpha/15.0.175)

PR #6074, opened 2021-12-21
fix(developer): work around devDependencies bug in npm

mcdurdin:
Relates to https://github.com/npm/cli/issues/2921

This issue arises after moving from npm 6.x to 7.x or later. As the build agents are still on 6.x, we have not yet seen this issue there.

As part of the release build for Keyman Developer, we bundle up some files and do some work in a temp folder to prepare folders for release. Even though we are only working with `dependencies` (e.g. using `npm install --production`), packages under `devDependencies` are still wrongly being checked by NPM. As we have two modules present in `devDependencies` that are only available when in the repo path, we need to remove them before attempting to continue. Yuck!

ref: https://github.com/npm/cli/issues/3975#issuecomment-985305678
ref: https://github.com/npm/cli/issues/2921

@keymanapp-test-bot skip

### DIFF
--- a/developer/js/bundle.sh
+++ b/developer/js/bundle.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -eu
+set -x
 
 # $1 = keyman_wix_build temp root
 
--- a/developer/js/bundle.sh
+++ b/developer/js/bundle.sh
@@ -1,0 +1,105 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# $1 = keyman_wix_build temp root
+
+display_usage() {
+  echo "Usage: $0 --build-path path"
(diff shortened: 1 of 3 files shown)


developer/js/bundle.sh
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# $1 = keyman_wix_build temp root
```
rc-swag, line 5: This can be removed now?

rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.172-alpha](https://keyman.com/downloads/releases/alpha/15.0.172)

PR #6082, opened 2022-01-05
chore(android/samples): Add -no-daemon flag to KMSample2 build script

### LINKED ISSUE #6075: chore(android): disable linting for kmsample2
The lint tool is crashing but only on kmsample2, so let's just disable it there for now?

For example: https://build.palaso.org/viewLog.html?buildId=301805&buildTypeId=KeymanAndroid_TestSamplesAndTestProjects&tab=buildLog

darcywong00:
Fixes #6075 and follow-on to #603

#603 added a `-no-daemon` build flag for the sample projects to build on CI, but KMSample2/build.sh inadvertently wasn't updated.

This PR updates KMSample2 so hopefully will stop the CI from hanging.
Will need to 🍒-pick to stable-14.0

## User Testing
* TEST_KMSAMPLE2 - Tests KMSample2 project still runs
1. Load the test build and launch KMSample2
2. Using the menus, set KMSample2 as a system IME
3. Launch a separate app (e.g. Chrome) and select a text field to type
4. When KMSample2 is the selected IME, verify it can type

### DIFF
--- a/android/Samples/KMSample2/build.sh
+++ b/android/Samples/KMSample2/build.sh
@@ -1,10 +1,61 @@
-#!/bin/sh
+#!/bin/bash
 # Build KMSample2
+
+set -e
+set -u
+
+display_usage ( ) {
+    echo "build.sh [-no-daemon] [-debug]"
+    echo
+    echo "Build KM Sample 2"
+    echo "  -no-daemon              Don't start the Gradle daemon. Use for CI"
+    echo "  -debug                  Compile only Debug variant"
+    exit 1
+}
+
+echo Build KMSample2
 
 #
 # Prevents 'clear' on exit of mingw64 bash shell
 #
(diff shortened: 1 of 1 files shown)


android/Samples/KMSample2/build.sh
```diff
@@ -1,10 +1,61 @@
-#!/bin/sh
+#!/bin/bash
```
mcdurdin, line 1:
```suggestion
#!/usr/bin/env bash
```

This gets the newer version of bash on macOS when installed with homebrew and works correctly on other platforms also.

mcdurdin (approved):
LGTM

MakaraSok:
* **TEST_KMSAMPLE2: PASSED** The KMSample2 does type and output the expected characters when selected and type in Chrome on Android 10.0.

PR #6091, opened 2022-01-10
fix(developer): bundle.sh must succeed

mcdurdin:
Fixes a build failure after #6074 was merged. The final grep statement in the line

  cat package.json | grep -v "@keymanapp/models-templates" | grep -v "@keymanapp/models-wordbreakers" > package.json

would fail to find anything to exclude, and return an error value, so the script would fail, as we use `set -e`. We didn't care that it didn't find anything to exclude... so we will ignore the return value.

@keymanapp-test-bot skip

### DIFF
--- a/developer/js/bundle.sh
+++ b/developer/js/bundle.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 set -eu
-set -x
 
 display_usage() {
   echo "Usage: $0 --build-path path"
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.175-alpha](https://keyman.com/downloads/releases/alpha/15.0.175)

PR #6100, opened 2022-01-12
fix(web): Use regex to determine display layer and functional layers

### LINKED ISSUE #6089: bug(android): longpress with Shift+RightAlt not working on touch
**Describe the bug**

A longpress key with reference to a key with Shift+RightAlt in the computer layout does not output anything. 
It works just fine though when testing on Chrome emulator.

This was encountered when testing [Ukwuani keyboard](https://drive.google.com/drive/folders/1M-HUkJhfvLgb9_G4d-WVuo2tMlqRNl-7).

**To Reproduce**

1. On the current stable version of Keyman (14.0.283) install the keyboard above.
2. Go the Shift layer and then press and hold on key <kbd>E</kbd> to output the capital E with a dot below
3. See error - nothing is output.

**Expected behavior**

The longpress with such reference should work as normal, just like how it works in Chrome emulator.

---

**Keyman for Android:**
 - Device: Pixel XL
 - OS: Android 10.0
 - Target application: Keyman

**KeymanWeb:**
 - OS: Windows 10
 - Browser: Chrome
 - KeymanWeb Version: ?
 - URL: N/A (test it locally using the Ukwuani keyboard project above.

---

**Additional context**

This works OK in iOS when testing with Keyman 14.0.284.

darcywong00:
fixes #6089

> A longpress key with reference to a key with Shift+RightAlt in the computer layout does not output anything.
It works just fine though when testing on Chrome emulator.

I initially started porting the changes in #5651 from stable-14.0, but Makara reported seeing the issue in 14.0 too. Thus, portions of this PR will need to be 🍒 picked back to stable-14.0

The complication in kmwembedded.ts comes from parsing `keyName` which contains:
[display layer]-[virtual key name]+[optional function layer] where display layer and function layer can contain multiple `-` delimiters.
For example: 'ctrl-shift' or 'rightalt-shift'

The virtual key name  can also be T_* , K_* , U_*, ~or ISO 9995 format~

In @MakaraSok 's issue, the displayLayer is 'shift' while the functional layer is 'rightalt-shift' from the longpress key.

## User Testing
Setup:
For each platform test, install the following keyboards (some were used in testing #5651):
1. ukwuani (zip of [source](https://github.com/keymanapp/keyman/files/7845806/ukwuani.zip))
2. sil_madi
3. sil_cameroon_qwerty

* **TEST_ANDROID**
1. Launch Keyman
2. Test and verify sil_euro_latin keyboard:
    a. Use the globe key to switch to sil_euro_latin keyboard
    b. On the "default" layer, longpress a key and select
    c. Verify the output character matches the longpress key
    d. Repeat for "shift" and "symbol" layers
3. Test and verify ukwuani keyboard:
    a. Use the globe key to switch to ukwuani keyboard
    b. On the default layer, Longpress on the <kbd>o</kbd>, <kbd>i</kbd>, <kbd>u</kbd>, <kbd>e</kbd>, or <kbd>n</kbd> keys and select the longpress key
    c. Verify the output character matches the longpress key
    d. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    e. Longpress on the <kbd>O</kbd>, <kbd>I</kbd>, <kbd>U</kbd>, <kbd>E</kbd>, or <kbd>N</kbd> keys and select the longpress key
    f. Verify the output character matches the longpress key
4. Test and verify sil_madi keyboard:
    a. Use the globe key to switch to sil_madi keyboard
    b. On the default layer, longpress on a vowel and select it
    c. Verify the correct character is output
    d. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    e. Longpress on the <kbd>A</kbd> and select one of the caps longpresses.
    f. Verify the correct character is output
5. Test and verify sil_cameroon_qwerty (verifies application of nextlayer)
    a. Use the globe key to switch to sil_cameroon_qwerty
    b. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    c. Longpress on the <kbd>W</kbd> or <kbd>E</kbd> keys and select one of the caps longpresses
    d. Verify the correct character is output
    e. Verify the keyboard switches to the to the default layer
    f. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    g. Press the spacebar
    h. Verify the layer remains on "Shift"
    i. For each of the following:
        * switch to the symbolic layer
        * longpress on the base key and select any of the longpress keys
        * verify the correct character is output
        * verify the keyboard **switches** to the default layer
        **Base Key**
        <kbd><<</kbd> 
        <kbd>>></kbd> 
        <kbd>'</kbd> (single quote)
        <kbd>"</kbd> (double quote)
    j. For each of the following:
        * switch to the symbolic layer
        * longpress on the base key and select any of the longpress keys
        * verify the correct character is output
        * verify the keyboard **remains** on the symbolic layer
        **Base Key**
        <kbd>(</kbd> 
        <kbd>)</kbd> 
        <kbd>?</kbd> (Note, the basekey <kbd>?</kbd> will switch to the default layer
        <kbd>!</kbd> (Note, the basekey <kbd>!</kbd> will switch to the default layer

* **TEST_IOS**
1. Repeat the same steps as ANDROID

### DIFF
--- a/android/KMEA/app/src/main/assets/keyboard.html
+++ b/android/KMEA/app/src/main/assets/keyboard.html
@@ -248,7 +248,8 @@
         var keyPos = x.toString() + ',' + y.toString();
         for(i=0; i<obj.length; i++)
         {
-          s=s+obj[i].layer+'-'+obj[i].coreID;
+          // elementID contains the layer and coreID
+          s=s+obj[i].elementID;
           if(obj[i].sp == 1 || obj[i].sp == 2) shift = true;
           if(typeof(obj[i].text) != 'undefined' && obj[i].text != null && obj[i].text != '') s=s+':'+toHex(obj[i].text);
           if(i < (obj.length -1)) s=s+';'
--- a/web/source/kmwembedded.ts
+++ b/web/source/kmwembedded.ts
@@ -365,24 +365,18 @@ namespace com.keyman.text {
       // Changes for Build 353 to resolve KMEI popup key issues
       keyName=keyName.replace('popup-',''); //remove popup prefix if present (unlikely)
 
-      let displayLayer = keymanweb.core.keyboardProcessor.layerId;
-
       // Regex for 'display layer'-'virtual key name'+'optional functional layer'
       // Can't just split on '-' because some layers like ctrl-shift contain it.
-      // Virtual key can be T_, U_, K_, or ISO 9995
-      let rx = new RegExp("(.*)-([TKU]_[^+]*|[A-E]\\d\\d)\\+?(.*)?");
-      let matches = keyName.match(rx);
+      // Virtual key name starts with T_, K_, or U_
+      // matches[1]: displayLayer (not used)
+      // matches[2]: keyId
+      // matches[3]: optional functionalLayer
+      let matches = keyName.match(/^(.+)-([TKU]_[^+]+)\+?(.+)?$/);
       if (matches == null) {
         return false;
       }
-      displayLayer = matches[1];
-      keyName = matches[2];
-      let functionalLayer = '';
-      if (matches[3]) {
-        functionalLayer = matches[3];
-        keyName += '+' + functionalLayer;
-      }
-
+      keyName = matches[2] + (matches[3] ? '+' + matches[3] : '');
+ 
       // Note:  this assumes Lelem is properly attached and has an element interface.
       // Currently true in the Android and iOS apps.
       var Lelem=keymanweb.domManager.lastActiveElement;
--- a/web/source/kmwembedded.ts
+++ b/web/source/kmwembedded.ts
@@ -366,23 +366,21 @@ namespace com.keyman.text {
       keyName=keyName.replace('popup-',''); //remove popup prefix if present (unlikely)
 
       let displayLayer = keymanweb.core.keyboardProcessor.layerId;
-      let functionalLayer = '';
 
       // Regex for 'display layer'-'virtual key name'+'optional functional layer'
       // Can't just split on '-' because some layers like ctrl-shift contain it.
       // Virtual key can be T_, U_, K_, or ISO 9995
       let rx = new RegExp("(.*)-([TKU]_[^+]*|[A-E]\\d\\d)\\+?(.*)?");
       let matches = keyName.match(rx);
+      if (matches == null) {
+        return false;
+      }
       displayLayer = matches[1];
       keyName = matches[2];
+      let functionalLayer = '';
       if (matches[3]) {
-        // Optional function layer
-        functionalLayer = matches[3]
+        functionalLayer = matches[3];
         keyName += '+' + functionalLayer;
-      }
-
-      if(displayLayer == 'undefined') {
-        displayLayer=keymanweb.core.keyboardProcessor.layerId;
       }
 
       // Note:  this assumes Lelem is properly attached and has an element interface.
--- a/web/source/kmwembedded.ts
+++ b/web/source/kmwembedded.ts
@@ -349,7 +349,8 @@ namespace com.keyman.text {
  /**
    *  Accept an external key ID (from KeymanTouch) and pass to the keyboard mapping
    *  
-   *  @param  {string}  keyName   key identifier
+   *  @param  {string}  keyName   key identifier which could contain a display layer and a "functional" layer
+   *                              e.g: 'shift-K_E+rightalt-shift'
    **/            
   keymanweb['executePopupKey'] = function(keyName: string) {
       let osk = keymanweb.osk;
@@ -361,14 +362,30 @@ namespace com.keyman.text {
       /* Clear any pending (non-popup) key */
       osk.vkbd.keyPending = null;
 
-      // Changes for Build 353 to resolve KMEI popup key issues      
+      // Changes for Build 353 to resolve KMEI popup key issues
       keyName=keyName.replace('popup-',''); //remove popup prefix if present (unlikely)
 
+      // Determine display layer (and ignore functional layer if it exists).
       // Can't just split on '-' because some layers like ctrl-shift contain it.
-      let separatorIndex = keyName.lastIndexOf('-');
-
+      let originalKeyName = keyName;
+      let displayLayer = keymanweb.core.keyboardProcessor.layerId;
+      let functionalLayerSeparatorIndex = keyName.indexOf('+');
+      let separatorIndex = (functionalLayerSeparatorIndex > 0) ?
+        keyName.substring(0, functionalLayerSeparatorIndex).lastIndexOf('-') :
+        keyName.lastIndexOf('-');
       if (separatorIndex > 0) {
+        displayLayer = keyName.substring(0, separatorIndex);
         keyName = keyName.substring(separatorIndex+1);
+      }
+      if(displayLayer == 'undefined') {
+        displayLayer=keymanweb.core.keyboardProcessor.layerId;
+      }
+
+      // Determine the "functional" layer from the coreID
+      let layer = displayLayer;
+      separatorIndex = originalKeyName.lastIndexOf('+');
+      if (separatorIndex > 0) {
+        layer = originalKeyName.substring(separatorIndex+1);
       }
 
       // Note:  this assumes Lelem is properly attached and has an element interface.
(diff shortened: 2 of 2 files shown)


MakaraSok:
- **TEST_ANDROID: PASSED** All things work as indicated.
- **TEST_IOS: PASSED** All things work as indicated.

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,27 @@ namespace com.keyman.text {
...
+      if (matches[3]) {
+        // Optional function layer
+        functionalLayer = matches[3]
+        keyName += '+' + functionalLayer;
+      }
 
-      if (separatorIndex > 0) {
-        keyName = keyName.substring(separatorIndex+1);
+      if(displayLayer == 'undefined') {
+        displayLayer=keymanweb.core.keyboardProcessor.layerId;
```
rc-swag, line 385: What is display layer used for once it has been set?

darcywong00: It had more use in 14.0 code (#5651) where it was used to determine `layer` and then get the modifier state
```
// Determine the "functional" layer
      let layer = displayLayer;
      separatorIndex = keyName.lastIndexOf('+');
      if (separatorIndex > 0) {
        layer = keyName.substring(separatorIndex+1);
      }

      // Note:  this assumes Lelem is properly attached and has an element interface.
      // Currently true in the Android and iOS apps.
      var Lelem=keymanweb.domManager.getLastActiveElement(),keyShiftState=com.keyman.text.KeyboardProcessor.getModifierState(layer);
```

Lastly in 14.0, it was used to:
```
      // Check the virtual key 
        Lcode: Codes.keyCodes[keyName],
        LisVirtualKey: true,
        kName: keyName,
        kLayer: layer,
        kbdLayer: displayLayer,
        ...
```

darcywong00:
Putting this back into draft. I'm gonna take another look in case the regex fails (displayLayer undefined)

darcywong00:
> I'm gonna take another look in case the regex fails (displayLayer undefined)

Since TEST_IOS passed, I guess the displayLayer will always be defined.

rc-swag (approved):
LGTM

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,25 @@ namespace com.keyman.text {
...
 
-      // Can't just split on '-' because some layers like ctrl-shift contain it.
-      let separatorIndex = keyName.lastIndexOf('-');
+      let displayLayer = keymanweb.core.keyboardProcessor.layerId;
 
-      if (separatorIndex > 0) {
-        keyName = keyName.substring(separatorIndex+1);
+      // Regex for 'display layer'-'virtual key name'+'optional functional layer'
+      // Can't just split on '-' because some layers like ctrl-shift contain it.
+      // Virtual key can be T_, U_, K_, or ISO 9995
```
mcdurdin, line 372: ISO9995 identifiers are not valid in the touch layout. The regex should not test for these.

Furthermore, ISO9995 identifiers in .kmn are converted into the base key ids at compile time -- they aren't in the compiled keyboard.

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,25 @@ namespace com.keyman.text {
...
+      let rx = new RegExp("(.*)-([TKU]_[^+]*|[A-E]\\d\\d)\\+?(.*)?");
```
mcdurdin, line 373: Six things:

1. This regex should explicitly match the entire identifier -- we need to include `^` and `$`.
2. Why not use the normal Javascript regex enclosing with `/`, which avoids double-escaping?
3. Rather than use `.*` for the layer components, you should use `.+` -- we don't want a zero-length layer identifier!
4. I've removed the ISO9995 component.
5. A key identifier must be `T_`, `K_` or `U_` with one-or-more, not zero-or-more chars.
6. Use `const` where possible (or eliminate variables).

```suggestion
      const rx = /^(.+)-([TKU]_[^+]+)\+?(.+)?$/;
```

Or:

```
      const matches = keyName.match(/^(.+)-([TKU]_[^+]+)\+?(.+)?$/);
```

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,25 @@ namespace com.keyman.text {
       /* Clear any pending (non-popup) key */
       osk.vkbd.keyPending = null;
 
-      // Changes for Build 353 to resolve KMEI popup key issues      
+      // Changes for Build 353 to resolve KMEI popup key issues
       keyName=keyName.replace('popup-',''); //remove popup prefix if present (unlikely)
 
-      // Can't just split on '-' because some layers like ctrl-shift contain it.
-      let separatorIndex = keyName.lastIndexOf('-');
+      let displayLayer = keymanweb.core.keyboardProcessor.layerId;
```
mcdurdin, line 368: This value is never used.

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,25 @@ namespace com.keyman.text {
...
+      let matches = keyName.match(rx);
+      if (matches == null) {
+        return false;
+      }
+      displayLayer = matches[1];
+      keyName = matches[2];
+      let functionalLayer = '';
+      if (matches[3]) {
+        functionalLayer = matches[3];
+        keyName += '+' + functionalLayer;
```
mcdurdin, line 383: Why not simply `keyName += '+' + matches[3];` and remove the reference to `functionalLayer` altogether?

android/KMEA/app/src/main/assets/keyboard.html
```diff
@@ -248,7 +248,8 @@
         var keyPos = x.toString() + ',' + y.toString();
         for(i=0; i<obj.length; i++)
         {
-          s=s+obj[i].layer+'-'+obj[i].coreID;
+          // elementID contains the layer and coreID
+          s=s+obj[i].elementID;
```
mcdurdin, line 252: I don't understand why we are making this change? Is this a separate fix bundled in?

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,25 @@ namespace com.keyman.text {
...
-        keyName = keyName.substring(separatorIndex+1);
+      // Regex for 'display layer'-'virtual key name'+'optional functional layer'
+      // Can't just split on '-' because some layers like ctrl-shift contain it.
+      // Virtual key can be T_, U_, K_, or ISO 9995
+      let rx = new RegExp("(.*)-([TKU]_[^+]*|[A-E]\\d\\d)\\+?(.*)?");
+      let matches = keyName.match(rx);
+      if (matches == null) {
+        return false;
+      }
+      displayLayer = matches[1];
```
mcdurdin, line 378: This is never used.

web/source/kmwembedded.ts
```diff
@@ -361,14 +362,25 @@ namespace com.keyman.text {
...
+      keyName = matches[2];
+      let functionalLayer = '';
+      if (matches[3]) {
+        functionalLayer = matches[3];
+        keyName += '+' + functionalLayer;
       }
```
mcdurdin, line 384:
```suggestion
      keyName = matches[2] + (matches[3] ? '+' + matches[3] : '');
```

Additional optimisation possible: if we include the `+` in the final regex capture group, we can simplify to:

```
      const matches = keyName.match(/^(.+)-([TKU]_[^+]+)\+?(.+)?$/);
      if (matches == null) {
        return false;
      }
      // matches[1]: displayLayer
      // matches[2]: keyId
      // matches[3]: functionalLayer
      keyName = matches[2] + matches[3];
```

Note I am including the extra detail in comments so that it is compiled out to reduce the size of the compiled code -- critical for web.

mcdurdin (changes requested):

android/KMEA/app/src/main/assets/keyboard.html
darcywong00: It's from #5651 that corresponds @jahorton 's DM to me about activeLayout.ts
https://github.com/keymanapp/keyman/blob/6c17ecbd5f3f4ccd47d2330944a513125d06af1a/common/core/web/keyboard-processor/src/keyboards/activeLayout.ts#L112-L118

> This should be s=s+obj[i].elementID .

web/source/kmwembedded.ts
darcywong00: I've incorporated all your comments above in the latest commit.

@keymanapp/test-bot retest

mcdurdin (approved):

darcywong00:
Trying to trigger retest
@keymanapp-test-bot retest

MakaraSok:
- **TEST_ANDROID: PASSED** Everything is still working as expected.
- **TEST_IOS: PASSED** Everything is still working as expected.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.180-alpha](https://keyman.com/downloads/releases/alpha/15.0.180)

PR #6111, opened 2022-01-16
fix(developer): repatch #6074

mcdurdin:
The previous patch in #6091 was not _quite_ correct. Unfortunately, this code path is not executed when we do a test build, but only on release builds, because we don't want to be publishing to npm for test builds...

@keymanapp-test-bot skip

### DIFF
--- a/developer/js/bundle.sh
+++ b/developer/js/bundle.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -eu
+set -x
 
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.179-alpha](https://keyman.com/downloads/releases/alpha/15.0.179)

PR #6122, opened 2022-01-17
feat(windows): emit keystrokes from imx callbacks

### LINKED ISSUE #6099: feat(windows):  Emit keystrokes from IMX dll call backs - keystroke processing optimization  🥑
The Simplified Chinese Keyboard wants to generate an "Enter" key using KEY_DOWN and KEY_UP actions, to work with the core integration we need to convert this to an emit keystroke action. 
This however is hard to do with the IMX dll (3rd party library) and the way the windows API works. Windows wants to know on the first test of keystroke if the Keyman will consume the keystroke or not. Currently, we can't know this without actually calling back to the IMX dll. However, we can't do that twice.  -- Essentially we will use this to make an optimisation in which the core processing of each keystroke will only be done once and cache the result. Actions will be applied on the second pass off the keystroke.

rc-swag:
Fixes #6099


rc-swag:
This had way to many conflicts with the branch it was cut from even a rebase was going to take to longer. It was quicker to cut a new branch.

PR #6123, opened 2022-01-18
fix(web): Use regex to determine display layer and functional layers 🍒 

darcywong00:
Follow-on to #5651 and 🍒 -pick of #6100 to stable-14.0

This uses a regex to parse `keyName` which contains:
[display layer]-[virtual key name]+[optional function layer].

The 14.0 code doesn't have the KMW refactoring so `displayLayer` and `functionalLayer` are separately used later in kmwembedded.ts (ll. 439-440).

 (The original PR just used the `matches[]` components without saving off `displayLayer` / `functionalLayer`)

## User Testing
Setup:
For each platform test, install the following keyboards (some were used in testing #5651):
1. ukwuani (zip of [source](https://github.com/keymanapp/keyman/files/7845806/ukwuani.zip))
2. sil_madi
3. sil_cameroon_qwerty

* **TEST_ANDROID**
1. Launch Keyman
2. Test and verify sil_euro_latin keyboard:
    a. Use the globe key to switch to sil_euro_latin keyboard
    b. On the "default" layer, longpress a key and select
    c. Verify the output character matches the longpress key
    d. Repeat for "shift" and "symbol" layers
3. Test and verify ukwuani keyboard:
    a. Use the globe key to switch to ukwuani keyboard
    b. On the default layer, Longpress on the <kbd>o</kbd>, <kbd>i</kbd>, <kbd>u</kbd>, <kbd>e</kbd>, or <kbd>n</kbd> keys and select the longpress key
    c. Verify the output character matches the longpress key
    d. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    e. Longpress on the <kbd>O</kbd>, <kbd>I</kbd>, <kbd>U</kbd>, <kbd>E</kbd>, or <kbd>N</kbd> keys and select the longpress key
    f. Verify the output character matches the longpress key
4. Test and verify sil_madi keyboard:
    a. Use the globe key to switch to sil_madi keyboard
    b. On the default layer, longpress on a vowel and select it
    c. Verify the correct character is output
    d. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    e. Longpress on the <kbd>A</kbd> and select one of the caps longpresses.
    f. Verify the correct character is output
5. Test and verify sil_cameroon_qwerty (verifies application of nextlayer)
    a. Use the globe key to switch to sil_cameroon_qwerty
    b. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    c. Longpress on the <kbd>W</kbd> or <kbd>E</kbd> keys and select one of the caps longpresses
    d. Verify the correct character is output
    e. Verify the keyboard switches to the to the default layer
    f. Press <kbd>Shift</kbd> to switch to the "Shift" layer
    g. Press the spacebar
    h. Verify the layer remains on "Shift"
    i. For each of the following:
        * switch to the symbolic layer
        * longpress on the base key and select any of the longpress keys
        * verify the correct character is output
        * verify the keyboard **switches** to the default layer
        **Base Key**
        <kbd><<</kbd> 
        <kbd>>></kbd> 
        <kbd>'</kbd> (single quote)
        <kbd>"</kbd> (double quote)
    j. For each of the following:
        * switch to the symbolic layer
        * longpress on the base key and select any of the longpress keys
        * verify the correct character is output
        * verify the keyboard **remains** on the symbolic layer
        **Base Key**
        <kbd>(</kbd> 
        <kbd>)</kbd> 
        <kbd>?</kbd> (Note, the basekey <kbd>?</kbd> will switch to the default layer
        <kbd>!</kbd> (Note, the basekey <kbd>!</kbd> will switch to the default layer

* **TEST_IOS**
1. Repeat the same steps as ANDROID

### DIFF
--- a/web/source/kmwembedded.ts
+++ b/web/source/kmwembedded.ts
@@ -357,7 +357,8 @@ namespace com.keyman.text {
  /**
    *  Accept an external key ID (from KeymanTouch) and pass to the keyboard mapping
    *
-   *  @param  {string}  keyName   key identifier
+   *  @param  {string}  keyName   key identifier which could contain a display layer and a "functional" layer
+   *                              e.g: 'shift-K_E+rightalt-shift'
    **/
   keymanweb['executePopupKey'] = function(keyName: string) {
       let core = (<KeymanBase> keymanweb).core;
@@ -373,27 +374,25 @@ namespace com.keyman.text {
       // Changes for Build 353 to resolve KMEI popup key issues
       keyName=keyName.replace('popup-',''); //remove popup prefix if present (unlikely)
 
+      // Regex for 'display layer'-'virtual key name'+'optional functional layer'
       // Can't just split on '-' because some layers like ctrl-shift contain it.
-      let separatorIndex = keyName.lastIndexOf('-');
-      var displayLayer = core.keyboardProcessor.layerId;
-      if (separatorIndex > 0) {
-        displayLayer = keyName.substring(0, separatorIndex);
-        keyName = keyName.substring(separatorIndex+1);
+      // Virtual key starts with T_, K_, or U_
+      // matches[1]: displayLayer
+      // matches[2]: keyId
+      // matches[3]" optional functionLayer
+      let matches = keyName.match(/^(.+)-([TKU]_[^+]+)\+?(.+)?$/);
+      if (matches == null) {
+        return false
       }
-      if(displayLayer == 'undefined') {
-        displayLayer=core.keyboardProcessor.layerId;
-      }
+      let displayLayer = matches[1];
+      keyName = matches[2] + (matches[3] ? '+' + matches[3] : '');
 
-      // Determine the "functional" layer
-      let layer = displayLayer;
-      separatorIndex = keyName.lastIndexOf('+');
-      if (separatorIndex > 0) {
-        layer = keyName.substring(separatorIndex+1);
-      }
+      // Determine the functional layer
+      let functionalLayer = (matches[3]) ? matches[3] : displayLayer;
 
       // Note:  this assumes Lelem is properly attached and has an element interface.
       // Currently true in the Android and iOS apps.
-      var Lelem=keymanweb.domManager.getLastActiveElement(),keyShiftState=com.keyman.text.KeyboardProcessor.getModifierState(layer);
+      var Lelem=keymanweb.domManager.getLastActiveElement(),keyShiftState=com.keyman.text.KeyboardProcessor.getModifierState(functionalLayer);
 
       keymanweb.domManager.initActiveElement(Lelem);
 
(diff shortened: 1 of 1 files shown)


MakaraSok:
- **TEST_ANDROID: PASSED** Everything is working as expected.
- **TEST_IOS: PASSED** Everything is working as expected.

web/source/kmwembedded.ts
```diff
@@ -373,27 +374,25 @@ namespace com.keyman.text {
...
-      if(displayLayer == 'undefined') {
-        displayLayer=core.keyboardProcessor.layerId;
+      // Virtual key starts with T_, K_, or U_
+      // matches[1]: displayLayer
+      // matches[2]: keyId
+      // matches[3]" optional functionLayer
+      let matches = keyName.match(/^(.+)-([TKU]_[^+]+)\+?(.+)?$/);
+      if (matches == null) {
+        return false
       }
```
mcdurdin, line 386: This is slightly different behaviour than the previous code. In the previous code, if there was no `displayLayer`, it would use `core.keyboardProcessor.layerId`, rather than bailing out. Do we need to maintain this distinction? (My preference is yes unless we can demonstrate that this never happens through further code analysis)

mcdurdin (changes requested):
This actually LGTM apart from the one ❓ I have regarding change in behaviour. I wonder if we should consider that same question for the 15.0 change also?

web/source/kmwembedded.ts
darcywong00: I considered using a second regex if displayLayer was undefined. I just don't know what all the possible combinations of displayLayer, keyId, and functional layer there really are.

Would be something like
```
let matches = keyName.match(/^(.+)-([TKU]_[^+]+)\+?(.+)?$/);
if (matches == null) {
   matches = keyName.match(/^()([TKU]_[^+]+)\+?(.+)?$/);   // can I use an empty group so matches[1] isn't used?
   if (matches == null) {
      // both regex failed so exit
      return false
   }
}
let displayLayer = matches[1] ? matches[1]  : core.keyboardProcessor.layerId;
keyName = matches[2] + (matches[3] ? '+' + matches[3] : '');

// Determine the functional layer
let functionalLayer = (matches[3]) ? matches[3] : displayLayer;
```

      // Determine the functional layer
      let functionalLayer = (matches[3]) ? matches[3] : displayLayer;

mcdurdin (approved):
> I just don't know what all the possible combinations of displayLayer, keyId, and functional layer there really are.

Okay, I've reviewed the web core code, specifically `elementID()` in activeLayout.ts:

https://github.com/keymanapp/keyman/blob/87b580d998be7dec76e353d85fbb4da2b3ccc7e9/common/core/web/keyboard-processor/src/keyboards/activeLayout.ts#L94-L118

and `coreID()`:

https://github.com/keymanapp/keyman/blob/87b580d998be7dec76e353d85fbb4da2b3ccc7e9/common/core/web/keyboard-processor/src/keyboards/activeLayout.ts#L62-L92

These two snippets show that we can rely on the `layer-keyid[+functionalLayer]` pattern (a key without an id per line 85 above is not going to be interactive, so we can ignore that unlikely scenario).

So what you already have should be sufficient.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.285](https://keyman.com/downloads/releases/stable/14.0.285)

PR #6134, opened 2022-01-20
chore(linux): Revert workaround for Python bug

ermshiperete:
A new version of python3-requests-cache is now available in Debian and Ubuntu that fixes the bug, so we can revert the
workaround.

This reverts commit 87d2361dbe19613985ea059adca1aed340c12ebd.

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.286](https://keyman.com/downloads/releases/stable/14.0.286)

PR #6141, opened 2022-01-23
fix(developer): handle Core failure to load

mcdurdin:
If Core fails to load when starting the debugger, avoid a crash and report details of the error back to the user.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.187-alpha](https://keyman.com/downloads/releases/alpha/15.0.187)

PR #6145, opened 2022-01-23
fix(developer): compile package fileVersion from keyboard FileVersion

### LINKED ISSUE #5992: feat(developer): Compile keyboards w/ new features with min v15
#5894 added Developer support for U_xxxx_yyyy keys. However, versions of Keyman for Android 14.0 and earlier will crash if attempting to render those subkeys.

Per @mcdurdin the compiler  should tag those keyboards as min v15.
Also applies for NewContext, PostKeystroke, &LayerChanged, and readonly group.

I'd need to check if Keyman for Android will prevent installing packages with versions > than itself.

mcdurdin:
Fixes #5992.

Updates package fileVersion property to be minimum keyboard version, to avoid issues with app crashing if functionality is not supported.

As far as I can tell, only Keyman for Windows currently checks the package fileVersion.

This could be quite nasty, except that the online keyboard repository build process also extracts the min Keyman version from the keyboard .kmx and .js, and records it in the .keyboard_info, preventing older versions of Keyman from seeing unsupported keyboards in the cloud repository.

Thus, the only gap is for peer distributed keyboards. So we should fix this gap but it is less urgent than it might otherwise be.

# User Testing 

## SUITE_COMPILER: Test build of package

**TEST_KMX:** Validate that the package compiler sets the right FileVersion from .kmx.

1. Create a keyboard, set `store(&version) "15.0"`, compile it and include only the resulting keyboard .kmx in a package. Compile the package.
2. Rename the resulting .kmp to a .zip, and look at the kmp.json file in it. It should have, among other content:

   ```json
   "system": {
     "fileVersion": "15.0"
   },
   ```

3. Look at the .kmp.inf file inside the same .kmp. It should have, among other content:

   ```
   [Package]
   Version=15.0
   ```
   
**TEST_JS:** Validate that the package compiler sets the right FileVersion from .js.

1. Create a keyboard, set `store(&version) "15.0"`, compile it and include only the resulting keyboard .js in a package. Compile the package.
2. Rename the resulting .kmp to a .zip, and look at the kmp.json file in it. It should have, among other content:

   ```json
   "system": {
     ...
     "fileVersion": "15.0"
   },
   ```

3. Look at the .kmp.inf file inside the same .kmp. It should have, among other content:

   ```
   [Package]
   Version=15.0
   ```

## SUITE_APPS: Test installation of package

I am expecting four out of five of these tests to fail. Do not be concerned :grin:

Take the package you created in SUITE_COMPILER, but add both the .js and the .kmx to it and rebuild it. Attempt to install it on each platform:

**TEST_KEYMAN_ANDROID:** Install the package on Keyman 14. It should not install (but probably will...)

**TEST_KEYMAN_IOS:** Install the package on Keyman 14. It should not install (but probably will...)

**TEST_KEYMAN_LINUX:** Install the package on Keyman 14. It should not install (but probably will...)

**TEST_KEYMAN_MAC:** Install the package on Keyman 14. It should not install (but probably will...)

**TEST_KEYMAN_WINDOWS:** Install the package on Keyman 14. It should not install.

### DIFF
--- a/windows/src/global/delphi/general/CompilePackage.pas
+++ b/windows/src/global/delphi/general/CompilePackage.pas
@@ -57,7 +57,8 @@ uses
   Keyman.System.PackageInfoRefreshLexicalModels,
 
   RedistFiles,
-  TempFileManager;
+  TempFileManager,
+  VersionINfo;
 
 type
   TCompilePackage = class
@@ -282,19 +283,38 @@ begin
 
     kmpinf.RemoveFilePaths;
 
-    if kmpinf.LexicalModels.Count = 0
-      then kmpinf.Options.FileVersion := SKeymanVersion70
-      else kmpinf.Options.FileVersion := SKeymanVersion120;
+    //
+    // Find the minimum package version based on keyboard version
+    // See also MergeKeyboardInfo.pas
+    //
 
-    if kmpinf.Options.FileVersion = SKeymanVersion70 then
+    if kmpinf.LexicalModels.Count = 0 then
     begin
+      kmpinf.Options.FileVersion := SKeymanVersion70;
+
+      for i := 0 to kmpinf.Keyboards.Count - 1 do
+      begin
+        if CompareVersions(kmpinf.Options.FileVersion, kmpinf.Keyboards[i].MinKeymanVersion) > 0 then
+          kmpinf.Options.FileVersion := kmpinf.Keyboards[i].MinKeymanVersion;
+      end;
+
       psf := TPackageContentFile.Create(kmpinf);
       psf.FileName := 'kmp.inf';
       psf.Description := 'Package information';
       psf.CopyLocation := pfclPackage;
 
       kmpinf.Files.Add(psf);
+    end
+    else
+    begin
+      // TODO: min model version is always 12.0 currently
+      // but we may need to support this in future
+      kmpinf.Options.FileVersion := SKeymanVersion120;
     end;
+
+    //
+    // Prepare and save
+    //
 
     psf := TPackageContentFile.Create(kmpinf);
     psf.FileName := 'kmp.json';
--- a/windows/src/developer/TIKE/compile/MergeKeyboardInfo.pas
+++ b/windows/src/developer/TIKE/compile/MergeKeyboardInfo.pas
@@ -479,7 +479,7 @@ var
   v: TJSONValue;
   alangs: TJSONArray;
   olangs, o: TJSONObject;
-  pair: TJSONPair; 
+  pair: TJSONPair;
   i: Integer;
   id: string;
 begin
@@ -858,7 +858,8 @@ begin
       MinVersion := FPackageKMXFileInfos[i].Info.FileVersion;
 
   FJSMinVersionString := '';
-  with TRegEx.Match(FJSFileInfo.Data, 'this\.KMINVER=([''"])([^''"]+)(\1)') do
+  // See also Keyman.System.KeyboardJSInfo.pas, CompilePackage.pas
+  with TRegEx.Match(FJSFileInfo.Data, 'this.KMINVER\s*=\s*([''"])(.*?)\1') do
   begin
     if Success then
     begin
--- a/windows/src/global/delphi/general/KeymanVersion.pas
+++ b/windows/src/global/delphi/general/KeymanVersion.pas
@@ -30,6 +30,7 @@ const
   TIER_STABLE = 'stable';
 
 const
+  SKeymanVersion150 = '15.0';
   SKeymanVersion140 = '14.0';
   SKeymanVersion130 = '13.0';
   SKeymanVersion120 = '12.0';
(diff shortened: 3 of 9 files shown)


windows/src/global/delphi/general/CompilePackage.pas
```diff
@@ -282,20 +283,39 @@ function TCompilePackage.BuildKMP: Boolean;
...
 
-    if kmpinf.Options.FileVersion = SKeymanVersion70 then
+    if kmpinf.LexicalModels.Count = 0 then
     begin
+      kmpinf.Options.FileVersion := SKeymanVersion70;
+
+      for i := 0 to kmpinf.Keyboards.Count - 1 do
+      begin
+        if CompareVersions(kmpinf.Options.FileVersion, kmpinf.Keyboards[i].MinKeymanVersion) > 0 then
+          kmpinf.Options.FileVersion := kmpinf.Keyboards[i].MinKeymanVersion;
```
darcywong00, line 298: I see it in the code here. Should we also have the user test verify that if a package contains multiple keyboards, the package fileVersion will be the minimum `MinKeymanVersion` among the keyboards? (e.g. among 7.0, 12.0, 15.0)

mcdurdin: It should be the maximum, not the minimum; I think the code is clear enough and I don't see an urgent need to user test it -- tester has enough to do as it is I think.

darcywong00 (approved):
lgtm

MakaraSok:
## SUITE_COMPILER: Test build of package
- **TEST_KMX: FAILED** There is an error when trying to compile the keyboard package of just the .kmx file.
```
pr6145.kps (47): Error: 0001 Validation error: Element 'MinKeymanVersion' is unexpected according to content model of parent element 'Keyboard'.   
pr6145.kps: Failure: Package C:\Users\Keyman\Documents\Keyman Developer\Projects\pr6145\source\pr6145.kps had validation errors.
```
Here is the keyboard project for testing.
[pr6145.zip](https://github.com/keymanapp/keyman/files/7939816/pr6145.zip)


- **TEST_JS: BLOCKED**


## SUITE_APPS: Test installation of package
- **TEST_KEYMAN_ANDROID: BLOCKED
- **TEST_KEYMAN_IOS: BLOCKED
- **TEST_KEYMAN_LINUX: BLOCKED
- **TEST_KEYMAN_MAC: BLOCKED
- **TEST_KEYMAN_WINDOWS: BLOCKED

mcdurdin:
I have corrected the logic around writing out the keyboard package source files. You should recreate the package from scratch as the existing one is now invalid.

We should also ensure that packages for 14.0 and earlier compiled with the 15.0 compiler still install on earlier versions of Keyman -- so I've added a test for that.

@keymanapp-test-bot retest

MakaraSok:
## SUITE_COMPILER: Test build of package

* **TEST_KMX: PASSED** both the kmp.json and kmp.inf do contain the expected lines.
* **TEST_JS: PASSED** both the kmp.json and kmp.inf do contain expected the expected lines.

## SUITE_APPS: Test installation of package

* **TEST_KEYMAN_ANDROID: PASSED** the install got through and the keyboard works just fine.
* **TEST_KEYMAN_IOS: BLOCKED** No stable build working to test this yet.
* **TEST_KEYMAN_WINDOWS: FAILED** fail as expected (tested in Windows 11 x64).

mcdurdin:
@MakaraSok wrote:
> * **TEST KEYMAN_WINDOWS: FAILED** fail as expected (tested in Windows 11 x64).

Given that Keyman for Windows 14 refuses to install a package intended for Keyman 15, I call that a 'pass' :grin:

SUITE_APPS
TEST_KEYMAN_WINDOWS PASSED

mcdurdin:
> * ✅ **TEST KEYMAN_ANDROID ([PASSED](https://github.com/keymanapp/keyman/pull/6145#issuecomment-1023943158))**: the install got through and the keyboard works just fine.

Okay, so this should actually be a fail, because it _shouldn't_ install! :grin:

SUITE_APPS
TEST_KEYMAN_ANDROID FAILED - Keyman for Android 14 should refuse to install a package intended for Keyman 15.

MakaraSok:
## SUITE_APPS: Test installation of package
* **TEST_KEYMAN_ANDROID: PASSED** The keyboard package got installed without any issue on the latest stable version (14.0.286) in Android 11.
* **TEST_KEYMAN_IOS: PASSED** The keyboard package got installed without any issue on the latest stable (14.0.286) in iOS 15.2.
* **TEST_KEYMAN_LINUX: BLOCKED** The stable doesn't get installed successfully. Each time the stable was installed, the alpha (15.0.169) gets in the way. The package got installed succesfully though. Mark this as BLOCKED for now.
* **TEST_KEYMAN_MAC: FAILED** The package doesn't get installed, no error shown and the keyboard doesn't appear in the keyboard menu either.

bharanidharanj:
---SUITE_APPS_14: Test installation of 14.0 package into 14.0

- **TEST_KEYMAN_WINDOWS: PASSED**  The Keyman Package 14 has been installed and working as expected. 

- **TEST_KEYMAN_LINUX: BLOCKED**  Not able to install Keyman 14 since it only accepts the current version of Keyman. (ie., Keyman 15)

PR #6154, opened 2022-01-25
feat(windows): testhost project for debugging keyman32

mcdurdin:
Scratching an itch. It's very painful to debug Keyman Core on Windows. This PR is for @rc-swag.

This helps a little. It allows you to step through the code without loading keyman32.dll into every process. It's not perfect, because when actions are sent, they tend to land in the debugger because SendInput does not care about the active window! However, if you don't care about that too much (or if you are willing to try a remote debug setup), then this should really simplify debugging.

@keymanapp-test-bot skip

Note: this is added into engine.sln for access through Visual Studio but is never built otherwise. It assumes that KEYMAN_ROOT is set and that you have a debug build of keyman32.dll / keyman64.dll. It may work in either 32 or 64 bit, but I have only tested 32 bit at this stage.


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.185-alpha](https://keyman.com/downloads/releases/alpha/15.0.185)

PR #6157, opened 2022-01-25
feat(windows): handle key output in testhost

mcdurdin:
In order to avoid the test host interfering too badly with the debugger, we will bypass SendInput and instead post all input to the local edit window. This can result in out-of-sequence input for rapid typing, but for a test host this is not important -- it's better that the input doesn't end up in the debugger window due to focus changes!

@keymanapp-test-bot skip


rc-swag (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.185-alpha](https://keyman.com/downloads/releases/alpha/15.0.185)

PR #6179, opened 2022-01-28
fix(linux): Don't crash if we lack permissions

### LINKED ISSUE #6171: bug(linux): Crash trying to add .json file to shared directory
When initializing km-config we go through the list of installed keyboards and mark all that are missing .json files. Later on we download and store the missing information. This fails for keyboards stored in system directories.

Sentry Issue: [KEYMAN-LINUX-V](https://sentry.io/organizations/keyman/issues/2918936111/?referrer=github_integration)

```
PermissionError: [Errno 13] Permission denied: '/usr/local/share/keyman/khmer_angkor/khmer_angkor.json'
  File "keyman_config/install_kmp.py", line 255, in process_keyboard_data
    with open(os.path.join(packageDir, keyboardID + '.json'), 'w') as outfile:
  File "keyman_config/get_info.py", line 31, in run
    process_keyboard_data(kmp['packageID'], packageDir)
  File "threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
```

ermshiperete:
Also deal with other file I/O related exceptions and report them as warnings. Also improve general robustness if we encounter a corrupt file.

This fixes #6171.

This will be tested as part of the pre-beta acceptance testing.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/install_kmp.py
+++ b/linux/keyman-config/keyman_config/install_kmp.py
@@ -249,12 +249,19 @@ def extract_kmp(kmpfile, directory):
 def process_keyboard_data(keyboardID, packageDir):
     kbdata = get_keyboard_data(keyboardID)
     if kbdata:
-        if not os.path.isdir(packageDir):
-            os.makedirs(packageDir)
+        if not os.path.isdir(packageDir) and os.access(os.path.join(packageDir, os.pardir), os.X_OK | os.W_OK):
+            try:
+                os.makedirs(packageDir)
+            except Exception as e:
+                logging.warning('Exception %s creating %s %s', type(e), packageDir, e.args)
 
-        with open(os.path.join(packageDir, keyboardID + '.json'), 'w') as outfile:
-            json.dump(kbdata, outfile)
-            logging.info("Installing api data file %s.json as keyman file", keyboardID)
+        if os.access(packageDir, os.X_OK | os.W_OK):
+            try:
+                with open(os.path.join(packageDir, keyboardID + '.json'), 'w') as outfile:
+                    json.dump(kbdata, outfile)
+                    logging.info("Installing api data file %s.json as keyman file", keyboardID)
+            except Exception as e:
+                logging.warning('Exception %s writing %s/%s.json %s', type(e), packageDir, keyboardID, e.args)
     # else:
     # 	message = "install_kmp.py: error: cannot download keyboard data so not installing."
     # 	rmtree(kbdir)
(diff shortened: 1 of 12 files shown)


linux/keyman-config/keyman_config/install_kmp.py
```diff
@@ -249,12 +249,19 @@ def extract_kmp(kmpfile, directory):
...
+            except Exception as e:
+                logging.warning('Exception %s creating %s %s', type(e), packageDir, e.args)
+
+        if os.access(packageDir, os.X_OK | os.W_OK):
+            try:
+                with open(os.path.join(packageDir, keyboardID + '.json'), 'w') as outfile:
+                    json.dump(kbdata, outfile)
+                    logging.info("Installing api data file %s.json as keyman file", keyboardID)
+            except Exception as e:
+                logging.warning('Exception %s writing %s/%s.json %s', type(e), packageDir, keyboardID, e.args)
```
mcdurdin, line 264: What happens in the failure case here?

linux/keyman-config/keyman_config/install_window.py
```diff
@@ -80,8 +80,8 @@ def __init__(self, kmpfile, online=False, viewkmp=None, language=None):
                 self.kbname = keyboardid
             self.checkcontinue = True
 
-            if installed_kmp_ver and info and 'version' in info and 'description' in info['version']:
-                if info['version']['description'] == installed_kmp_ver:
+            if installed_kmp_ver and secure_lookup('version', 'description'):
```
mcdurdin, line 83:
```suggestion
            if installed_kmp_ver and secure_lookup(info, 'version', 'description'):
```

linux/keyman-config/km-package-install
```diff
@@ -137,17 +140,18 @@ def main():
...
-            if kbdata['version'] == installed_kmp_ver:
+        kbdata_version = secure_lookup(kbdata, 'version')
+        if installed_kmp_ver and kbdata_version:
+            if kbdata_version == installed_kmp_ver:
                 logging.error(
                     "km-package-install: The %s version of the %s keyboard package is already installed.",
                     installed_kmp_ver, args.package)
                 sys.exit(1)
-            elif float(kbdata['version']) > float(installed_kmp_ver):
+            elif float(kbdata_version) > float(installed_kmp_ver):
```
mcdurdin, line 150: I was not aware that version numbers were being cast as floats -- that's not safe as version number strings can contain multiple sections, e.g. `1.0.1`.

mcdurdin (changes requested):
I like the additional safety checks.

I'm a little uncomfortable with the failure modes for caught exceptions -- what is the flow on effect of handling the exception instead of failing? If a kmp.json file fails to parse, we need to abort and not attempt to install, etc.

A couple of other things noted -- one not new, just spotted when I was reviewing changes, the other just a typo.

mcdurdin (approved):

PR #6183, opened 2022-01-31
chore(common/core/desktop): Improve meson build file 🏗️

ermshiperete:
Based on the suggestions of @eli-schwartz (in #5797).

Unfortunately we currently still have to support meson 0.45 (because that's what ships with Ubuntu 18.04), so we can't
implement everything suggested.

Also fixed some warnings triggered by the changes.

@keymanapp-test-bot skip

### DIFF
--- a/common/core/desktop/build.bat
+++ b/common/core/desktop/build.bat
@@ -103,8 +103,9 @@ rem ----------------------------------
 
 rem Standalone build, so we'll make the environment available to the caller
 rem Also setup
+rem Note: Visual Studio 2022 doesn't provide vcvarsall.bat, so we'll have to find a different solution
 endlocal
-for /f "usebackq tokens=*" %%i in (`..\..\..\resources\build\vswhere -latest -requires Microsoft.Component.MSBuild -find **\vcvarsall.bat`) do (
+for /f "usebackq tokens=*" %%i in (`..\..\..\resources\build\vswhere -version [15^,17^) -latest -requires Microsoft.Component.MSBuild -find **\vcvarsall.bat`) do (
   set VCVARSALL="%%i"
 )
 %VCVARSALL% %1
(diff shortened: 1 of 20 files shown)


mcdurdin:
context-api.cpp is failing on the last line of this, which suggests an encoding conversion issue (perhaps source file encoding? context-api.cpp file not being read as UTF-8 by compiler)?

https://github.com/keymanapp/keyman/blob/d2fdb653d1e3c0b397ca521b89b3dd3f505b74cf/common/core/desktop/tests/unit/kmnkbd/context_api.cpp#L80-L83

ctxt4 is initialized from:

```c++
  std::string const     initial_u8_bmp_context = u8"Hello, အရှောက်, मानव अधिकारों की सार्वभौम घोषणा",
                        initial_u8_smp_context = u8"😀😁😂😃😄😅😆😇😈😉😊😋😌😍😎😏";
  ...
  try_status(km_kbp_context_items_from_utf8(initial_u8_smp_context.data(), &ctxt4));
```

And state-api.cpp failing with a similar issue on L137:

https://github.com/keymanapp/keyman/blob/d2fdb653d1e3c0b397ca521b89b3dd3f505b74cf/common/core/desktop/tests/unit/kmnkbd/state_api.cpp#L131-L137

darcywong00:
Should this get rebased to `beta` or will it get 🍒 picked later?

ermshiperete:
In a [discussion on Slack](https://keymanapp.slack.com/archives/C6Q9WS09G/p1646853377150979?thread_ts=1646837610.797189&cid=C6Q9WS09G) we decided that this will only be on `master`:

```
marc
Yes, let's keep this targeting master .  Seems like it's still a bit wobbly, too!
From a thread in general | Mar 9th | View reply
```

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.3-alpha](https://keyman.com/downloads/releases/alpha/16.0.3)

PR #6184, opened 2022-01-31
chore(linux): stop releasing KMFL together with Keyman

### LINKED ISSUE #6163: chore(linux): stop releasing KMFL together with Keyman
From @ermshiperete:

Currently we release KMFL in sync with Keyman although we longer actively develop it. There have been no changes since v14 other than packaging related commits.

This raises the question if we should stop releasing KMFL. The Linux users who still need it would still have access to it since the v14 packages won’t go away.

Not releasing it with v15 would save time with packaging, especially when uploading to Debian since there are 3 less packages that would need to be reviewed each time we upload a new version to Debian.

KMFL is self-contained, so any changes to Keyman won’t affect it. This means we don’t gain anything from regularly building packages for KMFL, and there are no benefits to the user. In the unlikely event that we’d have to fix a bug in KMFL we’d still be able to do that and manually release packages.

ermshiperete:
Fixes #6163.

@keymanapp-test-bot skip

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -7,5 +7,5 @@
 keymanPackaging {
   distributionsToPackage = 'bionic focal impish jammy'
   arches = 'amd64 i386'
-  packagesToBuild = ['keyman', 'kmflcomp', 'libkmfl', 'ibus-kmfl']
+  packagesToBuild = ['keyman']
 }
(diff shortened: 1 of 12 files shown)


mcdurdin (approved):
LGTM

There are a lot of places where the same logic is being applied in script. I wonder if it makes sense to have a projects.inc.sh script (this could be left for another day)?

Something along the lines of:

```shell
projects=keyman
legacy_projects=""
if [ -n "$BUILD_LEGACY" ]; then
    legacy_projects="kmflcomp libkmfl ibus-kmfl"
fi
all_projects="$projects $legacy_projects"
```

(but I think it'd be a bit more complex than that)

PR #6197, opened 2022-02-03
feat(oem/fv/ios): Dictionary Support

sgschantz:
Adds dictionary/lexical model support to FirstVoices iOS app.
Tapping a keyboard in the list causes transition to a detail page where the keyboard is displayed with other settings, including downloadable dictionaries, if available.

keyboard table

image

# User Testing

The changes to the app are substantial and tests amount to a regression test

* TEST_INSTALL: Verify that a FirstVoices keyboard is visible in iOS

1. Go to iOS Settings -> General -> Keyboard -> Keyboards -> Add New Keyboard... and choose FirstVoices under the heading Third-Party Keyboards.
2. Go to the Messages app
3. Tap in the text edit field to begin typing a new message so that the system keyboard will appear
4. Tap on the globe icon
5. Verify that the keyboard has changed to 'EuroLatin (SIL)', available due to the FirstVoices install

* TEST_TURN_ON_KEYBOARD: Verify that another FV keyboard can be added

1. Open the FirstVoices app and tap the 'here' button on the main instructions page
2. Tap the first keyboard in the list which will probably be 'Inuvialuktun' -- tap anywhere but on the information icon
3. This should cause a transition to the settings page for Inuvialuktun
4. Tap the uppermost switch to turn on this keyboard
5. Tap the back arrow in the upper left arrow to return to the keyboards page
6. On the keyboards page, verify that Inuvialuktun now has a checkmark next to it
7. Switch to messages and verify that the Inuvialuktun keyboard is available from the globe menu

* TEST_KEYBOARD_HELP: Verify that the help page is available

1. For any keyboard listed in the keyboards page, tap the information icon 
2. Verify that a web page loads with helpful information about the keyboard that was selected

* TEST_INSTALL_DICTIONARY: Verify that a dictionary can be installed and used

1. On the keyboards page, scroll down and tap SENĆOŦEN under the BC Coast region heading
2. Tap the switch to activate the SENĆOŦEN keyboard
3. Verify that a switch is now available to activate the dictionary 'SENĆOŦEN (Saanich Dialect)'
4. Tap the switch and choose Install on the resulting confirmation dialog
5. Verify that related dictionary settings Suggest Predictions and Suggest Corrections have defaulted to the On position
6. Open Messages and after switching to the SENĆOŦEN keyboard, verify that words are suggested while typing
image

* TEST_STATE_SAVED: Verify that the user's settings are stored safely between app invocations

1. Make changes to various settings within the app, such as turning on certain keyboards and specifying different dictionary settings for the SENĆOŦEN keyboard
2. Quit the app and restart it
3. Verify that the previous settings are preserved

* TEST_SWITCH_DEPENDENCIES: Verify that switches are only available as appropriate

1. On the keyboards page, scroll down and tap SENĆOŦEN under the BC Coast region heading
2. When the keyboard switch is in the Off position, verify that all other switches are Off and disabled (cannot be toggled)
3. Tap the 'SENĆOŦEN keyboard switch so that it is in the On position and verify that the 'SENĆOŦEN (Saanich Dialect)' switch is now enabled
4. Tap the 'SENĆOŦEN (Saanich Dialect)' switch so that it is now in the On position and verify (after dictionary download) that the Suggest Predictions and Suggest Corrections switches automatically default to the On position
5. Tap the Suggest Predictions switch to the off position and verify that Suggest Corrections also moves to the off position where it is now disabled


darcywong00:
UX question: Are the defaults on "Suggest Predictions" and "Suggest Corrections" off on Keyman for iPhone and iPad?

I think on Android, we default to "On" (less clicking after the user installs a lexical-model)

sgschantz:
In Keyman for iOS, I installed a new keyboard, and the switches for Predictions and Corrections were active and enabled by default. That was even with no dictionary installed.

That doesn't make much sense as those options should not be available until there is a dictionary, right? After a dictionary is installed, we can activate them and default them as enabled. Until then they should be disabled and inactive.

sgschantz:
A few notes on the details page:
1. I labeled the switches Suggest Predictions and Suggest Corrections, using 'Suggest' instead of 'Enable' to try to be a little more descriptive. Switches always enable and disable things, so instead I wanted to describe what was being enabled or disabled. If there is a better word than 'suggest', let me know.
2. After changing the enabling/graying out of controls, I think the Downloadable Dictionaries section should be moved up above the Suggest switches because the Suggest switches are dependent upon whether there is an active dictionary.
3. Now we are calling the Suggest switches 'Language Settings' -- should they be Dictionary Settings instead? I think that would reinforce their purpose and make it more obvious why they are disabled when there is no dictionary.

The above changes differ from what we have in Keyman, but I think it would improve clarity of the UI.

As for how the state of the controls depend on each other, there is a pretty clear hierarchy:
1. If the keyboard switch is off, then all other controls are off and disabled.
2. If there is no dictionary downloaded or there is a dictionary and it is turned off, then both Suggest switches are off and disabled.
3. If the Suggest Predictions switch is off, then the Suggest Corrections switch is off and disabled. (This is currently enforced in Keyman. Keyman also hides the switch, but that is not necessary.)

oem/firstvoices/ios/FirstVoices/ViewController.swift
```diff
@@ -62,7 +62,8 @@ class ViewController: UIViewController, UIWebViewDelegate {
   func webView(_ webView: UIWebView, shouldStartLoadWith request: URLRequest, navigationType: UIWebView.NavigationType) -> Bool {
     let fragment: String? = request.url!.fragment
     if fragment?.range(of: "showKeyboards") != nil {
-      self.performSegue(withIdentifier: "Keyboards", sender: nil)
+      self.performSegue(withIdentifier: "keyboardList", sender: nil)
+ //     self.performSegue(withIdentifier: "Keyboards", sender: nil)
```
darcywong00, line 66: Can this be removed?

oem/firstvoices/ios/FirstVoices/LexicalModelRepository.swift
```diff
@@ -0,0 +1,83 @@
...
+    do {
+      let lexicalModelData = try Data(contentsOf: keymanApiUrl, options: NSData.ReadingOptions())
+
+      do {
+          // make sure this JSON is in the format we expect
+          if let jsonArray = try JSONSerialization.jsonObject(with: lexicalModelData, options: []) as? [[String : Any]] {
+              // try to read array
+            if !jsonArray.isEmpty {
+              let jsonModelMap = jsonArray.first
+              let name = jsonModelMap!["description"] as! String
```
darcywong00, line 43: Shouldn't name be?
```suggestion
              let name = jsonModelMap!["name"] as! String
Write Preview
```

oem/firstvoices/ios/FirstVoices/LexicalModelRepository.swift
```diff
@@ -0,0 +1,83 @@
...
+ * Class to call the Keyman API to query about lexical models for a given language.
+ *
+ */
+
+import UIKit
+import KeymanEngine
+
+class LexicalModelRepository {
+
+  private static let keymanLexicalModelApiUrl = "https://api.keyman.com/model?q=bcp47:"
```
darcywong00, line 20: I wonder if some of this should use existing methods in

https://github.com/keymanapp/keyman/blob/master/ios/engine/KMEI/KeymanEngine/Classes/Queries/Queries%2BLexicalModel.swift#L45

darcywong00:
iOS is using an old script to download fv_all.kmp
https://github.com/keymanapp/keyman/blob/7e43e3782dce134dcfe0b69b09541898203773cb/oem/firstvoices/ios/build.sh#L9-L10

We can remove `build_keyboards.sh` if you include the following
```bash
. "$KEYMAN_ROOT/resources/build/build-download-resources.sh"
```
and then do
https://github.com/keymanapp/keyman/blob/7e43e3782dce134dcfe0b69b09541898203773cb/oem/firstvoices/android/build.sh#L95-L101

oem/firstvoices/ios/FirstVoices/LexicalModelRepository.swift
sgschantz: Good question. I used the description field because it had the more succinct name:
        "description": "SEN\u0106O\u0166EN (Saanich Dialect)",
vs
        "name": "SEN\u0106O\u0166EN (Saanich Dialect) Lexical Model"

Not sure what I can expect from each field of the API, but this text goes in the switch name, so I need something brief. Is the lexical model definition incorrect?

mcdurdin: You should be using `name`. `description` is a long form which may be several lines long. It just so happens that the SNECOTEN model has a overly short description:

https://github.com/keymanapp/lexical-models/blob/fb306c0768caaf6f25cbb7abf9d7e4eb3ca74ef3/release/nrc/nrc.str.sencoten/nrc.str.sencoten.model_info#L3

vs name:

https://github.com/keymanapp/lexical-models/blob/fb306c0768caaf6f25cbb7abf9d7e4eb3ca74ef3/release/nrc/nrc.str.sencoten/source/nrc.str.sencoten.model.kps#L10

See for example https://api.keyman.com/model?q=bcp47:cmo-Khmr:

```json
[
    {
        "license": "mit",
        "languages": [
            "cmo-khmr"
        ],
        "description": "Bunong Wordlist is based on the Bible wordlist of 1,157 unique entries with their frequencies.",
        "id": "sil.cmo.bw",
        "name": "Bunong Wordlist",
        "authorName": "[object Object]",
        "lastModifiedDate": "2022-02-18T18:41:02.203Z",
        "packageFilename": "https://keyman.com/go/package/download/model/sil.cmo.bw?version=1.1&update=1",
        "packageFileSize": 107384,
        "jsFilename": "https://downloads.keyman.com/models/sil.cmo.bw/1.1/sil.cmo.bw.model.js",
        "jsFileSize": 286398,
        "packageIncludes": [],
        "version": "1.1",
        "minKeymanVersion": "12.0",
        "sourcePath": "release/sil/sil.cmo.bw"
    }
]
```

mcdurdin (approved):
This looks really good, apart from that one nit that Darcy caught. 

> I labeled the switches Suggest Predictions and Suggest Corrections, using 'Suggest' instead of 'Enable' to try to be a little more descriptive. Switches always enable and disable things, so instead I wanted to describe what was being enabled or disabled. If there is a better word than 'suggest', let me know.

I like it 😻

> After changing the enabling/graying out of controls, I think the Downloadable Dictionaries section should be moved up above the Suggest switches because the Suggest switches are dependent upon whether there is an active dictionary.

Agreed.

> Now we are calling the Suggest switches 'Language Settings' -- should they be Dictionary Settings instead? I think that would reinforce their purpose and make it more obvious why they are disabled when there is no dictionary.

Yep.

> The above changes differ from what we have in Keyman, but I think it would improve clarity of the UI.

Those changes should probably feed back to Keyman. Worth considering feeding the same changes to Keyman for Android (@darcywong00 do you concur?)

> As for how the state of the controls depend on each other, there is a pretty clear hierarchy:

Yep, this is very clear. I'm enjoying your UX design skills!

MakaraSok:
# Test Results

Tested with the build retrieved from the link provided in the Test Results above.

* **TEST_INSTALL (PASSED):** The "English - EuroLatin (SIL)" does appear following the instructions given.


* **TEST_TURN_ON_KEYBOARD (FAILED):** The "Inuvialuktun" keyboard doesn't show up when either pressing on the globe key or longpress on the globe key. The FV keyboard is stuck with "English - EuroLatin (SIL)".


* **TEST_KEYBOARD_HELP (PASSED):** The keyboard help page does show up.
image:** The "SENĆOŦEN" keyboard doesn't show up when either pressing on the globe key or longpress on the globe key. The FV keyboard is stuck with "English - EuroLatin (SIL)".


* **TEST_STATE_SAVED (BLOCKED):** The FV keyboard doesn't show anything else beside "English - EuroLatin (SIL)". 

Note: At one point, iMessage crashes when pressing on the FV globe key > Emoji key:



* **TEST_SWITCH_DEPENDENCIES (BLOCKED):** The FV keyboard doesn't show anything else beside "English - EuroLatin (SIL)".

sgschantz:
@MakaraSok I haven't been able to reproduce your trouble installing a keyboard, but I made one change to at least display an alert if it fails. Also, the switch should not change state now if the install fails. Please try again and let me know if the behavior is any different. Is FV Settings set to full access?

@mcdurdin What should I do with any error information that I get back from KeymanEngine so that we can investigate this? Without Sentry in FV, I'm not sure where to send it. Is there any place that I can retrieve error text when Makara encounters an issue? I could pass it back and display in this new alert, but that doesn't seem too user friendly.

mcdurdin:
> What should I do with any error information that I get back from KeymanEngine so that we can investigate this? Without Sentry in FV, I'm not sure where to send it. Is there any place that I can retrieve error text when Makara encounters an issue? I could pass it back and display in this new alert, but that doesn't seem too user friendly.

I think we aim to add Sentry to FV (I have created the fv-ios Sentry project at https://sentry.io/keyman/fv-ios/getting-started/apple-ios/). Otherwise, it is difficult, and I am certainly not keen on creating a new error-reporting mechanism for this.

If adding Sentry is too big a task for now, then we can temporarily add an alert while diagnosing these issues and then remove it again?

darcywong00:
> Worth considering feeding the same changes to Keyman for Android

Yeah, we can consider that. I had a question on the SENCOTEN settings: "Keyboard Information" section. If there's multiple keyboards available for a language, which one would "Version" correspond to?

sgschantz:
@MakaraSok I've added some code temporarily for troubleshooting the issue you are have activating the keyboard. This new version will display an alert with error information. Please retest and tell me what you see. I cannot reproduce this, so let me know if you have any more details that will isolate the issue.

@keymanapp-test-bot retest [TEST_TURN_ON_KEYBOARD]

sgschantz:
@darcywong00 Good question on the version number. I think it's broken. We could display the version of the active dictionary if there is one, but that's kind of weak and still not obvious. I'm not sure what the best solution is.

mcdurdin:
> Good question on the version number. I think it's broken. 

This was obviously designed way back in the day with a one-keyboard-per-language pattern. Do we have any languages with more than one keyboard in the list? If not, then this is a brokenness that we can just ignore for now :grin:  If we do, then I suggest for now just adding a version row for each keyboard and differentiating with the keyboard name in the caption (e.g. SENCOTEN version -> 9.1.1).

The detail is of course important but we also need to balance that with addressing more serious issues... so I am comfortable with a less than 100% elegant answer for this edge case.

darcywong00:
> Do we have any languages with more than one keyboard in the list?

Looking at keyboards.csv, I see "ikt-Latn: Inuinnuaqtun (Latin)" is used by fv_inuvialuktun and fv_uummartmitutun.

But I', thinking this should be fine because the FV apps separate Region --> Keyboard --> 

so each keyboard has its own row...

MakaraSok:
# Test Results

* **TEST_TURN_ON_KEYBOARD (FAILED):** The keyboard enabled and ticked, but it doesn't show up when switch to and fro from the iOS system keyboard. EuroLatin (SIL) keyboard is always there, not expected one which have previously been selected.

https://user-images.githubusercontent.com/28331388/158724337-44331887-0b15-45fb-be31-fcf34641cc96.mov

sgschantz:
@mcdurdin So Makara is not seeing the alert that I added which implies that Keyman Engine is not throwing an error and it looks like a success to FV. Do you know of any other way to troubleshoot this in Makara's environment?
@MakaraSok can you describe what environment you're running this in or provide any more info that might help me reproduce it?

mcdurdin:
> @mcdurdin So Makara is not seeing the alert that I added which implies that Keyman Engine is not throwing an error and it looks like a success to FV. Do you know of any other way to troubleshoot this in Makara's environment?

I guess running Console on his machine and looking at logs from the simulator may provide some enlightenment?

sgschantz:
@MakaraSok @darcywong00 @mcdurdin I just realized that this problem Makara is seeing is the same as #6226 which appeared with the old FV iOS code. The change of Darcy's should not have broken the iOS FV app, should it? Or maybe that's why there was an iOS test case on #6226?

mcdurdin:
> I just realized that this problem Makara is seeing is the same as #6226 which appeared with the old FV iOS code. The change of Darcy's should not have broken the iOS FV app, should it? Or maybe that's why there was an iOS test case on #6226?

Yes, I think that this problem is not new to this PR and is specific to the simulator.

MakaraSok:
# Retest all failed/blocked

I've retested this again today and it doesn't work as expected like what was experienced prior. Note that the test was done on a simulator with its "all content and settings" erased being installing the .

- **TEST_TURN_ON_KEYBOARD: FAILED** Please see the screen recording below.

https://user-images.githubusercontent.com/28331388/159197472-9e930029-c823-4024-94db-f73566fe8dff.mov

- **TEST_INSTALL_DICTIONARY: FAILED** The very attempt to install the dictionary disable itself after clicking on the "install" button. Please see the screen recording below.


https://user-images.githubusercontent.com/28331388/159198672-6246bc9e-a291-4459-864d-c3bde4df96ad.mov

The second attempt to enable the dictionary got through without bouncing back, but it still doesn't show up when summoned. Please see the screen recording below.


https://user-images.githubusercontent.com/28331388/159199192-e3482bb9-e753-4d52-8003-5c0b3c6ca467.mov

The blocked ones stay "Blocked".

MakaraSok:
FYI: The build from here was used https://build.palaso.org/buildConfiguration/Keyman_iOS_TestPullRequests/310924?buildTab=artifacts#%2Fupload.

sgschantz:
Ready for retest, but need to have a way to either test a release version of FV on the simulator or a real device. Debug FV on the simulator works well from Xcode. @mcdurdin any ideas on how to change the relase build so that it will behave the same on simulator and device?
@keymanapp-test-bot retest [TEST_TURN_ON_KEYBOARD]
@keymanapp-test-bot retest [TEST_INSTALL_DICTIONARY]

mcdurdin:
> ny ideas on how to change the relase build so that it will behave the same on simulator and device?

It's really hard. The simulator does not appear to behave identically to devices in all circumstances -- particularly that boundary between apps and app extensions seems different.

The best solution may be to merge to beta and do a beta release build which is then available in TestFlight. Otherwise, unless you are in the same room (where you can plug the device into your computer and do a build and run from XCode), I don't know how it can be done.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.224-beta](https://keyman.com/downloads/releases/beta/15.0.224)

MakaraSok:
# Test Results

* **TEST_TURN_ON_KEYBOARD (FAILED):** Right after a keyboard was selected/installed, it doesn't appear in the keyboard menu unless restarting the device.
* **TEST_INSTALL_DICTIONARY (FAILED):** The same behavior mentioned above.

_Note that when removing a keyboard from the FV app, every keyboard was removed with it leaving only EuroLatin keyboard to choose from._

MakaraSok:
# Test Results

* **TEST_TURN_ON_KEYBOARD (PASSED):** no issue found
* **TEST_INSTALL_DICTIONARY (PASSED):** no issue found
* **TEST_STATE_SAVED (PASSED):** no issue found
* **TEST_SWITCH_DEPENDENCIES (PASSED):** no issue found

PR #6202, opened 2022-02-03
feat(linux): Allow to run ibus tests on Wayland 🛤️

ermshiperete:
This change modifies the `run-tests.sh` script so that tests run on both Wayland (if supported) and X11 unless `--no-wayland` or `--no-x11` is specified.

This change also adds the new argument `--wayland` to the `ibus-keyman-tests` executable which runs the tests under
Wayland.

Note that running Wayland tests is only possible in Ubuntu 21.10+ because it requires a version of `mutter` that supports the headless option which got added in mutter 40.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -275,11 +275,13 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
   GdkDisplay *gdkDisplay = gdk_display_get_default();
 #ifdef GDK_WINDOWING_X11
   if (GDK_IS_X11_DISPLAY(gdkDisplay)) {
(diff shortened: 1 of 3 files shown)


keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.216-beta](https://keyman.com/downloads/releases/beta/15.0.216)

ermshiperete:
Note that running the tests with Wayland requires to use Ubuntu 21.10 or newer.

darcywong00:
Should we document the test arguments or mutter version in the readme?
Maybe at
https://github.com/keymanapp/keyman/tree/master/linux#linux-requirementssetup

Issue #6216, opened 2022-02-07
bug(android): failing to push to Play Store


mcdurdin:
**Describe the bug**

https://build.palaso.org/viewLog.html?buildId=306168&buildTypeId=KeymanAndroid_Build&tab=buildLog&_focus=1304

```
[05:11:17][Step 5/6] 
[05:11:17][Step 5/6] FAILURE: Build failed with an exception.
[05:11:17][Step 5/6] 
[05:11:17][Step 5/6] * What went wrong:
[05:11:17][Step 5/6] Execution failed for task ':commitEditForComDotFirstvoicesDotKeyboards'.
[05:11:17][Step 5/6] > A failure occurred while executing com.github.triplet.gradle.play.tasks.CommitEdit$Committer
[05:11:17][Step 5/6]    > 400 Bad Request
[05:11:17][Step 5/6]      POST https://androidpublisher.googleapis.com/androidpublisher/v3/applications/com.firstvoices.keyboards/edits/05954002417277635341:commit
[05:11:17][Step 5/6]      {
[05:11:17][Step 5/6]        "code": 400,
[05:11:17][Step 5/6]        "errors": [
[05:11:17][Step 5/6]          {
[05:11:17][Step 5/6]            "domain": "global",
[05:11:17][Step 5/6]            "message": "Changes cannot be sent for review automatically. Please set the query parameter changesNotSentForReview to true. Once committed, the changes in this edit can be sent for review from the Google Play Console UI.",
[05:11:17][Step 5/6]            "reason": "badRequest"
[05:11:17][Step 5/6]          }
[05:11:17][Step 5/6]        ],
[05:11:17][Step 5/6]        "message": "Changes cannot be sent for review automatically. Please set the query parameter changesNotSentForReview to true. Once committed, the changes in this edit can be sent for review from the Google Play Console UI.",
[05:11:17][Step 5/6]        "status": "INVALID_ARGUMENT"
[05:11:17][Step 5/6]      }
```

---

**Keyman for iPhone/iPad/Android:**
 - Keyman version: 15.0.193-alpha

mcdurdin:
Possibly related: https://github.com/Triple-T/gradle-play-publisher/issues/956

We may need to upgrade gradle-play-publisher to 3.5.0 or later? https://github.com/Triple-T/gradle-play-publisher/releases:

> **3.5.0**
> * A Google API change caused publishing to fail if the app was under review. This is now fixed.

PR #6217, opened 2022-02-08
chore(android): Update Play publishing plugin to 3.5.0

darcywong00:
This is an attempt to address #6216 where the FV Android app failed to publish to the Play Store.
Will see after the next alpha build goes to the Play Store.

Upgrading to 3.6.0+ will involve more pain since those 
> release are only compatible with the Android Gradle Plugin v7.x.

@keymanapp-test-bot skip


mcdurdin (approved):

mcdurdin:
Please 🍒 to stable-14.0 😁

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.194-alpha](https://keyman.com/downloads/releases/alpha/15.0.194)

PR #6219, opened 2022-02-08
fix(windows): add flag to track core process event

### LINKED ISSUE #6205: bug(windows): keyman outputs characters twice for apps using TSF
Found in pre-beta acceptance testing: #6192.
Effects Keyman for windows 15.0 .189 running the core processor.

When using apps that support the text services framework keyman for windows will output each key pressed twice if it is a standard character. For example, using the Euro Latin SIL keyboard and pressing "j" will result in "jj" in the app.
This occurs in FieldWorks and in the windows search on the taskbar.
Investigation Update: It WAS introduced when keyman core integration was modified to only call the core processor once for each keystroke and cache the actions.  

When using the TSF vs Other apps
The TSF never calls the `OnTestKeyDown` or `OnTestKeyUp` first as a test rather it calls  OnKeyDown. Or OnPreservedKey once.

However, for the OnPreserved Key, KMTIP implementation calls  `_KeymanProcessKeyStroke` up to two times. First with `updateable=False`, and this if it returns `True` call it again to actually do the work of consuming the keystroke..

Solution:
   Add a new bool to the thread. On `updateable=false` we will call the core's process event and set the new bool (already processed by core). Then when we get the `updateable=true` if the `alreadyprocessed` flag is set we processed the cached actions otherwise we send the keystroke to core to be processed and process the actions straight away.

rc-swag:
Fixes #6205

For the Text Services Frame (TSF) applications in most cases for keystrokes, there is no test parse with updatable=false. In these cases, we need to call the core process event function and process the actions in one call. A bool flag has been added `CoreProcessEventRun` to track if the core process event has been called for the current keystroke. 
In the case of legacy apps or the `OnPreservedKey` for TSF. That is in the case where it is called twice once with `updateable=false` the core process event is called in the first pass and the bool `CoreProcessEventRun` is set. 

This means on any call with `updateable=true` if `CoreProcessEventRun=false` we know we need to call the core processor to process the keystroke. 
The Processing of the core actions will clear  `CoreProcessEventRun`.

### DIFF
--- a/windows/src/engine/keyman32/kmprocess.cpp
+++ b/windows/src/engine/keyman32/kmprocess.cpp
@@ -98,11 +98,10 @@ char *getcontext_debug() {
 
 static BOOL
 Process_Event_Core(PKEYMAN64THREADDATA _td) {
-  PWSTR contextBuf            = _td->app->ContextBufMax(MAXCONTEXT);
+  PWSTR contextBuf = _td->app->ContextBufMax(MAXCONTEXT);
   km_kbp_context_item *citems = nullptr;
   ContextItemsFromAppContext(contextBuf, &citems);
-  if (KM_KBP_STATUS_OK !=
-      (km_kbp_status_codes)km_kbp_context_set(km_kbp_state_context(_td->lpActiveKeyboard->lpCoreKeyboardState), citems)) {
+  if (KM_KBP_STATUS_OK != km_kbp_context_set(km_kbp_state_context(_td->lpActiveKeyboard->lpCoreKeyboardState), citems)) {
     km_kbp_context_items_dispose(citems);
     return FALSE;
   }
@@ -110,9 +109,9 @@ Process_Event_Core(PKEYMAN64THREADDATA _td) {
   SendDebugMessageFormat(
       0, sdmGlobal, 0, "ProcessEvent: vkey[%d] ShiftState[%d] isDown[%d]", _td->state.vkey,
       static_cast<uint16_t>(Globals::get_ShiftState() & K_MODIFIERFLAG), (uint8_t)_td->state.isDown);
-  if (KM_KBP_STATUS_OK != (km_kbp_status_codes)km_kbp_process_event(
-                              _td->lpActiveKeyboard->lpCoreKeyboardState, _td->state.vkey,
-                              static_cast<uint16_t>(Globals::get_ShiftState() & K_MODIFIERFLAG), (uint8_t)_td->state.isDown)) {
+  if (KM_KBP_STATUS_OK != km_kbp_process_event(
+    _td->lpActiveKeyboard->lpCoreKeyboardState, _td->state.vkey,
+    static_cast<uint16_t>(Globals::get_ShiftState() & K_MODIFIERFLAG), (uint8_t)_td->state.isDown)) {
     SendDebugMessageFormat(0, sdmGlobal, 0, "ProcessEvent CoreProcessEvent Result:False %d ", FALSE);
     return FALSE;
   }
@@ -178,8 +177,11 @@ BOOL ProcessHook()
     // if we say yes it will call a second time to actually do the work.
     // We call the core process event only once and use the core's queued actions
     // on the second pass.
-    // For the TSF in most cases kmtip (except OnPreservedKey) will not call the non-updateable test pass.
+    // For the TSF in most cases kmtip (except OnPreservedKey) will not call the non-updateable test parse.
     // Therfore the core process event will need to be called before processing the actions.
+
+    // CoreProcessEventRun would be a sufficient test however testing TIPFUpdateable defines
+    // the status of the keystroke processing more precisely.
     if (!_td->TIPFUpdateable || !_td->CoreProcessEventRun) {
       if (!Process_Event_Core(_td)) {
         return FALSE;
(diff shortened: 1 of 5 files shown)


rc-swag:
# User Testing

To run these tests the EuroLatin (SIL) Keyboard from [here](https://keyman.com/keyboards/sil_euro_latin).

## TEST_SEARCHBAR

1. Start Keyman
2. Select the EuroLatin Keyboard
3. Type in to the windows search bar `a` `b` `c` `d` <kbd>Backspace</kbd>
4. Expected reasult is `abc`
 
Search_bar

bharanidharanj:
* **TEST_SEARCHBAR: PASSED** Installed EuroLatin Keyboard from Keyman 15.0.194-alpha and it is working as expected. 

  * **TEST_NOTEPAD: PASSED** Installed EuroLatin Keyboard from Keyman 15.0.194-alpha and it is working as expected.

windows/src/engine/keyman32/kmprocess.cpp
```diff
@@ -89,6 +89,37 @@ char *getcontext_debug() {
...
+ * @return  BOOL  Return TRUE if keystroke event is passed successfully
+ */
+
+static BOOL
+Process_Event_Core(PKEYMAN64THREADDATA _td) {
+  PWSTR contextBuf            = _td->app->ContextBufMax(MAXCONTEXT);
+  km_kbp_context_item *citems = nullptr;
+  ContextItemsFromAppContext(contextBuf, &citems);
+  if (KM_KBP_STATUS_OK !=
+      (km_kbp_status_codes)km_kbp_context_set(km_kbp_state_context(_td->lpActiveKeyboard->lpCoreKeyboardState), citems)) {
```
mcdurdin, line 105: Do we need the `(km_kbp_status_codes)` cast? It's ugly 😢

mcdurdin (approved):
LGTM. The refactor is good.

Do we need to test IMX functionality given that code path also changed?

windows/src/engine/keyman32/kmprocess.cpp
rc-swag: One of the compilers complained about this. I can remove it and see which one it was if it works everywhere I will leave it out.


Why is here an enum `km_kbp_status_codes` defined but we then use a `int` as I return code instead?

mcdurdin: Yeah I'm not sure why we have `int` instead of `km_kbp_status_codes` as a return type...

windows/src/engine/keyman32/kmprocess.cpp
```diff
@@ -142,27 +173,20 @@ BOOL ProcessHook()
...
-                                                    (uint8_t)_td->state.isDown)) {
-        SendDebugMessageFormat(0, sdmGlobal, 0, "ProcessEvent CoreProcessEvent Result:False %d ",FALSE);
+    // For applications not using the TSF kmtip calls this function twice for each keystroke,
+    // first to determine if we are doing processing work (TIPFUpdateable == FALSE),
+    // if we say yes it will call a second time to actually do the work.
+    // We call the core process event only once and use the core's queued actions
+    // on the second pass.
+    // For the TSF in most cases kmtip (except OnPreservedKey) will not call the non-updateable test pass.
+    // Therfore the core process event will need to be called before processing the actions.
+    if (!_td->TIPFUpdateable || !_td->CoreProcessEventRun) {
```
rc-swag, line 183: @marc I was lying in bed the other night and thought hey I don't really even need to check if `!_td->TIPFUpdateable` in this if statement because the coreProcessEventRun is not set then that is sufficient reason to run.
I know it is sad that I was lying in bed thinking about it.  Anyway part of me wants to leave it there just to make sure. It also helps the developer follow what is going on but I did leave a pretty massive comment. Your thoughts?

mcdurdin: I like the explicitness of the test. Perhaps we should add a comment relating to your in-bed thoughts just for future code readers?

rc-swag:
@bharanidharanj Thank you for the initial testing of this I have added some more tests at the suggestion of Marc.

bharanidharanj:
> @bharanidharanj Thank you for the initial testing of this I have added some more tests at the suggestion of Marc.

You are Welcome. I will test and add my comments. Thanks.

bharanidharanj:
##  SUITE_SIMPLIFIED_CHINESE:

  * **TEST_SIMPLIFIED_SINGLE: PASSED** It shows the expected result in the Notepad application.
  * **TEST_SIMPLIFIED_MULTIPLE: PASSED** It shows the expected result in the Notepad application.
  * **TEST_SIMPLIFIED_BACKSPACE_1: PASSED** It shows the expected result in the Notepad application.

  


  * **TEST_SIMPLIFIED_BACKSPACE_2: PASSED** It shows the expected result in the Notepad application.

rc-swag:
> @rc-swag You might be sleeping now. @mcdurdin Could you help and look into the test cases above? The results have been posted, but the bot doesn't seem to understand them.

@MakaraSok  and @bharanidharanj    This was error in the test headings by me. It initially it was just two tests and Bharandidharanj tested them. I then added some more tests with the SUITE header but did not go back and add SUITE to the first lot of tests.  So I have just added the header in for those tests and it has fixed it all.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.196-alpha](https://keyman.com/downloads/releases/alpha/15.0.196)

PR #6222, opened 2022-02-09
fix(windows): create Keyman/Diag folder in redirected profile

### LINKED ISSUE #6221: bug(windows): when setup is run as a 32 bit app in SYSTEM profile, Keyman fails to install
**Describe the bug**

This issue arose from a manual install run in SCCM / Intune. The install would appear to start, running under SYSTEM profile. We ran a trace with procmon and found that setup.log was being written to
`C:\Windows\SysWOW64\config\systemprofile\AppData\Local\Keyman\Diag\setup.log` (note that this is a 32 bit profile folder). Examining this file showed us that Windows Installer was failing to create its log file. Further examination of the procmon trace showed us that Windows Installer was attempting to write to `C:\Windows\System32\config\systemprofile\AppData\Local\Keyman\Diag\keymandesktop-xxxx.log` (note how this is the 64 bit profile folder). However the `Keyman\Diag` folder does not exist in the 64 bit profile.

**Diagnosis**

1. Keyman Setup is a 32 bit app.
2. So Windows' folder redirection for 32 bit apps on WOW64 is redirecting its `C:\Windows\System32\config\systemprofile\AppData\Local\` (aka `%LOCALAPPDATA%`) to `C:\Windows\SysWOW64\config\systemprofile\AppData\Local\`.
3. Keyman creates the `Keyman\Diag` folders in what it believes is `...\System32\...` successfully (but actually, they are really in `...\SysWOW64\...`).
4. Keyman then tells Windows Installer to write its log to `C:\Windows\System32\config\systemprofile\AppData\Local\Keyman\Diag\keymandesktop-xxxx.log`. But as Windows Installer is a 64 bit app, it doesn't do folder redirection, and so the folder which Keyman has faithfully created is not there for Windows Installer.

**Resolution**

1. We could add a command line parameter to specify logfile location -- i.e. work around the problem
2. If we are running in the SYSTEM profile, we could try turning off folder redirection, create the target folder, then turn it back on again. (See [Wow64DisableWow64FsRedirection](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-wow64disablewow64fsredirection)

---

 - OS: Windows 10 x64 

mcdurdin:
Fixes #6221.

If Keyman is being installed under the SYSTEM profile, e.g. when using tools such as SCCM or Intune, then the installer runs with a redirected folder for %LocalAppData%, under C:\Windows\SysWow64\config, but is given the name C:\Windows\System32\config\....

Keyman Setup then creates %LocalAppData%\Keyman\Diag, which it thinks is in C:\Windows\System32\config\... But in reality it is in SysWow64\... Keyman Setup passes a file under this folder to Windows Installer, running as 64 bit, which immediately falls over because the path does not exist to it, in the real System32\config folder.

This patch disables redirection temporarily just to create the folder under both 32 and 64 bit versions of the profile. It has no effect on normal user accounts (apart from verifying that the folder is present twice rather than once).

# User Testing

TEST_NORMAL: Verify that Keyman for Windows installs on a computer which has NEVER had Keyman installed previously (brand new Windows install essential).

TEST_REINSTALL: Verify that Keyman for Windows installs on a computer which has an earlier version of Keyman already installed.

TEST_SUPER: Verify that Keyman for Windows installs with the SYSTEM profile.

This must be run on a brand new 64-bit install of Windows (VM fine), which has never had Keyman installed previously.

1. Download [psexec](https://docs.microsoft.com/en-us/sysinternals/downloads/psexec).
2. Download the keyman-14.0.VER.exe installer from this PR (replace `VER` with appropriate version).
3. Start an elevated command prompt and run `psexec -s cmd`. This starts a command prompt with the SYSTEM profile.
4. A new command prompt will start, probably in C:\Windows\System32.
5. Locate `keyman-14.0.VER.exe` that you downloaded previously, and run it from in this SYSTEM command prompt, with the `-s -o` parameters, something like: `c:\Users\foo\Downloads\keyman-14.0.VER.exe -s -o`.
6. Wait a minute for the install to run, silently.
7. Verify that Keyman has been installed; try to start it.
8. Using Windows Explorer, navigate to `C:\Windows\System32\config\systemprofile\AppData\Local\Keyman\Diag` and verify that a `keymandesktop-xxxx.log` file exists. (Note, you may not be able to paste that path -- you may have to navigate manually in for permission to access the folder to be granted). This verifies that Keyman actually installed under the SYSTEM account.

### DIFF
--- a/windows/src/desktop/setup/RunTools.pas
+++ b/windows/src/desktop/setup/RunTools.pas
@@ -934,6 +934,25 @@ begin
   end;
 end;
 
+//
+// See #6221.
+//
+// If Keyman is being installed under the SYSTEM profile, e.g. when using tools
+// such as SCCM or Intune, then the installer runs with a redirected folder for
+// %LocalAppData%, under C:\Windows\SysWow64\config, but is given the name
+// C:\Windows\System32\config....
+//
+// Keyman Setup then creates %LocalAppData%\Keyman\Diag, which it thinks is in
+// C:\Windows\System32\config... But in reality it is in SysWow64... Keyman
+// Setup passes a file under this folder to Windows Installer, running as 64
+// bit, which immediately falls over because the path does not exist to it, in
+// the real System32\config folder.
+//
+// This patch disables redirection temporarily just to create the folder under
+// both 32 and 64 bit versions of the profile. It has no effect on normal user
+// accounts (apart from verifying that the folder is present twice rather than
+// once).
+//
 procedure TRunTools.ApplyWow64SystemProfilePatch;
 type
   TWow64DisableWow64FsRedirection = function(out cookie: PVOID): BOOL; stdcall;
--- a/windows/src/desktop/setup/RunTools.pas
+++ b/windows/src/desktop/setup/RunTools.pas
@@ -96,6 +96,7 @@ type
     procedure WaitFor(hProcess: THandle; var Waiting, Cancelled: Boolean);  // I3349
     procedure WriteToLog(const msg: string);
     procedure RunVersion11To13Upgrade(const KMShellPath: WideString);
+    procedure ApplyWow64SystemProfilePatch;
   public
     destructor Destroy; override;
     procedure CheckInternetConnectedState;
@@ -537,7 +538,8 @@ begin
     { Log the install to the diag folder }
 
     FLogFileName := TKeymanPaths.ErrorLogPath(ChangeFileExt(ExtractFileName(msiLocation.Path), ''));  // I1610 // I2755 // I2792
-    //ForceDirectories(GetErrLogPath);  // I2768
+
+    ApplyWow64SystemProfilePatch; // #6221
 
     MsiEnableLogW(INSTALLLOGMODE_VERBOSE, PWideChar(FLogFileName), 0);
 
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

MakaraSok:
- **TEST_NORMAL: PASSED** Keyman did get installed from the command prompt and the log file is found in the location specified. (see the screenshot for how it's shown)

  image

- **TEST_REINSTALL: PASSED** Keyman dit get installed from the command prompt and the log file is found in the location specified.

   image

- **TEST_SUPER: BLOCKED** Specific instructions needed to run the test.

mcdurdin:
> * **TEST SUPER: BLOCKED** Specific instructions needed to run the test.

Where are you getting stuck?

MakaraSok:
Instructions on how to test this case (TEST_SUPER), i.e. how to "Verify that Keyman for Windows installs with the SYSTEM profile".

MakaraSok:
What makes this test essentially different from the second test case?

MakaraSok:
Should the previous Keyman version be removed and then installing the version of this PR following the same exact set of the instructions?

mcdurdin:
> What makes this test essentially different from the second test case?

This is running under the SYSTEM account, so it has a different profile. This reproduces the environment of the install that was failing. 

> Should the previous Keyman version be removed and then installing the version of this PR following the same exact set of the instructions?

No, we need to run this install on a brand new installation of Windows (hopefully from a VM snapshot?) in order to reproduce the original conditions.

mcdurdin:
* **TEST_SUPER: PASSED** I believe this is working fine as the user reporting the issue is able to work with the patched 14.0 version. Am comfortable with passing this on that basis.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.287](https://keyman.com/downloads/releases/stable/14.0.287)

Issue #6244, opened 2022-02-11
feat(common): macosh.sh developer machine setup should turn on echo


srl295:
**Is your feature request related to a problem? Please describe.**
`resources/devbox/macos/macos.sh` should  echo what it's doing

**Describe the solution you'd like**
some output of what is done

**Describe alternatives you've considered**
reading the script myself

---

**Keyman for Windows/macOS/Linux:**
 - OS: macOS

PR #6246, opened 2022-02-11
feat(common): echo in macos.sh setup script

srl295:
- turn on echoing when we start doing stuff
- it is not too noisy, and tells the user what is really happening
- also, note the script name which added certain lines to ~/.bashrc
- fix appending to the ~/.bashrc

@keymanapp-test-bot skip

Fixes: #6244

### DIFF
--- a/resources/devbox/macos/macos.sh
+++ b/resources/devbox/macos/macos.sh
@@ -150,6 +150,9 @@ BREW_IOS="swiftlint carthage"
 BREW_MACOS="carthage cocoapods"
 BREW_ANDROID="openjdk@8 android-sdk android-studio ant gradle maven"
 
+# Turn on verbosity
+set -x
+
 brew install $BREW_ALL
 $REQUIRE_ANDROID && brew install $BREW_ANDROID
 $REQUIRE_IOS && brew install $BREW_IOS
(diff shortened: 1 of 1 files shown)


keyman-server:

mcdurdin:
@srl295 thanks for the PR. I see it's still in draft, just wanted to make a couple of notes:

1. For automatic builds, create branches on the keymanapp/keyman repo rather than on a fork -- we disallow automatic builds from forks for security reasons. Of course, for this PR, a build is not required.
2. To stop the test bot from complaining, annotate your PR description with "@keymanapp-test-bot skip" (see https://github.com/keymanapp/keyman/wiki/User-Testing-Workflows#instructions-for-developers for all the gory details).

srl295:
Still a couple of rough edges I want to sort out, then I will take this out of draft… Thanks…

resources/devbox/macos/macos.sh
```diff
@@ -150,6 +150,9 @@ BREW_IOS="swiftlint carthage"
 BREW_MACOS="carthage cocoapods"
 BREW_ANDROID="openjdk@8 android-sdk android-studio ant gradle maven"
 
+# Turn on verbosity
+set -x
```
darcywong00, line 154: Can move this to the top with the other Bash flags (ll. 3-4)

srl295: I thought about that, but I think it will add too much noise… I placed this line, right where I thought the real actions start taking place.

mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.202-alpha](https://keyman.com/downloads/releases/alpha/15.0.202)

PR #6248, opened 2022-02-12
fix(web): Non-integer OSK sizes

mcdurdin:
If the OSK had a non-integer width or height stored in the OSK cookie, then the width and height would be parsed incorrectly and the OSK would fail to be sized correctly on first load; this also caused a script error and made the OSK impossible to interact with.

@keymanapp-test-bot skip This would be somewhat difficult to test, involving manual editing of cookies...

### DIFF
--- a/web/source/osk/oskView.ts
+++ b/web/source/osk/oskView.ts
@@ -414,13 +414,13 @@ namespace com.keyman.osk {
         return;
       }
 
-      if(Number.isInteger(width as number)) {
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.199-alpha](https://keyman.com/downloads/releases/alpha/15.0.199)

PR #6278, opened 2022-02-22
fix(linux): Properly compare versions in km-package-install

### LINKED ISSUE #6182: bug(linux): km-package-install might fail if casting version string to float
`km-package-install` contains code to compare two versions. The way it is done is by casting the version string to float and then comparing that. However, a version string can contain multiple sections, e.g. '1.0.1', so the cast will fail.

See https://github.com/keymanapp/keyman/pull/6179/files#r795229563.

ermshiperete:
This change will fix a bug with comparing version numbers with more than two sections (e.g. '1.2.3').

Fixes #6182.

# User Testing

## Preparations

- The Linux version doesn't matter for these tests
- Install build artifacts of this PR
- Download and unzip the attached **gff_amharic.kmp** file
- Install the unzipped gff_amharic keyboard in the attached older version:

  ```bash
  cd Downloads # (or wherever you unzipped the .kmp file)
  km-package-install -f gff_amharic.kmp
  ```

- run `km-package-list-installed` and verify that the "GFF Amharic Keyboard" shows up with version 1.9.1 and id `gff_amharic`.

## Tests

- **TEST_UPDATE**: Update to new version of gff_amharic
  - from the command line, run:
    ```bash
    km-package-install -p gff_amharic
    ```
  - verify that this outputs a message saying that a newer version is available and that the old version 1.9.1 will be uninstalled and the new version 2.2 downloaded and installed.
  - run `km-package-list-installed` and verify that the previous "GFF Amharic Keyboard" line got replaced by a line "Amharic", version 2.2 (or newer) and id `gff_amharic`.

- **TEST_SAMEVERSION**: Reinstall same version of the keyboard
  - from the command line, run:
    ```bash
    km-package-install -p gff_amharic
    ```
  - verify that this outputs a message that version 2.2 is already installed

## Attachments

[gff_amharic_1.9.1.zip](https://github.com/keymanapp/keyman/files/8125210/gff_amharic_1.9.1.zip)

### DIFF
--- a/linux/keyman-config/km-package-install
+++ b/linux/keyman-config/km-package-install
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 
 import argparse
+from distutils.version import StrictVersion
 import logging
 import sys
 import os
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
LGTM

bharanidharanj:
* **TEST_UPDATE: PASSED**  Tested this issue in Ubuntu 20.04 Linux OS and verified that the amaharic new version 2.2 is downloaded and installed. Also, Running the second command would replaced Amharic, version 2.2 with showing id. 


 

* **TEST_SAMEVERSION: PASSED**   Tested this issue in Ubuntu 20.04 Linux OS and verified that the output message showing version 2.2 was already installed.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.202-alpha](https://keyman.com/downloads/releases/alpha/15.0.202)

PR #6279, opened 2022-02-22
chore(ios): separate ios-host.js 🎢

keyman-server:
This is part of the fix chain for #5853. Split ios-host.js out from keyboard.html in order to make it easier to debug and mock. No change to the content of the sscript, just moving it here.

@keymanapp-test-bot skip (We'll do proper tests in the follow-up PR)


rc-swag (approved):
lgtm

sgschantz (approved):
lgtm

PR #6285, opened 2022-02-24
fix(linux): Gracefully handle keyboard download failure

ermshiperete:
In the past we already gracefully handled keyboard download failures when the user triggered the download from the km-config UI. This change deals with the command line side of things, most often when the user clicks on the "Install keyboard" button on keyman.com. If the download fails we already output an error message and now don't try to install the non-existing .kmp file.

Fixes #6284.

@keymanapp-test-bot skip

PR #6287, opened 2022-02-24
fix(linux): Gracefully handle makedirs failures

ermshiperete:
This change catches the exception we're getting if we try to create a directory but e.g. there is already a file with that name. In this case we now output an error message. This change also gracefully handles failures when linking files.

Fixes #6285.

# User Testing

## Preparations

- The Linux version doesn't matter for these tests.
- Install build artifacts of this PR
- Simulate the corrupt environment by running:
  ```bash
  mv ~/.local/share/fonts ~/.local/share/fonts.bak
  touch ~/.local/share/fonts
  ```

After running the tests if you don't restore a snapshot, you might want to run the following command to get things back into a working state:
```bash
rm ~/.local/share/fonts
mv ~/.local/share/fonts.bak ~/.local/share/fonts
```

## Tests

**TEST_KB_INSTALL**: Install a keyboard

- On the command line run `km-config`, click "Download", enter search for "Khmer", and install the "Khmer Angkor" keyboard
- Verify that the "Khmer Angkor" keyboard got installed and appears in the list of keyboards in the "Keyman Configuration" dialog.
- Verify that the terminal window shows 4 lines that start with `ERROR:Can't create directory /home/...`

### DIFF
--- a/linux/keyman-config/keyman_config/install_kmp.py
+++ b/linux/keyman-config/keyman_config/install_kmp.py
@@ -101,8 +101,8 @@ class InstallKmp():
         self.kmpdocdir = get_keyman_doc_dir(area, self.packageID)
         self.kmpfontdir = get_keyman_font_dir(area, self.packageID)
 
-        if not os.path.isdir(self.packageDir):
-            os.makedirs(self.packageDir)
+        if not self._safeMakeDirs(self.packageDir):
+            return
 
         if not os.path.isfile(inputfile):
             message = _("File {kmpfile} doesn't exist").format(kmpfile=inputfile)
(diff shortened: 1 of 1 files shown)


linux/keyman-config/keyman_config/install_kmp.py
```diff
@@ -101,8 +101,8 @@ def _install_kmp(self, inputfile, online, language, area):
         self.kmpdocdir = get_keyman_doc_dir(area, self.packageID)
         self.kmpfontdir = get_keyman_font_dir(area, self.packageID)
 
-        if not os.path.isdir(self.packageDir):
-            os.makedirs(self.packageDir)
+        if not self._safeMakeDirs(self.packageDir):
```
darcywong00, line 104: This used to check a boolean. Is the same as 
`if self._safeMakeDirs(self.packageDir) is None`?

mcdurdin (approved):

bharanidharanj:
* **TEST_KB_INSTALL (PASSED):** Tested this issue in Impish 20.04 Linux OS. The output line shows an error message after installing the Keyboard.

keyman-server:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.205-alpha](https://keyman.com/downloads/releases/alpha/15.0.205)

ermshiperete:
Oops, got the issue# wrong. This actually fixes #6286.

PR #6300, opened 2022-02-28
chore(developer): disable upload for server except localhost

### LINKED ISSUE #6299: chore(developer): disable upload on Server for non-localhost
* Hide Upload button
* Disable drag-drop if not on localhost
* Don't forget to document this!

Note that server-side /upload is already blocked

mcdurdin:
Fixes #6299.

Moved some of the init to test.js from packages.js. If not on localhost, then the UI for upload is disabled -- note that backend for upload is already disabled for non-localhost.

This is necessary to prevent non-local site users uploading arbitrary files to the cache, which could lead to security bypass.

On localhost:

image

When running not on localhost, the Upload menu item, and the file drop target, are both disabled.

@keymanapp-test-bot skip

### DIFF
--- a/developer/server/src/routes.ts
+++ b/developer/server/src/routes.ts
@@ -27,10 +27,15 @@ export default function setupRoutes(app: express.Express, upload: multer.Multer,
     next();
   });
 
+  function isLocalhost(req: express.Request) {
+    return (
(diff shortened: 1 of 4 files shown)


rc-swag (approved):
lgtm

PR #6310, opened 2022-03-01
fix(web): stop repeated CSS injection ✂

### LINKED ISSUE #6072: bug(web): kmwosk.css is repeatedly reloaded when creating osk
**Describe the bug**

This is easiest to see in the keymanweb keyboard debugger in Keyman Developer. Swapping between device types causes kmwosk.css to be reloaded (by new InlinedOSKView -> new OSKView -> Util.linkStyleSheet). This should be a one-time operation.

---

**KeymanWeb:**
 - KeymanWeb Version: 15.0-alpha

mcdurdin:
Fixes #6072.

Stops the kmwosk.css stylesheet from being repeatedly inserted into the DOM.

# User Testing

* TEST_SINGLE_STYLESHEET: Test that the stylesheet is only inserted once.

(This test may be better performed by a developer?)

1. Open Keyman Developer web test window in Chrome.
2. Press F12 to view the Developer Console.
3. Use the Arrow tool in Developer Console (Ctrl+Shift+C) to select any part of the On Screen Keyboard.
4. In the Styles pane of the Developer Console, note how it shows styles coming from kmwosk.css.
5. There should be only one copy of each style (see image).
6. Switch to a different device in the KeymanWeb test window (perhaps do this two or three times).
7. Select an element of the On Screen Keyboard using the Arrow tool.
8. There should still be only one copy of each style.

You should see no styles repeating, although the style may not be identical to what you see in this image:

image

Before the bug was fixed, after repeatedly switching devices, you would see the styles repeating like this:

image

### DIFF
--- a/web/source/kmwutils.ts
+++ b/web/source/kmwutils.ts
@@ -613,6 +613,16 @@ namespace com.keyman {
      * @param   {string}  s   path to stylesheet file
      */
     linkStyleSheet(s: string): void {
+      try {
+        if(document.querySelector("link[href="+JSON.stringify(s)+"]") != null) {
+          // We've already linked this stylesheet, don't do it again
+          return;
+        }
+      } catch(e) {
+        // We've built an invalid href, somehow?
+        return;
+      }
+
       var headElements=document.getElementsByTagName('head');
       if(headElements.length > 0) {
         var linkElement=document.createElement('link');


rc-swag (approved):
lgtm

rc-swag:
Test Results: 
TEST_SINGLE_STYLESHEET: PASS

PR #6313, opened 2022-03-01
chore(developer): remove Package for Distribution action ✂

### LINKED ISSUE #4168: feat(developer): rename "Package for Distribution" button

The button "Package for Distribution" on the keyboard editor is a little unclear because it is really about creating a new package, and should really only be used once in the process of developing a keyboard.

Also, if you use a template project, which is recommended for all new users and new keyboard projects, there will likely already be a package in the project. That makes this button somewhat less than useful in those cases.

Either rename the button, e.g. "Create new package from keyboard", or else make it smarter -- so that it can detect if there is a package in the project associated with the keyboard (or both).

Or perhaps delete the button altogether or move it somewhere else in the UI?

image

---

**Keyman Developer:**
 - Keyman Developer version: 14.0.210-beta

mcdurdin:
Fixes #4168.

This button no longer makes sense when we have template projects, as it encourages keyboard developers to create multiple packages. The template project should already have a package.

It is also not hard to create a package if you don't already have one; this does not reduce the complexity significantly, apart from the mini feature of adding both the .js and the .kmx in a single step (the .kvk is added also by virtue of the .kmx being added, but that is done in any case if you manually add a .kmx to a package).

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/developer/TIKE/actions/dmActionsKeyboardEditor.dfm
+++ b/windows/src/developer/TIKE/actions/dmActionsKeyboardEditor.dfm
@@ -21,12 +21,6 @@ object modActionsKeyboardEditor: TmodActionsKeyboardEditor
       OnExecute = actKeyboardIncludeDebugInformationExecute
       OnUpdate = actKeyboardIncludeDebugInformationUpdate
     end
-    object actKeyboardCreatePackage: TAction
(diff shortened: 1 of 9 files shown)


rc-swag (approved):
lgtm

keyman-server:

PR #6316, opened 2022-03-01
fix(web): error in floating osk view when no keyboard active ✂

mcdurdin:
This fixes an issue that arises when loading the Keyman Developer Server home page and there are registered keyboards, but no keyboards active. The floating osk view would throw an exception because it would attempt to set the title of the view from the active keyboard, which is `null`.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

keyman-server:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.205-alpha](https://keyman.com/downloads/releases/alpha/15.0.205)

PR #6324, opened 2022-03-02
fix(developer): prevent execution point breaking editor location ✂

mcdurdin:
When using Keyman Developer Debugger, each key pressed would reset the cursor location to the top of the .kmn file. Super annoying!

This fixes that! (This was also completed during The Great Tasmanian Internet Outage of 2022, so it gets the ✂ theme)

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #6325, opened 2022-03-02
fix(developer): project should not steal focus on load ✂

mcdurdin:
When starting Keyman Developer, you might try and switch back to another app, only to have it jump back to Developer as the Project window finishes loading. How rude!

This fixes that.

(This was also completed during The Great Tasmanian Internet Outage of 2022, so it gets the ✂ theme)

@keymanapp-test-bot skip


darcywong00 (commented):
lgtm

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.209-beta](https://keyman.com/downloads/releases/beta/15.0.209)

PR #6337, opened 2022-03-07
chore(linux): Debug why Beta build didn't get triggered

ermshiperete:
This change adds temporary debug code to track down why the beta build didn't get triggered by outputting the OUTPUT string.

It also counts the lines we're outputting so that we can output a newline in case we're not outputting anything.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.207-beta](https://keyman.com/downloads/releases/beta/15.0.207)

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.209-beta](https://keyman.com/downloads/releases/beta/15.0.209)

PR #6282, opened 2022-02-24
feat(oem/fv/android): Add keyboard and dictionary controls

darcywong00:
This adds a feature to the FirstVoices Android app for keyboard and dictionary management. 

Several of these "new" files are direct copies from the Keyman for Android app (kmapro) to handle the kmp download, installation, and settings menus. The dictionary kmp installations are hard-coded to "silent" install (doesn't display help documentation).

Changed workflow:
The current FV app has checkboxes in the keyboard list to add/remove keyboards during setup. These have been replaced with forward arrows that go to keyboard settings menu. Enabling the toggle there will add the keyboard and check for associated dictionary to download. If there's a matching dictionary from api.keyman.com, the kmp will download in the background and silently install.

This also incorporates changes from #2255 and #4696 to disable suggestions when entering password field, and re-enabling when exiting.

Tagging @madebybridget

## Screenshots

Region menu - no change
regions

BC Coast menu - left blue checkbox shows currently installed keyboards; checkboxes replaced with forward arrows
bc coast

Haisla keyboard settings
haisla

Sencoten keyboard settings (dictionary name updates after it's installed)
sencoten

Model Picker menu (copied from kmapro)
str dictionaries

str dictionary info (also copied)
dictionary info

## User Testing
* **TEST_FV_ANDROID**
1. On an Android device/emulator, install and load the PR build of FV Android
2. When the FirstVoices app loads, click the "Select keyboards" button
3. Select "BC Coast" region --> forward arrow --> "Sencoten"
4. When the Sencoten keyboard setting menu is displayed, toggle the enable button to add the keyboard
5. Verify a notification appears on the bottom about checking for an associated dictionary to install
6. Allow time for the associated dictionary to download and install in the background. This may take minutes depending on network connection
7. Click on the Dictionary row
8. Verify the "Straits Salish (Latin) dictionaries" menu appears with SENCOTEN dictionary and a blue checkbox appear
9. Click the forward arrow
10. Verify the Sencoten dictionary info menu appears and shows
    * version
    * help link
    * Uninstall dictionary
11. Click the back arrow several times to return to the main menu
12. Click the "Setup" button to finish setting up FirstVoices as a system keyboard
13. Launch a separate app (e.g. Chrome)
14. In the Chrome browser searchbox, select the FirstVoices keyboard
15. Verify the Sencoten keyboard and dictionary appear
16. Return to the FV app and navigate to the Sencoten Settings menu
17. Disable predictions and verify corrections are also disabled
[...]

PR #6346, opened 2022-03-09
fix(oem/fv/android): Check keyboard exists before registering a model

darcywong00:
Fixes pre-launch report crash for FV Android and follow-on to #6282 

The Predictions toggle was checking if the active keyboard matched the current language before registering the lexical model.
During setup before a keyboard is enabled, this would cause a null pointer exception.

Note, a separate  follow-on PR will disable predictions by default until a dictionary is installed.

## User Testing
* **TEST_NO_ACTIVE_KEYBOARD**
1. Install the PR build of "FirstVoices for Android" and launch the app
2. On the main Setup menu, click "Select keyboards" --> BC Coast --> SENCOTEN
3. Before enabling the Sencoten keyboard, disable the toggle for "Enable predictions"
4. Verify the app does not crash


mcdurdin (approved):

bharanidharanj:
* **TEST_NO_ACTIVE_KEYBOARD (PASSED):** Tested this issue as per the instructions and it is working as expected.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.211-beta](https://keyman.com/downloads/releases/beta/15.0.211)

PR #6347, opened 2022-03-09
fix(developer): warn on ncaps usage inconsistency

mcdurdin:
If a developer has two rules where one includes CAPS and the other does not include NCAPS, the CAPS rule may not be correctly matched. This change warns them when this is the case.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #6351, opened 2022-03-09
fix(web): caps state tracking for touch

mcdurdin:
Selecting the new Caps Lock `caps` layer was not correctly setting the layer state in the key event to `caps`, which caused Caps Lock rules to fail to match.

# User Testing

This requires some testing of existing KeymanWeb keyboards + modifier keys. It would be good to do some experimentation with modifier keys aside from the tests below and just see if the experience matches your expectations. (note that while you can press the Caps Lock key on the desktop OSK, it does not change the hardware state of the key -- this is a browser limitation.)

* TEST_CAPS_DESKTOP: Verify that modifier keys work correctly both on hardware keyboard and desktop On Screen Keyboard. Use unminified test page, and Obolo_chwerty keyboard (under English). Test Shift, Unshifted, and Caps Lock. The Q key is a good choice because it has three states for the `ch` it emits: `ch` unshifted, `Ch` shifted, and `CH` caps lock.
* TEST_CAPS_MOBILE: Verify that modifier keys work correctly both on hardware keyboard and mobile touch keyboard. You can test in Chrome mobile emulator. Use unminified test page, and Obolo_chwerty keyboard (under English). Test Shift, Unshifted, and Caps Lock. The Q key is a good choice because it has three states for the `ch` it emits: `ch` unshifted, `Ch` shifted, and `CH` caps lock.


common/core/web/keyboard-processor/src/text/keyboardProcessor.ts
```diff
@@ -448,25 +445,10 @@ namespace com.keyman.text {
...
-          }
-        }
-      } else if(d) {
-        keyShiftState |= v;
-
-        for(i=0; i < lockNames.length; i++) {
-          if(v & Codes.stateBitmasks[lockNames[i]]) {
-            this.stateKeys[lockKeys[i]] = true;
-          }
-        }
```
mcdurdin, line 463: Per my analysis, this code path was never called (would require calling with `_UpdateVKState(null, v, 1)`, which never happened)

common/core/web/keyboard-processor/src/text/keyboardProcessor.ts
```diff
@@ -448,25 +445,10 @@ namespace com.keyman.text {
...
-      } else {
-        keyShiftState &= ~v;
```
mcdurdin, line 465: No-op. `keyShiftState` is `0` at this point.

common/core/web/keyboard-processor/src/text/keyboardProcessor.ts
```diff
@@ -448,25 +445,10 @@ namespace com.keyman.text {
...
-
-        for(i=0; i < lockNames.length; i++) {
-          if(v & Codes.stateBitmasks[lockNames[i]]) {
-            this.stateKeys[lockKeys[i]] = false;
```
mcdurdin, line 469: This was also a no-op. The only value passed in for `v` (at least when `e` was `null`) was `15` (`0x0F`), but `Codes.stateBitmasks` contains only values in the bitmask `0x3F00`.

web/source/osk/browser/pendingMultiTap.ts
```diff
@@ -111,6 +111,8 @@ namespace com.keyman.osk {
       //       description.
       let e = text.KeyEvent.constructNullKeyEvent(this.vkbd.device);
       e.kNextLayer = 'caps';
+      e.Lstates = text.Codes.stateBitmasks.CAPS;
+      e.LmodifierChange = true;
```
mcdurdin, line 115: These two values were required to allow KeymanWeb to track the shift to the Caps Lock state. My feeling is that they *shouldn't* be required, but there's still a bit of legacy fluff around modifier / lock state that is complicating things. Easiest to add them in for now.

ermshiperete (approved):

MakaraSok:
# Test Results

* **TEST_CAPS_DESKTOP (PASSED):** Tested on Chrome 99.0.4844.51 in macOS.

The keyboard works as expected, but the OSK doesn't show according to what's triggered by the CapsLock key on both the hardware keyboard and OSK itself.

The OSK shown is consistent with the function key pressed, i.e. <kbd>shift</kbd>. The OSK shown all shifted version of the character on the key cap; but when not when <kbd>caps lock</kbd> is pressed, what's shown on the OSK doesn't reflect what's output.


https://user-images.githubusercontent.com/28331388/158099246-bbf752d2-9841-453a-9d8e-6faa0a36de50.mov



* **TEST_CAPS_MOBILE (PASSED):** Work as expected on Chrome 99.0.4844.51 in macOS -- developer mode.

mcdurdin:
> The keyboard works as expected, but the OSK doesn't show according to what's triggered by the CapsLock key on both the hardware keyboard and OSK itself.

This is a limitation in the current Keyman .kvks format. Going forward, we hope to deprecate that format and use .keyman-touch-layout which will support this.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.213-beta](https://keyman.com/downloads/releases/beta/15.0.213)

Issue #5815, opened 2021-10-07
chore(common): centralize test keyboards and build them as artifacts


ermshiperete:
Currently test keyboards are scattered all over the keyman repo as well as in https://github.com/ermshiperete/TestKeyboards.

@mcdurdin wrote on Slack:

>  I think I'd prefer to add test keyboards into the main repo and have a process to build them all as artifacts, perhaps along with the Keyman Developer build. That way we always get current versions to play with and have the ability to contrast with older builds also as required.
But I am not quite ready to do the path factoring required on the repository because right now we have keyboards scattered all over the place for tests in various platforms.

ermshiperete:
Part of that was done as part of #5613, at least for the `common/core/desktop/tests/unit/kmx` tests.

mcdurdin:
Postponing this for now -- we have the `common/test/keyboards` path for any new test keyboards. These deploy for test and release for all platforms except web, which probably needs to keep its test keyboards within the `web/testing/` path for ease of access.

mcdurdin:
Closing as we have done a lot of consolidation. Any more work on this should have specific keyboards referenced as well as some justification for why we need to do it.

PR #6352, opened 2022-03-09
feat(common): build test keyboards via script

mcdurdin:
Relates to #5815.

build.sh is a new script that builds any or all of the keyboards in the common/test/keyboards folder. build.bat is a wrapper for build.sh, just for Windows CMD.

Assumes that kmcomp.exe has already been built or allows you to specify your own kmcomp.exe if necessary.

These keyboards will be built as part of the Windows/Developer build and be available as artifacts on that build. Later, may add them to downloads.keyman.com.

@keymanapp-test-bot skip

### DIFF
--- a/common/test/keyboards/build.sh
+++ b/common/test/keyboards/build.sh
@@ -1,0 +1,171 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+
+## START STANDARD BUILD SCRIPT INCLUDE
+# adjust relative paths as necessary
+THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BASH_SOURCE[0]}")"
+. "$(dirname "$THIS_SCRIPT")/../../../resources/build/build-utils.sh"
+. "$KEYMAN_ROOT/resources/shellHelperFunctions.sh"
+## END STANDARD BUILD SCRIPT INCLUDE
+THIS_DIR="$(dirname "$THIS_SCRIPT")"
+
+display_usage() {
+  echo "usage: build.sh [build options] [targets]"
+  echo
+  echo "Build options:"
+  echo "  --clean, -c       Clean instead of build"
+  echo "  --debug, -d       Debug build"
+  echo "  --silent, -s      Suppress information messages"
+  echo "  --keyboard, -k    Build only keyboards (not packages)"
+  echo "  --kmcomp path     Specify path to kmcomp.exe, defaults"
+  echo "                    to windows/bin/developer/kmcomp.exe"
+  echo
+  echo "Targets (all unless specified):"
+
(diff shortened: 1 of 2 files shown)


common/test/keyboards/build.sh
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+
+## START STANDARD BUILD SCRIPT INCLUDE
+# adjust relative paths as necessary
+THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BASH_SOURCE[0]}")"
+. "$(dirname "$THIS_SCRIPT")/../../../resources/build/build-utils.sh"
+. "$KEYMAN_ROOT/resources/shellHelperFunctions.sh"
```
rc-swag, line 10: I presume KEYMAN_ROOT is set up an all OSes that will be using this build test script.

mcdurdin: `KEYMAN_ROOT` is declared by `build-utils.sh` -- that's why this is a standard boilerplate for all our scripts (some older ones haven't been updated yet). It works cross-platform and deals with the irregularities. Once we got it right, the boilerplate has avoided so many hassles.

rc-swag (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.210-beta](https://keyman.com/downloads/releases/beta/15.0.210)

common/test/keyboards/build.sh
```diff
@@ -0,0 +1,171 @@
...
+      # wine-4.12.1 works; wine-5.0, wine-5.7 do not.
+      # retrieve these from:
+      # `brew tap gcenx/wine && brew install --cask --no-quarantine wine-crossover`
+      # may also need to `sudo spctl --master-disable`
+      KMCOMP_LAUNCHER=wine64
+      KMCOMP="$(dirname $KMCOMP)/kmcomp.x64.exe"
+    fi
+    ;;
+  *)
+    KMCOMP_LAUNCHER=wine
```
ermshiperete, line 103: We should check for a `kmcomp` script (that hopefully launches kmcomp.exe) and only if that fails set `KMCOMP_LAUNCHER=wine` (see common/core/desktop/tests/unit/kmx/meson.build)

mcdurdin: Yeah... I'm not sure it gets us very far; this script is not going to be ideal on platforms other than Windows for now, because these keyboards should really be built with the version of kmcomp found in the current branch (which is difficult currently on linux, macos; this will be improved once kmcompx lands). I don't want to search for another `kmcomp` script on the path automatically.

But I've added this to #6358: if you pass `--kmcomp` on the command line, you are then responsible for also including `wine`.

PR #6392, opened 2022-03-17
fix(windows): framework for switching of pipeline

### LINKED ISSUE #5996: bug(windows): emoji input panel in Windows 11 does not respond to keystrokes if Keyman is running
**Describe the bug**

If Keyman is running, even if a Keyman keyboard is not selected, it does not appear to be possible to use the keyboard in the emoji input panel.

Ref: https://community.software.sil.org/t/cant-search-for-emojis-or-gifs-in-the-windows-emoji-panel-if-keyman-is-running/5559

Note: this appears to be an issue only with the Windows 11 emoji input panel; Windows 10 appears to work fine.

I have reproduced this issue in Windows 11 Insider Preview on a VM.

**Workaround**

Disable Keyman's serialized input:

1. Open Keyman Configuration
2. Select Options / Keyman System Settings
3. For `engine.compatibility.serialize_input`, enter the value `0`.
4. Restart Keyman

---

 - OS: Windows 11 
 - Keyman version: 14.0.283 
 - Target Application: emoji input panel

rc-swag:
Fixes #5996 
We want to be able to switch off the serialized pipeline if a keyman
keyboard is not active. To do this we need to store this information
in a way the serialized thread can get the information.


windows/src/global/inc/keymankeyboard_id.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * Keyman is copyright (C) SIL International. MIT License.
+ *
+ * Name:             KeymanKeyboardIID
+ * Description:      Header file to include the keyman keyboard GUIDs
```
rc-swag, line 5: @marc I am not sold on the name I have chosen for this header file `keymankeyboard_id.h`. I was also wondering if it would fit under an existing header such `keymanversion.h`

windows/src/global/inc/keymankeyboard_id.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+#ifndef _keymankeyboard_id_h
+#define _keymankeyboard_id_h
+
+//const c_clsidKMTipTextService: TGUID = '{FE0420F1-38D1-4B4C-96BF-E7E20A74CFB7}';  // version 10.0
+//const c_clsidKMTipTextService_90: TGUID = '{487EB753-DB93-48C5-9E6A-4398E777C61D}';   // I3663   // I4248
+//const c_clsidTextServicesFramework: TGUID = '{529A9E6B-6587-4F23-AB9E-9C7D683E3C50}';
```
rc-swag, line 12: I found these other GUIDs mentioned in `utiltsf.pas`  however they are not referenced anywhere else.  Does this mean I should just define the main one `c_clsidKMTipTextService` as I have already done?

rc-swag:
# User Testing

Start with windows 11 VM (do we have a base name for this yet)
Install the keyman build associated with this PR.

## TEST_WINDOWS_KBD
* Start Keyman
* Select English as the input method (not a keyman keyboard).
* Open Notepad 
* Type some text on the physical keyboard `Hello`
* Press the <kbd>&#x229E; Win</kbd> + <kbd>.</kbd> to bring up the emoji and gif window.
* In the search box type `cat` 
* The emoji window should now display some cat emoji
* Select a cat emoji.
* Final output should be `Hello 🐈` 

## TEST_KEYMAN_KBD
* Start Keyman
* Select a keyman keyboard any language
* Open Notepad 
* Type some text on the physical keyboard. 
* Press the <kbd>&#x229E; Win</kbd> + <kbd>.</kbd> to bring up the emoji and gif window.
* In the search box type some characters.
* Does it allow input? The search will not work as it works against English input. But need to check input still works
* Select an emoji character.
* Final output should be the text type plus the emoji character.

windows/src/engine/keyman32/k32_lowlevelkeyboardhook.cpp
```diff
@@ -222,10 +222,12 @@ LRESULT _kmnLowLevelKeyboardProc(
...
   */
 
-  if (hs->dwExtraInfo != 0 || hs->scanCode == SCAN_FLAG_KEYMAN_KEY_EVENT || hs->vkCode == VK_PROCESSKEY || hs->vkCode == VK_PACKET) {
+  if (hs->dwExtraInfo != 0 || hs->scanCode == SCAN_FLAG_KEYMAN_KEY_EVENT ||
+      hs->vkCode == VK_PROCESSKEY || hs->vkCode == VK_PACKET || !isKeymanKeyboardActive) {
     // This key event was generated by Keyman, so pass it through
     // dwExtraInfo is set to 0x4321DCBA by mstsc which does prefiltering. So we ignore for anything where dwExtraInfo!=0 because it
-    // probably is not hardware generated and may cause more issues to filter it
+    // probably is not hardware generated and may cause more issues to filter it/
+    // We also ignore if a keyman keyboard is not currently active.
```
rc-swag, line 230:
```suggestion
    // probably is not hardware generated and may cause more issues to filter it.
    // We also ignore if a Keyman keyboard is not currently active.
```

windows/src/engine/keyman32/globals.h
```diff
@@ -305,4 +307,7 @@ BOOL Reg_GetDebugFlag(LPSTR pszFlagRegistrySetting, BOOL bDefault);
 
 void WINAPI SetCustomPostKeyCallback(CUSTOMPOSTKEYCALLBACKPROC proc);
 
+extern BOOL isKeymanKeyboardActive; // is the active keyboard a Keyman keyboard
+
+
```
rc-swag, line 312:
```suggestion
```

windows/src/engine/kmtip/globals.h
```diff
@@ -106,7 +107,7 @@ extern CRITICAL_SECTION g_cs;
 
 extern DWORD g_dwDeepIntegration;   // I4375
 
-extern const CLSID c_clsidKMTipTextService;
+//extern const CLSID c_clsidKMTipTextService;
```
rc-swag, line 110:
```suggestion
```

windows/src/global/inc/keymankeyboard_id.h
mcdurdin: I would call it kmtip_guids.h

windows/src/global/inc/keymankeyboard_id.h
mcdurdin: Yeah, the others won't be needed here at present -- one is for an old version of the Keyman TSF addin, and the other is the CLSID for the Text Services Framework itself, which is available already if we need it.

windows/src/global/inc/keymankeyboard_id.h
```diff
@@ -0,0 +1,16 @@
...
+/* FE0420F1-38D1-4B4C-96BF-E7E20A74CFB7 Keyman keyboard guid */
```
mcdurdin, line 13:
```suggestion
/* FE0420F1-38D1-4B4C-96BF-E7E20A74CFB7 kmtip (Keyman Text Input Processor) text service CLSID */
```

mcdurdin (changes requested):

windows/src/engine/keyman32/globals.h
```diff
@@ -305,4 +307,6 @@ BOOL Reg_GetDebugFlag(LPSTR pszFlagRegistrySetting, BOOL bDefault);
 
 void WINAPI SetCustomPostKeyCallback(CUSTOMPOSTKEYCALLBACKPROC proc);
 
+extern BOOL isKeymanKeyboardActive; // is the active keyboard a Keyman keyboard
```
mcdurdin, line 310:
```suggestion
/**
 * Is the active keyboard a Keyman keyboard? This value is
 * only valid in the master controller thread, and is used
 * by serialized input to ensure that we only trap input
 * while a Keyman keyboard is active.
 **/
extern BOOL isKeymanKeyboardActive;
```

windows/src/engine/keyman32/k32_lowlevelkeyboardhook.cpp
```diff
@@ -222,10 +222,12 @@ LRESULT _kmnLowLevelKeyboardProc(
 
   */
 
-  if (hs->dwExtraInfo != 0 || hs->scanCode == SCAN_FLAG_KEYMAN_KEY_EVENT || hs->vkCode == VK_PROCESSKEY || hs->vkCode == VK_PACKET) {
+  if (hs->dwExtraInfo != 0 || hs->scanCode == SCAN_FLAG_KEYMAN_KEY_EVENT ||
+      hs->vkCode == VK_PROCESSKEY || hs->vkCode == VK_PACKET || !isKeymanKeyboardActive) {
```
mcdurdin, line 226:
```suggestion
  if (hs->dwExtraInfo != 0 || 
      hs->scanCode == SCAN_FLAG_KEYMAN_KEY_EVENT ||
      hs->vkCode == VK_PROCESSKEY || 
      hs->vkCode == VK_PACKET || 
      !isKeymanKeyboardActive) {
```

Easier to spot that they are all separate conditions if we split across multiple lines

windows/src/engine/keyman32/kmhook_getmessage.cpp
```diff
@@ -233,6 +234,13 @@ LRESULT _kmnGetMessageProc(int nCode, WPARAM wParam, LPARAM lParam)
 		return CallNextHookEx(Globals::get_hhookGetMessage(), nCode, wParam, lParam);
 	}
 
+  if (mp->message == wm_keyman_control)
+  {
+      SendDebugMessageFormat(0, sdmInternat, 0, "GetMessage: wm_keyman_control hwnd=%x %x %x", mp->hwnd, mp->wParam, mp->lParam);
+      ProcessWMKeymanControl(mp->wParam, mp->lParam);
+      return CallNextHookEx(Globals::get_hhookGetMessage(), nCode, wParam, lParam);
+  }
+
```
mcdurdin, line 243: New code should probably match the new coding style ... although I am not 100% sold on mixing brace style. Certainly indents should be 2 chars though!

```suggestion
  if (mp->message == wm_keyman_control) {
    SendDebugMessageFormat(0, sdmInternat, 0, "GetMessage: wm_keyman_control hwnd=%x %x %x", mp->hwnd, mp->wParam, mp->lParam);
    ProcessWMKeymanControl(mp->wParam, mp->lParam);
    return CallNextHookEx(Globals::get_hhookGetMessage(), nCode, wParam, lParam);
  }

```

windows/src/engine/keyman32/kmhook_getmessage.cpp
```diff
@@ -412,6 +420,32 @@ void ProcessWMKeymanControlInternal(HWND hwnd, WPARAM wParam, LPARAM lParam)
...
+void
+ProcessWMKeymanControl(WPARAM wParam, LPARAM lParam) {
+
+  switch (wParam) {
+  case KMC_PROFILECHANGED:
+    if (!Globals::IsControllerThread(GetCurrentThreadId())) {
+      break;
+    }
+    WORD wAtom = HIWORD(lParam);
+    char atomStr[128];
```
mcdurdin, line 432: Can we move these outside the `switch` statement as their scope is rubbery otherwise...

windows/src/engine/keyman32/kmhook_getmessage.cpp
```diff
@@ -412,6 +420,32 @@ void ProcessWMKeymanControlInternal(HWND hwnd, WPARAM wParam, LPARAM lParam)
...
+    if (GetAtomName(wAtom, atomStr, 128)) {
+      // Is this a keyman keyboard check c_clsidKMTipTextService
+      WCHAR clsidstr[40];
+      StringFromGUID2(c_clsidKMTipTextService, clsidstr, _countof(clsidstr));
+      PWSTR atomWstr = strtowstr(atomStr);
```
mcdurdin, line 437: This is leaking atomWstr - needs to be deleted after use

mcdurdin, line 437: Given `c_clsidKMTipTextService` is a constant, can we declare this as a constant WCHAR and avoid doing this conversion every keystroke? (That then means the leak I noted moments ago will be a non-issue too)

mcdurdin (changes requested):
Looking good. A few notes, all minor :grin:

windows/src/engine/keyman32/kmhook_getmessage.cpp
mcdurdin: My bad, my note was a little unclear and slightly wrong.

1. We can declare an additional `const char cs_clsidKMTipTextService[] = "{XXXX-XXXX...}";` in the kmtip_guids header/cpp. (check the format the StringFromGUID2 returns and match that, including casing!)
2. Then we can just use the equiv of l.438: `if (strstr(atomstr, cs_clsidKMTipTextService)) {` and avoid all the conversion and allocation code :grin:

rc-swag: For 1. We could assign the  cs_clisdKMtipTextService[] in the `kmtip_guids.cpp` I think using the `StringFromGUID2` to ensure it matches.... but that is probably overkill.  Seeing as you are never going to update the guid are you only add new ones

mcdurdin: Yep, overkill. One time code review check when we update kmtip_guids.cpp!

windows/src/engine/keyman32/kmhook_getmessage.cpp
```diff
@@ -412,6 +419,24 @@ void ProcessWMKeymanControlInternal(HWND hwnd, WPARAM wParam, LPARAM lParam)
 	}
 }
 
+void
+ProcessWMKeymanControl(WPARAM wParam, LPARAM lParam) {
+  WORD wAtom = HIWORD(lParam);
```
mcdurdin, line 424: This will work fine but technically the `wAtom` is only available with `KMC_PROFILECHANGED`, so we should leave the declaration here and move the assignment to just above line 433 (`GetAtomName`) for clarity and to avoid misleading future devs...

windows/src/global/vc/kmtip_guids.cpp
```diff
@@ -0,0 +1,4 @@
+#include "pch.h"
+
+const CLSID c_clsidKMTipTextService = {0xFE0420F1, 0x38D1, 0x4B4C, {0x96, 0xBF, 0xE7, 0xE2, 0x0A, 0x74, 0xCF, 0xB7}};
+const char cs_clsidKMTipTextService[] = "FE0420F1-38D1-4B4C-96BF-E7E20A74CFB7";
```
mcdurdin, line 4:
```suggestion
const char cs_clsidKMTipTextService[] = "{FE0420F1-38D1-4B4C-96BF-E7E20A74CFB7}";
```

Given the following from [StringFromGUID2 docs](https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-stringfromguid2):

> `[out] lpsz`: A pointer to a caller-allocated string variable to receive the resulting string. The string that represents rguid includes enclosing braces.

mcdurdin (approved):

windows/src/global/vc/kmtip_guids.cpp
rc-swag: hmm I check the logging on this.  I guess it is working because it is still a subset. with or without the brackets.

windows/src/engine/keyman32/kmhook_getmessage.cpp
rc-swag: So that is sort of what I had before. I didn't actually get this statement I just did it anyway. 

>  Can we move these outside the switch statement as their scope is rubbery otherwise... 

I am not sure why it all can't just be inside the KMC_PROFILECHANGED as it is the only scope it makes sense.

bharanidharanj: @rc-swag Shall I test this PR in Windows 10 OS and post my comment? (since I have only Windows 10 OS in my machine) Your Suggestion, Please.

rc-swag: > test this PR in Windows 10 OS and post my comment? (since I have only Windows 10 OS in my machine) Your Suggestion, Please.

@bharanidharanj  unfortunately this problem only occurred in windows 11. So 
@MakaraSok  do you have a win 11 VM?
Otherwise @mcdurdin  I can I test ( I already have my own PR) on this occasion?

mcdurdin: > Otherwise @mcdurdin can I test ( I already have my own PR) on this occasion?

Sure. There's no real restriction on who does tests -- Makara and Bharani are typically setup to run tests faster and it's good to get a fresh set of eyes on it but I'm all for reducing friction!

MakaraSok:
I have Win11. I can take this.

MakaraSok:
# Test Results

* **TEST_WINDOWS_KBD (PASSED):** Tested on Windows 11 Pro v10.0.22000 (VM). Everything is OK. No error encountered. _Note: the physical cannot be used to trigger the emoji window in a VM running Mac. Hence, the alternative is to use the Soft Keyboard available in VirtualBox to test this. If it's not sufficient, please let me know.

image

 
* **TEST_KEYMAN_KBD (FAILED):** It does allow input, but the text typed is not shown in the language of the selected keyboard. All texts produced are English letters instead of Laos.

image

rc-swag:
> It does allow input, but the text typed is not shown in the language of the selected keyboard. All texts produced are English letters instead of Laos.

I think this has to do with the input type the emoji search window has. Even though it displays English characters in the search it is like it doesn't recognise them fully. For example, typing "cat" returns window switches to searching for "gifs" and finds cat results. So it appears to be searching recognising the english characters. However, if I click on the emojis tab it show no results. If I switch the keyboard to english(AU) and again type "cat" it does return "cat" results.

1. Full description of steps. Install Khmer Angkor and assign a hot key to it. Assing a hotkey to English (Au) keyboard.
2. Select Khmer Angkor open notepad type some characters
3. Press <kbd>&#x229E; Win</kbd> + <kbd>.</kbd>
4. Note that Khmer Angkor is still selected in the windows status bar.
5. Type the key caps `c` `a` `t`  it switched to the gif tab and has results for cats. The english letters `cat` is shown in text field. And no input is sent through to the notepad app.
6. Select the emojis tab clear the search and again type `c` `a` `t`  - no results found
7. Use the hot key to switch to english (AU) 
8. Type `c` `a` `t`  results for cats are shown.

Obviously, the best way to work with this as a user is to use the hotkeys to switch the keyboard after bringing up the emoji search window then using the hotkeys to switch back once finished. 


I think I will need to discuss this with @mcdurdin .  I think from a keyman perspective we are doing all we can do. However, I would like some insight into this behaviour of showing English characters in the search window.

mcdurdin:
> Obviously, the best way to work with this as a user is to use the hotkeys to switch the keyboard after bringing up the emoji search window then using the hotkeys to switch back once finished.
> 
> I think I will need to discuss this with @mcdurdin . I think from a keyman perspective we are doing all we can do. However, I would like some insight into this behaviour of showing English characters in the search window.

Yes, at this time, the objective is to make the emoji input panel accessible while Keyman keyboard is not active. We are not attempting yet to make it interop completely with a Keyman keyboard -- that may not be possible given the emoji input panel is a kind of input method itself.

rc-swag:
@MakaraSok  I have updated the test slighlty. Basically the behaviour you witnessed is a pass. The input is actually the US keyboard characters, the main thing is that it doesn't stop input. The user can use hotkeys to switch between keyboards. See my previous comment and Marcs reply [here](https://github.com/keymanapp/keyman/pull/6392#issuecomment-1085505911)

@keymanapp-test-bot retest TEST_KEYMAN_KBD

MakaraSok:
- **TEST_KEYMAN_KBD: (PASSED)** things are working as indicated.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.230-beta](https://keyman.com/downloads/releases/beta/15.0.230)

PR #6403, opened 2022-03-21
fix(developer): detect changes in dropdowns in builder

### LINKED ISSUE #6386: bug(developer): select the dropdown hint on the key ID doesn't sync the label on the key unless retype in full
**Describe the bug**

https://user-images.githubusercontent.com/28331388/158528616-744339b7-f0e7-4704-b675-155567e7588f.mov

**Expected behavior**

When select the key ID from the dropdown hint, the key label should be updated with it. It should not require one to retype the key ID in full following the hint seen on the screen.

---

**Keyman Developer:**
 - OS: Windows 10 Pro 21H2 x64
 - Keyman Developer version: 15.0.211 - beta

mcdurdin:
Fixes #6386.

The touch layout editor would not reliably detect changes if using the mouse to select an item in a dropdown 'autocomplete'.

# User Testing

**TEST_SELECTOR:** Verify that selecting an item in a dropdown works correctly.

1. In the Touch Layout Editor, select a key, select the ID edit control.
2. Type `K_S`. Use the mouse to choose an item from the dropdown that is presented.
3. Verify that the selected item is shown in small letters at the bottom of the selected on the key cap.

See the video in #6386 for comparison with the original issue.

### DIFF
--- a/windows/src/developer/TIKE/xml/layoutbuilder/builder.js
+++ b/windows/src/developer/TIKE/xml/layoutbuilder/builder.js
@@ -1446,11 +1446,18 @@ $(function () {
     builder.updateKeyId(builder.selectedKey());
   }, {saveOnce: true});
 
+  const wrapInstant = function(f) {
+    return function() {
+      window.setTimeout(f.bind(this), 0);
+    }
+  }
+
   $('#inpKeyName')
     .change(inpKeyNameChange)
     .autocomplete({
       source: builder.lookupKeyNames,
-      change: inpKeyNameChange
+      change: inpKeyNameChange,
+      select: wrapInstant(inpKeyNameChange)
     })
     .on('input', inpKeyNameChange)
     .blur(function () {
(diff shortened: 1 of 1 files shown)


bharanidharanj:
* **TEST_SELECTOR (FAILED):** Tested this with the attached Keyman Developer 15.0.218.0 and the letters are not changed to small letters in the Touch Layout Editor.

mcdurdin:
> 3\. shown in small letters

> * letters are not changed to small letters in the Touch Layout Editor.

Oops! I think that's my mistake in the test instructions -- in this case, it just meant a small font size, not lower case. So I think we can call this passed 😀 

TEST_SELECTOR (PASSED)

windows/src/developer/TIKE/xml/layoutbuilder/builder.js
```diff
@@ -1446,11 +1446,18 @@ $(function () {
     builder.updateKeyId(builder.selectedKey());
   }, {saveOnce: true});
 
+  const wrapInstant = function(f) {
+    return function() {
+      window.setTimeout(f.bind(this), 0);
+    }
+  }
+
```
rc-swag, line 1454: So what is this doing? avoiding a race condition?

mcdurdin: This is needed because the autocomplete `select` event happens before the text is changed in the target element, to give the callback a chance to block the selection. By adding a `setTimeout`, this delays the update to the next tick. The next tick is essentially immediate, but will always happen after the currently executing code returns control to the JavaScript event loop.

rc-swag (approved):
lgtm

PR #6407, opened 2022-03-21
fix(ios): remove busy 100ms waitForFonts timer on iOS

mcdurdin:
Observed while debugging another issue on iOS. The VisualKeyboard.waitForFonts function sets up a busy timer to wait for font loading. This functionality is not even needed on the iOS app, because it's really only used for the touch alias elements.

However, it was being called, potentially multiple times, leading to the timer never being removed. On Android, the timer was correctly removed.

This aligns the iOS code path with the Android code path.

@keymanapp-test-bot skip

### DIFF
--- a/web/source/kmwembedded.ts
+++ b/web/source/kmwembedded.ts
@@ -34,7 +34,7 @@ namespace com.keyman.osk {
       }
     } else if(keyName.indexOf('K_ROPT') >= 0) {
       if(keyDown) {
-        this.highlightKey(e,false);            
+        this.highlightKey(e,false);
         if(typeof keyman['hideKeyboard'] == 'function') {
           keyman['hideKeyboard']();
         }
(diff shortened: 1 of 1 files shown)


web/source/kmwembedded.ts
```diff
@@ -78,11 +78,11 @@ namespace com.keyman.osk {
         this.keytip = new osk.embedded.KeyTip(window['oskCreateKeyPreview'], window['oskClearKeyPreview']);
       }
     };
+  }
```
rc-swag, line 81: So all you did was remove the wait for fonts function out of `com.keyman.osk`  namespace?

mcdurdin: Not quite. I moved the `waitForFonts` function override up a level so it was not in the Android-specific code path.

rc-swag (approved):
lgtm

PR #6408, opened 2022-03-22
fix(developer): structure sizes for kmcomp x64

mcdurdin:
~~I am not sure how this was ever working.~~ Okay, looked at history, and this was broken by #5963 (commit e24b7561c6a69b9cdd5f16b2c1858f1a5ce159be), of all things, which added structure size sanity checks... but neglected to remember that struct sizes may differ between 32-bit and 64-bit code.

This just adds the specific structure sizes for x64 and verifies that the Delphi sizes are the same as the reported sizes from C++ (as reported by `kcframe.exe --sizeof`)

Minor tweak to printf format args for kcframe to assist in this, sorting a compiler warning.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.221-beta](https://keyman.com/downloads/releases/beta/15.0.221)

PR #6409, opened 2022-03-22
chore(mac): update build env for monterey

mcdurdin:
Python2.7 is excluded from macOS Monterey, and is not available directly via brew any longer. We still need it for now. pyenv does the job...

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #6428, opened 2022-03-25
fix(ios): reset context when osk is resized

### LINKED ISSUE #6416: bug(ios): switching keyboards does not trigger `newContext`
# Test Results

Run the Keyman app, either in iPhone simulator or real device. Version of iOS is not important. Test the System keyboard and In-App keyboard.

**in-app:** 👎 

The keyboard still doesn't show the Shift layer right off. 

**system:** 👍 

The keyboard shows Shift layer right off after switching to the keyboard.

Clicking in the text area doesn't enable the Shift layer either. Note also

https://user-images.githubusercontent.com/28331388/159616240-e171c65c-50cf-43b4-9e5b-c144e4a7e85c.mov

**in-app:** 👎 

The keyboard doesn't switch to the Shift layer at the start of a text field; however, it does after full stop (.), exclamation (!) or question mark (?) and one or two spaces (e.g. . ). 

**system:** 👍 
The keyboard does switch to the Shift layer at the start of a text field; however, it does after full stop (.), exclamation (!) or question mark (?) and one or two spaces (e.g. . ). 

_Originally posted by @MakaraSok in https://github.com/keymanapp/keyman/issues/6366#issuecomment-1075880631_

keyman-server:
Fixes #6416.

The OSK is being rebuilt when we change size in some cases, which causes the default layer to be incorrectly set.

# User Testing

* **TEST_DEFAULT_LAYER:** With the obolo_chwerty_3151 keyboard, test that the default layer is set correctly when selecting the keyboard, and in as many circumstances as you can find, both in-app and system. Be sure to test the scenario described in #6416.

### DIFF
--- a/ios/engine/KMEI/KeymanEngine/resources/Keyman.bundle/Contents/Resources/ios-host.js
+++ b/ios/engine/KMEI/KeymanEngine/resources/Keyman.bundle/Contents/Resources/ios-host.js
@@ -80,6 +80,7 @@ function setBannerHeight(h) {
 
     // Refresh KMW's OSK
     kmw.correctOSKTextSize();
+    doResetContext();
 }
 
 function setDeviceType(deviceType) {
@@ -95,7 +96,8 @@ function setOskHeight(height) {
         kmw.core.activeKeyboard.refreshLayouts();
     }
     kmw.osk.show(true);
-    kmw['correctOSKTextSize']();
+    kmw.correctOSKTextSize();
+    doResetContext();
 }
 
 function setOskWidth(width) {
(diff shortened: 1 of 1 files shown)


bharanidharanj:
* **TEST_DEFAULT_LAYER (PASSED):** Tested this with updated iOS simulator along with Obolo_Chwerty#6351 (hence, I did not find the other Keyboard ( obolo_chwerty_3151) as mentioned by Marc. However, this PR is passed and it is working as expected. ie., while switching from some other other keyboard to obolo_chwerty keyboard, the default layer is set correctly (eg., in the shift layer). Seems to be working as expected. 

In-App Keyboard: (the default layer of the obolo_chwerty keyboard)
 

System Keyboard: (the default layer is in the shift layer)

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.225-beta](https://keyman.com/downloads/releases/beta/15.0.225)

PR #6432, opened 2022-03-27
fix(web): *Shifted* showed unshifted key cap

mcdurdin:
Changes `*Shifted*` from image (same as `*Shift*`) to image.

While this was done in code as a global override, probably following Apple stylistic changes, it is better to leave this to the keyboard developer to decide on.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #6435, opened 2022-03-28
feat(developer): improve Touch Layout Editor and Character Map integration

### LINKED ISSUE #4854: feat(developer): insert code to the key code & character to the key cap

Currently, Keyman Developer would only allow insert character to the Touch layout.

A proposed solution from [Ibrahima](https://community.software.sil.org/t/feature-request-on-keyman-developer/4606):
> Dragging or double-clicking on character on Unicode Map should populate both key and key code fields respectively with character and character code. Right now, only character is transferred to those if we drag or double-click the character, no matter what mode we select (character/Code/Name).
> 
> I find it easier and we can avoid errors while typing codes like U_1FB4.
> But, we have to make sure that the U_ is included instead of the U+ (which I understood was not supported. If I remember)
> 
> Here’s a summary:
image

N/A

mcdurdin:
Fixes #4854.

Adds some intelligence to the drag + drop integration with the Touch Layout Editor, so that we always drop a character onto a key cap, we always drop a code onto a ID field, and can drop both character + code together (if the key id is T_new_..., or holding Ctrl key).

Hovering over any key for 0.25 sec while dragging will select that key. You can then continue dragging down to the subkey array and drop onto any of those keys. Note that you do not need to hover for 0.25 sec before you drop onto a key -- the drop operation will select the key under the cursor when dropping.

Double-clicking will update the selected key or subkey; Ctrl will do both the key cap and the ID.

Demo video: https://youtu.be/qoLpuah72kw

# User Testing

This feature is conceptually fairly simple, but has a lot of nuance to make it work nicely. So we want to test all the edges cases if we can!

For the double-click tests, select a key or subkey before double-clicking.

## SUITE_EXISTING_KEY: test with a key already on the touch layout

TEST_DRAG_DROP: Drag a character onto a key. When you drop it, it should replace the existing character. The ID should not be modified.

TEST_DRAG_DROP_WITH_CTRL: Drag a character onto a key. When you drop it, it should replace the existing character. The ID should be updated to the appropriate character code such as `U_1234`.

TEST_DRAG_DROP_WITH_SHIFT: Drag a character onto a key. When you drop it, it should be added after the existing character. The ID should not be modified.

TEST_DOUBLE_CLICK: Select a key, then double-click a character in the Charmap to insert it onto the key. The ID should not be modified.

TEST_ID: Drag a character to the ID field for the key. When you drop it, it should generate a value such as `U_1234`. Do this again with a character higher than `U+10000` to verify supplementary plane characters.

## SUITE_NEW_KEY: test with a key just added to the touch layout

Before starting these tests, add a new key to the layout using one of the green plus buttons. It should have no key cap, and an ID like `T_new_...'.

TEST_DRAG_DROP: Drag a character onto the new key. When you drop it, it should be placed on the key, and the ID should be updated to the appropriate character code such as `U_1234`.

TEST_DOUBLE_CLICK: Select the new key. Double-click a character on the Character Map. It should be placed on the key, and the ID should be updated to the appropriate character code such as `U_1234`.

## SUITE_EXISTING_SUB_KEY: test with a subkey

For the drag+drop tests, you should drag over a key which has subkeys, wait 0.25 sec for the subkey array to appear, and then drag to the subkey.

For the double-click tests, select a subkey before double-clicking.

TEST_DRAG_DROP: Drag a character onto a subkey. When you drop it, it should replace the existing character. The subkey ID should not be modified.

TEST_DRAG_DROP_WITH_CTRL: Drag a character onto a subkey. When you drop it, it should replace the existing character. The ID should be updated to the appropriate character code such as `U_1234`.

TEST_DRAG_DROP_WITH_SHIFT: Drag a character onto a subkey. When you drop it, it should be added after the existing character. The ID should not be modified.

TEST_DOUBLE_CLICK: Select a subkey, then double-click a character in the Charmap to insert it onto the subkey. The ID should not be modified.

TEST_ID: Drag a character to the ID field for the subkey. When you drop it, it should generate a value such as `U_1234`. Do this again with a character higher than `U+10000` to verify supplementary plane characters.

## SUITE_NEW_SUB_KEY:

First, create a new subkey.

For the drag+drop tests, you should drag over the parent key which has the new subkey, wait 0.25 sec for the subkey array to appear, and then drag to the new subkey.

For the double-click tests, select the new subkey before double-clicking.

TEST_DRAG_DROP: Drag a character onto the new subkey. When you drop it, it should be placed on the subkey, and the ID should be updated to the appropriate character code such as `U_1234`.

TEST_DOUBLE_CLICK: Select the new subkey. Double-click a character on the Character Map. It should be placed on the subkey, and the ID should be updated to the appropriate character code such as `U_1234`.

### DIFF
--- a/windows/src/developer/TIKE/xml/layoutbuilder/builder-charmap.js
+++ b/windows/src/developer/TIKE/xml/layoutbuilder/builder-charmap.js
@@ -1,0 +1,188 @@
+//
+// Character map drag+drop and double-click insertion
+//
+
+builder.dragDrop = {};
+(function(builder, dragDrop) {
+  dragDrop.isSubKey = function(key) {
+    return key.parent().attr('id') == 'sk';
+  };
+
+  dragDrop.getKeyElementFromElement = function(elem) {
+    if(elem.classList.contains('key')) {
+      return elem;
+    }
+
+    if(!elem.parentElement) {
+      return null;
+    }
+
+    if(elem.parentElement.classList.contains('key')) {
+      return elem.parentElement;
+    }
+
+    return null;
+  };
+
+  dragDrop.getKeyElementFromDragTarget = function(target) {
+    let key = dragDrop.getKeyElementFromElement(target);
+    let isIdField = false;
+
+    if(key) {
+      key = $(key);
+    } else {
+      if(target.id == 'inpKeyCap') {
+        key = builder.selectedKey();
+      } else if(target.id == 'inpKeyName') {
+        key = builder.selectedKey();
+        isIdField = true;
+      } else if(target.id == 'inpSubKeyCap') {
+        key = builder.selectedSubKey();
+      } else if(target.id == 'inpSubKeyName') {
+        key = builder.selectedSubKey();
+        isIdField = true;
+      }
+
+      if(!key) {
+        return null;
+      }
+    }
+
+    return { key: key, isIdField: isIdField };
+  };
+
+  dragDrop.isHighSurrogate = function(character) {
+    let codeUnit = character.charCodeAt(0);
+    return codeUnit >= 0xD800 && codeUnit <= 0xDBFF;
+  };
+
+  dragDrop.isLowSurrogate = function(character) {
+    let codeUnit = character.charCodeAt(0);
+    return codeUnit >= 0xDC00 && codeUnit <= 0xDFFF;
+  };
+
+  dragDrop.surrogatePairToValue = function(character) {
+    let h = character.charCodeAt(0), l = character.charCodeAt(1);
+    return 0x10000 + (h - 0xD800) * 0x400 + (l - 0xDC00);
+  };
+
+  dragDrop.charToUnicodeValue = function(s) {
+    let v;
+    if(s.length == 2 && dragDrop.isHighSurrogate(s) && dragDrop.isLowSurrogate(s.substring(1))) {
+      v = builder.surrogatePairToValue(s);
+    } else if(s.length == 1) {
+      v = s.charCodeAt(0);
+    } else {
+      return null;
+    }
+
+    let ch = v.toString(16);
+    while(ch.length < 4) ch = '0' + ch;
+    return ch;
+  };
+
+  dragDrop.keySelectTimeout = 0;
+
+  dragDrop.clearKeySelectTimeout = function() {
+    if(dragDrop.keySelectTimeout) {
+      window.clearTimeout(dragDrop.keySelectTimeout);
+      dragDrop.keySelectTimeout = 0;
+    }
+  };
+
+  builder.charmapDragOver = function(o) {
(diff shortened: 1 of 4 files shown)


windows/src/developer/TIKE/xml/layoutbuilder/builder-charmap.js
```diff
@@ -0,0 +1,188 @@
...
+    let codeUnit = character.charCodeAt(0);
+    return codeUnit >= 0xD800 && codeUnit <= 0xDBFF;
+  };
+
+  dragDrop.isLowSurrogate = function(character) {
+    let codeUnit = character.charCodeAt(0);
+    return codeUnit >= 0xDC00 && codeUnit <= 0xDFFF;
+  };
+
+  dragDrop.surrogatePairToValue = function(character) {
```
darcywong00, line 64: Does Windows have any shared utility code for handling surrogate pairs?

mcdurdin: Well, this is JS code. It's currently completely isolated from all the other JS in Keyman. Not keen on refactoring to that level at this point. If necessary, we can add a TODO to move shared code around, but it's not going to win us much...

darcywong00 (approved):
lgtm

windows/src/developer/TIKE/xml/layoutbuilder/builder-charmap.js
mcdurdin: And ... that surrogate pair code is all gone now! Yay!

bharanidharanj:
## SUITE_EXISTING_KEY: test with a key already on the touch layout

* **TEST_DRAG_DROP (PASSED):** Tested this with the Keyman Developer (attached PR build) and it is working as expected. 
 


* **TEST_DRAG_DROP_WITH_CTRL (PASSED):** Tested this with the Keyman Developer (attached PR build) and it is working as expected. 
 


* **TEST_DRAG_DROP_WITH_SHIFT (PASSED):** Tested this with the Keyman Developer (attached PR build) and it is working as expected. 
 


* **TEST_DOUBLE_CLICK (PASSED):** Tested this with the Keyman Developer (attached PR build) and it is working as expected. 
 


* **TEST_ID (PASSED):** Tested this with the Keyman Developer (attached PR build) and it is working as expected.

bharanidharanj:
## SUITE_NEW_KEY: test with a key just added to the touch layout

* **TEST_DRAG_DROP (PASSED):** notes
* **TEST_DOUBLE_CLICK (PASSED):** notes

bharanidharanj:
## SUITE_NEW_SUB_KEY: 

* **TEST_DRAG_DROP (PASSED):** notes
* **TEST_DOUBLE_CLICK (PASSED):** notes

PR #6440, opened 2022-03-29
feat(developer): hint on unreachable code

### LINKED ISSUE #355: feat(developer): Generate warnings when mixed-cases of key names found
Key names are not case-sensitive, so a layout like
```
+ [T_a_U_0300] > 'a' U+0330
+ [T_A_U_0330] > 'A' U+0330
```
will generate a touch layout where only `T_a_U_0330` is defined.

If possible, Developer should generate a compilation warning if a layout attempts to mix cases in the key names.

### LINKED ISSUE #1710: feat(developer): Warn on unreachable code
If there are two identical rules, then the second rule will never be reached.  For example:

```
+ [K_A] > 'foo'
...
+ [K_A] > 'bar'  c this rule will never be matched
```

Warn or hint on these.

**Product context**
- Keyman Developer Compiler

mcdurdin:
Fixes #355.
Fixes #1710.

This is structured as a feature so it is tested as such, but was part of the set of fixes scheduled for 15.0 beta. Any rule in a group that Keyman Developer assesses as 'unreachable' will receive a hint message. I have set this up as a hint message because there will likely be many keyboards which have unreachable rules, and we don't want to break the build for those.

The current rule match is fairly simplistic -- it only does an exact comparison -- so duplication which is indirect, through `any(store)` references in the context would still result in unreachable rules that are not picked up.

While working on this, I discovered that the Keyman Developer unit tests were not being run (oops!). So I have fixed that as well, when I added the tests for this.

I have also refactored some of the error values for KMCOMP so that we have an INFO category (replacing the useless MEMORY category -- a MEMORY error just becomes a FATAL error). A good activity at some point would be to rework kmcomp's error reporting to allow more flexibility in error strings, e.g. placeholder variables.

Another future feature for kmcomp / tike would be to selectively disable certain hints/warnings (define in the .kpj I think).

@keymanapp-test-bot skip


windows/src/developer/kmcmpdll/UnreachableRules.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  for (i = 0; i < gp->cxKeyArray; i++, kp++) {
+    std::wstring key = MakeHashKeyFromFileKey(kp);
+    if (map.count(key) > 0) {
+      FILE_KEY const & k1 = map.at(key);
+      if (kp->Line != k1.Line && reportedLines.count(kp->Line) == 0) {
+        reportedLines.insert(kp->Line);
+        currentLine = kp->Line;
+        wsprintf(ErrExtra, "Overridden by rule on line %d", k1.Line);
+        AddWarning(CINFO_UnreachableRule);
```
darcywong00, line 39: Do you want to add this?
```
return CERR_INFO;
```

Currently, it looks like `VerifyUnreachableRules()` always returns `CERR_None`

windows/src/test/unit-tests/kmcomp/test_5963_start_of_sentence.kmn
```diff
@@ -21,7 +21,7 @@ store(digit) '0'..'9'
     if(&newLayer = "") if(&layer = 'numeric') any(digit) > context
 
     c no other changes, so detect sentence or layer change
-    if(&newLayer = 0") > use(nextLayer)
+    if(&newLayer = "") > use(nextLayer)
```
darcywong00, line 24: Does `&newLayer` depend on `store(&VERSION) '15.0'`?

https://help.keyman.com/developer/language/reference/newlayer#toc-version-history

windows/src/developer/kmcmpdll/Compiler.rc
```diff
@@ -228,6 +228,7 @@ BEGIN
     CWARN_NulNotFirstStatementInContext       "nul must be the first statement in the context"
     CWARN_IfShouldBeAtStartOfContext          "if, platform and baselayout should be at start of context (after nul, if present])"
```
ermshiperete, line 229:
```suggestion
    CWARN_IfShouldBeAtStartOfContext          "if, platform and baselayout should be at start of context (after nul, if present)"
```

windows/src/developer/kmcmpdll/UnreachableRules.cpp
mcdurdin: `CERR_INFO` is a bitmask value that is not valid on its own.

Returning `CERR_None` is correct. Returning a value other than `CERR_None` aborts the compile, and suggests that the `VerifyUnreachableRules()` function had a failure. Given it only ever tags rules with info values (despite `AddWarning` it's actually a info/hint), we don't want to abort the compile.

windows/src/test/unit-tests/kmcomp/test_5963_start_of_sentence.kmn
mcdurdin: This is an interesting one. We could make it require version 15.0, as the newContext/postKeystroke functionality won't work on earlier versions, but you could legitimately run the keyboard with reduced functionality on earlier versions. What's more, version 15.0 will pick up the new functionality correctly even when the keyboard is tagged with an earlier version number -- it's not a breaking change.

(The min version flag is really around breaking changes -- data that may cause an earlier version of Keyman to fail to load the keyboard or cause other errors after load. This was not 100% consistent through v6.0-v9.0, but it has been pretty consistent in recent versions.)

Given this, the compiler says, hey, that's fine. This keyboard will still work with earlier versions of Keyman and won't cause errors. So it assigns v9.0 to the keyboard. I think that's appropriate behaviour.

But it's a little misleading for keyboard authors!

darcywong00 (approved):
lgtm

keyman-server:

PR #6331, opened 2022-03-04
fix(web): stop masking error

mcdurdin:
We are seeing this error thrown in some logs but it masks the underlying error.

@keymanapp-test-bot skip

PR #6442, opened 2022-03-29
fix(web): track load error detail

mcdurdin:
This is a follow-up for #6331. We still did not have enough detail to be able to trace the root cause of the load errors. This adds a little more diagnostics.

@keymanapp-test-bot skip

### DIFF
--- a/web/source/keyboards/kmwkeyboards.ts
+++ b/web/source/keyboards/kmwkeyboards.ts
@@ -484,7 +484,7 @@ namespace com.keyman.keyboards {
         //
         // In case p's rejection is never caught, throwing this error will generate logs that shows up
         // in Sentry or in the console, with useful information for debugging either way.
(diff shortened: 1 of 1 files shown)


web/source/keyboards/kmwkeyboards.ts
```diff
@@ -729,15 +731,16 @@ namespace com.keyman.keyboards {
...
 
-        reject();
-      }, false);
+        reject(msg);
+      });
 
 
       // The load event will activate a newly-loaded keyboard if successful and report an error if it is not.
-      Lscript.addEventListener('load', function() {
+      Lscript.addEventListener('load', function(ev) {
```
darcywong00, line 743: What is `ev` used for?

mcdurdin: nothing, in the end! removed :)

darcywong00 (approved):
lgtm

PR #5933, opened 2021-11-18
fix(mac): use xcframework to support m1

mcdurdin:
@keymanapp-test-bot skip﻿

This allows us to build Keyman for macOS on an M1

PR #6443, opened 2022-03-29
fix(web): track load error detail 🍒

mcdurdin:
Cherry-pick of #6442

This is a follow-up for #6331. We still did not have enough detail to be able to trace the root cause of the load errors. This adds a little more diagnostics.

@keymanapp-test-bot skip

### DIFF
--- a/web/source/keyboards/kmwkeyboards.ts
+++ b/web/source/keyboards/kmwkeyboards.ts
@@ -391,7 +391,7 @@ namespace com.keyman.keyboards {
         //
         // In case p's rejection is never caught, throwing this error will generate logs that shows up
         // in Sentry or in the console, with useful information for debugging either way.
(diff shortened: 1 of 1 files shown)


darcywong00:
A hiccup in the iOS build. Do we just try again?

mcdurdin:
> A hiccup in the iOS build. Do we just try again?

Nope. 14.0-stable build is broken because we had to update our build agents to XCode 13. It requires us to change all the dependencies to xcframeworks, which we did for 15.0-alpha in #5107, #5155, #5164 and #5933. Not sure if the effort is worth it, given we are about to go to 15.0-stable in a few weeks.

darcywong00 (approved):
lgtm.

This has the original commit with `ev`. But that's fine.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.289](https://keyman.com/downloads/releases/stable/14.0.289)

PR #6445, opened 2022-03-30
feat(developer): check filename case consistency when compiling keyboard

### LINKED ISSUE #194: Warn if filenames have different case to the reference when compiling, for better cross-platform stability
<a href="https://github.com/mcdurdin"></img></a> **Issue by [mcdurdin](https://github.com/mcdurdin)**
_Tuesday Jun 13, 2017 at 03:57 GMT_
_Originally opened as https://github.com/sillsdev/keyman/issues/58_

----

mcdurdin:
Fixes #194.

Adds a check for all the various file types that may be referenced by a .kmn file to ensure that the case of the reference matches the case in the filesystem, to ensure that files are transportable across to case-sensitive filesystems.

The output is provided as a new "Hint" level message so that it does not fail keyboard builds.

**Hint vs Info:** There was confusion in the compiler between INFO and HINT type messages. The problem is that INFO messages will be suppressed by `-s`/`-ss` in kmcomp, but we actually still want HINT messages to be emitted. So I have gone ahead and renamed `CERR_INFO` to `CERR_HINT` and updated all the places.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #6463, opened 2022-04-03
feat(developer): check for duplicate store/group names

### LINKED ISSUE #6462: feat(developer): Generate error on duplicate store or group definitions
Would the following also be considered unreachable code?
https://github.com/keymanapp/keyman/issues/6089#issuecomment-1008536489

From
```
begin Unicode > use(main)

group(main) using keys
group(main) using keys
```

to
```
begin Unicode > use(main)

group(main) using keys
c group(main) using keys)
```

I encountered this while troubleshooting one of @MakaraSok test keyboards.

_Originally posted by @darcywong00 in https://github.com/keymanapp/keyman/issues/1710#issuecomment-1084082678_

mcdurdin:
Fixes #6462.

The compiler will now produce an error if a store() or group() name is re-used, as this is invalid.

@keymanapp-test-bot skip


windows/src/global/inc/Compfile.h
```diff
@@ -79,7 +79,7 @@ struct FILE_STORE {
   BOOL fIsOption;
   BOOL fIsDebug;
   BOOL fIsCall;
-  int line;
+  int line; // TODO: int vs dword, line vs Line (see FILE_KEY, FILE_GROUP)
```
rc-swag, line 82: Do you miss updating this, to be DWORD?

mcdurdin: I added it as a future TODO. It's one of those little things -- casing is easy enough, but changing the type needs to be checked carefully for boundary conditions, so didn't want to put that into this PR.

darcywong00 (approved):
lgtm

windows/src/developer/kmcmpdll/kmcmpdll.h
```diff
@@ -6,6 +6,8 @@
 BOOL AddCompileString(LPSTR buf);
 BOOL AddCompileMessage(DWORD msg);
 
+// TODO: These macros can return FALSE in functions that expect a DWORD CERR_x
+//       return value type. This is just plain wrong!
```
rc-swag, line 10: So is there a plan to address this?  Just having this comment warns the developer for future use, but what about for current use cases?

mcdurdin: I haven't opened an issue on this. I think we should, at some point, revisit the compiler error messages, but it's lowpri so for now I just want to leave a TODO in the code. (We are not seeing egregious failures due to this, so in some ways it's a curiosity.)

From my perspective, a TODO in the code is something that we can turn into an issue for resolution as/when we have time -- without overwhelming our issue lists. I try not to create new TODOs for new code but add them to existing code when maintaining it -- particularly when in a beta cycle, TODO may be a preference over a fix.

rc-swag: No worries. Definitely don't want the extra work in the beta cycle.  Are we really ever going to go through the TODO list. A quick search shows there are 3700 + TODOs. A number of those are blank from the title block but still could safely say there in the 100s.

rc-swag (approved):
lgtm

windows/src/developer/kmcmpdll/kmcmpdll.h
mcdurdin: > Are we really ever going to go through the TODO list. A quick search shows there are 3700 + TODOs. A number of those are blank from the title block but still could safely say there in the 100s.

This is a job for an intern!

keyman-server:

PR #6465, opened 2022-04-03
feat(developer): add Clear All button to icon editor

### LINKED ISSUE #2749: feat(developer): Icon editor could use a clear all button
I've struggled with the icon editor. Even to create a simple letter on solid background is hard work. On of the issues is that it's hard to get a clean slate. Because the default text that is there is anti-aliased, there's lots of different colored pixels. So the paint bucket tool doesn't help much. One has to resort to using the pencil to change each pixel. Later I learned I could use the filled in square, but it's not obvious. 
If there was a clear all button that would be very helpful.

mcdurdin:
Fixes #2749.

Adds a Clear All button which does the same as Edit/Clear Selection.

At the same time, fixes the Clear code for transparent background colours; previously, clear all with transparent background gave a black fill.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/developer/TIKE/main/UframeBitmapEditor.dfm
+++ b/windows/src/developer/TIKE/main/UframeBitmapEditor.dfm
@@ -6,8 +6,8 @@ object frameBitmapEditor: TframeBitmapEditor
   TabOrder = 0
   TabStop = True
   object imgTransparent: TImage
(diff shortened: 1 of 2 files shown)


darcywong00 (approved):
lgtm

keyman-server:

PR #6466, opened 2022-04-04
fix(developer): icon editor font color now follows fgcolor

### LINKED ISSUE #2748: feat(developer): Icon tool font color not intuitive
I was really struggling to get a colored font letter. I expected the letter to come out in the foreground color that I selected, but it always came out black. It took me a long time to figure out that I had to open the font properties dialog and then select the color of the text. Can we change it that the text will be in whatever foreground color is chosen?

mcdurdin:
Fixes #2748.

Note that transparent text is not supported for ClearType or antialiased modes.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #6479, opened 2022-04-05
fix(iOS): check version during package install

### LINKED ISSUE #6255: bug(ios): test fileVersion during package install
> @mcdurdin I built a test keyboard with Developer 15 that has a longpress of U_xxxx_yyyy on a key. kmp.json has
> 
> ```
>     "fileVersion": "7.0"
> ```
> 
> Should fileVersion be 15.0 for Keyman 15 features?

Yes, with #6145, the `fileVersion` will be `"15.0"`. From the testing in that PR, we can see that we have to fix several platforms:

4. iOS - Ignores fileVersion (@sgschantz)

I think we should try and fix this in 15.0 for better file safety going forward. Linux and macOS *should* test the .kmx version anyway at runtime, but it really should be blocked at install time.

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/issues/5992#issuecomment-1037584657_

sgschantz:
Fixes #6255

When a package is being installed, checks the fileVersion field to find the minimum version of Keyman that the package supports. If this minimum version is greater than the version of Keyman , then the package installation fails and an alert is presented to the user.

# User Testing

* **TEST_FUTURE_VERSION:** Try installing package [the_99.kmp](https://downloads.keyman.com/developer/beta/15.0.219/keyboards/the_99.kmp). It should not fail to install and display the following alert:
image

* **TEST_KHMER:** Try installing Khmer Angkor -- it should install successfully

### DIFF
--- a/ios/engine/KMEI/KeymanEngine/Classes/KeymanPackage.swift
+++ b/ios/engine/KMEI/KeymanEngine/Classes/KeymanPackage.swift
@@ -22,6 +22,7 @@ public enum KMPError : String, Error {
   //        to provide better clarity.
   case wrongPackageType = "kmp-error-wrong-type"
   case resourceNotInPackage = "kmp-error-missing-resource"
+  case unsupportedKeymanVersion = "kmp-error-unsupported-keyman-version"
 
   var localizedDescription: String {
     return engineBundle.localizedString(forKey: self.rawValue, value: nil, table: nil)
@@ -397,6 +398,10 @@ public class KeymanPackage {
         throw KMPError.noMetadata
       }
 
+      let packageName = metadata.info?.name?.description ?? ""
+      let minVersion = metadata.system.fileVersion
+      try validateVersionSupported(packageName: packageName, minSupportedVersion: minVersion)
+
       var package: KeymanPackage
 
       switch metadata.packageType {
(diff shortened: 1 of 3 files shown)


bharanidharanj:
* **TEST_FUTURE_VERSION (PASSED):** Tested this with the latest Keyman 15.0.228-beta build in iPhone 13 Pro Simulator (iOS 15.2) and installing the 99_kmp Keyboard leads to show an alert message. 


* **TEST_KHMER (PASSED):** Tested this with the attached Keyman 15.0.228-beta build in iPhone 13 Pro Simulator (iOS 15.2)  and successfully installed the Khmer_Angkor Keyboard.

ios/engine/KMEI/KeymanEngine/Classes/KeymanPackage.swift
```diff
@@ -418,6 +423,20 @@ public class KeymanPackage {
     return nil
   }
 
+  /*
+   * Checks whether the version of Keyman being used is greater than or equal to the minimum
+   * Keyman version required by the package, and, if not, throws error.
+   */
```
mcdurdin, line 429:
```suggestion
  /**
   * Checks whether the version of Keyman being used is greater than or equal to the minimum
   * Keyman version required by the package, and, if not, throws error.
   *
   * This function only verifies the major component of the version number, not the 
   * minor or patch versions. This meets the Keyman design requirement that breaking
   * file format changes can only occur in major version updates.
   */
```

mcdurdin (approved):

mcdurdin:
Please ignore the failed iOS build here -- this is due to changes in the build configuration for #6486, and will not be an issue going forward.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.229-beta](https://keyman.com/downloads/releases/beta/15.0.229)

PR #6486, opened 2022-04-06
feat(ios): testflight builds for PRs

keyman-server:
This change enables PR builds on TestFlight. These will be available only to internal testers in the "PR Builds" group. If you want access, contact @mcdurdin. They will not be automatically installed but can be selected from the TestFlight app on iDevices.

This makes a semantic change to the "build number". iOS builds have two version numbers, both of which are technically dotted triple integers (`1.2.3`) (with some weird undocumented edges, such as 18 char maximum length, additional components allowed but ignored in some situations etc etc):

1. The version number, stored in `CFBundleShortVersionString`, shown to end users, identifying a released version of Keyman.
2. The build number, stored in `CFBundleVersion`. There may be multiple build numbers for each version number, but they must be unique for a given version number.

For build numbers, I have opted to go with:

1. A release build (alpha, beta or stable) will always have a build number of `1.0`.
2. Any test build on a PR will have a build number of `0.<PR#>.<TeamCity-build.counter>`. This gives us the useful property of linking to a specific PR, and fulfils the uniqueness requirement.

Previously, build numbers were identical to version numbers, but this is not necessary.

Version numbers have not been changed.

One downside of this change is that the upload to Apple takes about 10 minutes, which adds 30% to the time of an iOS build. We'll monitor to see how much of a problem this is (the iOS test and release builds are already the slowest of any single platform, largely due to the appstore being very slow).

Very rough build stats (low and high values for test builds and release builds):

  | Test - low | Test - high | Release - low | Release - high
-- | -- | -- | -- | --
Android | 0:03:56 | 0:08:00 | 0:08:00 | 0:15:00
iOS | 0:25:00 | 0:35:00 | 0:47:00 | 1:25:00
Linux | 0:17:00 | 0:20:00 | 0:02:00 | 0:06:00
macOS | 0:07:00 | 0:09:00 | 0:07:00 | 0:12:00
Web | 0:05:00 | 0:18:00 | 0:02:00 | 0:06:00
Windows + Dev | 0:30:00 | 0:42:00 | 0:26:00 | 1:00:00

This was helpful in understanding the craziness: https://medium.com/xcblog/agvtool-automating-ios-build-and-version-numbers-454cab6f1bbe (although I avoided use of agvtool at this point)

Going along with this is an additional step in the Test-14.0 iOS TeamCity build configuration, to make the fastlane call to upload the build. This step is only run if the build is run on a PR branch (which is indicated by the `env.TEAMCITY_PR_NUMBER` variable being an integer (as opposed to `beta`, `stable-14.0` etc)

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.229-beta](https://keyman.com/downloads/releases/beta/15.0.229)

PR #6491, opened 2022-04-07
fix(windows): Don't reset defaults on Keyman upgrade

### LINKED ISSUE #1864: bug(windows): Hotkeys are lost on upgrade
**Describe the bug**
At least in some situations, keyboard hotkey settings are lost on upgrading from 9.0 to 11.0.

rc-swag:
Fixes #1864
There was a time,  up to version 11 when the Product version GUID was the same between versions. That is no longer the case.
Now we only set defaults if no previous installed version exists.

There were a few snags along the way if the MSI installer requests a reboot after installing a previous version of Keyman, then on reboot the `FInstallInfo.InstalledVersion.Version` will be empty. 
Added switch of `-d` to the command line which if it is set the defaults will be set. 
When msi installer requests a reboot the command to run after reboot can be set to include `-d`.

It would be good to move the checking of the `FInstallInfo.InstalledVersion.Version` into `bootstrapmain.pas`  however the version is not available at this point. So it needs to be checked in the `RunTool.pas` however whether it is a first run or continuing install after a reboot needs to passed through with the flag `ContineSetup`.

### DIFF
--- a/windows/src/desktop/setup/RunTools.pas
+++ b/windows/src/desktop/setup/RunTools.pas
@@ -80,7 +80,7 @@ type
     function CacheMSIFile(msiLocation: TInstallInfoFileLocation): WideString;
     procedure FinishCacheMSIFile(msiLocation: TInstallInfoFileLocation;
       InstallSuccess: Boolean);
-    function InstallMSI(msiLocation: TInstallInfoFileLocation; InstallDefaults: Boolean): Boolean;
+    function InstallMSI(msiLocation: TInstallInfoFileLocation; var InstallDefaults: Boolean; ContinueSetup: Boolean): Boolean;
     procedure ConfigFirstRun(StartKeyman,StartWithWindows,
       CheckForUpdates,StartDisabled,StartWithConfiguration,InstallDefaults,
       AutomaticallyReportUsage: Boolean);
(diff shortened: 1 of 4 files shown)


rc-swag:
# User Testing

# TEST_UPDATE_KM_VERSION
1. On a windows machine can be VM. Download and install Keyman for windows version 14. You could get a copy [here](https://keyman.com/downloads/)
2. Open Keyman Configuration and select the Hotkeys tabs
3. Change the default hotkeys assigned under `General Hotkeys` for `Open Keyboard Menu`
4. Change the default hotkey for `Show On Screen Keyboard Pane`
5. Close the configuration.
6. Download the Keyman for windows PR build from this PR.
7. Double click on the file in windows explorer to start the installation.
8. Follow the Installer questions and install.
9. After installation is complete go into Keyman Configuration and verify that the Hotkeys for ``General Hotkeys` and `Open Keyboard Menu` are still the values changed to in steps 3&4.

darcywong00 (approved):
lgtm

> Added switch of -d to the command line which if it is set the defaults will be set.

Does this new command-line switch need to be documented anywhere?

bharanidharanj:
* **TEST_UPDATE_KM_VERSION (PASSED):** Tested this as per the instructions in Windows 11 OS (VirtualBox) and it is working as expected. 
 


* **TEST_CLEAN_INSTALL (PASSED):** Tested this as per the instructions in Windows 11 OS (VirtualBox) and it is working as expected.

rc-swag:
> Does this new command-line switch need to be documented anywhere?

Yes the Keyman help files #6600

PR #6524, opened 2022-04-12
chore(android/app): Add description for referrer response error

darcywong00:
For investigating [Sentry report](https://sentry.io/organizations/keyman/issues/2700298447/?project=5983520&query=is%3Aunresolved&statsPeriod=14d)

```
onInstallReferrerSetupFinished: Unexpected code: 4
```
This was added in [Play Install Referrer API Library 2.2](https://developer.android.com/google/play/installreferrer/release-notes#2_2) which added the response `PERMISSION_ERROR`

Unfortunately the documentation is sparse on how to fix non-OK responses (the Android [sample code](https://developer.android.com/google/play/installreferrer/library#connecting) isn't even updated to include the new response).

All the non-OK responses seem to get lumped in the same Sentry issue.

@keymanapp-test-bot skip

### DIFF
--- a/android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
@@ -79,6 +79,9 @@ public class CheckInstallReferrer {
             // Connection couldn't be established; log to Sentry
             KMLog.LogError(TAG, "onInstallReferrerSetupFinished:SERVICE_UNAVAILABLE");
             break;
+          case InstallReferrerClient.InstallReferrerResponse.PERMISSION_ERROR:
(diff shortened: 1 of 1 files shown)


darcywong00:
Since the Play Store referrer API calls could fail for reasons beyond our control, I wonder if should remove all the Sentry logging and NOOP (like the sample code)

mcdurdin (approved):

mcdurdin:
> Since the Play Store referrer API calls could fail for reasons beyond our control, I wonder if we should remove all the Sentry logging and NOOP (like the sample code)

I think it is still helpful for us at present to understand sources of errors with the install referrer as that supports bundled installation of keyboards...

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.233-beta](https://keyman.com/downloads/releases/beta/15.0.233)

PR #6548, opened 2022-04-20
chore(linux): Fix failure in Debian reproducibility testing 🐧

ermshiperete:
Debian reproducibility testing runs a slightly different package build that seemingly doesn't set the `DEB_BUILD_MULTIARCH` variable. 
This change introduces our own environment variable to detect package builds so that we can skip the ibus-keyman integration tests that require running an X server.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.238-beta](https://keyman.com/downloads/releases/beta/15.0.238)

PR #6581, opened 2022-04-29
chore(android/engine): Ignore DownloadManager if ID not found

darcywong00:
Cleans up [Sentry log](https://sentry.io/organizations/keyman/issues/2699309698/events/eeb999c5f4eb46e3831205385b38990a/?project=5983520&query=is%3Aunresolved) involving DownloadManager() notifications we can ignore.

From the comment in 
https://github.com/keymanapp/keyman/blob/30c3276ce851e014463bc65a2d870a23fd4c51d4/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDownloadMgr.java#L92-L99

it appears we get notifications from DownloadManager whenever a download finishes. They may not be Keyman downloads, so we ignore those download IDs.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #6584, opened 2022-05-01
chore(web): report on file size

mcdurdin:
@keymanapp-test-bot skip

Relates to #3796.

Adds a script that checks the file size of minified keymanweb.js, so we can keep track of file size changes more proactively. It will report a status check on any PR, shown like this:

image

If the file size grows by less than 1kb, the message will be a warning (but still a passing check):

```
⚠️ Warning: This build of keymanweb.js is %d bytes, %D bytes (%d%%) larger than build $REMOTE_VERSION.
```

And if by more than 1kb, the check will fail with the following:

```
❌ Oh dear! This build of keymanweb.js is %d bytes, %D bytes (%d%%) larger than build $REMOTE_VERSION.
```


jahorton (approved):

keyman-server:

PR #6597, opened 2022-05-10
chore(web): fixup keyboard-processor and input-processor tests 🎡

mcdurdin:
Uses poor man's module mode (i.e. appending .js files) to avoid module  issues with the changeover from Lerna to NPM and tsc -b for builds. We will revisit this once the builds have stabilised.

Note a separate fix to workaround a breaking change in tsc 3.9. Ref https://github.com/microsoft/TypeScript/pull/32264#issuecomment-677718191

We need getters for the ActiveLayout classes to be enumerable so that we can copy them across in the polyfill functions.

@keymanapp-test-bot skip


jahorton (approved):
As #6632 will resolve the "poor man's modules" approach (temporarily used here) in the near future, I'm fine giving this the go-ahead for now.

PR #6622, opened 2022-05-16
chore(web): fixup languageProcessor text unit tests 🎡

mcdurdin:
Uses "Poor Man's Modules" until we have ES6, just so we can get passing
tests.

@keymanapp-test-bot skip

### DIFF
--- a/developer/js/bundle.sh
+++ b/developer/js/bundle.sh
@@ -108,7 +108,6 @@ mv kmtypes.tgz "$KEYMAN_WIX_TEMP_MODELCOMPILER"
 cd "$KEYMAN_WIX_TEMP_MODELCOMPILER"
 # package-lock.json wasn't bundled; this is needed to keep dependency versions
 # consistent.
-# cp "$KEYMAN_MODELCOMPILER_TEMP/package-lock.json" "$KEYMAN_WIX_TEMP_MODELCOMPILER/"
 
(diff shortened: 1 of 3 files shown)


developer/js/bundle.sh
```diff
@@ -108,7 +108,7 @@ mv kmtypes.tgz "$KEYMAN_WIX_TEMP_MODELCOMPILER"
 cd "$KEYMAN_WIX_TEMP_MODELCOMPILER"
 # package-lock.json wasn't bundled; this is needed to keep dependency versions
 # consistent.
-cp "$KEYMAN_MODELCOMPILER_TEMP/package-lock.json" "$KEYMAN_WIX_TEMP_MODELCOMPILER/"
+# cp "$KEYMAN_MODELCOMPILER_TEMP/package-lock.json" "$KEYMAN_WIX_TEMP_MODELCOMPILER/"
```
darcywong00, line 111: Can this be removed?

jahorton (approved):
As https://github.com/keymanapp/keyman/pull/6632 will resolve the "poor man's modules" approach (temporarily used here) in the near future, I'm fine giving this the go-ahead for now.

developer/js/bundle.sh
mcdurdin:
```suggestion
```

PR #6686, opened 2022-06-01
fix(web): crash in highlightKey

mcdurdin:
Fixes [KEYMAN-WEB-X](https://sentry.io/organizations/keyman/issues/2846706970/?project=5983524&query=is%3Aunresolved).

Relates to #6541.

`TypeError: Cannot read properties of undefined (reading 'indexOf')`

@jahorton I have not identified a root cause for this. There may be something deeper going on? But this mitigates the crash report I think.

~~Once this is approved I will cherry-pick to beta (which will then flow through to alpha eventually).~~

@keymanapp-test-bot skip (I cannot think of any applicable tests)

### DIFF
--- a/web/source/osk/visualKeyboard.ts
+++ b/web/source/osk/visualKeyboard.ts
@@ -1921,7 +1921,7 @@ namespace com.keyman.osk {
      **/
     highlightKey(key: KeyElement, on: boolean) {
       // Do not change element class unless a key
-      if(!key || (key.className == '') || (key.className.indexOf('kmw-key-row') >= 0)) return;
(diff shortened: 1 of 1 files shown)


jahorton (approved):
If memory serves, that should 100% be fine.  I can't find the reference right off-hand, but I remember noting that this got fixed in a very similar manner during 15.0's InlinedOSK work out of necessity, and it's served us well there.

I think I even asked at the time if it should be 🍒'd, and we elected not to at that time.  Though... I could be wrong about that.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 14.0.294](https://keyman.com/downloads/releases/stable/14.0.294)

PR #6687, opened 2022-06-01
chore(developer): move browse4folder to Developer source 🎢

mcdurdin:
browse4folder is used only by Keyman Developer IDE.

@keymanapp-test-bot skip


ermshiperete (approved):
Looks reasonable, but then I don't speak Delphi :smile:

PR #6691, opened 2022-06-01
fix(developer): install keyboard call should quote kmshell.exe path

mcdurdin:
In many cases, this would have worked fine. But if there was pathing ambiguity, it could have failed with strange errors.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

PR #6712, opened 2022-06-05
chore(windows): move JCL and JVCL files to Developer 🎢

mcdurdin:
Removes majority of extraneous file from JCL + JVCL, and moves the remainder to Keyman Developer, where they are used only for the docking components in TIKE. This is still a large dependency, but at least it's just in the one place now.

@keymanapp-test-bot skip


ermshiperete (commented):
I'm surprised that we still have files in `windows/src/ext/jedi` after this change?

ermshiperete:
> I'm surprised that we still have files in `windows/src/ext/jedi` after this change?

Ahh, I see. #6713 moves those files!

ermshiperete (approved):

PR #6743, opened 2022-06-10
chore(windows): split sentry-upload-difs.sh to Developer 🎢

mcdurdin:
Splits the script into separate scripts for each platform. Note that this is only ever called from TeamCity builds.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tools/sentry-upload-difs.sh
+++ b/developer/src/tools/sentry-upload-difs.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+#
+# Upload Debug Information Files for this developer folder
+#
+# We only want .pdb and .sym files at this time -- we should not need the
+# symtabs that PE files give us as the full debug symbols give us better
+# coverage anyway.
+#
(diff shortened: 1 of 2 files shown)


developer/src/tools/sentry-upload-difs.sh
```diff
@@ -0,0 +1,44 @@
...
+set -e
+set -u
+
+## START STANDARD BUILD SCRIPT INCLUDE
+# adjust relative paths as necessary
+THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BASH_SOURCE[0]}")"
+. "$(dirname "$THIS_SCRIPT")/../../../resources/build/build-utils.sh"
+## END STANDARD BUILD SCRIPT INCLUDE
+
+cd "$KEYMAN_ROOT/windows/src"
```
ermshiperete, line 25: Shouldn't this be `developer/src` ?

ermshiperete (changes requested):

PR #6751, opened 2022-06-11
chore(developer): move developer-specific headers 🎢

mcdurdin:
Moves the following four header files:
* debugstore.h -> kmcmpdll (51952c224)
* compiler.h -> common, renamed to legacy_kmx_file.h (e5b31669a)
* compfile.h -> kmcmpdll (e295e3ef7)
* comperr.h -> kmcmpdll (2fe84c7ea)

And fixes up references as required.

This may be more easily reviewed commit-by-commit; see each commit's message for more details on the moves.

@keymanapp-test-bot skip


developer/src/kmcmpdll/keyman64.h
```diff
@@ -30,7 +30,7 @@
 #include <windows.h>
 #include <assert.h>
 
-#include "compiler.h"
+#include "../../../../common/windows/cpp/include/legacy_kmx_file.h"
```
ermshiperete, line 33: I think that's one `../` too much

ermshiperete (changes requested):

developer/src/kmcmpdll/keyman64.h
mcdurdin: You're right, but weirdly enough it builds?!  If I add an extra `../` it fails to build. But it builds when it has one too many `../` components... Is there some bizarre pathing standard I am not aware of here?

Anyway, I have corrected it!

ermshiperete (approved):

PR #6752, opened 2022-06-11
chore(windows): merge vkeys.h and move to common 🎢

mcdurdin:
@keymanapp-test-bot skip

Moves vkeys.h to common, turns it into a proper header and adds corresponding vkeys.cpp so that we don't include the data multiple times in a project.

Removes duplication of vkeys.h in kmcmpdll and global/inc.

Updates all projects to reference the common files.

For kmdecomp, adds a precompiled header so that we can simply #include "pch.h" in the common vkeys.h, and updates refs to vkeys.h.


developer/src/kmcmpdll/kmcmpdll.vcxproj.filters
```diff
@@ -88,6 +67,23 @@
     <ClInclude Include="CheckFilenameConsistency.h">
       <Filter>Header files</Filter>
     </ClInclude>
+    <ClInclude Include="$(KEYMAN_ROOT)\developer\src\ext\json-schema-validator\json-schema.hpp" />
+    <ClInclude Include="$(KEYMAN_ROOT)\developer\src\ext\json\json.hpp" />
```
ermshiperete, line 71: No header files?

ermshiperete (changes requested):

developer/src/kmcmpdll/kmcmpdll.vcxproj.filters
mcdurdin: Visual Studio moved these out I think? I didn't update this file manually.

Fixed.

PR #6755, opened 2022-06-12
chore(windows): cleanup keyman64.h usage 🎢

mcdurdin:
keyman64.h is now only used by keyman32/keyman64. Shared components have been moved to legacy_kmx_memory.h - KMX structures in-memory, as used by kmanalyze, kmdecomp.

Removed unnecessary references to keyman64.h from kmtip.

Also removed aiDebugger.* which are obsolete files.

One day in the future, we could merge keyman64.h with keymanengine.h.

@keymanapp-test-bot skip


ermshiperete (approved):
Funny that the `keyman32` directory contains `keyman64.h` :smile:

mcdurdin:
> Funny that the `keyman32` directory contains `keyman64.h` 😄

History comes back to bite sometimes: this rename happened as a part of the addition of 64-bit support.

Separate low-priority wish: I would like to unify the keyman32 and keyman64 folders -- we should be able to build both from the same .vcxproj really as the differences are minor.

PR #6780, opened 2022-06-15
chore(web): shorten check/web/file-size check message

mcdurdin:
@keymanapp-test-bot skip

### DIFF
--- a/web/tools/check-build-size.sh
+++ b/web/tools/check-build-size.sh
@@ -120,18 +120,18 @@ if (( $LOCAL_FILE_SIZE <= $REMOTE_FILE_SIZE )); then
   # We are good -- the new build is smaller than the old one
   RESULT_STATE=success
   REPORT_CODE=0
-  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION"
+  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION, at $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
 elif (( $DIFF < $SIZE_THRESHOLD )); then
   # Warning, larger, but less than 1kb larger
   RESULT_STATE=success
   REPORT_CODE=1
-  RESULT_MESSAGE="⚠️ Warning: keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION"
+  RESULT_MESSAGE="⚠️ Warning: keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION, at $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
 else
   # Failure - new build is significantly larger
   DIFF=$(($LOCAL_FILE_SIZE - $REMOTE_FILE_SIZE))
   RESULT_STATE=failure
   REPORT_CODE=2
-  RESULT_MESSAGE="❌ Oh dear! keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION"
+  RESULT_MESSAGE="❌ Oh dear! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION, at  $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
 fi
 
 echo "$RESULT_MESSAGE"
--- a/web/tools/check-build-size.sh
+++ b/web/tools/check-build-size.sh
@@ -120,18 +120,18 @@ if (( $LOCAL_FILE_SIZE <= $REMOTE_FILE_SIZE )); then
   # We are good -- the new build is smaller than the old one
   RESULT_STATE=success
   REPORT_CODE=0
-  RESULT_MESSAGE="✅ Excellent! This build of keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than build $REMOTE_VERSION."
+  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION"
 elif (( $DIFF < $SIZE_THRESHOLD )); then
   # Warning, larger, but less than 1kb larger
   RESULT_STATE=success
   REPORT_CODE=1
-  RESULT_MESSAGE="⚠️ Warning: This build of keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than build $REMOTE_VERSION."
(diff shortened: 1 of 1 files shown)


web/tools/check-build-size.sh
```diff
@@ -120,18 +120,18 @@ if (( $LOCAL_FILE_SIZE <= $REMOTE_FILE_SIZE )); then
   # We are good -- the new build is smaller than the old one
   RESULT_STATE=success
   REPORT_CODE=0
-  RESULT_MESSAGE="✅ Excellent! This build of keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than build $REMOTE_VERSION."
+  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION"
```
jahorton, line 123: Is there a need to keep the lead word - "Excellent" / "Warning" / "Oh dear" - in place, given the emojis before them that also indicate the same sentiment?

Also, why are we nixing the end period?

jahorton: Also... if the point is to help us see the important details first...  maybe...

```suggestion
  RESULT_MESSAGE="✅ keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION at $(printf "%'d" $LOCAL_FILE_SIZE) bytes."
```

With similar changes to the other two messages.

------

The filesize delta is most important, after which comes the version we're comparing against, I'd think.

mcdurdin: > Is there a need to keep the lead word - "Excellent" / "Warning" / "Oh dear" - in place, given the emojis before them that also indicate the same sentiment?

I wanted the lead words because this is a happy little friendly bot which is just trying to be helpful. Without the lead words, it sounds harsh. 😁 

Happy to reorder the message as you suggest -- agree delta in bytes is probably most important. Although I do wonder if the overall file size is something that perhaps should be visible to scare us -- because a .js that's over 100kb is 😱 **scary** 😱!

jahorton: > because a .js that's over 100kb is 😱 scary 😱!

If it were an individual file, pre-compilation... sure.  But, for something as large-scale as KMW, I think a little leniency may be necessary.

For your consideration: https://github.com/keymanapp/keyman/blob/master/common/models/wordbreakers/src/default/data.ts

image

82.2 kB.  For a _single_ one of the source files.

Sure, that's pre-minification..., but still.  (And I'm curious how much minification gets applied / can safely _be_ applied, given how buried & wrapped in the internals it is.)

mcdurdin: > If it were an individual file, pre-compilation... sure. But, for something as large-scale as KMW, I think a little leniency may be necessary.

Yes, this is a reality we are living with. KMW is going to be large for a while. But I think it's grown larger than it needs to because we haven't been watching this. So now we will push a bit harder on file size.

Many other client-side JS libraries are nowhere near as big as this. jquery does as heck of a lot in 88kb. 

If we care about file size, that specific file could be much, much, much smaller just by using shorter identifiers -- or by using a trie! See an example of compressing similar (more) data to just 9.6kb at https://github.com/foliojs/unicode-properties, pre-gzip.

jahorton: I mean, that's what Closure's "advanced minification" is all about.  Notably, it's probably more feasible to enable than it's been in a long while thanks to the 🎡 refactor set.

That said, I imagine it'd still take a fair bit of work to fully enable without some significant breakages in places.  If we wanted to invest a significant amount of time to get it right, that's something we _could_ do.

mcdurdin: Yes, but there are ways we can apply CS to this apart from just minification. Minification is a bit of a big dumb hammer (despite all the smarts in it), and tricks like using a trie to encode a large data bundle like this fall outside its remit.

Not saying we are prioritising this over other things just now. Just something that is _more_ important than it was :grin:.

web/tools/check-build-size.sh
```diff
@@ -120,18 +120,18 @@ if (( $LOCAL_FILE_SIZE <= $REMOTE_FILE_SIZE )); then
   # We are good -- the new build is smaller than the old one
   RESULT_STATE=success
   REPORT_CODE=0
-  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION"
+  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION, at $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
```
ermshiperete, line 123: Now it's more ambiguous if the size is the new or old size

mcdurdin: So it is... 🤣🤣🤣

It's not critical though! I will do a one word change and then pretty please can I get an approval? I'd like to merge this administrivial PR!

mcdurdin:
```suggestion
  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION, now $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
```

web/tools/check-build-size.sh
```diff
@@ -120,18 +120,18 @@ if (( $LOCAL_FILE_SIZE <= $REMOTE_FILE_SIZE )); then
...
   RESULT_STATE=success
   REPORT_CODE=0
-  RESULT_MESSAGE="✅ Excellent! This build of keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than build $REMOTE_VERSION."
+  RESULT_MESSAGE="✅ Excellent! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) smaller than $REMOTE_VERSION, now $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
 elif (( $DIFF < $SIZE_THRESHOLD )); then
   # Warning, larger, but less than 1kb larger
   RESULT_STATE=success
   REPORT_CODE=1
-  RESULT_MESSAGE="⚠️ Warning: This build of keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than build $REMOTE_VERSION."
+  RESULT_MESSAGE="⚠️ Warning: keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION, at $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
```
mcdurdin, line 128:
```suggestion
  RESULT_MESSAGE="⚠️ Warning: keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION, now $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
```

web/tools/check-build-size.sh
```diff
@@ -120,18 +120,18 @@ if (( $LOCAL_FILE_SIZE <= $REMOTE_FILE_SIZE )); then
...
 else
   # Failure - new build is significantly larger
   DIFF=$(($LOCAL_FILE_SIZE - $REMOTE_FILE_SIZE))
   RESULT_STATE=failure
   REPORT_CODE=2
-  RESULT_MESSAGE="❌ Oh dear! This build of keymanweb.js is $(printf "%'d" $LOCAL_FILE_SIZE) bytes, $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than build $REMOTE_VERSION."
+  RESULT_MESSAGE="❌ Oh dear! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION, at  $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
```
mcdurdin, line 134:
```suggestion
  RESULT_MESSAGE="❌ Oh dear! keymanweb.js is $(printf "%'d" $DIFF) bytes ($PERCENT_DIFF%) larger than $REMOTE_VERSION, now $(printf "%'d" $LOCAL_FILE_SIZE) bytes"
```

ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.20-alpha](https://keyman.com/downloads/releases/alpha/16.0.20)

PR #6835, opened 2022-06-23
chore(ios): log script actions and fixup export settings to avoid version being clobbered

mcdurdin:
@keymanapp-test-bot skip

Fixes the clobbering of build_version values. For some reason, this was only happening on ba-macos-ext-hba-01 and not on ba-macos-ext-hba-03, despite there being little difference between the two agents. However, writing `manageAppVersionAndBuildNumber` to the export plist makes the behaviour consistent between the two agents. `manageAppVersionAndBuildNumber` does not appear to be documented anywhere that I can find on apple.com -- only discussion in forums and [StackOverflow](https://stackoverflow.com/questions/68237292/xcode-manage-version-and-build-number-option#comment122517504_68237292).

Thank you yet again Apple for introducing a yet another new undocumented build behaviour with the most recent forced xcode update...

This also rewrites some of the scripts for iOS (and a little bit for macOS) so that we can capture output from script phases in xcodebuild.

I will cherry-pick the export plist options to stable-15.0, but propose not back-porting the script changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -168,3 +168,6 @@ resources/environment.sh
 
 **/node_modules/
 
+# Temporary file for logging scripts in xcode runs, see build-utils.sh for
+# details
+/xcodebuild-scripts.log
--- a/ios/exportAppStore.plist
+++ b/ios/exportAppStore.plist
@@ -19,5 +19,7 @@
     </dict>
     <key>signingStyle</key>
     <string>manual</string>
+    <key>manageAppVersionAndBuildNumber</key>
+    <false/>
 </dict>
 </plist>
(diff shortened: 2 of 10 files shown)


jahorton (approved):
> but propose not back-porting the script changes.

Why not?  I see little reason to be concerned here; the iOS build process is currently identical for both `stable-15.0` and `master`; it hasn't undergone the dramatic changes that Web has and that Windows & Developer are undergoing.  (I'd understand for those.)

That said. LGTM.

mcdurdin:
> Why not?

Our normal policy is not to back-port changes that are not required for stable to function correctly. The script changes introduce risk. If I was to back port them, I'd want to wait until we'd run with the alpha build for a few days at least. There certainly should be no reason to block #6837 for those script changes in any case (if we decide to do it, it could always go in a new PR).

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.19-alpha](https://keyman.com/downloads/releases/alpha/16.0.19)

PR #6847, opened 2022-06-26
fix(common): trigger builds correctly for stable builds

mcdurdin:
@keymanapp-test-bot skip

stable-15.0 builds were not triggering for test PRs. I think this is because a test build had never run on that branch in TeamCity, and so its locator didn't match. Removing the buildType appears to have resolved this.


ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.24-alpha](https://keyman.com/downloads/releases/alpha/16.0.24)

PR #6865, opened 2022-06-28
fix(developer): readonly groups should never emit output

### LINKED ISSUE #6846: bug(web): selecting text when NewContext entry point is present deletes text in mobile simulation view

mcdurdin:
Retrying #6848.

Fixes #6846. (Note #6799 is related but fixed by subsequent PR.)

Note: this fix and the following PR will be cherry-picked to stable-15.0 once testing and review is complete.

  * `use` other readonly groups

# User Testing

Use the keyboard [sil_euro_latin.zip](https://github.com/keymanapp/keyman/files/8985670/sil_euro_latin.zip) for testing. All tests should be performed in the Keyman Developer web debugger:

3. Open sil_euro_latin.kmn
4. In the **Build** tab, click **Compile Keyboard**, then click **Test Keyboard on web**.
5. Select 'http://localhost:8008' from the list of web addresses and click **Open in browser**.

* **TEST_DELETE_SELECTION_DESKTOP:** Using the web debugger, select 'Windows' from the **Device** menu. Click in the text area, and type a few words using the hardware keyboard. Select one of the words you just typed using the mouse. For this test to pass, the selected word should not be deleted as soon as you select it!

* **TEST_DELETE_SELECTION_IPHONE:** Using the web debugger, select 'iPhone 5S' from the **Device** menu. Click in the text area, and type a few words using the iPhone keyboard. Select one of the words you just typed using the mouse.  For this test to pass, the selected word should not be deleted as soon as you select it!

If you want to compare with the previous (wrong) behaviour, you can run the same tests using [version 15.0.263](https://keyman.com/go/download/keyman-developer?version=15.0.263).

### DIFF
--- a/developer/src/tike/compile/CompileKeymanWeb.pas
+++ b/developer/src/tike/compile/CompileKeymanWeb.pas
@@ -1024,7 +1024,10 @@ begin
 
   if IsKeyboardVersion10OrLater() and (pwsz^ <> #0) then
   begin
-    Result := Result + nlt+Format('k.KDC(%d,t);', [ len ] );   // I3681
+    if not fgp.fReadOnly then
+    begin
+      Result := Result + nlt+Format('k.KDC(%d,t);', [ len ] );   // I3681
+    end;
     len := -1;
   end;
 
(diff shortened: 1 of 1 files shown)


MakaraSok:
# Test Results

All tests done in a VM running Windows 10 Pro 21H2 x64.

* **TEST_DELETE_SELECTION_DESKTOP (PASSED):** The selected word doesn't get deleted.
* **TEST_DELETE_SELECTION_IPHONE (PASSED):** The selected word doesn't get deleted.

jahorton (approved):

PR #6873, opened 2022-06-29
fix(developer): readonly groups should never emit output 🍒

### LINKED ISSUE #6846: bug(web): selecting text when NewContext entry point is present deletes text in mobile simulation view

mcdurdin:
Fixes #6846.

Cherry-pick of #6865.

  * `use` other readonly groups

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:

PR #6891, opened 2022-07-05
chore(web): update readme

mcdurdin:
@keymanapp-test-bot skip

Leftover work from last time I booted to macos ;-)

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -2,11 +2,12 @@
 The Original Code is (C) SIL International
 
 ## Prerequisites
-See [build configuration](../docs/build/index.md) for details on how to configure your build environment.
+See [build configuration](../docs/build/index.md) for details on how to
+configure your build environment.
 
 **********************************************************************
 
-The following folders contain the distribution for KeymanWeb:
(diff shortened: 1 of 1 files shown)


jahorton (approved):
Another one of those "why has it been this way since 10.0" cases ("KeymanWeb" vs "Keyman Engine for Web")?

🤷 I guess it is technically a bit more clean & proper with the change, though.

mcdurdin:
> Another one of those "why has it been this way since 10.0" cases ("KeymanWeb" vs "Keyman Engine for Web")?

There's always more work to do? 

> 🤷 I guess it is technically a bit more clean & proper with the change, though.

Yep, as we go forward, keep trying to tidy up and clarify the distinctions.

For web, technically this repo only contains Keyman Engine for Web. KeymanWeb is the app keymanweb.com. It's a funny distinction but I think it helps others if we can try to be clear on these. Helps me too...

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.25-alpha](https://keyman.com/downloads/releases/alpha/16.0.25)

PR #6895, opened 2022-07-05
fix(linux): Fix post-install script :cherries: 

### LINKED ISSUE #6893: bug(linux): Installation sometimes shows error
Installing ibus-keyman occasionally fails with an error:

```
Setting up ibus-keyman (15.0.265-1~sil1~focal) …
error: user name does not exist
```

Depending on settings this error gets ignored or can fail the installation.

This can happen if during the installation the environment variable `SUDO_USER` isn't set. I'm not sure under what circumstances this happens, but I've seen it, and there's an [issue on the community site](https://community.software.sil.org/t/keyman-linux-dependency-errors/6419).

ermshiperete:
This change fixes the ibus-keyman post install script which previously failed if for whatever reason `SUDO_USER` is not set.

Fixes #6893

(:cherries:-picked from #6894 )

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/ibus-keyman.postinst
+++ b/linux/debian/ibus-keyman.postinst
@@ -37,7 +37,7 @@ case "$1" in
       fi
 
       # Verify that it's running now
-      ! ibusdaemon=`ps --user ${SUDO_USER} -o s= -o cmd | grep --regexp="^[^ZT] ibus-daemon .*--xim.*"`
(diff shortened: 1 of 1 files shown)


keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.266](https://keyman.com/downloads/releases/stable/15.0.266)

PR #6904, opened 2022-07-06
fix(web): improve `console.error()` reporting 🍒

mcdurdin:
Cherry-pick of #6890 and #6905.

Only change is to add `//@ts-ignore` for Sentry calls, required for 15.0 branch, not needed in 16.0.

@keymanapp-test-bot skip

PR #6905, opened 2022-07-07
chore(android): disable sendError for js errors

mcdurdin:
@keymanapp-test-bot skip

This is a follow-up to #6890 and already cherry-picked to #6904.

Turns out that we were getting *three* reports of JS errors on Android. These are also pretty useless for the same reason as for `sendKMWError`, and in fact conflated the JS errors with other kinds of errors. This makes it very hard to pick out the non-JS errors from the noise.


ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.28-alpha](https://keyman.com/downloads/releases/alpha/16.0.28)

PR #6906, opened 2022-07-07
fix(developer): set contextDevice in web debugger 🖇️

mcdurdin:
With the fix in fix/web/context-only-device, we need to ensure that the debugger correctly tells KeymanWeb which device is active.

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:

PR #6908, opened 2022-07-07
fix(developer): add no-ops for symbol builds for components

mcdurdin:
Fixes the release build where we do `nmake symbols` for developer/src, which was failing on calls to these two sets of components. (In windows/src, the build structure is different due to an additional layer of folders, so we end up with no `nmake symbols` call on the common components).

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.28-alpha](https://keyman.com/downloads/releases/alpha/16.0.28)

PR #6918, opened 2022-07-09
feat(developer): compiler regression support (build kmcomp-$version.zip during release build) 🚦

mcdurdin:
Replaces the script building kmcomp-$version.zip in TeamCity. Will be used for more automated keyboard testing shortly.

Means kmcomp-$version.zip will be available as an artifact in all test builds.

I have adjusted the TeamCity release build configuration for Developer to only build kmcomp-$version.zip if it needs to. Once this merges, I will cleanup the old code in that build configuration as it is 16.0+ (the older build is part of Keyman for Windows 14.0 build and that won't change for now).

Note: also moves the distribution point for Keyman Developer Server from src/server/build to bin/server (where it really belongs and can be included in the archives more easily).

@keymanapp-test-bot skip

### DIFF
--- a/common/windows/cpp/src/keymansentry.cpp
+++ b/common/windows/cpp/src/keymansentry.cpp
@@ -222,6 +222,11 @@ void keyman_sentry_report_start() {
 }
 
 int keyman_sentry_main(bool is_keyman_developer, const char *logger, int argc, char *argv[], int (*run)(int, char**)) {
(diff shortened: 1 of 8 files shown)


mcdurdin:
kmcomp-$version.zip looks good:

image

rc-swag (approved):
lgtm

PR #6920, opened 2022-07-11
feat(developer): Option to skip embed of compiler version 🚦

mcdurdin:
If `-no-compiler-version` is specified, then we skip the embedding of the `TSS_COMPILEDVERSION` and `TSS_KEYMANCOPYRIGHT` stores. This is intended for use with regression test builds of keyboards when we want to verify changes to the compiler.

This patch introduces the new export `SetCompilerOptions` in kmcmpdll, which should make it easier to include new options like this in future versions.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

Issue #6167, opened 2022-01-26
chore: add other non-printing characters to OSK font


mcdurdin:
> **Is there any way to mark special characters, such as: ZWSP, ZWNJ, ZWJ visually on the OSK?**

See https://github.com/silnrsi/font-keymanweb-osk/issues/3

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyboards/issues/1710#issuecomment-1022704548_

darcywong00:
KMW OSK font updated in #9469 so remaining work is to use the lebels (*ZWSP* etc)

mcdurdin:
Calling this done, enough for now.

PR #6936, opened 2022-07-12
feat(developer): add hint support to Touch Layout Editor and schema 🙊

mcdurdin:
@keymanapp-test-bot skip

I ended up with all the changes in one PR to support hints in Keyman Developer and the schema.

I reworked the touch layout editor to support hints, including moving the main control toolbar so that additional properties could be neatly fit into the design.

image

Green coloured hint text indicates that the hint has been customised:
image

I moved the delete button off the key cap itself so that it doesn't overlap the hint text:
image

I moved the controls off the top and into a left-hand-side toolbar, moving some functions into Edit dialogs so we have more flexibility for future functionality:
image

Hint text shows a placeholder of the default hint (in this case from the longpress) if no custom hint has been applied:
image

Platform properties dialog:
image

View options dialog:
image

Note that the "Show all modifier options" check box would be rarely used so it's good to get it out of the main UI.

### DIFF
--- a/common/schemas/keyman-touch-layout/README.md
+++ b/common/schemas/keyman-touch-layout/README.md
@@ -84,16 +84,26 @@ optionally contain three additional properties: `font`, `fontsize`, and
 : A boolean value, if `true`, then the base layout's key cap will be displayed
   in the top left of the key cap. Defaults to `false`.
 
-* `displayHint`
+* `defaultHint`
 
-: A string value, one of `"none"`, `"dot"`, or `"text"`. Defaults to `"dot"`.
-  Controls the display of the hint on the key cap:
+: A string value. Default is `"dot"`.  Controls the display of the hint text on
+  the key cap, which is shown at the top right of the key cap.
 
   * `"none"`: show no hint on the key cap.
-  * `"dot"`: show a dot on top-right of key cap if a longpress menu
-    is available, default and only option for Keyman 15.0 and earlier.
-  * `"text"`: show the text `hint` property if present, often on
-    top-right of key cap.
+  * `"dot"`: show a dot on top-right of key cap if a longpress menu is
+    available, default. Matches behavior for Keyman 15.0 and earlier.
+  * `"longpress"`: show the default longpress key cap, if it exists.
+  * `"multitap"`: show the first multitap key cap, if it exists.
+  * `"flick"`: show the first flick key cap, selected in clockwise order,
+    starting from North.
+  * `"flick-n"`: show the key cap from the North flick, if it exists.
+  * `"flick-ne"`: show the key cap from the North-East flick, if it exists.
+  * `"flick-e"`: show the key cap from the East flick, if it exists.
+  * `"flick-se"`: show the key cap from the South-East flick, if it exists.
+  * `"flick-s"`: show the key cap from the South flick, if it exists.
+  * `"flick-sw"`: show the key cap from the South-West flick, if it exists.
+  * `"flick-w"`: show the key cap from the West flick, if it exists.
+  * `"flick-nw"`: show the key cap from the North-West flick, if it exists.
 
   Introduced in Keyman 16.0.
 
@@ -219,6 +229,9 @@ object](#sk-object). A `key` object must include either an `id` or `sp` property
 
 : String. If present, a small hint, often shown at top right of key cap.
 
+  Introduced in Keyman 16.0.
+
+
 * `layer`
 
 : String. If specified, overrides the modifier associated with the layer (this
@@ -309,7 +322,9 @@ object](#sk-object). A `key` object must include either an `id` or `sp` property
 
 : If present, defines additional keys that may be selected when using a
   flick gesture on the key, with a [`flick` object](#flick-object). Flicks
-  should not be used on modifier keys, special keys, or the Spacebar. Keyman 16+.
+  should not be used on modifier keys, special keys, or the Spacebar.
+
+  Introduced in Keyman 16.0.
 
 * `multitap`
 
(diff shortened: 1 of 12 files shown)


common/schemas/keyman-touch-layout/README.md
```diff
@@ -84,6 +84,19 @@ optionally contain three additional properties: `font`, `fontsize`, and
...
+* `displayHint`
+
+: A string value, one of `"none"`, `"dot"`, or `"text"`. Defaults to `"dot"`.
+  Controls the display of the hint on the key cap:
+
+  * `"none"`: show no hint on the key cap.
+  * `"dot"`: show a dot on top-right of key cap if a longpress menu
+    is available, default and only option for Keyman 15.0 and earlier.
+  * `"text"`: show the text `hint` property if present, often on
+    top-right of key cap.
```
jahorton, line 96: "Often on"?  What decides whether or not it is shown there?

Keeping in #4879 and #5079 both in mind, and the fact that we have traditionally displayed underlying key caps on the top-_left_ of a key, what's the plan for this behavior?  Will we drop the "top-left" part when using hints for underlying key caps, assuming the latter part of that clause?  Or will there be a way to select where it is displayed?

mcdurdin: Repositioning hints will be possible once we support theming (17.0 currently). It's all just CSS :grin: This text could just say 'on top-right' and leave the word 'often' out but figured it didn't hurt too much.

mcdurdin: And note that hints != underlying key. These are two separate strings that may *both* be displayed. The one thing that does go when we use hints is the "." marker indicating a longpress (see `displayHints` for how we control this).

jahorton (commented):
Design's a definite LGTM, and most of the code is as well.  I do have a few concerns re: the documentation, though.

common/schemas/keyman-touch-layout/README.md
```diff
@@ -292,19 +322,24 @@ object](#sk-object). A `key` object must include either an `id` or `sp` property
...
-  should not be used on modifier keys, special keys, or the Spacebar. Keyman 16+.
+  should not be used on modifier keys, special keys, or the Spacebar.
+
+  Introduced in Keyman 16.0.
 
 * `multitap`
 
 : If present, defines an array of keys that may be selected sequentially when
   tapping multiple times on the key, an array of [`sk` objects](#sk-object).
   Multitaps generally should not be used on modifier keys, special keys, or the
```
mcdurdin, line 333:
```suggestion
  Multitaps may not be used on layer switch keys, special keys such as the globe key, or the
```

common/schemas/keyman-touch-layout/README.md
```diff
@@ -292,19 +322,24 @@ object](#sk-object). A `key` object must include either an `id` or `sp` property
...
   Spacebar. Beware of mixing multitaps with layer switching, as this can lead to
-  a confusing user experience. Keyman 16+.
+  a confusing user experience.
```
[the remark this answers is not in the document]
mcdurdin: Yep.

common/schemas/keyman-touch-layout/README.md
```diff
@@ -292,19 +322,24 @@ object](#sk-object). A `key` object must include either an `id` or `sp` property
...
 
   Keyman 15+: the Shift key has a predefined multitap sequence to access the
   `caps` layer, which is available if the `caps` layer is present. This is the
-  only multitap sequence supported in Keyman 15.
+  only multitap sequence supported in Keyman 15. This multitap sequence cannot
+  be overridden.
```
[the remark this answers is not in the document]
mcdurdin:
```suggestion
  be overridden, and is the sole case where multitap input may cause a layer switch.
```

jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.38-alpha](https://keyman.com/downloads/releases/alpha/16.0.38)

MayuraVerma:
How to disable default "." for hint?
Is adding "space" a good way?
How to display "ZWJ" or "ZWNJ" symbol along with text?

mcdurdin:
> How to disable default "." for hint?

In version 17+, this can be done in the touch layout editor in the Platform properties dialog:

image

> Is adding "space" a good way?

I don't really understand this question!

> How to display "ZWJ" or "ZWNJ" symbol along with text?

* For ZWNJ, uou can use `*ZWNJ*` on the keycap, e.g. which will look like this:

  image

  Full list: https://help.keyman.com/developer/16.0/guides/develop/creating-a-touch-keyboard-layout-for-amharic-the-nitty-gritty#toc-key-text

* ZWJ is not yet supported, see #6167

MayuraVerma:
This "Default Hint Source" option is not present in latest developer
image

"Is adding "space" a good way?
I don't really understand this question!"
I am using "Space" as custom hint to clear the "dot".  Looks like there will be a new option to set "No default hint". However this feature is still not avaiable in 17.0.185 alpha developer.
when the option is implemented. If user wants a custom text on few keys, then he/she sets the option to "No default hint" and add custom text only to keys needed? Is it the idea?

Two Unicode fields could be confusing. It could be "Hint_Text" and "Hint_Unicode", instead.
image

Is custom hint implemented in

MayuraVerma:
How to show *ZWNJ* symbol along with a text for "Key display text"

mcdurdin:
> This "Default Hint Source" option is not present in latest developer

This is a platform-level property, not a global option for the touch layout. It should be present in 17.0.185-alpha

> If user wants a custom text on few keys, then he/she sets the option to "No default hint" and add custom text only to keys needed? Is it the idea?

Yes, that's correct.

> Two Unicode fields could be confusing. It could be "Hint_Text" and "Hint_Unicode", instead.

Good point. Can we you this in a separate issue?

> How to show _ZWNJ_ symbol along with a text for "Key display text"

Currently, we don't support `*ZWNJ* text` -- the symbol must be the entire key cap text of `*ZWNJ*`. If this is a feature you'd like, can you open an issue for it so we can track it?

PR #6942, opened 2022-07-14
fix(developer): crash when switching back a tab 🎾

mcdurdin:
In rare circumstances, Keyman Developer can crash during startup if you try and switch back to a tab which is not yet available (Ctrl+Shift+Tab).

@keymanapp-test-bot skip


keyman-server:

PR #6947, opened 2022-07-14
fix(developer): crash when switching back a tab 🎾 🍒

mcdurdin:
Cherry-pick of #6942.

In rare circumstances, Keyman Developer can crash during startup if you try and switch back to a tab which is not yet available (Ctrl+Shift+Tab).

@keymanapp-test-bot skip


ermshiperete (approved):

keyman-server:

PR #6997, opened 2022-07-26
chore(linux): Add script to upload to debian

ermshiperete:
- Create and sign source package
- Upload to mentors.debian.net
- Create (and optionally push) commit with updated changelog file based on stable branch
- Create (and optionally push) commit with updated changelog file based on master branch

@keymanapp-test-bot skip

### DIFF
--- a/linux/scripts/upload-to-debian.sh
+++ b/linux/scripts/upload-to-debian.sh
@@ -1,18 +1,31 @@
-#!/bin/bash
-# Upload the latest release to mentors.debian.net. Call from stable-* branch.
-# Usage: upload-to-debian.sh -k <Debian signing key>
-# Requires an entry in ~/.dput.cf:
-# ```
-# [mentors]
-# fqdn = mentors.debian.net
-# incoming = /upload
-# method = https
-# allow_unsigned_uploads = 0
-# progress_indicator = 2
-# allowed_distributions = .*
-# ```
+#!/usr/bin/env bash
+set -eu
 
-set -e
+usage()
+{
+    cat - <<- END
+
+${0} -k <Debian signing key> [-n] [--push]
+
+Create source package for latest stable release and upload to Debian,
+and update (and optionally push) changelog file.
+
+    -k <key>    The PGP key used to sign the source package
+    -n          Simulate upload
+    --push      Push changelog changes to GitHub
+    --help      Display usage help
+
+Requires an entry in ~/.dput.cf:
+    [mentors]
+    fqdn = mentors.debian.net
+    incoming = /upload
+    method = https
+    allow_unsigned_uploads = 0
+    progress_indicator = 2
+    allowed_distributions = .*
+END
+}
+
 
 ## START STANDARD BUILD SCRIPT INCLUDE
 # adjust relative paths as necessary
@@ -20,73 +33,49 @@ THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BA
 . "$(dirname "$THIS_SCRIPT")/../../resources/build/build-utils.sh"
 ## END STANDARD BUILD SCRIPT INCLUDE
 
-PROGRAM_NAME="$(basename "$0")"
-RED='\033[0;31m'
-GREEN='\033[0;32m'
-NC='\033[0m' # No Color
+. "$KEYMAN_ROOT/resources/shellHelperFunctions.sh"
 
-error()
-{
-	echo -e "${RED}$PROGRAM_NAME: $1${NC}" >&2
-}
-
-log()
-{
-	echo -e "${GREEN}$PROGRAM_NAME: $1${NC}"
-}
-
-usage()
-{
-    echo ""
-    echo "Usage:"
-    echo "${0} -k <Debian signing key> [-n] [--push]"
-    echo ""
-    echo "Create source package for latest stable release and upload to Debian,"
-    echo "and update (and optionally push) changelog file."
-    echo ""
-    echo "    -k <key>    The PGP key used to sign the source package"
-    echo "    -n          Simulate upload"
-    echo "    --push      Push changelog changes to GitHub"
-    echo "    --help      Display usage help"
-}
+NOOP=
+PUSH=
+DEBKEYID=
 
 while (( $# )); do
     case $1 in
-        -k) shift; debkeyid=$1;;
+        -k) shift; [ ! $# -eq 0 ] && DEBKEYID=$1 || fail "Error: The last argument is missing a value. Exiting.";;
         -n) NOOP=: ;;
         --help) usage ; exit 0 ;;
         --push) PUSH=1 ;;
-        *) error "Error: Unexpected argument \"$1\". Exiting." ; exit 1 ;;
+        *) fail "Error: Unexpected argument \"$1\". Exiting." ;;
     esac
-    shift || (error "Error: The last argument is missing a value. Exiting."; false) || exit 2
+    shift || fail "Error: The last argument is missing a value. Exiting."
 done
 
-if [ -z "$debkeyid" ]; then
+if [ -z "$DEBKEYID" ]; then
     usage
-    exit 3
+    exit 2
 fi
 
 if ! git diff --quiet; then
-    error "You have changed files in your git working directory. Exiting."
-    exit 4
+    fail "You have changed files in your git working directory. Exiting."
 fi
 
+echo_heading "Fetching latest changes"
+git fetch -p origin
 stable_branch=$(git branch -r | grep origin/stable- | sort | tail -1)
 stable_branch=${stable_branch##* }
 git checkout ${stable_branch#origin/}
 
 cd $KEYMAN_ROOT/linux
-log "Building source package"
+echo_heading "Building source package"
 DIST=unstable scripts/debian.sh
 cd debianpackage/
-log "Signing source package"
-debsign -k$debkeyid --re-sign *.changes
-log "Uploading packages to mentors.debian.net"
+echo_heading "Signing source package"
+debsign -k$DEBKEYID --re-sign *.changes
+echo_heading "Uploading packages to mentors.debian.net"
 $NOOP dput mentors *.changes
 cd ..
 
-log "Updating changelog"
-git fetch -p origin
+echo_heading "Updating changelog"
 git checkout -B chore/linux/changelog $stable_branch
 cp debianpackage/keyman-*/debian/changelog debian/
 git add debian/changelog
@@ -100,3 +89,6 @@ git cherry-pick -x chore/linux/changelog
 if [ -n "$PUSH" ]; then
     $NOOP git push origin chore/linux/cherry-pick/changelog
 fi
+
+echo_heading "Finishing"
+git checkout master
--- a/linux/scripts/upload-to-debian.sh
+++ b/linux/scripts/upload-to-debian.sh
@@ -65,7 +65,7 @@ stable_branch=$(git branch -r | grep origin/stable- | sort | tail -1)
 stable_branch=${stable_branch##* }
 git checkout ${stable_branch#origin/}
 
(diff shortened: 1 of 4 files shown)


linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+# [mentors]
+# fqdn = mentors.debian.net
+# incoming = /upload
+# method = https
+# allow_unsigned_uploads = 0
+# progress_indicator = 2
+# allowed_distributions = .*
+# ```
+
+set -e
```
mcdurdin, line 15:
```suggestion
set -eu
```

For consistency in our bash scripts, and also because it helps trap typos. But does make working with unset variables much more tedious -- and you'd have to fix up all of your refs to your param vars `NOOP`, `PUSH`, `debkeyid`. So consider this optional...

I note also that `fail()` in shellHelperFunctions.sh -- see later comments -- can barf with `-u` so one day should fix that too...

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
mcdurdin, line 1:
```suggestion
#!/usr/bin/env bash
```

While this is not necessary for Linux, it doesn't hurt, and I'd like to try and keep our bash script boilerplate consistent for ease of maintenance.

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+
+error()
+{
+	echo -e "${RED}$PROGRAM_NAME: $1${NC}" >&2
+}
+
+log()
+{
+	echo -e "${GREEN}$PROGRAM_NAME: $1${NC}"
+}
```
mcdurdin, line 36: I think, source shellHelperFunctions.sh and use its defined colors:

https://github.com/keymanapp/keyman/blob/be7c9c58a9e81f596a64065cbbb0a58b786d9446/resources/shellHelperFunctions.sh#L85-L114

But we can probably improve what we have in shellHelperFunctions.sh!

Rename `PROGRAM_NAME` to `THIS_SCRIPT_NAME`, put that definition into build-utils.sh (this matches `THIS_SCRIPT` and `THIS_SCRIPT_PATH`.

Then, some of this may belong in our shellHelperFunctions.sh include, although probably with less generic names:

* rename `log` to `log_message` (or you may choose to use `echo_heading`?)
* rename `error` to `log_error`
* add `log_warning` (don't change `warn()` because it is used elsewhere)

```bash
log_message() {
  echo "${COLOR_GREEN}$THIS_SCRIPT_NAME: $*${COLOR_RESET}"
}

log_error() {
  echo "${COLOR_RED}$THIS_SCRIPT_NAME: $*${COLOR_RESET}"
}

log_warning() {
  echo "${COLOR_YELLOW}$THIS_SCRIPT_NAME: $*${COLOR_RESET}"
}
```

Note: we already have a chore to unify build-utils.sh and shellHelperFunctions.sh -- don't worry about doing that for now.

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+
+if [ -z "$debkeyid" ]; then
+    usage
+    exit 3
+fi
+
+if ! git diff --quiet; then
+    error "You have changed files in your git working directory. Exiting."
+    exit 4
+fi
```
mcdurdin, line 72: I wonder if we should be using ``if [[ `git status --porcelain` ]]; then`` instead?

https://stackoverflow.com/a/25149786/1836776

So many ways to skin a cat but `git diff` will miss untracked files, won't it?

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+
+stable_branch=$(git branch -r | grep origin/stable- | sort | tail -1)
```
mcdurdin, line 74: This assumes that your remote tracking branches are up-to-date. Which means this may track the wrong release the first time the script is run after a stable release, if you haven't `git fetch`ed?

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+stable_branch=${stable_branch##* }
+git checkout ${stable_branch#origin/}
```
mcdurdin, line 76: You're checking out the local stable branch here -- but further down on line 90 you refer to the remote tracking branch because you don't strip `origin/`. Is that your intention?

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+
+cd $KEYMAN_ROOT/linux
```
mcdurdin, line 78:
```suggestion
cd "$KEYMAN_ROOT/linux"
```

In Case Of Space.

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+# Requires an entry in ~/.dput.cf:
+# ```
+# [mentors]
+# fqdn = mentors.debian.net
+# incoming = /upload
+# method = https
+# allow_unsigned_uploads = 0
+# progress_indicator = 2
+# allowed_distributions = .*
+# ```
```
mcdurdin, line 13: Can you move the `usage()` function up here and merge the details between the comment and the function -- i.e. include all the detail in `usage()`'s output? That way we don't need them twice.

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+cd debianpackage/
+log "Signing source package"
+debsign -k$debkeyid --re-sign *.changes
+log "Uploading packages to mentors.debian.net"
+$NOOP dput mentors *.changes
+cd ..
+
+log "Updating changelog"
+git fetch -p origin
+git checkout -B chore/linux/changelog $stable_branch
```
mcdurdin, line 90: If `chore/linux/changelog` branch already exists, this is going to fail here so it'd be good to check for that branch before you start this process.

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+
+while (( $# )); do
+    case $1 in
+        -k) shift; debkeyid=$1;;
+        -n) NOOP=: ;;
+        --help) usage ; exit 0 ;;
+        --push) PUSH=1 ;;
+        *) error "Error: Unexpected argument \"$1\". Exiting." ; exit 1 ;;
+    esac
+    shift || (error "Error: The last argument is missing a value. Exiting."; false) || exit 2
```
mcdurdin, line 61:
```suggestion
    shift || fail "Error: The last argument is missing a value. Exiting."
```

Let's use our `fail` helper since we have it. We lose the specific exit code but it's probably not important? ;-)

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+    echo ""
+    echo "    -k <key>    The PGP key used to sign the source package"
+    echo "    -n          Simulate upload"
+    echo "    --push      Push changelog changes to GitHub"
+    echo "    --help      Display usage help"
+}
+
+while (( $# )); do
+    case $1 in
+        -k) shift; debkeyid=$1;;
```
mcdurdin, line 55:
```suggestion
        -k) shift; DEBKEYID=$1;;
```

I wonder if we should upcase this given our other vars are upper case?

linux/scripts/upload-to-debian.sh
```diff
@@ -0,0 +1,102 @@
...
+git commit -m "chore(linux): Update debian changelog"
+if [ -n "$PUSH" ]; then
+    $NOOP git push origin chore/linux/changelog
+fi
+
+git checkout -B chore/linux/cherry-pick/changelog origin/master
+git cherry-pick -x chore/linux/changelog
+if [ -n "$PUSH" ]; then
+    $NOOP git push origin chore/linux/cherry-pick/changelog
+fi
```
mcdurdin, line 102: Do you want to `git checkout master` at the end of this process? Not required, but leaves the repo state a little tidier?

mcdurdin (changes requested):
Well... I had fun nitpicking here 😁  Take these comments for what you will -- but hoping that some of them help us with less painful future maintenance and consistency!

linux/scripts/upload-to-debian.sh
ermshiperete: No, it doesn't fail :smile: That's what the uppercase `-B` is for. It resets the branch if it already exists - which would be ok since an existing branch should have a version of `changelog` that's either part of the new `changelog` or superseded. Unless you use that branch for something completely unrelated, in which case you're out of luck.

linux/scripts/upload-to-debian.sh
ermshiperete: I don't think we care about untracked files in this script. We only modify `changelog`, which already exists. There could be some untracked files which would be overwritten by switching branches, but then `git checkout` will fail. And there might be untracked files which won't be overwritten, but might be in the workspace for a reason. 

We could use `if [[ `git status --porcelain --untracked-files=no` ]]; then` to ignore untracked files, but than that's not much different from what we have, so I'll leave that.

resources/shellHelperFunctions.sh
```diff
@@ -90,16 +90,26 @@ if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; th
     COLOR_BLUE=$(tput setaf 4)
     COLOR_YELLOW=$(tput setaf 3)
     COLOR_RESET=$(tput sgr0)
+    SETMARK='\x1b]1337;SetMark\x07'
```
mcdurdin, line 93:
```suggestion
    # e.g. VSCode https://code.visualstudio.com/updates/v1_69#_setmark-sequence-support
    HEADING_SETMARK='\x1b]1337;SetMark\x07'
```

I didn't know what this did!

Thought we should make the variable name slightly less generic?

resources/shellHelperFunctions.sh
```diff
@@ -90,16 +90,26 @@ if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; th
...
 else
     COLOR_RED=
     COLOR_GREEN=
     COLOR_BLUE=
     COLOR_YELLOW=
     COLOR_RESET=
+    SETMARK=
```
mcdurdin, line 100:
```suggestion
    HEADING_SETMARK=
```

resources/shellHelperFunctions.sh
```diff
@@ -90,16 +90,26 @@ if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; th
...
 fi
 
 echo_heading() {
-  echo "${COLOR_BLUE}$*${COLOR_RESET}"
+  echo -e "${SETMARK}${COLOR_BLUE}$*${COLOR_RESET}"
```
mcdurdin, line 104:
```suggestion
  echo -e "${HEADING_SETMARK}${COLOR_BLUE}$*${COLOR_RESET}"
```

mcdurdin (approved):
Changes LGTM!

linux/scripts/upload-to-debian.sh
ermshiperete: yes, that's intentional. We need to be at the stable branch so that `scripts/debian.sh` picks up the correct version. But when we create the changelog branch we want to base it on the remote branch.

mcdurdin: Can we document that in the code?

ermshiperete: Done in #7009

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.40-alpha](https://keyman.com/downloads/releases/alpha/16.0.40)

PR #7003, opened 2022-07-28
fix(developer): increase maximum file size for server 🎾

mcdurdin:
Some models are larger than 10MB.

Ref: https://community.software.sil.org/t/text-suggestion-not-working-in-some-cases/6490

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.42-alpha](https://keyman.com/downloads/releases/alpha/16.0.42)

PR #7013, opened 2022-07-29
fix(developer): compiler emitting garbage for readonly groups 🎾

mcdurdin:
Picked up while reviewing code for conversion to C++. The compiler could, in some circumstances emit garbage code for readonly groups because we weren't testing all scenarios correctly. In practice, this would have been rare as readonly groups don't emit characters, but it should be fixed!

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tike/compile/CompileKeymanWeb.pas
+++ b/developer/src/tike/compile/CompileKeymanWeb.pas
@@ -1038,7 +1038,8 @@ begin
     begin
 			if InQuotes then
       begin
-        Result := Result + '");';
+        if not fgp.fReadOnly then
+          Result := Result + '");';
         InQuotes := False;
       end;
 
@@ -1223,9 +1224,9 @@ begin
         InQuotes := True; len := -1;
       end;
 
-      if rec.ChrVal in [Ord('"'), Ord('\')] then Result := Result + '\';
       if not fgp.fReadOnly then
       begin
+        if rec.ChrVal in [Ord('"'), Ord('\')] then Result := Result + '\';
         Result := Result + Javascript_String(rec.ChrVal);  // I2242
       end;
     end;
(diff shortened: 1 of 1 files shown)


jahorton (commented):
I'd personally find it helpful to see a 'before' and 'after' for an affected keyboard... and ideally a unit test based on that.  Mostly because it's a bit hard to envision the problem & how it fixes it from just the context shown by GitHub.

mcdurdin:
This is a followup to #6865. Further detail on the `fReadonly` logic is available there.

> I'd personally find it helpful to see a 'before' and 'after' for an affected keyboard... and ideally a unit test based on that. Mostly because it's a bit hard to envision the problem & how it fixes it from just the context shown by GitHub.

So I just tried to do this and discovered that it's actually not possible for the compiler to hit these locations, barring a bug in the KMX compiler. The C++ KMX compiler will block the build for these code paths, for readonly groups, e.g. with the given code:

```
group(readonly) readonly
if(&platform='touch') > 'a'
```

the kmx compiler emits the following error:

```
keymanweb_7013.kmn (15): Error: 406E Output is not permitted in a readonly group
```

So, all that this is doing, is keeping the logic consistent with other places where in the same function where it is important to keep the output from being emitted (e.g. CODE_USE, CODE_CONTEXT).

Given that, do you think I should abandon the 15.0 cherry-pick because this is really just about code quality and consistency rather than an achieveable bug -- and just leave this in 16.0?

Here is an example of what I am trying to achieve, all other places where we have calls to KeymanWeb functions that emit output, these calls are wrapped in a test for `not fgp.fReadOnly`, so this is just keeping that pattern:

https://github.com/keymanapp/keyman/blob/5895d26e648dd5e45e4821eb9bd1cf00dfd15cc0/developer/src/tike/compile/CompileKeymanWeb.pas#L1081-L1089

jahorton (approved):

keyman-server:

PR #7014, opened 2022-07-29
fix(developer): compiler emitting garbage for readonly groups 🎾 🍒

mcdurdin:
Picked up while reviewing code for conversion to C++. The compiler could, in some circumstances emit garbage code for readonly groups because we weren't testing all scenarios correctly. In practice, this would have been rare as readonly groups don't emit characters, but it should be fixed!

@keymanapp-test-bot skip


ermshiperete (approved):

ermshiperete:
By the way, :cherries:-picked from #7013 :smile:

keyman-server:

PR #7016, opened 2022-07-29
fix(developer): rebuilding a model would not refresh it in server 🎾 🍒

mcdurdin:
Cherry-pick of #7015.

When rebuilding a model, Keyman Developer would query for the registered model, but a missing simplifyId call meant that it never matched the model name (due to `.` vs `_`).

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.268](https://keyman.com/downloads/releases/stable/15.0.268)

PR #7018, opened 2022-07-29
chore(common): build-utils.sh minor tweaks

mcdurdin:
Fixing up a number of minor bits and pieces identified during demo:

* invalid action or :target would not be picked up if paired with a valid one (e.g.: build:foo should not have worked)
* colors defined in both shellHelperFunctions and build-utils were problematic; able to remove them from shellHelperFunctions because all references included both scripts
* builder_has_action and builder_report now allow both `action:target` and `action :target`
* keyman-version/build.sh did not have working test action...

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/build-utils.sh
+++ b/resources/build/build-utils.sh
@@ -270,21 +270,31 @@ replaceVersionStrings_Mkver() {
 # Standard build script functions for managing command line, actions and targets
 ################################################################################
 
-# TODO: colors are defined here and in shellHelperFunctions.sh
 # The following allows coloring of warning and error lines, but only works if there's a
 # terminal attached, so not on the build machine.
-if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; then
+builder_use_color() {
+  if $1; then
     COLOR_RED=$(tput setaf 1)
     COLOR_GREEN=$(tput setaf 2)
     COLOR_BLUE=$(tput setaf 4)
     COLOR_YELLOW=$(tput setaf 3)
     COLOR_RESET=$(tput sgr0)
-else
+    # e.g. VSCode https://code.visualstudio.com/updates/v1_69#_setmark-sequence-support
+    HEADING_SETMARK='\x1b]1337;SetMark\x07'
+  else
     COLOR_RED=
     COLOR_GREEN=
     COLOR_BLUE=
     COLOR_YELLOW=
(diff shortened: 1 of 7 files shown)


resources/build/build-utils.sh
```diff
@@ -270,21 +270,31 @@ replaceVersionStrings_Mkver() {
 # Standard build script functions for managing command line, actions and targets
 ################################################################################
 
-# TODO: colors are defined here and in shellHelperFunctions.sh
 # The following allows coloring of warning and error lines, but only works if there's a
 # terminal attached, so not on the build machine.
-if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; then
+builder_use_color() {
```
ermshiperete, line 275: I don't think this method is intended to be used outside of this file, is it?

```suggestion
_builder_use_color() {
```

resources/build/build-utils.sh
```diff
@@ -270,21 +270,31 @@ replaceVersionStrings_Mkver() {
...
     COLOR_YELLOW=
     COLOR_RESET=
+    HEADING_SETMARK=
+  fi
+}
+
+if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; then
+  builder_use_color true
+else
+  builder_use_color false
```
ermshiperete, line 297:
```suggestion
  _builder_use_color true
else
  _builder_use_color false
```

ermshiperete (approved):

ermshiperete (approved):

resources/shellHelperFunctions.sh
```diff
@@ -82,32 +82,9 @@ get_platform_folder() {
...
-}
-
-if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; then
-  shell_helper_color true
-else
-  shell_helper_color false
-fi
+# Note: COLOR_ and HEADING_ variables are defined in build-utils.sh
+# which is already required for scripts so these should be safe.
+# (part of TODO: consolidate these two scripts)
```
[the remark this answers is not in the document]
ermshiperete: yes, with the consolidate source package it should now be possible to use the standard include script.

jahorton (approved):

keyman-server:

PR #7023, opened 2022-08-01
chore(core): get wasm core building again

mcdurdin:
Updates search for emscripten if not on path, and fixes up unit test infrastructure for wasm builds for new tests.

Note that Linux config instructions do not currently list details on how to configure dependencies such as emscripten. This is a separate project!

@keymanapp-test-bot skip

### DIFF
--- a/docs/build/macos.md
+++ b/docs/build/macos.md
@@ -71,6 +71,11 @@ These dependencies are also listed below if you'd prefer to install manually.
   brew install node emscripten openjdk@8
   ```
 
+  Note: if you install emscripten with brew on macOS, only emscripten binaries
+  are added to the path via symlinks. This makes it reasonably safe to have
+  emscripten on the path, unlike on other platforms where emscripten also ends
+  up adding its versions of node, python, and other binaries to the path.
+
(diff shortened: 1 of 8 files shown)


docs/build/macos.md
```diff
@@ -71,6 +71,11 @@ These dependencies are also listed below if you'd prefer to install manually.
   brew install node emscripten openjdk@8
   ```
 
+  Note: if you install emscripten with brew on macOS, only emscripten binaries
+  are added to the path via symlinks. This makes it reasonably safe to have
+  emscripten on the path, unlike on other platforms where emscripten also ends
+  up adding its versions of node, python, and other binaries to the path.
+
```
srl295, line 78: Brew has a minimalist and compartmentalized approach.

srl295 (approved):

docs/build/macos.md
mcdurdin: Seems like everything has an opinion 🤣... should it be on the path? should it be off the path? bah!

keyman-server:

PR #7031, opened 2022-08-03
fix(developer): stack overflow when compiling non-web keyboard 🎾 

### LINKED ISSUE #7030: bug(developer): building a keyboard in Developer can sometimes induce a stack overflow
**Describe the bug**

Reported by @DavidLRowe.

https://sentry.io/organizations/keyman/issues/3396183413/events/4dc840c6fdcd41a5b746ef585196d15a/?project=5983519

**To Reproduce**

The keyboard compiler can enter into a loop state with the following steps:

1. A keyboard has a 'web' target
2. The keyboard has been tested on web (meaning it was listed in the web debugger)
3. The 'web' target (and any web-related targets) are removed from the keyboard
4. Press F7 to compile the keyboard within the Keyboard Editor.

In this situation, the web debug interface detects that the keyboard is already listed in web debugger and looks to recompile it because the keyboard source file is newer. But because the keyboard .js is no longer built, it ends up out of date, which means that when it finishes compiling, it tests again ... and goes into a loop.

---

**Keyman Developer:**
 - OS:                       any 

mcdurdin:
Fixes #7030.

Prevents a stack overflow / loop when the web target is removed from a keyboard and it has already been tested on web, and the user presses the Compile button.

# User Testing

* **TEST_COMPILE:** Make sure the compiler no longer goes into a loop and crashes.

  1. Create a new keyboard project, targeting 'all'
  2. Compile the keyboard, and start a web test on it.
  3. Change the `store(&targets)` line to `store(&targets) 'windows'`
  4. Press Compile again.
  5. The compiler should only compile once!

  (Note: you may want to test this procedure in 15.0.267 to observe it going wrong!)

### DIFF
--- a/developer/src/tike/project/Keyman.Developer.UI.Project.kmnProjectFileUI.pas
+++ b/developer/src/tike/project/Keyman.Developer.UI.Project.kmnProjectFileUI.pas
@@ -67,6 +67,7 @@ uses
   UfrmKeymanWizard,
   UfrmKeyboardFonts,
   UfrmMDIEditor,
+  UKeymanTargets,
   UmodWebHttpServer,
   Keyman.Developer.System.ServerAPI,
   Keyman.Developer.UI.ServerUI,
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

bharanidharanj:
* **TEST_COMPILE (PASSED):** Tested this as per the instructions in Keyman Developer 16.0.43-alpha-test-7031 in Windows and it seems that it compiled once.

keyman-server:

PR #7034, opened 2022-08-03
fix(developer): stack overflow when compiling non-web keyboard 🎾 🍒

### LINKED ISSUE #7030: bug(developer): building a keyboard in Developer can sometimes induce a stack overflow

mcdurdin:
Cherry-pick of #7031.

Fixes #7030.

Prevents a stack overflow / loop when the web target is removed from a keyboard and it has already been tested on web, and the user presses the Compile button.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/developer/TIKE/project/Keyman.Developer.UI.Project.kmnProjectFileUI.pas
+++ b/windows/src/developer/TIKE/project/Keyman.Developer.UI.Project.kmnProjectFileUI.pas
@@ -67,6 +67,7 @@ uses
   UfrmKeymanWizard,
   UfrmKeyboardFonts,
   UfrmMDIEditor,
+  UKeymanTargets,
   UmodWebHttpServer,
   Keyman.Developer.System.ServerAPI,
   Keyman.Developer.UI.ServerUI,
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.268](https://keyman.com/downloads/releases/stable/15.0.268)

Issue #7043, opened 2022-08-05
spec(common): KMX+ binary format supporting LDML


srl295:
## Introduction

Document an optimized binary format for use by the process_event #5015  and other consumers of LDML data.
It will be written into the KMX binary file.

## Goals

- new section is 64-bit aligned
- new section and header uses 64-bit offsets

## implementation

- kmx_file.h - version16, kmx+ file header
- kmx+ optimized binary format
- in-memory structures for LDML keyboard

## other

- json or xml streamers (dev use)
- binary streamer <-> into .kmx file (prod)
- use bson?

srl295:
# KMX Plus Binary Format #7043

- Authors: MD SL

## C7043.0 Introduction

This document discusses the binary format for KMXPlus, which contains keyboards converted from source LDML data.

Draft spec PR: <https://github.com/unicode-org/cldr/pull/1847>

## C7043.1 Principles

- The data described here is located at byte offset `dpKMXPlus`.
- All integer values are unsigned 32-bit little-endian unless otherwise specified.
- All strings are UTF-16LE unless otherwise specified. (See the `strs` section.) String data items are identified with `str:`, indicating a 32 bit index into the `strs` table.
- All offsets are 32-bit little-endian values.  For all sections except for the `'sect'` section (which see), offsets are relative to the beginning of each section.

## C7043.2 Sections

- Data is divided into several sections. The very first section is the `'sect'` section which is the table of contents.
- All sections, including the first, begin on a 64-bit boundary with zero padding as needed.
- All sections begin with a 32-bit (four character) section identifier, and a 32-bit section size.
- Other than the `sect` table itself, the rest of the sections follow in binary order in the file.  In other words, the binary ordering of the section identifiers determines the order of the file layout.

### C7043.2.1 `sect`—Section Table of contents

The very first section is a table of contents listing the rest of the sections.  The table of contents does not list itself.

This is the only section where all byte offsets are relative to the value of `dpKMXPlus`.

| ∆ | Bits | Name    | Description                              |
|---|------|---------|------------------------------------------|
| 0 |  32  | ident   | `sect`                                   |
| 4 |  32  | size    | int: Length of section                   |
| 8 |  32  | total   | int: KMXPlus entire length               |
|12 |  32  | count   | int: Number of following section headers |

Then for `count` repetitions:

| ∆ | Bits | Name    | Description                                       |
|---|------|---------|---------------------------------------------------|
[...]

PR #7055, opened 2022-08-08
feat(core): constant generator between core and developer 🙀

srl295:
#7043

build is not optimal, but works

@keymanapp-test-bot skip


srl295:
todos:
- [ ] meson calls build.sh
- [ ] generate the .h
- [ ] get the module stuff to actually work

srl295:
.h is really basic but it's something:

```c
//
// Generated File - do not edit
// This file is generated by core/tools/ldml-const-builder/build.sh
// based on core/include/ldml/keyboardprocessor_ldml.ts

#pragma once
#define LDML_meta_settings_fallback_omit 1
#define LDML_meta_settings_transformFailure_omit 2
#define LDML_meta_settings_transformPartial_hide 4
#define LDML_section_keys ((uint32_t)'keys')
#define LDML_section_loca ((uint32_t)'loca')
#define LDML_section_meta ((uint32_t)'meta')
#define LDML_section_sect ((uint32_t)'sect')
#define LDML_section_strs ((uint32_t)'strs')
#define LDML_strs_flags_extend 1
#define LDML_version 1.0
```

mcdurdin:
Our cunning scheme to involve node, npm and typescript to do the shared constant generation in the Keyman Core build may come to naught. @ermshiperete are we likely to run into trouble here with the Debian builds? (i.e. should we be avoiding dragging in the node+tsc dependencies?)

Even now we've got something not quite right; see the build log below:
```
[18:14:44][Step 2/4] npm WARN exec The following package was not found and will be installed: tsc
[18:14:44][Step 2/4] 
[18:14:44][Step 2/4]                                                                                
[18:14:44][Step 2/4]                 This is not the tsc command you are looking for                
[18:14:44][Step 2/4]                                                                                
[18:14:44][Step 2/4] 
[18:14:44][Step 2/4] To get access to the TypeScript compiler, tsc, from the command line either:
[18:14:44][Step 2/4] 
[18:14:44][Step 2/4] - Use npm install typescript to first add TypeScript to your project before using npx
[18:14:44][Step 2/4] - Use yarn to avoid accidentally running code from un-installed packages
```

I guess we need to do an `npm install` (or `npm ci`) because the build agent starts with a clean repo. We may then need to include a full path, something like `../../../node_modules/.bin/tsc` to run.

ermshiperete:
> are we likely to run into trouble here with the Debian builds? (i.e. should we be avoiding dragging in the node+tsc dependencies?)

Maybe. We'd have to add a dependency on `npm`. The one available for [Bionic is quite old (3.5)](https://packages.ubuntu.com/bionic/npm). As long as it is possible to build with that version I think we might be good.

core/include/ldml/keyboardprocessor_ldml.h
```diff
@@ -0,0 +1,18 @@
...
+#define LDML_META_SETTINGS_FALLBACK_OMIT 1
+#define LDML_META_SETTINGS_TRANSFORMFAILURE_OMIT 2
+#define LDML_META_SETTINGS_TRANSFORMPARTIAL_HIDE 4
+#define LDML_SECTION_KEYS ((uint32_t)'keys')
+#define LDML_SECTION_LOCA ((uint32_t)'loca')
+#define LDML_SECTION_META ((uint32_t)'meta')
+#define LDML_SECTION_SECT ((uint32_t)'sect')
+#define LDML_SECTION_STRS ((uint32_t)'strs')
+#define LDML_STRS_FLAGS_EXTEND 1
+#define LDML_VERSION 1.0
```
mcdurdin, line 18: I think we need to treat this as a string because we may end up with `"1.0.1"` in the future

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Constants shared between the data access layer and the compiler
+     */
+    export const constants = {
+        /**
+         * The version of the LDML processor
+         */
+        version: '1.0',
+
+        section_keys: 'keys',
```
mcdurdin, line 22: We should add a comment above these constants. Do we need any reference to kmxplus in this file?

core/tools/ldml-const-builder/build.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+#
+# Builds the include script for the current Keyman version.
```
mcdurdin, line 3:
```suggestion
# Builds /core/include/ldml/keyboardprocessor_ldml.h from /core/include/ldml/keyboardprocessor_ldml.ts
```

core/tools/ldml-const-builder/build.sh
```diff
@@ -0,0 +1,55 @@
...
+# TODO: build if out-of-date if test is specified
+# TODO: configure if npm has not been run, and build is specified
+
+if builder_has_action configure; then
+#   verify_npm_setup
+  builder_report success configure
+fi
+
+if builder_has_action clean; then
+#   npm run clean
```
mcdurdin, line 36:
```suggestion
```

core/tools/ldml-const-builder/build.sh
```diff
@@ -0,0 +1,55 @@
...
+builder_describe "Build and run the constant builder for LDML" configure clean build run
+builder_parse "$@"
+
+# TODO: build if out-of-date if test is specified
+# TODO: configure if npm has not been run, and build is specified
+
+if builder_has_action configure; then
+#   verify_npm_setup
+  builder_report success configure
+fi
```
mcdurdin, line 33:
```suggestion
```

core/tools/ldml-const-builder/build.sh
```diff
@@ -0,0 +1,55 @@
...
+. "$KEYMAN_ROOT/resources/shellHelperFunctions.sh"
+
+# This script runs from its own folder
+cd "$(dirname "$THIS_SCRIPT")"
+
+KBP_LDML_H_FILE="../../include/ldml/keyboardprocessor_ldml.h"
+
+################################ Main script ################################
+
+builder_describe "Build and run the constant builder for LDML" configure clean build run
```
mcdurdin, line 24:
```suggestion
builder_describe "Build and run the constant builder for LDML" clean build run
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -0,0 +1,50 @@
...
+//
+// If you update this file, you *must* be sure to re-run
+//
+//  core/tools/ldml-const-builder/build.sh clean build run
+//
+// To update keyboardprocessor_ldml.h, and commit the result.
+//
+// It is not updated automatically.
+
+// namespace com.keyman.core.ldml {
```
mcdurdin, line 12: We'll need to figure this out at some point, so perhaps:
```suggestion
// TODO-LDML: namespace com.keyman.core.ldml {
```

mcdurdin (approved):
LGTM. Some nits only. Process-wise, we don't normally require re-review after tidyup commits.

mcdurdin:
@ermshiperete fyi we decided to take this out of the automatic build for now. It's an infrequent task and not hard to do when we update constants. Avoids adding complications and dependencies to the build.

core/include/ldml/keyboardprocessor_ldml.ts
srl295: Adding a comment, and also better description up top

core/include/ldml/keyboardprocessor_ldml.h
srl295: good idea.

core/tools/ldml-const-builder/build.sh
srl295: great

PR #7060, opened 2022-08-11
chore(linux): Refactor `jenkins.sh`

ermshiperete:
- use standard Keyman include scripts
- remove package build support for legacy projects
- install dependencies from `debian/control` file instead of maintaining a separate list

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #7062, opened 2022-08-11
chore(core): refactor kmx_file.h to common 🚑

mcdurdin:
Part of the kmcompx refactor. We want the definitions in kmx_file.h rather than continuing to use legacy_kmx_file.h.

@keymanapp-test-bot skip

### DIFF
--- a/common/include/km_types.h
+++ b/common/include/km_types.h
@@ -1,0 +1,65 @@
+#pragma once
+
+/*
+#if defined(_WIN32) || defined(_WIN64)
+#define snprintf _snprintf
+#define vsnprintf _vsnprintf
+#define strcasecmp _stricmp
+#define strncasecmp _strnicmp
+#endif
+*/
+
+#if defined(__LP64__) || defined(_LP64)
+/* 64-bit, g++ */
+#define KMX_64BIT
+#endif
+
+#if defined(_WIN64) && !defined(USE_64)
+/* 64-bit, Windows */
+#define KMX_64BIT
+#endif
+
+typedef uint32_t            KMX_DWORD;
+typedef int32_t             KMX_BOOL;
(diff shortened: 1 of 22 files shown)


mcdurdin:
@ermshiperete can you help me get the Keyman packaging (Linux) build going?

ermshiperete:
Looks like we need to add `common/include` to the source package in `linux/scripts/dist.sh`, partially reverting dbca5712e71e7c48a9e40d87cd36ccec2591e23b.

common/include/km_types.h
```diff
@@ -0,0 +1,65 @@
...
+#define KMX_64BIT
+#endif
+
+typedef uint32_t            KMX_DWORD;
+typedef int32_t             KMX_BOOL;
+typedef uint8_t             KMX_BYTE;
+typedef uint16_t            KMX_WORD;
+
+#if defined(__cplusplus)
+typedef char16_t   km_kbp_cp;
```
srl295, line 28: hooray

srl295 (approved):
LGTM FWIW

common/include/km_types.h
mcdurdin: I'm not sure if I am happy or sad. `char16_t` is so unloved by the standard libraries!

srl295: use ICU4C instead!

Anyway, it's a happier type than wchar will ever be…

rc-swag (approved):
lgtm

common/include/km_types.h
mcdurdin: Let's have a discussion on using ICU4C in the Keyman Core. I think we stand to gain considerable benefit. Only concern is the additional weight for the WASM build.

keyman-server:

Issue #7046, opened 2022-08-05
feat(core): add 'flag' for process_event, touch


srl295:
#7042 and blocks #5015

- add a `flag` parameter to process_event()
- add a value of `touch` to distinguish from hardware events
- `flag` will be  ignored by current processors.

PR #7066, opened 2022-08-11
feat(core): add event_flags to process_event #7046 🙀 

srl295:
- Width: uint16_t
- defined `km_kbp_event_flags` to define the bitfield
- bitfield has value `KM_KBP_EVENT_FLAG_TOUCH` for touch events
- ignored everywhere, currently

Fixes: #7046 

@keymanapp-test-bot skip

### DIFF
--- a/core/include/keyman/keyboardprocessor.h
+++ b/core/include/keyman/keyboardprocessor.h
@@ -1052,6 +1052,7 @@ enum km_kbp_tech_value {
  * Bit flags to be used with the event_flags parameter of km_kbp_process_event
  */
 enum km_kbp_event_flags {
+  KM_KBP_EVENT_FLAG_DEFAULT = 0, // default value: hardware
   KM_KBP_EVENT_FLAG_TOUCH = 1, // set if the event is touch, otherwise hardware
 };
 
--- a/core/include/keyman/keyboardprocessor.h
+++ b/core/include/keyman/keyboardprocessor.h
@@ -1048,6 +1048,12 @@ enum km_kbp_tech_value {
   KM_KBP_TECH_LDML        = 1 << 2
 };
 
+/**
+ * Bit flags to be used with the event_flags parameter of km_kbp_process_event
+ */
+enum km_kbp_event_flags {
+  KM_KBP_EVENT_FLAG_TOUCH = 1, // set if the event is touch, otherwise hardware
+};
 
 /*
 ```
@@ -1088,6 +1094,7 @@ state is passed.
 - __modifier_state__:
 The combinations of modifier keys set at the time key `vk` was pressed, bitmask
 from the `km_kbp_modifier_state` enum.
+- __event_flags__: Event level flags, see km_kbp_event_flags
 
 ```c
 */
@@ -1096,7 +1103,8 @@ km_kbp_status
 km_kbp_process_event(km_kbp_state *state,
                      km_kbp_virtual_key vk,
                      uint16_t modifier_state,
-                     uint8_t is_key_down);
+                     uint8_t is_key_down,
+                     uint16_t event_flags);
 
 /*
 ```
--- a/core/python/keyman/keyboardprocessor/_api.py
+++ b/core/python/keyman/keyboardprocessor/_api.py
@@ -277,13 +277,20 @@ class Tech(IntFlag):
     KMN = 1
     LDML = 2
 
+class EventFlag(IntFlag):
+    KM_KBP_EVENT_FLAG_DEFAULT = 0
+    """default value: hardware"""
+    KM_KBP_EVENT_FLAG_TOUCH = 1
+    """set if the event is touch, otherwise hardware"""
+
 
 __method(None, 'get_engine_attrs', POINTER(Attributes))
 
 __method(None, 'process_event', Status,
          (State_p, Dir.IN, 'state'),
          (VirtualKey, Dir.IN, 'vkey'),
-         (c_uint16, Dir.IN, 'modifier_state'))
+         (c_uint16, Dir.IN, 'modifier_state'),
+         (c_uint16, Dir.IN, 'event_flags'))
 
 
 class Modifier(IntFlag):
--- a/core/python/keyman/keyboardprocessor/api.py
+++ b/core/python/keyman/keyboardprocessor/api.py
@@ -114,5 +114,6 @@ class State:
 def process_event(vk: VirtualKey,
                   modifier_state,
                   state: State,
-                  acts: ActionList):
+                  acts: ActionList,
+                  event_flags):
     pass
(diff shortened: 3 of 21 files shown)


mcdurdin:
This has failing builds for Windows:

```
[17:34:19][Step 1/3] "C:\BuildAgent\work\7ac43416c45637e9\keyman\windows\src\engine\keyman32\keyman32.vcxproj" (Rebuild target) (1) ->
[17:34:19][Step 1/3] (ClCompile target) -> 
[17:34:19][Step 1/3]   C:\BuildAgent\work\7ac43416c45637e9\keyman\windows\src\engine\keyman32\kmprocess.cpp(113,27): error C2660: 'km_kbp_process_event': function does not take 4 arguments [C:\BuildAgent\work\7ac43416c45637e9\keyman\windows\src\engine\keyman32\keyman32.vcxproj]
[17:34:19][Step 1/3] 
[17:34:19][Step 1/3]     0 Warning(s)
[17:34:19][Step 1/3]     1 Error(s)
```

and Developer: 

```
[17:44:30][Step 1/4] "C:\BuildAgent\work\7ac43416c45637e9\keyman\developer\src\tike\tike.dproj" (default target) (1) ->
[17:44:30][Step 1/4] (_PasCoreCompile target) -> 
[17:44:30][Step 1/4]   main\Keyman.System.KeymanCore.pas(312): error E2029: Identifier expected but ')' found [C:\BuildAgent\work\7ac43416c45637e9\keyman\developer\src\tike\tike.dproj]
[17:44:30][Step 1/4]   debug\Keyman.System.Debug.DebugCore.pas(13): error F2063: Could not compile used unit 'Keyman.System.KeymanCore.pas' [C:\BuildAgent\work\7ac43416c45637e9\keyman\developer\src\tike\tike.dproj]
[17:44:30][Step 1/4] 
```

srl295:
@mcdurdin looks like it was an out of date build that failed. I'll try rerunning…

srl295:
it actually probably could use some kind of test because it affects all of the process_event paths. but that could be done later, when further changes are made

srl295:
PR [56aeee3](https://github.com/keymanapp/keyman/pull/7066/commits/56aeee35c76de489bf8439b96cdfe6943a458897) was accidentally added to #7071  … it will come out in the wash.

srl295:
uf. 

https://build.palaso.org/buildConfiguration/KeymanDesktop_TestPullRequests/332386?buildTab=log&logView=flowAware&focusLine=2398&linesState=329

>     C:\BuildAgent\work\7ac43416c45637e9\keyman\windows\src\engine\keyman32\kmprocess.cpp(113,27): error C2660: 'km_kbp_process_event': function does not take 4 arguments [C:\BuildAgent\work\7ac43416c45637e9\keyman\windows\src\engine\keyman32\keyman32.vcxproj]

but yet looking at the build log:

https://github.com/keymanapp/keyman/commit/e8dff5cf4e015ce1e28e9b31ba0977112bc001b9#diff-8762b89391fb2aafa43506a684aedaae668bfff3551b906ac03ede86e7eb9141R115

https://github.com/keymanapp/keyman/blob/e8dff5cf4e015ce1e28e9b31ba0977112bc001b9/windows/src/engine/keyman32/kmprocess.cpp#L115

it's being called with 5 arguments not 4.

Is the build somehow out of date?

srl295:
> `), (uint8_t)_td->state.isDown), KM_KBP_EVENT_FLAG_DEFAULT)`

should be

> `> `), (uint8_t)_td->state.isDown, KM_KBP_EVENT_FLAG_DEFAULT))`

srl295:
Discussion with @rc-swag — will move the `is_key_down` parameter into a flag.   So `process_event` will go from 4 parameters to 4.

mcdurdin:
> Discussion with @rc-swag —
> 
> * [ ]  will move the `is_key_down` parameter into a flag.   So `process_event` will go from 4 parameters to 4.

I'm not terribly comfortable with this. The reason is, `process_event` is an API surface. It's internal, sure, but we still try to maintain API stability on internal APIs. If we change the _shape_ of the API, by adding a parameter, existing code will fail to compile*, but fixes will be straightforward -- in most cases we just need to `_kmn_unused(new_param)`. However, if we change the _meaning_ of an API parameter without changing the shape of the call, we have to (a) manually audit all usages, and (b) update every reference to work with the new meaning. Change-wise, this introduces greater risk, and also forces us to update all components at the same time.

* Caveat: as long as we update API declarations for all languages -- e.g. we may have API decls in Delphi and Python as well as C++!

For a public API, we would never change the shape or meaning of an API call -- we'd introduce an new API surface and mark the old one as deprecated. For an internal API, I am willing to have shape changes.

Now, I know that in semantic versioning, the idea is that major releases can change APIs. I personally hate this: a given API function, once published, should remain stable both in _shape_ and _meaning_ for as long as that is viable -- only changing, for example, if security flaw repairs require it. This greatly simplifies the task of updating dependencies for downstream devs, and in the end we all win.

rc-swag:
I agree with what you say in principle and maybe the ship has already sailed. It's just in the space of a couple of months we have just added two arguments which are just flags. Each time the API call is changing and breaking by doing that. With this change, it is more resilient to future updates - at least for simple flags. I suggested this because the `is_keydown` was recent change.
The ideal is to have an ordered struct with an encoder and decoder so that you can handle updates without breaking any existing consumers of the API call but that ship has also sailed, it also doesn't really match the pattern of the core.  To match the pattern of the core it would be array-like keyboard options and context.  Either way, it would be able to grow the information required in a `process_event` API call without having to break the API call just for the addition of new event data.

Happy to not have this change while better it is still not the ideal solution think we should look to better handling of event data across API boundaries in the future.

mcdurdin:
Yep, good points. We do have more flexibility on the API given this is internal. Currently, we have the following API consumers that I am aware of:

1. Tests (if they can be considered consumers)
2. Keyman for Linux
3. Keyman for Windows
4. Keyman Developer - debugger
5. Keyman Developer - OSK import from .kmx

And I'd forgotten about #5, FWIW. 

In terms of good API design, you are right too. I was concerned more with stability of existing APIs but it's a shame we didn't consider making key_down a bit flag from the beginning.

Happy to be flexible as long as we can be sure we verify the changes properly (user test may be required?)

mcdurdin (approved):
LGTM. I think the proposed flag change to process_event could be a follow-up PR at this point.

srl295:
I think this pr already hits all call sites. I'll merge and we can consider followon, it's still the feature branch.

srl295:
> However, if we change the meaning of an API parameter without changing the shape of the call, we have to (a) manually audit all usages, and (b) update every reference to work with the new meaning. Change-wise, this introduces greater risk, and also forces us to update all components at the same time.

to be clear I would keep the parameter count but change the type.  
Actually, 99% of the call sites (tests) ended in `…,1);` meaning "key-down event" . If `1` was defined as "the flag for a keydown event" then it would be a compatible call. If we're hitting all sites 

But i do hear you about silent failures, if it's too compatible then we won't catch the issues.

> as long as we update API declarations for all languages

this PR updates pascal and python already.

In any event I'm unblocked from further work, so I think I will proceed with the meat of the implementation.

Issue #5015, opened 2021-05-03
spec(core): LDML KeyboardProcessor 🙀 


mcdurdin:
## Introduction 

This is a holding issue for the Keyman Core required to support LDML keyboards. We'll be filling this in as we complete planning and design. Much of this content should be moved to separate issues as we expand the details and establish component boundaries.

**Objective:** build a keystroke processor that works with the existing Keyman Core design that supports LDML keyboards, including:

* load/save of LDML data
* keyboard metadata API (particularly to support OSK) (this may not be needed?)
* stateless keystroke transform

Intent is to build this in C++. It will need to cross compile to native Windows, macOS, Linux and WASM. The module will need to be standalone and not have runtime dependencies (static link of libraries is probably okay).

## Related Features

LDML implementation:
* #7042
* #7043 

Groundwork:
* #5011
* #5012
* #5013
* ~~#5069~~
* [Draft spec document](https://docs.google.com/document/d/1L-C5FFdsvbtfIV65rXomUR3Hg_Y8NCjFqjcmiY2xZzU/edit#) 

## General Library Properties

* no i/o, minimal/no dependencies. Why?
  * So that the WASM compiler/linker does not have to chase down file IO and deps
  * So that the lib is provably secure and portable
  * So that the lib can be tested in complete isolation from any deps or
    environmental concerns (outside data etc)
* no_std? - to consider whether this is appropriate or not.

* API boundary will be UTF-16 (std::u16string).
* Internal string form will also be std::u16string.
* Physical limits of input context and output transforms. 64?, 256 chars?
* Keyboard data storage: a binary (black-box, not necessarily optimized) format
  that the KMXPlus compiler will produce given input XML.

## C5015.1: Infrastructure

~~Moved to #5069.~~

* Define LDMLKeyboardProcessor folder, build scripts, basic files
* Add template unit tests

## C5015.2: KMX+ Binary Loader

* File Format: #7043
* Loads from the BLOB
* no i/o requirements so kbd processor can depend on this
* Provides metadata access (see §2.1, "Library for LDML access")
* used by:
  * LDML Keyboard Processor Library
  * unit test
  * clients (i.e. "list of keyboards" or "filtering keyboards", "get osk data")
* API:
  * ...
  * Metadata
  * Keyboard definition, transforms
  * OSK layout

## C5015.3: LDML Keyboard Processor Library

* no i/o
* no state besides context
* Depends on:
  * LDML Datablob Library
* API:
  * `constructor` function:
[...]

PR #7071, opened 2022-08-13
feat(core): add sample file, update union #5015 🙀

srl295:
- add tiny sample kmx
- update union
- add DTD

#5015 

@keymanapp-test-bot skip


core/src/kmx/kmx_file.h
```diff
@@ -340,11 +340,14 @@ struct COMP_KEYBOARD {
   KMX_DWORD dwBitmapSize;   // 003C size in bytes of the bitmaps
 };
 
-// TODO-LDML: fill this in
+/**
+ * Only valid if comp_keyboard.dwFlags&0x20
```
mcdurdin, line 344:
```suggestion
 * Only valid if comp_keyboard.dwFlags&KF_KMXPLUS
```

Feel like it's better docs to use the const name than the value.

core/tests/unit/ldml/001_tiny.xml
```diff
@@ -0,0 +1,20 @@
...
+<keyboard locale="mt" conformsTo="techpreview">
+  <info author="srl295" indicator="🙀" layout="qwerty" normalization="NFC" />
+
+  <names>
+    <name value="TestKbd" />
+  </names>
+
+  <keys>
+    <key id="hmaqtua" to="ħ" />
+    <key id="say" to="ថា" />
```
mcdurdin, line 12: Technically, `that` rather than `say` but very good 😆

core/tests/unit/ldml/ldmlKeyboard.dtd
```diff
@@ -0,0 +1,192 @@
+<!--
```
mcdurdin, line 1: I wonder if we can put this in resources/standards-data/ldml-keyboard along with a README.md detailing where it is sourced from and update strategy?

See for example https://github.com/keymanapp/keyman/blob/d4b8094b137ef2a1aac1eb60a6d0a01658bb525f/resources/standards-data/langtags/readme.md and its corresponding data file.

I realise that this may be problematic/annoying for DOCTYPE decls so feel free to push back on this.

mcdurdin (approved):

core/tests/unit/ldml/ldmlKeyboard.dtd
srl295: I think that's reasonable. Will need a long term strategy for this anyway. Should be versioned.

srl295: https://github.com/keymanapp/keyman/blob/master/resources/standards-data/unicode-character-database/readme.md this one is not so descriptive :)

srl295:
00041d6d0 belongs in PR #7066 instead

srl295:
> 👉  Note: rebased to pickup upstream PR changes

core/tests/unit/ldml/ldmlKeyboard.dtd
mcdurdin: Personally, I love the simplicity of that readme.

core/src/kmx/kmx_file.h
```diff
@@ -340,11 +340,18 @@ struct COMP_KEYBOARD {
...
-  };
+  KMX_DWORD dwKMXPlusSize;  // 0044 size in bytes of entire KMXPlus data
+};
+
+/**
+ * Only valid if comp_keyboard.dwFlags&KF_KMXPLUS
+ */
+struct COMP_KEYBOARD_EX {
+  COMP_KEYBOARD comp_keyboard;              // 0000 see COMP_KEYBOARD
+  COMP_KEYBOARD_KMXPLUSINFO comp_keyboard_kmxplusinfo;  // 0040 see COMP_KEYBOARD_EXTRA
```
mcdurdin, line 353:
```suggestion
  COMP_KEYBOARD             header;    // 0000 see COMP_KEYBOARD
  COMP_KEYBOARD_KMXPLUSINFO kmxplus;   // 0040 see COMP_KEYBOARD_EXTRA
```
I wonder if we can just keep them a bit shorter?

srl295:
just a rebase onto the feature branch

srl295:
What is the modelling layer error?

>  LMLayer using the trie model Regression tests.Chrome 100.0.4896.127 (Windows 10):

mcdurdin:
> What is the modelling layer error?

Lexical Models. You can ignore it. Being addressed by #7037.

PR #7083, opened 2022-08-17
feat(core): build out ldml structures / initial scaffolding implementation 🙀

srl295:
- checkin spec to repo
- new header: kmx_plus.h with data access structs and functions
- static asserts to validate header sizes
- early validation of data
- internal dump of kmx+ data

@keymanapp-test-bot skip


core/src/ldml/C7043_ldml.md
```diff
@@ -0,0 +1,165 @@
+# KMX Plus Binary Format #7043
```
srl295, line 1: i'm fine to move this file.  Just very convenient to do all my work in the repo.

core/tests/unit/ldml/test_kmx_plus.cpp
```diff
@@ -0,0 +1,10 @@
+#include <stdio.h>
+#include <assert.h>
+#include <kmx/kmx_plus.h>
+
+using namespace km::kbp::kmx;
+
+int main(int argc, const char *argv[]) {
+    printf("OKAY\n");
```
srl295, line 8: This is a stub for now, because I'm doing only static_asserts for now, but may add some more detailed unit tests of kmx_plus functions here.

srl295:
```
dump_kmxplus_data(): Got a PCOMP_KEYBOARD at 0x7fb724705f10
KMXPlus offset 0x100, KMXPlus size 0x200
sect: (73656374) size 0x20
sect: total 0x100
sect: count 0x4
strs sect#0: 73747273 @ A0
strs: (73747273) size 0x170
strs: count 0x8
#0x0: mt
#0x1: TestKbd
#0x2: srl295
#0x3: techpreview
#0x4: qwerty
#0x5: NFC
#0x6: U+D83D U+DE40 
#0x7: U+1790 U+17B6 
meta sect#1: 6D657461 @ 40
meta: (6D657461) size 0x24
 name:  #0x1
 author:        #0x2
 conform:       #0x3
 layout:        #0x4
 normalization: #0x5
 indicator:     #0x6
 settings:      0x0
loca sect#2: 6C6F6361 @ 30
loca: (6C6F6361) size 0x10010
 Locale #0: #0x0
keys sect#3: 6B657973 @ 70
keys: (6B657973) size 0x30
 count: #0x2
 #0x0
  vkey  0x51
  mod   0x0
  flags 0x0
         UTF-32:U+0127
 #0x1
  vkey  0x57
  mod   0x0
  flags 0x1
         Extend: String #0x7
```

dilemma
(I fixed the sample data)

srl295 (commented):
tests pass!

srl295:
> A16S9

OK, i'll keep updating this PR then.

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+    putchar((ident & 0xFF000000) >> 24);
+    ident <<= 8;
+  }
+}
+
+static void
+dump_kmxplus_header(const COMP_KMXPLUS_HEADER* hdr) {
+  if (hdr == NULL) {
+    printf("! dump_kmxplus_header: NULL header\n");
+  }
```
mcdurdin, line 31:
```suggestion
  if (hdr == NULL) {
    printf("! dump_kmxplus_header: NULL header\n");
    return;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+    printf("! could not load 'vkey' section\n");
+    return;
+  }
+  dump_kmxplus_header((const COMP_KMXPLUS_HEADER*)vkey);
+  // TODO-LDML
+}
+
+static void
+dump_kmxplus_strs(const uint8_t* /*data*/, const COMP_KMXPLUS_STRS* strs) {
+  dump_kmxplus_header((const COMP_KMXPLUS_HEADER*)strs);
```
mcdurdin, line 99:
```suggestion
  if(strs == NULL) {
    printf("! could not load 'strs' section\n");
    return;
  }
  dump_kmxplus_header((const COMP_KMXPLUS_HEADER*)strs);
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+  for (KMX_DWORD i=0; i<strs->count; i++) {
+    KMX_WCHAR buf[BUFSIZ];
+    printf("#0x%X: ", i);
+    PKMX_WCHAR str = strs->get(i, buf, BUFSIZ);
+    if (!str) {
+        printf("NULL/ERR\n");
+        continue;
+    }
+    for(int j=0; str[j] && j<0x30; j++) {
+        if (str[j] < 0x7F && str[j] != 0x0020 && str[j] > 0x20) {
```
mcdurdin, line 110:
```suggestion
        if (str[j] < 0x7F && str[j] > 0x20) {
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+            putchar(str[j]);
+        } else {
+            printf("U+%04X ", str[j]);
```
mcdurdin, line 113:
```suggestion
            printf(" U+%04X ", str[j]);
```

We'll end up with space being written as `U+0020`, is that your intention? Either way if we mix ascii with other characters, we will either have chars bumping up against U+xxxx (`abcU+0020 U+00C1 `) or excess spaces (`abc U+0020  U+00C1 `) with my fix ... which do you prefer?

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+    printf("\n");
+  }
+}
+
+static void
+dump_kmxplus_sect(const uint8_t* data, const COMP_KMXPLUS_SECT* sect) {
+  dump_kmxplus_header((const COMP_KMXPLUS_HEADER*)sect);
```
mcdurdin, line 122:
```suggestion
  if(sect == NULL) {
    printf("! could not load 'sect' section\n");
    return;
  }
  dump_kmxplus_header((const COMP_KMXPLUS_HEADER*)sect);
```

(note: we should probably be using `nullptr` instead of `NULL` throughout)

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+            dump_kmxplus_keys(data, as_kmxplus_keys(entrydata));
+            break;
+        case LDML_SECTION_LOCA:
+            dump_kmxplus_loca(data, as_kmxplus_loca(entrydata));
+            break;
+        case LDML_SECTION_META:
+            dump_kmxplus_meta(data, as_kmxplus_meta(entrydata));
+            break;
+        case LDML_SECTION_SECT:
+            printf("! Cowardly refusing to dump nested 'sect' section.\n");
```
mcdurdin, line 142: This should be an error, shouldn't it? We don't allow nested `sect` sections?

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+}
+
+KMX_DWORD COMP_KMXPLUS_SECT::find(KMX_DWORD ident) const {
+  for (KMX_DWORD i = 0; i < count; i++) {
+    if (ident == entries[i].sect) {
+        return entries[i].offset;
+    }
+  }
+  return 0;
+}
```
mcdurdin, line 202: Alternatively, on load we could fill `std::maps` with these and leverage existing C++ classes?

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+PKMX_WCHAR
+COMP_KMXPLUS_STRS::get(KMX_DWORD entry, PKMX_WCHAR buf, KMX_DWORD bufsiz) const {
+    assert(entry < count);
+    if (entry >= count) {
+        return NULL;
+    }
+    KMX_DWORD offset = entries[entry].offset;
+    KMX_DWORD length = entries[entry].length;
+    assert(bufsiz > (length+1)); // assert bufsiz big enough
```
mcdurdin, line 212: Given `bufsiz` is an input variable, perhaps we need a way of telling the caller what buffer size we really need rather than asserting on an unexpectedly large string?

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
...
+    assert(offset+((length+1)*2) <= header.size); // assert not out of bounds
+    const uint8_t* thisptr = reinterpret_cast<const uint8_t*>(this);
+    const KMX_WCHAR* start = reinterpret_cast<const KMX_WCHAR*>(thisptr+offset);
+    for(KMX_DWORD i=0;i<=length;i++) {
+        buf[i] = start[i];
+    }
```
mcdurdin, line 218: Or we could do a `memcpy` here for bonus perf points? Admittedly, a good compiler may figure that out for itself...

core/src/kmx/kmx_plus.cpp
```diff
@@ -0,0 +1,224 @@
+#include <kmx/kmx_plus.h>
```
mcdurdin, line 1: Let's add the copyright header

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -10,57 +10,90 @@
 // It is not updated automatically.
```
mcdurdin, line 10: BTW, we should add a copyright header at the top of this file.

core/src/ldml/C7043_ldml.md
mcdurdin: Fine by me -- can you update the issue to point to the appropriate location where the most up-to-date version of the spec can be found?

core/src/ldml/ldml_processor.cpp
```diff
@@ -9,8 +9,41 @@
...
+    if(**sourceStart & 0xFFFF0000) {
+      // Don't support supplemental chars yet
+      return conversionNotOk;
+    }
+    *((*targetStart)++) = (UTF16)((*(*sourceStart)++) & 0xFFFF); // BMP
+
+    assert(*sourceStart==sourceEnd);
+    assert(*targetStart <= targetEnd);
+    return(conversionOK);
+}
```
mcdurdin, line 44: Can we clarify that this is a temporary patch?

core/tests/unit/ldml/meson.build
```diff
@@ -33,7 +33,7 @@ ldml = executable('ldml',
     objects: lib.extract_all_objects())
 
 tests = [
-  '000_null_keyboard',
+  # '000_null_keyboard',
```
mcdurdin, line 36: We should still have a null_keyboard test that just loads and exits without any keystrokes, ideally?

mcdurdin (approved):
Some nits. LGTM.

core/src/kmx/kmx_plus.cpp
srl295: Right. This might become a validator but it's currently just a debug statement. 

Is there a good place for a "dump" facility? Separate tool?

core/src/kmx/kmx_plus.cpp
srl295: Yes. That's actually the next PR.

core/tests/unit/ldml/meson.build
srl295: The validator will eventually reject it as not containing any sections. I could update it to be valid but null.

core/src/kmx/kmx_plus.cpp
mcdurdin: Yes, we should have a `core/src/tools` folder (per our [repo filesystem layout doc](https://docs.google.com/document/d/1gWlgDA_re33i_ny4FiBrjuZLJLBJqlIAbspRxI55FQk/edit)). A `kmxplusdump` tool would belong there pretty neatly. Or else in `developer/src/tools` but I think under `core` makes sense given the C++ codebase.

Note that `core/tests` should probably be moving to `core/src/test` according to that doc (cleanup for another day?).

core/tests/unit/ldml/meson.build
mcdurdin: Yep, updating it to be a valid null keyboard would be good.

core/include/ldml/keyboardprocessor_ldml.ts
srl295: this would be a great task for an automated scanner.

srl295 (commented):
thanks. Updated w/ copyright notices

PR #7084, opened 2022-08-18
fix(linux): Improve setting context

ermshiperete:
This change gets the context from the surrounding text if the context is shorter. This helps when the user puts the IP after a certain
character. Previously we basically lost the context, with this change we restore the context so that it's like the user just typed
the previous characters.


ermshiperete:
@mcdurdin This seems like a good idea because it allows you to pick up where you left of before you switched to another app, but I'm not entirely sure it doesn't have side effects that I'm currently not seeing.

mcdurdin:
> @mcdurdin This seems like a good idea because it allows you to pick up where you left of before you switched to another app, but I'm not entirely sure it doesn't have side effects that I'm currently not seeing.

From my understanding, this is looking at picking up context when insertion point changes, where the app has surrounding-text support. We should certainly be doing this. However, I am not sure that this is quite right yet. A few thoughts, if the insertion point moves (apart from when we are inserting text), then:

1. We should always abandon any cached context, whether or not the app supports surrounding-text.
2. We don't need to compare the surrounding-text length to the cached context. Just throw away the cached context in this situation.
3. If surrounding-text is available, definitely read it in.
4. We should ensure we clear any deadkey state as well.

This insertion point movement applies to focus changes within the current app, mouse clicks, text selection, executing commands, hotkeys, etc. If the user switches apps, we should reset state in the same way: it is not generally helpful to track deadkey state as the user won't remember which deadkey they have pressed in the app anyway.

ermshiperete:
Thanks for your thoughts, @mcdurdin! The context now gets always reset in `reset_context` which should take care of the points you mentioned, and also simplifies the code :smile:

linux/ibus-keyman/src/engine.c
```diff
@@ -241,40 +241,33 @@ client_supports_surrounding_text(IBusEngine *engine) {
...
-        }
-        g_free(surrounding_text);
-        g_free(current_context_utf8);
-    }
-    else {
-        km_kbp_context_clear(context);
+    if (km_kbp_context_items_from_utf8(surrounding_text, &context_items) == KM_KBP_STATUS_OK) {
+      km_kbp_context_set(context, context_items);
     }
+    km_kbp_context_items_dispose(context_items);
```
mcdurdin, line 268:
```suggestion
    if (km_kbp_context_items_from_utf8(surrounding_text, &context_items) == KM_KBP_STATUS_OK) {
      km_kbp_context_set(context, context_items);
      km_kbp_context_items_dispose(context_items);
    }
```

linux/ibus-keyman/src/engine.c
```diff
@@ -241,40 +241,33 @@ client_supports_surrounding_text(IBusEngine *engine) {
...
-        current_context_utf8 = get_current_context_text(context);
+    ibus_engine_get_surrounding_text(engine, &text, &cursor_pos, &anchor_pos);
 
-        ibus_engine_get_surrounding_text(engine, &text, &cursor_pos, &anchor_pos);
-        context_pos = anchor_pos < cursor_pos ? anchor_pos : cursor_pos;
-        context_start = context_pos > MAXCONTEXT_ITEMS ? context_pos - MAXCONTEXT_ITEMS : 0;
-        surrounding_text = g_utf8_substring(ibus_text_get_text(text), context_start, context_pos);
-        g_message("new context is:%u:%s: cursor:%d anchor:%d", context_pos - context_start, surrounding_text, cursor_pos, anchor_pos);
+    context_start    = context_end > MAXCONTEXT_ITEMS ? context_end - MAXCONTEXT_ITEMS : 0;
+    context_end      = anchor_pos < cursor_pos ? anchor_pos : cursor_pos;
```
mcdurdin, line 260: These lines are out of order?

mcdurdin (changes requested):

ermshiperete:
This change still has a problem when the keyboard outputs a character + deadkey. In response to outputting the character we get a `ibus_keyman_engine_set_surrounding_text` call in which we call `reset_context` which removes the deadkey marker. Even if we wouldn't call `reset_context` and directly call `km_kbp_context_set` we would lose the deadkey. Or should we make `ibus_keyman_engine_set_surrounding_text` a no-op?

Not sure how to deal with this. Any suggestions @mcdurdin or @rc-swag ?

mcdurdin:
On Windows, we compare the new text context from the app to the textual part of the cached context. If it is different, then we reset the cached context and deadkeys. (In theory, we could have the same text in two places, but it's not particularly harmful in that rare situation anyway for the deadkey to 'transport' across text contexts.)

bharanidharanj:
### GROUP_BIONIC: Ubuntu 18.04 Bionic with Gnome Shell and X11

* **TEST_IPA_SAMEAPP (PASSED):** Installed Keyman 16.0.83-alpha build in Bionic Linux OS (VM) . Tested this as per the instructions and verified that after pressing > key, the letter n was changed into ŋ.

IPASample


* **TEST_IPA_TWOAPPS (PASSED):** Tested this as per the instructions in gedit and gnome terminal and after the switching I was able to change the n character into ŋ by pressing the > key.

IPATwoapps


* **TEST_DEADKEY_SAMEAPP (PASSED):** Tested this as per the instructions and verified the deadkeys are not working after position changes.

DeadKeysameapp



* **TEST_DEADKEY_TWOAPPS (FAILED):** Tested this as per the instructions and noticed thta the letter á did not change in the gnome terminal. seems to be an issue. 

DeadKeyTwoApps


* **TEST_GEDIT_k019 (PASSED):** Tested this as per the instructions and verified the expected output on the text editor.

Geditk019

* **TEST_TERM_k019 (PASSED):** Tested this as per the instructions and verified the expected output appears on the gnome-terminal.

Termk019

ermshiperete:
GROUP_BIONIC 
TEST_DEADKEY_TWOAPPS(PASSED): the result is correct, my instructions were wrong

GROUP_JAMMY_X11
TEST_DEADKEY_TWOAPPS(PASSED): the result is correct, my instructions were wrong

ermshiperete:
@bharanidharanj Please make sure you use the new ibus version (check version with `dpkg -l ibus`; should be version `*sil2.1*`) and reboot after installation of ibus. Then I think the remaining failing test will work.

@keymanapp-test-bot retest GROUP_JAMMY_X11 TEST_TERM_k019

linux/ibus-keyman/src/engine.c
```diff
@@ -241,40 +241,39 @@ client_supports_surrounding_text(IBusEngine *engine) {
...
+    ibus_engine_get_surrounding_text(engine, &text, &cursor_pos, &anchor_pos);
+
+    context_end      = anchor_pos < cursor_pos ? anchor_pos : cursor_pos;
+    context_start    = context_end > MAXCONTEXT_ITEMS ? context_end - MAXCONTEXT_ITEMS : 0;
+    surrounding_text = g_utf8_substring(ibus_text_get_text(text), context_start, context_end);
+    g_message("%s: new context is :%s: (len:%u) cursor:%d anchor:%d", __FUNCTION__,
+      surrounding_text, context_end - context_start, cursor_pos, anchor_pos);
+
+    current_context_utf8 = get_current_context_text(context);
+    if (!g_str_has_suffix(surrounding_text, current_context_utf8) || !g_utf8_strlen(current_context_utf8, -1)) {
```
mcdurdin, line 265:
```suggestion
    if (!g_str_has_suffix(surrounding_text, current_context_utf8) || g_utf8_strlen(current_context_utf8, -1) == 0) {
```

I know it has the same effect, but `g_utf8_strlen` is returning a length value, wheres `g_str_has_suffix` is returning a boolean, so the semantic meaning is different.

Note: `g_utf8_strlen(x,-1)` is costly for non-zero length strings. Is this better?

```suggestion
    if (*current_context_utf8 == 0 || !g_str_has_suffix(surrounding_text, current_context_utf8)) {
```

linux/ibus-keyman/src/engine.c
```diff
@@ -241,40 +241,39 @@ client_supports_surrounding_text(IBusEngine *engine) {
...
+      g_message("%s: setting context because it has changed from expected", __FUNCTION__);
+      if (km_kbp_context_items_from_utf8(surrounding_text, &context_items) == KM_KBP_STATUS_OK) {
+        km_kbp_context_set(context, context_items);
+        km_kbp_context_items_dispose(context_items);
+      }
```
mcdurdin, line 270: If this fails, I think we should `km_kbp_context_clear(context)` and log.
```suggestion
      if ((status = km_kbp_context_items_from_utf8(surrounding_text, &context_items)) == KM_KBP_STATUS_OK) {
        km_kbp_context_set(context, context_items);
        km_kbp_context_items_dispose(context_items);
      } else {
        km_kbp_context_clear(context);
        g_message("%s: setting context failed with %d", __FUNCTION__, status);
      }
```

mcdurdin (approved):
LGTM

bharanidharanj:
> @bharanidharanj Please make sure you use the new ibus version (check version with `dpkg -l ibus`; should be version `*sil2.1*`) and reboot after installation of ibus. Then I think the remaining failing test will work.
> 
> @keymanapp-test-bot retest GROUP_JAMMY_X11 TEST_TERM_k019

@ermshiperete Okay, Thanks.

bharanidharanj:
### GROUP_JAMMY_X11: Ubuntu 22.04 Jammy with Gnome Shell and X11

* **TEST_TERM_k019 (PASSED):** Retested this as per Eberhard's suggestion (ie., running the dpkg -l ibus command on the terminal) and I was able to see the expected output.

linux/ibus-keyman/src/engine.c
mcdurdin: See #7540.

linux/ibus-keyman/src/engine.c
mcdurdin: See #7541.

keyman-server:

Issue #7098, opened 2022-08-22
bug(core): ldml—Cхх compiler issues


srl295:
- use `nullptr` instead of `NULL`
- debug output comments on https://github.com/keymanapp/keyman/pull/7083#pullrequestreview-1079764223
- multichar constants: https://zipcon.net/~swhite/docs/computers/languages/c_multi-char_const.html - redefine as a string.
- zero sized array is not appreciated by windows

srl295:
zero sized array:  use `item entries[]` instead of `item entries[0]` (c99)

mcdurdin:
Ah, not actually LDML compiler ... but C++ compiler issues for Core 😆  I was devastated for a moment because I hadn't even pushed my first commit on the compiler!

srl295:
done with #7116 and #7100

PR #7054, opened 2022-08-08
epic: LDML keyboard support 🙀

mcdurdin:
* Implements #7042

issue label: [epic-ldml](https://github.com/keymanapp/keyman/labels/epic-ldml)

_Process: We'll use this PR comment as a central location for all issues and PRs relating to the epic-ldml project, so this comment our landing page for the project. All epic-ldml PRs will eventually merge into this branch before landing in `master`. Edit this comment rather than adding additional content_

# To-do Items

_Process: add to-do items here, convert to issues as appropriate._

- [ ] move IO out of Keyman Core into Keyman Engine
  - [ ] Linux
  - [ ] Windows
  - [ ] Developer
  - [ ] Tests

- [ ] #7310 

_Process: add all related issues and PRs here, categorised for ease of tracking_

# Specifications

### Overview / Common
* #7043

### Core
* #5015

### Developer
* #7047
* #5016 
* #5077 
* #5078

# Issues

### Overview / Common
* #7044
* #7045 

### Core
* #7046

### Developer
* #7532 

# Pull Requests

### Common
* #7049
* #7050

### Core
* #7048

### Developer

@keymanapp-test-bot skip
- we are not ready for user testing on this work yet.

PR #7100, opened 2022-08-23
fix(core): ldml: remove 0-length arrays 🙀

srl295:
- use C99 flexible array member instead

#7098

@keymanapp-test-bot skip


core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -19,7 +19,10 @@
 // TODO-LDML: namespace com.keyman.core.ldml {
 /**
  * Constants for the KMXPlus data format
- * These are shared between the data access layer and the compiler
+ * These are shared between the data access layer and the compiler.
+ * Note that the section IDs (section_keys etc.) are 32 bit hex
+ * values that are designed to appear as text when written in little endian
+ * format, so 0x7379656b = 'keys'
```
mcdurdin, line 25: Would it be possible to do this translation dynamically in our typescript code so we can have nice strings here in the source .ts file?

And then the .h file could have e.g.

```
#define LDML_SECTIONID_KEYS 0x7379656B
#define LDML_SECTIONNAME_KEYS "keys"
```

mcdurdin (changes requested):
Everything LGTM except I am a little uncomfortable with the change to numeric identifiers in the headers -- I feel like that's hard to read and maintain. I'd like to see us dynamically generating those in the TS code for use in C++ (so, having a helper for them in the TS header).

core/include/ldml/keyboardprocessor_ldml.ts
srl295: that makes sense as an approach.  I'll add a helper function in TS and use it in the generator.

srl295:
- [ ] MSVC choking on the printf

core/include/ldml/keyboardprocessor_ldml.ts
srl295: done

srl295:
@mcdurdin 
> Cannot open include file: '../kmx/kmx_file.h': No such file or directory

merge conflict?

mcdurdin:
> Cannot open include file: '../kmx/kmx_file.h': No such file or directory
> 
> merge conflict?

Which build is this from?

srl295:
@mcdurdin  see if 16b8d725fd0527ac21c67c0f70d1837c7e8b0b7b fixes it

srl295:
@mcdurdin ptal and go ahead and merge if this works

mcdurdin:
Jenkins build fails with:

image

@ermshiperete any ideas?

mcdurdin:
I'm going to go ahead and merge this and we'll solve the Jenkins build in a separate PR, because as it is this hiccup is block builds on all the other PRs in feature-ldml branch. (Isn't it always the way that we spend so much of our energy on silly build issues?)

mcdurdin (approved):

mcdurdin:
This same compile error is happening everywhere on #7054, so we'll need to dig deeper.

PR #7101, opened 2022-08-23
feat(developer): LDML keyboard compiler - infrastructure 🙀

mcdurdin:
Lays down infrastructure for the LDML keyboard compiler, including integration into Keyman Developer release build.

Compiler is written in node. Wrapper cmd script included with Keyman Developer, `kmldmlc.cmd`. On other platforms, you'll eventually be able to use `npm i -g @keymanapp/ldml-keyboard-compiler` and then `kmldmlc` in similar fashion.

Note that changes to build-utils.sh are also cherry-picked to a separate PR, #7103 to bring them out of the feature-ldml timeline. (I will keep these two PRs in sync.)

@keymanapp-test-bot skip We will do user testing once there is actually something to test!

PR #7103, opened 2022-08-23
feat(common): add parameter variable support to `builder_` functions

mcdurdin:
This PR is cherry-picked from #7101, to bring new build-utils.sh functionality into alpha outside of the feature-ldml branch timeline.

The changes here allows you to define variables for parameters, where you need to pass additional data on the command line.

First, use `builder_describe` to add an option, with a variable name, for example (cut down example here from the original PR):

```
builder_describe "Build Keyman LDML Keyboard Compiler kmldmlc" \
  "bundle                   Creates a bundled version of kmldmlc" \
  "--build-path=BUILD_PATH  Build directory for bundle"
```

Adding `=VAR` to the option definition means that the builder_ utils will expect a second parameter after the option, for example:

```
./build.sh bundle --build-path /tmp/foo
```

This new option will be made available in the environment variable `$VAR`, so in the above example, it would be `$BUILD_PATH`. The environment variable will only be defined if the option is passed on the command line (so use normal bash semantics to avoid errors, or test the option with `builder_has_option`).

You can also mix with short form of options, for example:

```
builder_describe "Build Keyman LDML Keyboard Compiler kmldmlc" \
  "bundle                      Creates a bundled version of kmldmlc" \
  "--build-path,-b=BUILD_PATH  Build directory for bundle"
```

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/build-utils.sh
+++ b/resources/build/build-utils.sh
@@ -27,6 +27,11 @@
 # Note: keep changes to version, tier and tag determination in sync with mkver (windows/src/buildutils/mkver)
 #
 
+#
+# Prevents 'clear' on exit of mingw64 bash shell
+#
+SHLVL=0
+
 # Setup variable for calling script's path and name
 if [ ! -z ${THIS_SCRIPT+x} ]; then
   THIS_SCRIPT_PATH="$(dirname "$THIS_SCRIPT")"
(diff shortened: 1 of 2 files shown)


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.51-alpha](https://keyman.com/downloads/releases/alpha/16.0.51)

PR #7077, opened 2022-08-17
fix: remove saving and restoring context kbd options

rc-swag:
Fixes: #7074

This PR removes the saving of context and keyboard options
on the non-updateable call to process hook when using the core processor.

The windows Text Input Processor calls twice for each keystroke the first call is not updatable. With the core integration we run
the processor only once - the actions and the context are available to be applied on the updatable call.
Saving and restoring the options meant that the stores values were overwritten and seeing as the process event call was not called again the stores changed value was lost.

PR #7107, opened 2022-08-24
fix(windows): remove saving and restoring context kbd options 🍒

rc-swag:
Cherry Pick from #7077
Remove the saving of context and keyboard options
on the non updateable call to process hook.

The windows Text Input Processor calls twice one call
is not updateable. With the core integeration we run
the processor only once, preserving the actions and
context to be applied on the updatable call.
@keymanapp-test-bot skip


ermshiperete (approved):

keyman-server:

PR #7089, opened 2022-08-22
fix(windows): Add invalidate context action to non-updatable parse

rc-swag:
Fixes #7078 

Rename ProcessActionsTestParse to NonUpdatable parse as it does more than just test. Add handling for the invalidate context action to this parse. In the case when an emit keystroke is in the action list it will be emitted and the Updateable round will never occur so we need to invalidate the context now in the non-updatable parse.

# User Testing

Install the keyman for windows build associated with this PR
The key referred to as "enter" also has key cap labelled "return" on many keyboards.
**TEST_ENTER_INVALIDATE**
1. Install gff_ethiopic_7 from Keyman.com [Ethiopic - (Language Neutral)](https://keyman.com/keyboards/gff_ethiopic_7) 
2. Open Notepad
3. Select  Ethiopic - (Language Neutral) as the current keyboard.
4. Type <kbd>a</kbd><kbd>s</kbd><kbd>d</kbd><kbd>f</kbd><kbd>Enter</kbd><kbd>a</kbd>

Expected Result: 
"አስድፍ
አ"

Keyboards for the following test are found in the zip file [7077_test_keyboards.zip](https://github.com/keymanapp/keyman/files/9371509/7077_test_keyboards.zip) . Reusing keyboards from PR#7077
**TEST_OUTPUT_KEYSTROKE**
1. Install [043 - output and keystroke.kmx](https://github.com/keymanapp/keyman/files/9371509/7077_test_keyboards.zip)
2. Open Notepad
3. Select  043 - output and keystroke.kmx as the current keyboard.
4. Type <kbd>1</kbd><kbd>2</kbd><kbd>3</kbd>
Expected Result: "abd3"

**TEST_OUTPUT_KEYSTROKE_INVALID_CONTEXT**

1. If not installed from the previous test - Install [043 - output and keystroke.kmx](https://github.com/keymanapp/keyman/files/9371509/7077_test_keyboards.zip)
2. Open Notepad
3. Select  043 - output and keystroke.kmx as the current keyboard.
4. Type <kbd>1</kbd><kbd>2</kbd><kbd>Enter</kbd><kbd>3</kbd>
Expected Result: 
"abc
3"

**TEST_DEADKEY_AND_CONTEXT**
1. Install [045 - deadkey and context.kmx](https://github.com/keymanapp/keyman/files/9371509/7077_test_keyboards.zip)
2. Open Notepad
3. Select  045 - deadkey and context.kmx as the current keyboard.
[...]

PR #7108, opened 2022-08-24
fix(windows): Add invalidate context action to non-updatable parse 🍒

rc-swag:
Cherry Pick from #7089 
Rename ProcessActionsTestParse to NonUpdatable parse as it does more than just test. Add handling for the invalidate context action to this parse. In the case when an emit keystroke is in the action list it will be emitted and the Updateable round will never occur so we need to invalidate the context now in the non-updatable parse.
@keymanapp-test-bot skip


ermshiperete (approved):

keyman-server:

PR #7116, opened 2022-08-24
fix(core): fixes for kmx_file rearrangement 🙀

srl295:
fix breaks in #7111 and prior

#7098

@keymanapp-test-bot skip

### DIFF
--- a/common/include/km_types.h
+++ b/common/include/km_types.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <cstdint>
+#include <stdint.h>
 
 /*
 #if defined(_WIN32) || defined(_WIN64)
--- a/common/include/km_types.h
+++ b/common/include/km_types.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cstdint>
+
 /*
 #if defined(_WIN32) || defined(_WIN64)
 #define snprintf _snprintf
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -4,6 +4,8 @@
   Implementation for the KMX Plus utilities
 */
 
+#include <km_types.h>
+#include <kmx_file.h>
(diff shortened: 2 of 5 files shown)


mcdurdin (approved):

common/include/km_types.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <stdint.h>
```
mcdurdin, line 3: Curious why this is needed?

srl295: @mcdurdin me too

srl295: i wonder if it used to be indirectly included somehow.  That's the danger of optimizing includes !

mcdurdin: So you found you needed it to build?

mcdurdin: Jenkins build is failing because of this line...

```
[2022-08-24T18:58:13.334Z] /<<PKGBUILDDIR>>/common/include/km_types.h:3:10: fatal error: cstdint: No such file or directory

[2022-08-24T18:58:13.334Z]  #include <cstdint>
```

srl295: for linux, yes.  I don't see a teamcity build that shows this, because it stopped on an earlier error. I can test that though

srl295: ^ @mcdurdin  that was old, i changed <cstdint> to <stdint.h> 

I opened https://github.com/keymanapp/keyman/pull/7117 to try skipping that line, to double check that it would fail on linux not just on srl-linux

srl295: i'll merge #7116 *unless* #7117  shows that the #include is unnecessary

srl295: ^ windows failed without that include… passed with.

srl295:
jenkins failing, but it was already failing on  f5f1bfd … merging

PR #7117, opened 2022-08-24
test: what if we did not have stdint 🙀 

srl295:
alternate to #7116


srl295:
@mcdurdin fails on windows

srl295:
breaks. Closing.  See #7116

PR #7120, opened 2022-08-25
feat(developer): Write kmxplus section 🙀

mcdurdin:
Adds support for writing kmxplus section of a kmx file. Extends `KMXFile` to a `KMXPlusFile` subclass. Adds `KMXPlusBuilder` which is called from `KMXBuilder` if `KMXBuilder` finds that the file is a `KMXPlusFile`.

This means that `KMXPlusBuilder` generates a buffer that could be stored independently of its parent .kmx file transport, if that is later considered desirable.

Adds new constants for sizes of various table components, which makes building the file much simpler.

@keymanapp-test-bot skip


developer/src/kmldmlc/src/keyman/kmx/kmx-builder.ts
```diff
@@ -195,20 +195,30 @@ export default class KMXBuilder {
...
-  calculateBitmapSize() {
+  buildBitmap() {
     // TODO
     return 0;
   }
 
-  calculateKMXPlusSize() {
-    // TODO
-    return 0;
+  buildKMXPlus(base: number) {
```
darcywong00, line 209: So both buildBitMap() and buidlKMXPlus() return a size?

```suggestion
  buildKMXPlus(base: number): number {
```

core/include/ldml/keyboardprocessor_ldml.h
```diff
@@ -18,11 +18,16 @@
...
+#define LDML_LENGTH_KEYS_ITEM 0x10
 #define LDML_LENGTH_LOCA 0x10
+#define LDML_LENGTH_LOCA_ITEM 0x4
 #define LDML_LENGTH_META 0x24
 #define LDML_LENGTH_SECT 0x10
+#define LDML_LENGTH_SECT_ITEM 0x8
 #define LDML_LENGTH_STRS 0x10
+#define LDML_LENGTH_STRS_ITEM 0x8
 #define LDML_LENGTH_VKEY 0x10
+#define LDML_LENGTH_VKEY_ITEM 0x8
```
darcywong00, line 30: Are these items auto-generated by core/tools/ldml-const-builder/build.sh? (l.12)

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
```diff
@@ -0,0 +1,319 @@
...
+      reserved: 0,
+      count: this.file.kmxplus.vkey.vkeys.length,
+      items: []
+    };
+
+    for(let item of this.file.kmxplus.vkey.vkeys) {
+      vkey.items.push({
+        vkey: item.vkey,
+        target: item.target
+      });
```
rc-swag, line 215: Will you need to consider padding here also as the `BUILDER_VKEY_ITEM` is 64 bits so may not align with a 128-bit boundry.

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
```diff
@@ -0,0 +1,319 @@
...
+  size: number;
+  _offset: number; // used only for building the output
+};
+
+interface BUILDER_SECT_ITEM {
+  sect: number;
+  offset: number; //? new r.VoidPointer(r.uint32le, {type: 'global'})
+};
+
+interface BUILDER_SECT extends BUILDER_SECTION {
```
rc-swag, line 16: I may be a little lost here but is this interface only ever used for TOC  could we call it `BUILDER_TOC`.  The name `BUILDER_SECTION` and `BUILDER_SECT` are similar. Just a readability comment.

developer/src/kmldmlc/src/keyman/kmx/kmx-builder.ts
mcdurdin: Yes, but we don't need to explicitly set the output type because Typescript can infer it.

core/include/ldml/keyboardprocessor_ldml.h
mcdurdin: Yes

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
mcdurdin: We don't need to pad individual entries in the array.

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
mcdurdin: `SECT` matches the section name of `sect`. It is a little confusing but not sure what else to do with it at present!

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
rc-swag: Not the individual items no. However, once the for loop `for(let item of this.file.kmxplus.vkey.vkeys)` ends you would need to check to see if the array has ended on 128-bit boundary, would you not? then just pad the last 64 bits. Or do you do this somewhere else?

mcdurdin: Yep, not doing section-level padding yet, that will come in a follow-up PR along with other optimisations such as sorting.

rc-swag (approved):
Lgtm

developer/src/kmldmlc/src/keyman/kmx/kmx-builder.ts
```diff
@@ -195,20 +195,30 @@ export default class KMXBuilder {
...
+    if(!(this.file instanceof KMXPlusFile)) {
+      return 0;
```
srl295, line 211: should this be an error? or does 0 imply error?

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
srl295: I realized that also.  `sect` is the array of SECTion offsets. 

Could add an extraneous comment just to clarify

```suggestion
// Builder for the 'sect' (Section table of contents) section
interface BUILDER_SECT extends BUILDER_SECTION {
```

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
```diff
@@ -0,0 +1,319 @@
...
+      });
+    }
+
+    return vkey;
+  }
+
+  private finalize_strs() {
+    this.sect_strs.count = this.sect_strs.items.length;
+    let offset = constants.length_strs + constants.length_strs_item * this.sect_strs.count;
+    // TODO: consider padding
```
srl295, line 224: maybe

```suggestion
    offset += (0x10-(offset%0x10))%0x10;
```

but probably should be a function, and should make sure to zero

srl295 (approved):

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder.ts
mcdurdin: I'll do padding in a follow-up PR, but will consider something like this.

developer/src/kmldmlc/src/keyman/kmx/kmx-builder.ts
mcdurdin: This is supposed to support non-kmxplus files as well, so this section would be zero-length, not an error, if not a kmxplus file.

srl295: OK wasn't sure if this would be called in that case.

srl295 (approved):

PR #7127, opened 2022-08-25
feat(developer): LDML keyboard XML Loader and Compiler 🙀

mcdurdin:
Implements basic LDML keyboard xml loader and compiler.

@keymanapp-test-bot skip


developer/src/kmldmlc/src/keyman/compiler/compiler.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    let y = -1;
+    for(let row of layer.row) {
+      y++;
+      if(y > USVirtualKeyMap.length) {
+        this.callbacks.reportMessage(0, `'hardware' layer has too many rows`);
+        break;
+      }
+
+      const keys = row.keys.split(' ');
```
srl295, line 40: looking good! 🚀

developer/src/kmldmlc/src/keyman/ldml-keyboard/ldml-keyboard-xml-reader.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default class LDMLKeyboardXMLSourceFileReader {
+
+  //private callbacks: CompilerCallbacks;
+  constructor (callbacks: CompilerCallbacks) {
+    //this.callbacks = callbacks;
+  }
+
+  /**
+   * xml2js will not place single-entry objects into arrays.
```
srl295, line 13: it's trying to 'help' you

srl295 (approved):
🚀

developer/src/kmldmlc/src/keyman/ldml-keyboard/ldml-keyboard-xml-reader.ts
mcdurdin: To be fair, given the ambiguity inherent in XML, we either put everything into an array or we end up doing something like this. So I think this is fine :smile:

That's why I used the `explicitArray: false` parameter on line 42. If I had set that to true, I'd have had the opposite problem: having to unbox all the single-element XML nodes :grin:

PR #7129, opened 2022-08-26
feat(core): use kmldmlc to build test keyboards 🙀

mcdurdin:
Removes .kmx from repo and builds them from the source .xml files using kmldmlc (building that if necessary also with a poor fella's dependency check for now), and generally a lot of cleanup of the meson.build files for the ldml tests.

Moves keyboards to a subfolder of the ldml unit test folder to keep them neat and tidy and avoid embedding huge files into ldml.js for test of WASM.

Moves invalid keyboards to invalid-keyboards subfolder -- we don't attempt to build these, just use the hand-crafted .kmx as-is.

Note: 000_null_keyboard does build successfully, but the core ldml keyboardprocessor currently fails on it, because it is missing keys and vkey sections.

Needed to get emscripten out of the path because it runs a different version of node than we do (poor form, emscripten, [to clobber path with private versions of node](https://github.com/emscripten-core/emscripten/issues/4848)!). Instead, we define an environment variable, `EMSCRIPTEN_BASE`, to point to the path where `emcc` may be found.

TODO: wasm builds are not currently enabled on mac build agents, probably because of the pathing issues we experienced previously. We should fix this at some point, will require updating the wasm defs for macos also.

@keymanapp-test-bot skip


mcdurdin:
We have a little bit of a dependency issue here for tests. We don't actually want to build the keyboards using kmldmlc if we are just doing a deployment e.g. a debian packaging build -- as we won't have a node.js environment there. That is, we really only want to build keyman-version and kmldmlc if we are running tests. But it looks we may not be able to do the specific calls we need in the version of meson we have available (0.45?!).

Currently the Linux build calls core/build.sh twice, once to configure, and once to build+test:

https://github.com/keymanapp/keyman/blob/fa28040643bfdbf65afb4a607838a56d353736ac/linux/scripts/build.sh#L29-L44

So one option is we disable tests here on line 43 and add another option, say, "no-node-deps". Ugly though.

Another possibility is we don't fail on missing node, but simply disable the keyboard build and corresponding tests? However, that's not ideal because if we have a broken environment elsewhere we would have silently skipped the tests.

mcdurdin:
image

Yay!

mcdurdin:
Will verify that all Keyman Core tests do run the ldml tests before marking ready-for-review

core/tests/unit/ldml/keyboards/meson.build
```diff
@@ -0,0 +1,43 @@
...
+# Description:  Cross platform build script to compile kmkbpldml API unit tests.
+# Create Date:  5 Aug 2022
+# Authors:      Marc Durdin
+#
+
+tests = [
+# disabling 000 and 002 until we have updates to core or to the keyboards so that they pass
+#  'k_000_null_keyboard',
+  'k_001_tiny',
+#  'k_002_null_invalid'
```
srl295, line 11: I wouldn't expect kmldmlc to generate an invalid .kmx.  So perhaps null_invalid should be preserved for now? Or maybe a directory full of invalid .kmxes (with a starting population of 1)? Eventually I'd like to do some fuzz testing to validate the validator.

srl295:
if we do this (require node for testing), wouldn't it also imply that we could run the ldml constant generator to build ldml? otherwise we'd have a situation where ldml is built but not testable, which seems odd.

mcdurdin:
No, because the build for debian packaging doesn't have access to node. We
can test in a separate CI process but we don't run tests during the
packaging build.

mcdurdin:
Yes I think I will set-up a separate folder for invalid KMX files. Ideally
they will be stored in a text format and compiled down so that we can read
them without a hex editor?

mcdurdin:
> Yes I think I will set-up a separate folder for invalid KMX files. Ideally they will be stored in a text format and compiled down so that we can read them without a hex editor?

I will do the text format as a follow-up PR.

rc-swag (approved):
lgtm

PR #7132, opened 2022-08-26
feat(core): ldml: improve file reading/validation 🙀

srl295:
- don't reparse structs on each keydown, read into memory
- improve validation
- move validation into constructor
- validate length of data
- don't retain the raw .kmx data at runtime

#5015 

and of course,
@keymanapp-test-bot skip

### DIFF
--- a/core/src/ldml/ldml_processor.cpp
+++ b/core/src/ldml/ldml_processor.cpp
@@ -102,15 +102,14 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t> 
   KMXPLUS_ASSERT(true, data.size() > sizeof(kmx::COMP_KEYBOARD_EX));
 
   // Locate the structs here, but still retain ptrs to the raw structs.
-  const kmx::PCOMP_KEYBOARD comp_keyboard = (kmx::PCOMP_KEYBOARD)data.data();
+  const kmx::COMP_KEYBOARD* comp_keyboard = (kmx::COMP_KEYBOARD*)data.data();
   KMXPLUS_ASSERT(true, !!(comp_keyboard->dwFlags & KF_KMXPLUS));
   const kmx::COMP_KEYBOARD_EX* ex = reinterpret_cast<const kmx::COMP_KEYBOARD_EX*>(comp_keyboard);
 
   // validate size and offset
-  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize >  LDML_LENGTH_SECT);
-  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize <  (data.size() - sizeof(kmx::COMP_KEYBOARD_EX)));
+  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize >= LDML_LENGTH_SECT);
   KMXPLUS_ASSERT(true, ex->kmxplus.dpKMXPlus     >= sizeof(kmx::COMP_KEYBOARD_EX));
-  KMXPLUS_ASSERT(true, ex->kmxplus.dpKMXPlus     <  data.size());
+  KMXPLUS_ASSERT(true, ex->kmxplus.dpKMXPlus + ex->kmxplus.dwKMXPlusSize <= data.size());
 
   // calculate pointers to start and end of kmxplus
   const uint8_t * const kmxplusdata   = data.data() + ex->kmxplus.dpKMXPlus;      // Start of + data
@@ -157,7 +156,7 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t> 
       KMX_WCHAR out[BUFSIZ];
       if (entry.flags && LDML_KEYS_FLAGS_EXTEND) {
         KMXPLUS_ASSERT(false, nullptr == strs->get(entry.to, out, BUFSIZ));
-        for(len=0; len<BUFSIZ && out[len]; len++);
+        for(len=0; len<BUFSIZ && out[len]; len++); // TODO-LDML: validate string length here, #7134 
       } else {
         UTF32 buf32[2];
         buf32[0] = entry.to; // UTF-32
--- a/core/src/ldml/ldml_processor.cpp
+++ b/core/src/ldml/ldml_processor.cpp
@@ -148,6 +148,35 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t> 
     KMXPLUS_ASSERT(true, keys != nullptr);
     // Specified data size fits in total
     KMXPLUS_ASSERT(true, (offset+keys->header.size) < ex->kmxplus.dwKMXPlusSize);
+
+    // read from keys
+    for (KMX_DWORD i=0; i<keys->count; i++) {
+      const kmx::COMP_KMXPLUS_KEYS_ENTRY &entry = keys->entries[i];
+      KMX_DWORD len = 0;
+      KMX_WCHAR out[BUFSIZ];
+      if (entry.flags && LDML_KEYS_FLAGS_EXTEND) {
+        KMXPLUS_ASSERT(false, nullptr == strs->get(entry.to, out, BUFSIZ));
+        for(len=0; len<BUFSIZ && out[len]; len++);
+      } else {
+        UTF32 buf32[2];
+        buf32[0] = entry.to; // UTF-32
+        buf32[1] = 0; // to avoid UMR warning
+        UTF32 *sourceStart = &buf32[0];
+        const UTF32 *sourceEnd = &buf32[1]; // Reference off the end. NULL to avoid UMR.
+        UTF16 *targetStart = (UTF16*)out;
+        const UTF16 *targetEnd = (UTF16*)out+BUFSIZ-1;
+        ConversionResult result = ::ConvertUTF32toUTF16(&sourceStart, sourceEnd, &targetStart, targetEnd, strictConversion);
+        KMXPLUS_ASSERT(conversionOK, result);
+        *targetStart = 0;
+        len = 1; // TODO=LDML calculate
+        // len = (targetStart - out);
+        KMXPLUS_ASSERT(true, len>=1 && len <= 2);
+      }
+      KMXPLUS_ASSERT(true, len>0 && len < BUFSIZ);
+      KMXPLUS_ASSERT(0, out[len]); // null termination
+      ldml_vkey_id vkey_id((km_kbp_virtual_key)entry.vkey, (uint16_t)entry.mod);
+      vkey_to_string[vkey_id] = std::u16string(out, len); // assign the string
+    }
   }
   KMXPLUS_PRINTLN("_valid = true");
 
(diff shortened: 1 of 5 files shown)


core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+  const kmx::COMP_KMXPLUS_STRS *strs = nullptr;
+  const kmx::COMP_KMXPLUS_KEYS *keys = nullptr;
+
   // TODO-LDML: load the file from the buffer (KMXPlus format)
   // Note: kb_path is essentially debug metadata here
-  assert(data.size() != 0);
+  KMXPLUS_ASSERT(true, data.size() > sizeof(kmx::COMP_KEYBOARD_EX));
 
+  // Locate the structs here, but still retain ptrs to the raw structs.
   const kmx::PCOMP_KEYBOARD comp_keyboard = (kmx::PCOMP_KEYBOARD)data.data();
```
mcdurdin, line 105:
```suggestion
  const kmx::COMP_KEYBOARD* comp_keyboard = (kmx::COMP_KEYBOARD*)data.data();
```

Can we use `reinterpret_cast` here also? Suggest using pointer type styling consistently (`COMP_FOO*` everywhere or `PCOMP_FOO` everywhere).

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+  KMXPLUS_ASSERT(true, !!(comp_keyboard->dwFlags & KF_KMXPLUS));
+  const kmx::COMP_KEYBOARD_EX* ex = reinterpret_cast<const kmx::COMP_KEYBOARD_EX*>(comp_keyboard);
+
+  // validate size and offset
+  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize >  LDML_LENGTH_SECT);
```
mcdurdin, line 110: What if there is only a `SECT` section? Technically structurally valid? This assumption may need discussion -- it impacts other parts of the keyboard load below.

```suggestion
  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize >= LDML_LENGTH_SECT);
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize <  (data.size() - sizeof(kmx::COMP_KEYBOARD_EX)));
+  KMXPLUS_ASSERT(true, ex->kmxplus.dpKMXPlus     >= sizeof(kmx::COMP_KEYBOARD_EX));
+  KMXPLUS_ASSERT(true, ex->kmxplus.dpKMXPlus     <  data.size());
```
mcdurdin, line 113:
```suggestion
  KMXPLUS_ASSERT(true, ex->kmxplus.dpKMXPlus + ex->kmxplus.dwKMXPlusSize <= data.size());
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+  KMXPLUS_ASSERT(true, data.size() > sizeof(kmx::COMP_KEYBOARD_EX));
 
+  // Locate the structs here, but still retain ptrs to the raw structs.
   const kmx::PCOMP_KEYBOARD comp_keyboard = (kmx::PCOMP_KEYBOARD)data.data();
+  KMXPLUS_ASSERT(true, !!(comp_keyboard->dwFlags & KF_KMXPLUS));
+  const kmx::COMP_KEYBOARD_EX* ex = reinterpret_cast<const kmx::COMP_KEYBOARD_EX*>(comp_keyboard);
+
+  // validate size and offset
+  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize >  LDML_LENGTH_SECT);
+  KMXPLUS_ASSERT(true, ex->kmxplus.dwKMXPlusSize <  (data.size() - sizeof(kmx::COMP_KEYBOARD_EX)));
```
mcdurdin, line 111: I think this can be skipped as less helpful or correct than my change to l.113?

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+  KMXPLUS_ASSERT(true, kmx::validate_kmxplus_data(kmxplusdata));
+
+  // Get out the SECT header
+  {
+    const KMX_DWORD offset = 0;
+    sect = kmx::as_kmxplus_sect(kmxplusdata+offset);
+    KMXPLUS_ASSERT(true, sect != nullptr);
+    // Specified data size fits in total
+    KMXPLUS_ASSERT(true, (offset+sect->header.size) < ex->kmxplus.dwKMXPlusSize);
+  }
```
mcdurdin, line 132: For unit testing and reusability, it would be good if these could all be extracted to separate functions, in a separate .cpp. Could even make them a separate class.

I'd like to see the common components of these in a single function:
* offset tests
* size tests
* ident tests

```
KMXPLUS_ASSERT(true, (offset+sect->header.size) < ex->kmxplus.dwKMXPlusSize);
```
This should be `<=` in all cases because the section may be the last thing in the kmxplus section.

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+    KMXPLUS_ASSERT(true, (offset+keys->header.size) < ex->kmxplus.dwKMXPlusSize);
+
+    // read all keys into array
+    for (KMX_DWORD i=0; i<keys->count; i++) {
+      const kmx::COMP_KMXPLUS_KEYS_ENTRY &entry = keys->entries[i];
+      KMX_DWORD len = 0;
+      KMX_WCHAR out[BUFSIZ];
+      if (entry.flags && LDML_KEYS_FLAGS_EXTEND) {
+        KMXPLUS_ASSERT(false, nullptr == strs->get(entry.to, out, BUFSIZ));
+        for(len=0; len<BUFSIZ && out[len]; len++);
```
mcdurdin, line 160: A malformed file could cause a run past the end of buffer here.

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+        UTF16 *targetStart = (UTF16*)out;
+        const UTF16 *targetEnd = (UTF16*)out+BUFSIZ-1;
+        ConversionResult result = ::ConvertUTF32toUTF16(&sourceStart, sourceEnd, &targetStart, targetEnd, strictConversion);
+        KMXPLUS_ASSERT(conversionOK, result);
+        *targetStart = 0;
+        len = 1; // TODO=LDML calculate
+        // len = (targetStart - out);
+        KMXPLUS_ASSERT(true, len>=1 && len <= 2);
+      }
+      KMXPLUS_ASSERT(true, len>0 && len < BUFSIZ);
```
mcdurdin, line 176: Is a zero length output possible for a key?

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+        UTF32 buf32[2];
+        buf32[0] = entry.to; // UTF-32
+        buf32[1] = 0; // to avoid UMR warning
+        UTF32 *sourceStart = &buf32[0];
+        const UTF32 *sourceEnd = &buf32[1]; // Reference off the end. NULL to avoid UMR.
+        UTF16 *targetStart = (UTF16*)out;
+        const UTF16 *targetEnd = (UTF16*)out+BUFSIZ-1;
+        ConversionResult result = ::ConvertUTF32toUTF16(&sourceStart, sourceEnd, &targetStart, targetEnd, strictConversion);
+        KMXPLUS_ASSERT(conversionOK, result);
+        *targetStart = 0;
```
mcdurdin, line 171: This conversion is a little painful. I wonder if we can use the unicode.h macros instead -- both faster and cleaner?

```
if(Uni_IsSMP(entry.to)) {
  out[0] = Uni_UTF32ToSurrogate1(entry.to);
  out[1] = Uni_UTF32ToSurrogate2(entry.to);
  out[2] = 0;
  len = 2;
} else {
  out[0] = (char16_t) entry.to;
  out[1] = 0;
  len = 1;
}
```

mcdurdin (changes requested):
This is looking good -- but I think it'd be good to restructure the file loading and validation here before it solidifies.

core/src/ldml/ldml_processor.cpp
srl295: That's a good question. I would think that would not be valid, but perhaps it should be.

core/src/ldml/ldml_processor.cpp
srl295: It is painful. I have a separate ticket for improving this.

core/src/ldml/ldml_processor.cpp
srl295: > This is looking good -- but I think it'd be good to restructure the file loading and validation here before it solidifies.

thanks for the comments.  Separate functions make sense.  Perhaps they can be part of the kmx_plus.h structs.

core/src/ldml/ldml_processor.cpp
```diff
@@ -59,23 +59,130 @@ namespace {
...
+    // Specified data size fits in total
+    KMXPLUS_ASSERT(true, (offset+keys->header.size) < ex->kmxplus.dwKMXPlusSize);
+
+    // read all keys into array
+    for (KMX_DWORD i=0; i<keys->count; i++) {
+      const kmx::COMP_KMXPLUS_KEYS_ENTRY &entry = keys->entries[i];
+      KMX_DWORD len = 0;
+      KMX_WCHAR out[BUFSIZ];
+      if (entry.flags && LDML_KEYS_FLAGS_EXTEND) {
+        KMXPLUS_ASSERT(false, nullptr == strs->get(entry.to, out, BUFSIZ));
```
rc-swag, line 159: I think this may have been asked but I missed the answer. Is the reason to just use a static buffer of size 512  because we know it will always be long enough?  Otherwise, we could make a get accept a nullptr for the buffer and return the required size.

core/src/ldml/ldml_processor.cpp
srl295: I'm not sure that makes sense.  It's a keyboard, so it should have keys, right? at least a 0-length array of keys.  If it was a generic container that had other content, then perhaps `keys` would be optional.

core/src/ldml/ldml_processor.cpp
srl295: https://github.com/keymanapp/keyman/issues/7134

core/src/ldml/ldml_processor.cpp
srl295: this is going to change to using a string buffer, https://github.com/keymanapp/keyman/issues/7134

core/src/ldml/ldml_processor.cpp
mcdurdin: A "null" keyboard is a valid keyboard -- useful if nothing else for unit tests. But also we may well see keyboards that use a US English layout but have transforms. So I do think I can see a case for skipping a `keys` section.

In any case, I think we want to be clear on what is structurally valid vs functional validity. Then, we need to clearly document which sections are required in the kmxplus spec file.

core/src/ldml/ldml_processor.cpp
mcdurdin: Again, structural vs functional validity. I think we should treat it as possible.

core/src/ldml/ldml_processor.cpp
srl295: Sounds good to me then.  No keys, just transforms. 

- [ ] update the spec to clarify validity

core/src/ldml/ldml_processor.cpp
srl295: - [ ] also note this in the spec.

core/src/ldml/ldml_processor.cpp
srl295:
```suggestion
```

core/src/ldml/ldml_processor.cpp
srl295: I'm going to do the string work in a separate PR, concentrating on validation here.

core/src/ldml/ldml_processor.cpp
srl295:
```suggestion
        for(len=0; len<BUFSIZ && out[len]; len++); // TODO-LDML: validate string length here, #7134 
```

core/src/ldml/ldml_processor.cpp
srl295: I don't think we should do any special work to handle the 0-length case, so I'm going to require a 0-length entry in the string table.  (Yes, it means that 12 bytes total maximum are added to the file if a keyboard contains null keys!)

mcdurdin: The compiler would end up doing that by default anyway, wouldn't it? 😆

mcdurdin (approved):

srl295:
I'm going to merge this because it improves validation.  I will make a followup PR to split out the validation.

srl295:
feature-ldml broke… on it…

mcdurdin:
> feature-ldml broke… on it…

What do you mean?

srl295:
> > feature-ldml broke… on it…
> 
> What do you mean?

between your PR and my PR, my validation now fails on 001 tiny.

srl295:
@mcdurdin  fix https://github.com/keymanapp/keyman/pull/7152

Issue #7102, opened 2022-08-23
feat(build): add a Dockerfile for linux builds


srl295:
https://gist.github.com/srl295/08ecb5e61ecc3351bd43296c501e9ad0 is a Dockerfile that is able to build keyman/core in a linux docker image.

Add this to the repository.

PR #7133, opened 2022-08-26
feat(linux): Dockerfile for linux builder

srl295:
- add Dockerfile in linux/Dockerfile, referencing linux/debian/control
- add docs in docs/build/linux-ubuntu.md

Fixes: #7102

@keymanapp-test-bot skip

### DIFF
--- a/docs/build/linux-ubuntu.md
+++ b/docs/build/linux-ubuntu.md
@@ -64,3 +64,23 @@ All dependencies are already installed if you followed the instructions under [P
 Building:
 
 * [Building Keyman Core](../../core/doc/BUILDING.md)
+
+## Docker Builder
+
+The Docker builder allows you to perform a linux build from anywhere Docker is supported.
+To build the docker image:
+
+```shell
+cd ../../linux
+docker pull ubuntu:latest
+docker build . -t keymanapp/keyman-linux-builder:latest
+```
+
+Once the image is built, it may be used to build parts of Keyman.
+
+```shell
+cd ../core
+# keep linux build artifacts separate
+mkdir -p build/linux
+docker run -it --rm -v $(pwd)/..:/home/build -v $(pwd)/build/linux:/home/build/core/build keyman-linux-builder:latest bash -c 'cd core; bash build.sh -d'
+```
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright (c) 2022 SIL International. All rights reserved.
+#
+# builder image for a linux build
+# see ../docs/build/linux-ubuntu.md
+
+FROM --platform=amd64 ubuntu:latest
+LABEL org.opencontainers.image.authors="SIL International."
+LABEL org.opencontainers.image.url="https://github.com/keymanapp/keyman.git"
+LABEL org.opencontainers.image.title="Keyman Linux Build Image"
+# We will switch to a build user after some installation
+USER root
+RUN useradd -c "Build user" -d $HOME -m build
+ENV HOME /home/build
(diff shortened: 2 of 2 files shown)


srl295:
I used the `keymanapp/` org found at https://hub.docker.com/u/keymanapp but there's no automation to push this build there.

I use this image a lot. It ought to be able to build all of keyman, But I've only tested it with core.

docs/build/linux-ubuntu.md
```diff
@@ -64,3 +64,23 @@ All dependencies are already installed if you followed the instructions under [P
...
+docker build . -t keymanapp/keyman-linux-builder:latest
+```
+
+Once the image is built, it may be used to build parts of Keyman.
+
+```shell
+cd ../core
+# keep linux build artifacts separate
+mkdir -p build/linux
+docker run -it --rm -v $(pwd)/..:/home/build -v $(pwd)/build/linux:/home/build/core/build keyman-linux-builder:latest bash -c 'cd core; bash build.sh -d'
```
ermshiperete, line 85: Do we need both `-v`s? I'd think the latter is included in the former.

ermshiperete (approved):
:+1:

docs/build/linux-ubuntu.md
srl295: no, by default it builds into `build/arch` which would stomp on your local build.   This creates a `build/linux` which is its own `build` subdirectory.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.53-alpha](https://keyman.com/downloads/releases/alpha/16.0.53)

mcdurdin:
Just thinking about this -- I think the Dockerfile should be in /resources/devbox/linux, as ideally we will have multiple Docker images for different platform builds -- one for each platform (although Windows will probably be a non-starter for a while due to Delphi). And this is about build environment rather than something specific to Keyman.

Currently discussing whether or not we can do a Docker image for macOS builds as well. We may use this for future macOS build agents.

srl295:
@mcdurdin OK, want a new PR to move it there?

mcdurdin:
> @mcdurdin OK, want a new PR to move it there?

Oh yes please :grin:

srl295:
> > @mcdurdin OK, want a new PR to move it there?
> 
> Oh yes please 😁

A wrinke here is that the Dockerfile depends on `linux/debian/control` - it actually is using the debian control file to pull in the deps.  Could have a script in `devbox` that copies over `linux/debian/control` into a gitignored location?

mcdurdin:
It's not possible to `ADD ../../../linux/debian/control /tmp/control`?

srl295:
> It's not possible to `ADD ../../../linux/debian/control /tmp/control`?

No, it's not possible to ADD outside of the docker context.  You could start from up three levels, but then it would send the entre repo as context.

srl295:
conversely, could a script in devbox do `docker build ../../../linux` ?

mcdurdin:
Yeah I guess so? Then we'll eventually end up with a Dockerfile in each of the primary top-level platform folders -- android/Dockerfile, developer/Dockerfile, ios/Dockerfile, linux/Dockerfile, mac/Dockerfile, web/Dockerfile, windows/Dockerfile?

Alternatively, just three: linux/Dockerfile, windows/Dockerfile, mac/Dockerfile, which contain all the toolchains needed for building any product that can be built on that platform. (Gosh what a wonderful world it would be if macOS would (legally) run in a VM on non-mac hardware...)

Windows is complicated because Delphi. macOS will be a massive image because XCode. So not sure this is yet ideal but step-by-step we improve our build environments I hope.

PR #7137, opened 2022-08-28
feat(core): hextobin tool 🙀

mcdurdin:
The hextobin tool is used to help us construct invalid .kmx files without having to either hack these into the compiler, or write them with a hex editor. It feels a little bit of overkill, but this makes it easy to (a) work with our intended file structures, and (b) review any changes to these invalid keyboard tests.

Note: I did do a cursory search for a hex2bin or hextobin that we could easily use but did not find a good simple cross-platform tool that we could consume (`xxd -r` would work but AFAICT constrains our format to a dump which is much harder to work with and read -- I wanted to be able to annotate with file structures.)

For an example, see ik_000_null_invalid.txt.

This means that building the invalid keyboards for tests requires node, but that should be fine given we need node to build the valid keyboards anyway.

I've put hextobin under a new common/tools because it will be used by the kmldmlc compiler unit tests in my next PR in this chain.

@keymanapp-test-bot skip


mcdurdin:
Failing tests are unrelated. Le sigh.

PR #7140, opened 2022-08-29
feat(developer): unit test framework for kmldmlc 🙀

mcdurdin:
Establishes a basic set of unit and e2e tests for the LDML keyboard compiler, giving me a place to put additional tests.

I'm using hextobin for the binary generation of .kmx for comparison for test. It's a lot of work to hand-craft. But doing it automatically or including binary fixtures has the downside of probably failing to actually check that the generated file is valid!

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_plus.h
+++ b/core/src/kmx/kmx_plus.h
@@ -39,8 +39,8 @@ struct COMP_KMXPLUS_SECT_ENTRY {
 struct COMP_KMXPLUS_SECT {
   COMP_KMXPLUS_HEADER header;
   KMX_DWORD total;                     // 0008 KMXPlus entire length
-  KMX_DWORD count;                     // 000B number of section headers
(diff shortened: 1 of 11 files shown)


core/src/kmx/kmx_plus.h
```diff
@@ -39,8 +39,8 @@ struct COMP_KMXPLUS_SECT_ENTRY {
 struct COMP_KMXPLUS_SECT {
   COMP_KMXPLUS_HEADER header;
   KMX_DWORD total;                     // 0008 KMXPlus entire length
-  KMX_DWORD count;                     // 000B number of section headers
-  COMP_KMXPLUS_SECT_ENTRY entries[];  // 0010 section entries
+  KMX_DWORD count;                     // 000C number of section headers
```
srl295, line 42: good catch

srl295 (approved):
LGTM

PR #7147, opened 2022-08-29
fix(linux): Allow downgrades for installing build deps

ermshiperete:
When install the keyman-build-deps_*.deb file it can happen that we end up with a lower version number than what we previously
built. This change allows downgrades so that the build dependencies can be installed.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.54-alpha](https://keyman.com/downloads/releases/alpha/16.0.54)

PR #7153, opened 2022-08-30
refactor(developer): each section compiler is a pure function 🙀

mcdurdin:
Makes each section compiler a pure function. Good for testing in the future.

Well, almost. There are still callback side-effects for messages. These may in future be returned as an array as well, eliminating the need for a message callback.

But restructures each section compiler so that it is self-contained and responsible only for writing its own section. Also tightens up the types for the sections constants.

Last, but not least, moves responsibility for writing the output .kmx out of compiler.ts and into its callers -- removing one dependency from the compiler. Reading input files still needs to be done in the compiler, as there may be multiple input files with `<import>` that we cannot know about until we start parsing xml.

@keymanapp-test-bot skip

### DIFF
--- a/core/include/ldml/keyboardprocessor_ldml.ts
+++ b/core/include/ldml/keyboardprocessor_ldml.ts
@@ -5,6 +5,29 @@
   to be shared between TypeScript and C++ via the generator (below)
 */
 
+// This defines the section identifiers and ensures that we include each and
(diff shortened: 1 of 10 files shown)


mcdurdin:
Jenkins build failure is fixed by #7147.

srl295 (approved):

PR #7156, opened 2022-08-30
fix(android/engine): Fix Backspace key to delete without errant subkeys

darcywong00:
Fixes #7069, fixes #7172, and fixes #7155 and follow-on to #6984

When the subkeys window is dismissed, the subKeysList isn't cleared, so holding backspace key would end up displaying the previous longpress menu.

# User Testing

We need to re-run the main test from #6984, first, to verify that this does not re-introduce new bugs.

Setup - A physical Android device. Don't use emulator because the scenario involves typing fast on the OSK (e.g. two thumbs)

* **TEST_POPUPS** - Verify popup keys don't get stuck on
1. On the Android device, install the PR build of Keyman for Android
2. Dismiss the "Get Started" menu
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Type on the OSK using the following scenarios and verify expected output:
    * Clicking a suggestion on the suggestion banner - should insert the suggestion
    * short-press a key and release - should insert the base key
    * long-press a key, select a long-press key, and release - should insert the long-press key (**note:** there should be a 0.5 second delay before the menu appears)
    * long-press a key, while keeping the finger down, move off the long-press options, and release - should **not** output
    * long-press a key, while keeping the finger down, move off the long-press options, then move back on a long-press option so it's highlighted, and release - should output the long-press key
    * quickly type a long paragraph (e.g. repeat the word "reply") - verify long-press keys don't get stuck (displayed when not touching a key)

* **TEST_UP_FLICK** - Verify that the up-flick gesture reliably opens the longpress menu
1. On the Android device, install the PR build of Keyman for Android
2. Dismiss the "Get Started" menu
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Touch a key such as `e` and immediately swipe upwards to access the long-press menu, keeping your finger down. This should show long-press options without the 0.5 second delay.
   * You should be able to select an option from the long-press menu.
   * You should be able to move back onto the base key to select the base key letter again.
   * You should also be able to move your finger away from the menu to cancel the input.
5. Please run this test a number of times to verify that the behaviour remains consistent.

* **TEST_HELD_BACKSPACE** - Verify spurious longpress menu doesn't appear while holding backspace key
1. On the Android device, install the PR build of Keyman for Android
2. Dismiss the "Get Started" menu
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Accept a suggestion
5. Type another letter
6. Touch and hold the backspace key
7. Verify the previous letter is deleted
[...]

Issue #7069, opened 2022-08-12
bug(android): Pressing Backspace key for a while make the cursor stuck on the text pane 


bharanidharanj:
**Describe the bug**
After entering a word in the text pane I tried to delete all the characters by long pressing the Backspace key leads to stuck the cursor on the text pane. 

**To Reproduce**
1. Install Keyman 16.0.45 alpha build. 
2. Open Keyman In-App. 
3. Enter a word in the text pane.
4. Try to delete all the characters in the word by long pressing the Backspace Key. 

Here, I noticed that the cursor on the text pane get stuck and I was not able to delete any one of the character. I have attached the Screenshot as well as the Video file for reference. 

**Expected behavior**
Long pressing with the Backspace key should delete all the characters one by one. 

**Screenshots**

<img width="500" src="https://user-images.githubusercontent.com/19683143/184368802-753dc848-d5d6-4611-a00d-c5be0cce3d1f.PNG">

**Video**

https://user-images.githubusercontent.com/19683143/184367700-2cd4fac0-3387-460c-8eb9-785096d43b83.mp4

**Related issues**
https://github.com/keymanapp/keyman/issues/6544

---

**Keyman for Android:**
   Device:      Android Mobile
 - OS:            Android 11.0 
 - Keyman version:    16.0.45-alpha
 - Target Application: Keyman In-App

---

**Keyboard**
 - Keyboard name:   English - EuroLatin(SIL)
 - Keyboard version: 2.0
 - Language name:   English

**Additional context**
1. I was able to reproduce this issue in API 29 / Android 10 emulator. 
2. I was not able to reproduce this issue in iOS 15.5 / iPhone 13 Pro Max Simulator.

MakaraSok:
Confirm that this issue does occur when testing with the latest alpha (v16.0.49) on Android 10 (API 29) and Android 11 (API 30).

rowbory:
I saw it affect an Android user with version 15.0.268 last week, so it may not be new in 16.x.

mcdurdin:
I believe this is fixed by #7156. Can you confirm @MakaraSok @bharanidharanj?

bharanidharanj:
@mcdurdin  Retested this issue with Keyman 16.0.54-alpha-test-7156 build in both Android emulator and Redmi Note 8 Pro device (Ver 11.0) and I was not able to reproduce it. Seems to be fixed and working fine with the #7156 PR build. Hence, I am closing the same. Thanks.

[...]

PR #7176, opened 2022-08-31
fix(android/engine): :cherries: Cleanup list of subkeys when dismissing window

darcywong00:
:cherries: pick of #7156 to stable-15.0

When the subkeys window is dismissed, the subKeysList isn't cleared, so holding backspace key or globe key would end up displaying the previous longpress menu.

# User Testing

We need to re-run the main test from #6984, first, to verify that this does not re-introduce new bugs.

Setup - A physical Android device. Don't use emulator because the scenario involves typing fast on the OSK (e.g. two thumbs)

* **TEST_POPUPS** - Verify popup keys don't get stuck on
1. On the Android device, install the PR build of Keyman for Android
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Type on the OSK using the following scenarios and verify expected output:
    * Clicking a suggestion on the suggestion banner - should insert the suggestion
    * short-press a key and release - should insert the base key
    * long-press a key, select a long-press key, and release - should insert the long-press key (**note:** there should be a 0.5 second delay before the menu appears)
    * long-press a key, while keeping the finger down, move off the long-press options, and release - should **not** output
    * long-press a key, while keeping the finger down, move off the long-press options, then move back on a long-press option so it's highlighted, and release - should output the long-press key
    * quickly type a long paragraph (e.g. repeat the word "reply") - verify long-press keys don't get stuck (displayed when not touching a key)

* **TEST_UP_FLICK** - Verify that the up-flick gesture reliably opens the longpress menu
1. On the Android device, install the PR build of Keyman for Android
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Touch a key such as `e` and immediately swipe upwards to access the long-press menu, keeping your finger down. This should show long-press options without the 0.5 second delay.
   * You should be able to select an option from the long-press menu.
   * You should be able to move back onto the base key to select the base key letter again.
   * You should also be able to move your finger away from the menu to cancel the input.
5. Please run this test a number of times to verify that the behaviour remains consistent.

* **TEST_HELD_BACKSPACE** - Verify spurious longpress menu doesn't appear while holding backspace key
1. On the Android device, install the PR build of Keyman for Android
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Accept a suggestion
5. Type another letter
6. Touch and hold the backspace key
7. Verify the previous letter is deleted
8. Verify longpress menus for that letter do **not** appear
Note, if the backspace key fails to delete, that's a separate issue #7069 not fixed on this PR

* **TEST_HELD_GLOBE** - Verify spurious longpress menu doesn't appear while holding globe key
1. On the Android device, install the PR build of Keyman for Android
3. In the Keyman app, start typing with the default sil_euro_latin.
4. Touch and hold the globe key
5. Verify longpress menus for other letters do **not** appear
6. Release the globe key and verify the Keyboard menu appears (default globe key action when only 1 Keyman keyboard is installed)

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -372,8 +372,10 @@ final class KMKeyboard extends WebView {
 
   public void dismissSubKeysWindow() {
     try {
-      if (subKeysWindow != null && subKeysWindow.isShowing())
+      if (subKeysWindow != null && subKeysWindow.isShowing()) {
         subKeysWindow.dismiss();
+      }
+      subKeysList = null;
     } catch (Exception e) {
       KMLog.LogException(TAG, "", e);
     }


mcdurdin (approved):

bharanidharanj:
* **TEST_POPUPS (PASSED):** Tested this with the attached PR build (15.0.270-test-7176) in the Android Mobile device Version 11.0 and it is working as expected. 
* **TEST_UP_FLICK (PASSED):** Tested this with the attached PR build (15.0.270-test-7176) in the Android Mobile device Version 11.0 and it is working as expected. 
* **TEST_HELD_BACKSPACE (PASSED):** Tested this with the attached PR build (15.0.270-test-7176) in the Android Mobile device Version 11.0 and it is working as expected. 
* **TEST_HELD_GLOBE (PASSED):** Tested this with the attached PR build (15.0.270-test-7176) in the Android Mobile device Version 11.0 and it is working as expected.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.270](https://keyman.com/downloads/releases/stable/15.0.270)

PR #7178, opened 2022-08-31
chore(android): Reduce Toast notification noise

### LINKED ISSUE #7170: bug(android): polish - toast notifications when there’s nothing to update are annoying
**Describe the bug**

From a team review of the Keyman for Android UX (#7161)

There is no need to show a toast about checking for updates, nor the outcome of the update check. Particularly because these toasts can appear when using the system keyboard, which is then completely out of context for the user.

We should reserve toasts only for user-initiated actions that have a delayed asynchronous result, and perhaps for serious errors.

---

darcywong00:
Fixes #7170

To minimize UX distractions, this PR removes Toast notifications when nothing new is happening (installed/updated).

The dictionary query on the Language Settings menu will still notify the app is "Checking for associated dictionary to download" to give the user feedback on the menu.

## User Testing
* **TEST_REDUCED_TOAST_NOISE** - Verifies the number of Toast notifications are reduced
2. Launch the app and observe the "Get Started" menu is displayed
3. Verify there's no Toast notification about checking the cloud catalog for updates
4. Dismiss the "Get Started" menu and  go to Keyman settings --> Installed Languages
5. Click on "+" and search for "afade" language keyboard to install (install a keyboard that has no associated dictionary)
6. Install sil_cameroon_azerty and install the keyboard for the "Afade" language"
7. After the keyboard package is installed, verify no Toast notifications appear about checking for associated dictionary
8. Go to Keyman Settings --> Installed Languages --> Afade
9. On the Afade language settings menu, click the bottom "Dictionary - Check for available dictionary"
10. Verify a Toast notification appears about checking for associated dictionary to download
11. Wait a few minutes and verify no additional toast notifications appear
12. Go to Keyman settings --> Installed Languages
13. Click on "+" and search for khmer_angkor (install a keyboard that has an associated dictionary)
14. Install khmer_angkor
15. After the keyboard package is installed, verify:
  * no Toast notifications appear about "checking for associated dictionary"
  * two Toast notifications do appear about a "dictionary being installed in the background" and eventually "Resources successfully updated" (this may take a while depending on network lag)

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudLexicalModelMetaDataDownloadCallback.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudLexicalModelMetaDataDownloadCallback.java
@@ -148,7 +148,7 @@ public class CloudLexicalModelMetaDataDownloadCallback implements ICloudDownload
             KMLog.LogException(TAG, "Error parsing lexical model from api.keyman.com. ", e);
           }
         } else {
-          BaseActivity.makeToast(aContext, R.string.no_associated_model, Toast.LENGTH_SHORT);
+          // Don't make a Toast notification about nothing to download
         }
       }
 
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudLexicalModelMetaDataDownloadCallback.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudLexicalModelMetaDataDownloadCallback.java
@@ -125,31 +125,12 @@ public class CloudLexicalModelMetaDataDownloadCallback implements ICloudDownload
    * @param aMetaDataResult the meta data results
    */
   private void processCloudResults(Context aContext, List<MetaDataResult> aMetaDataResult) {
-    for(MetaDataResult _r:aMetaDataResult) {
-      if (_r.returnjson.target== CloudApiTypes.ApiTarget.Keyboard) {
+    for (MetaDataResult _r : aMetaDataResult) {
+      if (_r.returnjson.target == CloudApiTypes.ApiTarget.Keyboard) {
         //handleKeyboardMetaData(_r);
       }
-      if(_r.returnjson.target== CloudApiTypes.ApiTarget.KeyboardLexicalModels) {
-        JSONArray lmData = _r.returnjson.jsonArray;
-        if (lmData != null && lmData.length() > 0) {
-          try {
-            JSONObject modelInfo = lmData.getJSONObject(0);
-
-            if (modelInfo.has("packageFilename") && modelInfo.has("id")) {
-              String _modelID = modelInfo.getString("id");
-              ArrayList<CloudApiTypes.CloudApiParam> urls = new ArrayList<>();
-              urls.add(new CloudApiTypes.CloudApiParam(
-                CloudApiTypes.ApiTarget.LexicalModelPackage,
-                modelInfo.getString("packageFilename")));
-              _r.additionalDownloadid = CloudLexicalPackageDownloadCallback.createDownloadId(_modelID);
-              _r.additionalDownloads= urls;
-            }
-          } catch (JSONException e) {
-            KMLog.LogException(TAG, "Error parsing lexical model from api.keyman.com. ", e);
-          }
-        } else {
-          // Don't make a Toast notification about nothing to download
-        }
+      if (_r.returnjson.target == CloudApiTypes.ApiTarget.KeyboardLexicalModels) {
+        processCloudResultForModel(aContext, _r);
       }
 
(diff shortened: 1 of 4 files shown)


bharanidharanj:
* **TEST_REDUCED_TOAST_NOISE (PASSED):** Tested with the attached PR build (16.0.55-alpha-test-7178) in my Android Mobile device Version 11.0 and I noticed that there is no Toast Notification during the the installation of the Cameroon Azerty and Khmer Angkor Keyboards. Seems to be working as expected.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudLexicalModelMetaDataDownloadCallback.java
```diff
@@ -148,7 +148,7 @@ private void processCloudResults(Context aContext, List<MetaDataResult> aMetaDat
             KMLog.LogException(TAG, "Error parsing lexical model from api.keyman.com. ", e);
           }
         } else {
-          BaseActivity.makeToast(aContext, R.string.no_associated_model, Toast.LENGTH_SHORT);
+          // Don't make a Toast notification about nothing to download
```
mcdurdin, line 151: Suggest we just delete the `else` block altogether -- we have commit history for the 'why'.

In fact, would be better to invert the test on line 134 to `continue`? Perhaps we should do a little cleanup? I would do something like this:

```java
  private void processCloudResults(Context aContext, List<MetaDataResult> aMetaDataResult) {
    for(MetaDataResult _r:aMetaDataResult) {
      if (_r.returnjson.target== CloudApiTypes.ApiTarget.Keyboard) {
        //handleKeyboardMetaData(_r);
      }
      if(_r.returnjson.target== CloudApiTypes.ApiTarget.KeyboardLexicalModels) {
        processCloudResultForModel(_r.returnjson.jsonArray);
      }
    }
  }

  private void processCloudResultForModel(JSONArray lmData) {
    if (lmData == null || lmData.length() == 0) {
      KMLog.LogError(TAG, "Error in lexical model metadata from api.keyman.com - zero or null");
      return;
    }

    try {
      JSONObject modelInfo = lmData.getJSONObject(0);
      if (!modelInfo.has("packageFilename") || !modelInfo.has("id")) {
        KMLog.LogError(TAG, "Error in lexical model metadata from api.keyman.com - missing metadata");
        return;
      }

      String _modelID = modelInfo.getString("id");
      ArrayList<CloudApiTypes.CloudApiParam> urls = new ArrayList<>();
      urls.add(new CloudApiTypes.CloudApiParam(
        CloudApiTypes.ApiTarget.LexicalModelPackage,
        modelInfo.getString("packageFilename")));
      _r.additionalDownloadid = CloudLexicalPackageDownloadCallback.createDownloadId(_modelID);
      _r.additionalDownloads= urls;
    } catch (JSONException e) {
      KMLog.LogException(TAG, "Error in lexical model metadata from api.keyman.com. ", e);
    }
  }
```

Not sure if LogError exists but if it doesn't it should. Logging something going wrong that we are catching to sentry -- we want to know if lm data from the server is bogus.

Ironically, I would suggest adding toasts to LogError and LogException if they are not there already, so the user knows something has gone wrong and can report the error to us (assuming they remember the message at all!)

android/KMEA/app/src/main/java/com/tavultesoft/kmea/data/CloudRepository.java
```diff
@@ -400,7 +400,7 @@ private void downloadMetaDataFromServer(@NonNull Context context, UpdateHandler
         BaseActivity.makeToast(context, R.string.catalog_download_is_running_in_background, Toast.LENGTH_SHORT);
       } else {
         updateIsRunning = true;
-        BaseActivity.makeToast(context, R.string.catalog_download_start_in_background, Toast.LENGTH_SHORT);
+        // Don't make a Toast notification about cloud update
```
mcdurdin, line 403:
```suggestion
```

No need for the comment 😉

mcdurdin (changes requested):
I mean, LGTM, but I would encourage a little bit of refactor cleanup in one block of code...

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudLexicalModelMetaDataDownloadCallback.java
darcywong00: > Ironically, I would suggest adding toasts to LogError and LogException if they are not there already, so the user knows something has gone wrong and can report the error to us (assuming they remember the message at all!)

I split that to a separate issue, since that may need more testing to see if adding Toasts to KMLog make the app too noisy. (Hopefully not)

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.55-alpha](https://keyman.com/downloads/releases/alpha/16.0.55)

PR #7179, opened 2022-08-31
chore(core): split out validators 🙀

### LINKED ISSUE #7136: feat(core): ldml: improve validator 🙀
- validate all section ranges
- validate metadata?
- more sample files, perhaps fuzzing

Furthering #7124 

#5015

srl295:
Fixes #7136

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -274,7 +274,7 @@ kmx_plus::kmx_plus(const COMP_KEYBOARD *keyboard)
   KMXPLUS_PRINTF(("kmx_plus(): KMXPlus offset 0x%X, KMXPlus size 0x%X\n", ex->kmxplus.dpKMXPlus, ex->kmxplus.dwKMXPlusSize));
   const uint8_t* rawdata = reinterpret_cast<const uint8_t*>(keyboard);
 
-  sect = section_from_bytes<COMP_KMXPLUS_SECT>(rawdata, ex->kmxplus.dwKMXPlusSize);
+  sect = section_from_bytes<COMP_KMXPLUS_SECT>(rawdata+ex->kmxplus.dpKMXPlus, ex->kmxplus.dwKMXPlusSize);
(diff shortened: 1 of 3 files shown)


srl295:
rebased

mcdurdin:
Note, perhaps use `Fixes #7136` or `Closes #7136` in your description so we get a 'fixing' relationship?

core/src/kmx/kmx_plus.cpp
```diff
@@ -46,33 +50,102 @@ validate_section_name(KMX_DWORD ident) {
...
-validate_kmxplus_keys(const uint8_t* /*data*/, const COMP_KMXPLUS_KEYS* keys) {
-  if(keys == nullptr) {
-    KMXPLUS_PRINTF(("! could not load 'keys' section\n"));
-    return false;
-  }
-  if(!validate_kmxplus_header((const COMP_KMXPLUS_HEADER*)keys)) {
+bool
+COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
+  KMXPLUS_PRINTF((" count: #0x%X\n", this->count));
+  if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
```
mcdurdin, line 142: Should this be a `!=`? Would there be a reason why `header.size` is greater than the sum of these components? Same applies in other `valid` functions :grin:

mcdurdin (approved):
LGTM. Let's have a discussion on forward binary compatibility at some point before we get too locked in.

core/src/kmx/kmx_plus.cpp
srl295: ooo yes.

srl295: forward compat would be the only reason for ≠

srl295: actually my hand assembled tiny would also have a ≠ because there's extra padding at the end of the strs.

PR #7184, opened 2022-09-01
feat(developer): use json schema to verify source file 🙀

### LINKED ISSUE #7181: chore(developer): kmldmlc - Ensure that DTD is used when verifying XML files 🙀
* Ensure that DTD is used when verifying XML files in kmldmlc

From #7177

mcdurdin:
Fixes #7181.

As we do not appear to have access to a good DTD validation library in Node.js, we will instead convert the DTD to a JSON schema file and use that for validation.

I have not verified every aspect of the JSON schema at this point, but will incrementally do so as I implement further aspects of the compiler.

This updates all the testing to use the schema. As the schema validation is slow, I have increased the 'normal' time for tests.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmldmlc/build.sh
+++ b/developer/src/kmldmlc/build.sh
@@ -46,6 +46,8 @@ fi
 
 if builder_has_action build; then
   npm run build
+  # We need the schema file at runtime and bundled
+  cp "$KEYMAN_ROOT/resources/standards-data/ldml-keyboards/techpreview/ldml-keyboard.schema.json" "$THIS_SCRIPT_PATH/build/src/"
   builder_report success build
 fi
 
--- a/developer/src/kmldmlc/test/ldml-keyboard/test-ldml-keyboard-xml-reader.ts
+++ b/developer/src/kmldmlc/test/ldml-keyboard/test-ldml-keyboard-xml-reader.ts
@@ -16,4 +16,15 @@ describe('ldml keyboard xml reader tests', function() {
     assert.equal(callbacks.messages.length, 1);
     assert.deepEqual(callbacks.messages[0], CompilerErrors.InvalidFile("data/keyboard must have required property 'names'"));
   });
+
+  it("should fail to load files with an invalid conformsTo", function() {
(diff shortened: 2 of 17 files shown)


developer/src/kmldmlc/test/ldml-keyboard/test-ldml-keyboard-xml-reader.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it("should fail to load files with an invalid conformsTo", function() {
+    const inputFilename = makePathToFixture('invalid-conforms-to.xml');
+    const callbacks = new CompilerCallbacks();
+    let reader = new LDMLKeyboardXMLSourceFileReader(callbacks);
+    const source = reader.loadFile(inputFilename);
+    assert.isNull(source);
+    assert.equal(callbacks.messages.length, 1);
+    assert.deepEqual(callbacks.messages[0], CompilerErrors.InvalidFile("data/keyboard/conformsTo must be equal to one of the allowed values"));
```
darcywong00, line 27: Is the validator able to determine what value was invalid?

mcdurdin: No, I don't think so -- but it's fairly clear what it should be changed to, so should be okay?

darcywong00 (approved):
lgtm

developer/src/kmldmlc/build.sh
```diff
@@ -46,6 +46,8 @@ fi
 
 if builder_has_action build; then
   npm run build
+  # We need the schema file at runtime and bundled
+  cp "$KEYMAN_ROOT/resources/standards-data/ldml-keyboards/techpreview/ldml-keyboard.schema.json" "$THIS_SCRIPT_PATH/build/src/"
```
mcdurdin, line 50: We should move this to the configure step and then use this file also in test/helpers/index.ts -- keeps things neat and tidy and avoids deeper cross-referencing. But will do this in a follow-up PR.

srl295:
https://unicode-org.atlassian.net/browse/CLDR-15990

PR #7186, opened 2022-09-02
feat(common): add size and offset calculators to hextobin 🙀

mcdurdin:
This makes it easier to write a binary file which contains offsets and size values. Using this for unit tests for the LDML keyboard compiler.

Example kmldmlc basic.txt is updated and included.

Future enhancement suggestion: include count() for repeated blocks and index() for block indices within these repeated blocks. Would be very useful for string tables in the example ref.

The real benefit can be seen in the follow-up PR #7187 with far fewer changes required to basic.txt fixture for an additional field.

@keymanapp-test-bot skip

### DIFF
--- a/common/tools/hextobin/package.json
+++ b/common/tools/hextobin/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "commander": "^3.0.0"
   },
-  "main": "build/index.js",
+  "main": "build/hextobin.js",
   "bin": {
     "hextobin": "build/hextobin.js"
   },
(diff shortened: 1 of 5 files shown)


common/tools/hextobin/package.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "commander": "^3.0.0"
   },
-  "main": "build/index.js",
+  "main": "build/hextobin.js",
```
mcdurdin, line 13:
```suggestion
  "main": "build/index.js",
```

This was a booboo

rc-swag (approved):
lgtm

srl295 (approved):

PR #7187, opened 2022-09-02
feat(core): add version field to meta 🙀

mcdurdin:
Also adds semver package for validating the version field.

@keymanapp-test-bot skip


common/tools/hextobin/package.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "commander": "^3.0.0"
   },
-  "main": "build/hextobin.js",
+  "main": "build/index.js",
```
mcdurdin, line 13: Mistake from previous PR...

core/src/kmx/kmx_plus.h
```diff
@@ -34,6 +34,8 @@ namespace kmx {
 #pragma warning ( disable : 4200 )
 #endif
 
+typedef KMX_DWORD KMXPLUS_STR;
```
mcdurdin, line 37: Has no real meaning but helps to clarify that these are offsets to strings rather than plain DWORDs 😁

rc-swag (approved):
LGTM

core/src/kmx/kmx_plus.h
srl295, line 37:
```suggestion
/**
 * Indicates an offset into the strs table (0 = zero length)
 */
typedef KMX_DWORD KMXPLUS_STR;
```

srl295 (approved):

PR #7190, opened 2022-09-02
fix(developer): compiler mismatch on currentLine

### LINKED ISSUE #7122: bug(developer): Debugger doesn't step through normalise rules in Khmer Angkor keyboard
When using the debugger it's possible to step through the rules after a keystroke and see which one applies. However when the processing gets to the normalise group the location shows up in line 227 between a comment and a rule in the main group instead of showing the normalise rule that gets applied.

- Keyman Developer 15.0.268
- Khmer Angkor keyboard from [keyboards repo](https://github.com/keymanapp/keyboards/tree/master/release/k/khmer_angkor)

mcdurdin:
Fixes #7122.

The compiler was updating currentLine when checking for unreachable rules, but not restoring it afterwards. This led to mismatches in the debugger and in compiler warnings.

# User Testing

TEST_SINGLE_STEP: Verify that single step in the debugger no longer ends up on the wrong line

1. Load khmer_angkor.kmn in Keyman Developer, and start debugging (<kbd>F5</kbd>).
2. Type <kbd>x</kbd><kbd>E</kbd><kbd>j</kbd>
3. Turn on the Single Step mode from Debug menu.
4. Type <kbd>m</kbd>, and step through with Debug/Step Forward.
5. Step forward through to the `normalise` group, and then verify that the next step shows the correct rule that is applied, not a blank line.

### DIFF
--- a/developer/src/kmcmpdll/DeprecationChecks.cpp
+++ b/developer/src/kmcmpdll/DeprecationChecks.cpp
@@ -25,7 +25,7 @@ BOOL CheckForDeprecatedFeatures(PFILE_KEYBOARD fk) {
       // Keyman 7
       #define TSS_WINDOWSLANGUAGES 29
   */
-  int currentLineBackup = currentLine;
+  int oldCurrentLine = currentLine;
   DWORD i;
   PFILE_STORE sp;
 
@@ -46,7 +46,7 @@ BOOL CheckForDeprecatedFeatures(PFILE_KEYBOARD fk) {
     }
   }
 
-  currentLine = currentLineBackup;
+  currentLine = oldCurrentLine;
 
   return TRUE;
 }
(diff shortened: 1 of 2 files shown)


mcdurdin:
@SabineSIL we will need to cherry-pick this fix into your kmcompx branch. Let me know if you would like assistance with that :grin:.

bharanidharanj:
* **TEST_SINGLE_STEP (PASSED):** Tested this with Keyman Developer 16.0.55-alpha-test-7190 build as per the instructions. Verified that the next step shows a rule and not showing any blank line in the window.

developer/src/kmcmpdll/DeprecationChecks.cpp
```diff
@@ -25,7 +25,7 @@ BOOL CheckForDeprecatedFeatures(PFILE_KEYBOARD fk) {
       // Keyman 7
       #define TSS_WINDOWSLANGUAGES 29
   */
-  int currentLineBackup = currentLine;
+  int oldCurrentLine = currentLine;
```
mcdurdin, line 28: Just making the backup variable name consistent -- same pattern used in three places. The currentLine change is made for error reporting.

rc-swag (approved):
lgtm

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.61-alpha](https://keyman.com/downloads/releases/alpha/16.0.61)

PR #7192, opened 2022-09-02
chore(linux): Fix ibus-keyman.postinst script

ermshiperete:
When installing ibus-keyman it could happen that we got an (ignored) error "user name does not exist". This change tries to work around this problem.

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/ibus-keyman.postinst
+++ b/linux/debian/ibus-keyman.postinst
@@ -37,8 +37,8 @@ case "$1" in
       fi
 
       # Verify that it's running now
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.56-alpha](https://keyman.com/downloads/releases/alpha/16.0.56)

PR #7194, opened 2022-09-02
feat(common): hextobin `index` command 🙀

mcdurdin:
Adds an `index` command as suggested by #7186.

> Future enhancement suggestion: include ... index() for block indices within these repeated blocks. Would be very useful for string tables in the example ref.

Turns out the also suggested `count()` can be done with `index()` so didn't add that.

Uses `index()` for string references and count of strings in basic.txt.

Reformat basic.txt to make it more readable as it will get less 'basic' over time!

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmldmlc/test/fixtures/basic.txt
+++ b/developer/src/kmldmlc/test/fixtures/basic.txt
@@ -48,30 +48,34 @@ block(kmxplusinfo)                  #  struct COMP_KEYBOARD_KMXPLUSINFO {
   diff(sect,eof)                    #    KMX_DWORD dwKMXPlusSize;  // 0044 size in bytes of entire KMXPlus data
                                     #  };
 
-block(sect)        #  struct COMP_KMXPLUS_SECT {
-  73 65 63 74      #    KMX_DWORD header.ident;   // 0000 Section name
-  sizeof(sect)     #    KMX_DWORD header.size;    // 0004 Section length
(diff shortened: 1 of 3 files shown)


developer/src/kmldmlc/test/fixtures/basic.txt
```diff
@@ -48,30 +48,34 @@ block(kmxplusinfo)                  #  struct COMP_KEYBOARD_KMXPLUSINFO {
...
 
-block(strs)          #  struct COMP_KMXPLUS_STRS {
-  73 74 72 73        #    KMX_DWORD header.ident;   // 0000 Section name - strs
-  diff(strs,endstrs) #    KMX_DWORD header.size;    // 0004 Section length
-  0B 00 00 00        #    KMX_DWORD count;          // 0008 count of str entries
-  00 00 00 00        #    KMX_DWORD reserved;       // 000C padding
+block(strs)                         #  struct COMP_KMXPLUS_STRS {
+  73 74 72 73                       #    KMX_DWORD header.ident;   // 0000 Section name - strs
+  diff(strs,endstrs)                #    KMX_DWORD header.size;    // 0004 Section length
+  index(strNull,endstrs,2)          #    KMX_DWORD count;          // 0008 count of str entries
```
srl295, line 76: nice

srl295 (approved):
LGTM

PR #7196, opened 2022-09-03
feat(core): add 'name' section 🙀

mcdurdin:
Updates compiler, headers and documentation for the 'name' section, which will always be present as the LDML keyboard spec requires at least one name in the `<names>` element.

@keymanapp-test-bot skip

### DIFF
--- a/core/src/ldml/C7043_ldml.md
+++ b/core/src/ldml/C7043_ldml.md
@@ -196,6 +196,31 @@ For each key:
 - `vkey`: Is the standard vkey, 0-255
 - `target`: Is the target (resolved) vkey, 0-255.
 
-### C7043.2.7 Transforms and friends
+### C7043.2.7 `name`—Names
+
+Defines the names of the keyboard as found in the source `<names>` element.
+While this section is optional in the binary format, in practice it will always
+be present, as the source format requires at least one name.
+
+| ∆ | Bits | Name    | Description                              |
+|---|------|---------|------------------------------------------|
+| 0 |  32  | ident   | `name`                                   |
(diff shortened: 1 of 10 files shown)


developer/src/kmcmpdll/kcframe/kcframe.vcxproj
```diff
@@ -114,7 +114,7 @@
     <ClCompile>
       <Optimization>MaxSpeed</Optimization>
       <InlineFunctionExpansion>OnlyExplicitInline</InlineFunctionExpansion>
-      <AdditionalIncludeDirectories>.;..\..\..\global\inc;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
mcdurdin, line 117: These changes just so we can build the C++ host app 'kcframe' for kmcmpdll (used for debug currently) after we rearranged developer source files for 16.0.

core/src/ldml/C7043_ldml.md
```diff
@@ -196,6 +196,31 @@ For each key:
...
+
+For each name in `count`:
+
+| ∆ | Bits | Name    | Description                              |
+|---|------|---------|------------------------------------------|
+|16+|  32  | name    | str: A name for the keyboard             |
+
+Note that the first name is repeated in the `meta` section. The remaining names
+are stored in source file order, and the semantic meaning of each name is not
+defined here.
```
srl295, line 222: ok, here it is.  Should we remove that name from `meta` Yes, breakage, but…

mcdurdin: * #7214

PR #7199, opened 2022-09-04
refactor(developer): kmldmlc - rename `CompilerErrors` to `CompilerMessages` 🙀

mcdurdin:
Just wanted to rename this to make it clearer. Also added a prefix to each message generator to indicate the type of message, as otherwise it is not obvious within the compiler which messages are going to cause the compiler to fail.

Note that the message generator function prefixes are Title Cased whereas the message identifier constant prefixes are CAPS. I guess this could be confusing but I think it works okay in practice.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #7202, opened 2022-09-05
refactor(developer): cleanup virtual key constants 🙀

mcdurdin:
Uses our `K_xxx` constants for the CLDR VKey Enum mapping. Renames us-virtual-keys to virtual-key-constants, better reflecting its purpose, and adds a unit test to verify that the codes in the CLDR VKey Enum mapping correspond to our expected Keyman virtual key codes.

@keymanapp-test-bot skip


developer/src/kmldmlc/src/keyman/ldml-keyboard/virtual-key-constants.ts
```diff
@@ -0,0 +1,197 @@
...
+  K_SLASH:191,
+  K_BKQUOTE:192,
+  K_LBRKT:219,
+  K_BKSLASH:220,
+  K_RBRKT:221,
+  K_QUOTE:222,
+  K_oE2:226,  // ISO B00, key to right of left shift, not on US keyboard
+  K_OE2:226,
+  k_oC1:193,  // ISO B11, ABNT-2 key to left of right shift, not on US keyboard
+  k_OC1:193,
```
mcdurdin, line 115: `k_oC1` is added in this PR

darcywong00: Does this also get added to common/web/keyboard-processor/src/text/codes.ts as a TODO?

srl295 (approved):
LGTM.  Should this be generated from the .h?

mcdurdin:
> LGTM. Should this be generated from the .h?

Nah, these constants are pretty much eternal now. We've got a unit test to verify the CLDR VKey Enum values against Keyman values and I'm happy with it as is. (We do have a separate task to consolidate some of these constants ... which is mostly configuration pain)

developer/src/kmldmlc/src/keyman/ldml-keyboard/virtual-key-constants.ts
mcdurdin: I think we can leave that to the consolidation TODO which is already in the file :grin:

PR #7203, opened 2022-09-05
chore(developer): add test for compiler messages and cleanup e2e test 🙀

mcdurdin:
The purpose of these tests is to ensure that our slightly WET constant definitions for compiler errors are consistent in their usage:

  * that names are consistent between the message functions and constants
  * that constants have unique codes
  * that constants have the right error mask for their type

This means we don't have to worry about DRYing out the constants, which would probably require us to have a preprocessing step, which has its own negatives.

Also, added more reporting to e2e test in case of failures.

@keymanapp-test-bot skip


srl295 (approved):
nice.  Do we track code coverage?

mcdurdin:
> nice. Do we track code coverage?

We do now :grin: #7213

PR #7204, opened 2022-09-05
feat(core): ldml keyboard transform elements 🙀

mcdurdin:
In draft for now until I confirm the new spec section layouts as I work on the compiler.

Adds definitions for LDML keyboard transform elements `transforms`, `backspaces`, and `reorders`, and the binary format sections `bksp`, `elem`, `finl`, `ordr`, and `tran`.

Minimally updates the compiler so that it will build, but no other changes to the compiler.

Compiler and Typescript updates to support building these sections will come in the next PR.

@keymanapp-test-bot skip


srl295 (approved):
LGTM … seems plausible.  I have not verified the entire semantics against the TR.

PR #7206, opened 2022-09-05
feat(windows): configuration UI polish

### LINKED ISSUE #3354: bug(windows): Can't remove language from keyboard if text overlaps X
The Keyman Desktop Configuration dialog shows a X next to a language for a keyboard that allows to remove the language. However, when the window is too small the X shows up behind the text, making it impossible to click.

image

---

**Keyman Desktop**
 - OS: Windows 10
 - Keyman version: 14.0.81

### LINKED ISSUE #6296: feat(windows): Keyman Configuration UI Polish
This is the specification for the UI polish to the Keyman Configuration in Windows.
Created from [16.0 Planning Document ](url)https://docs.google.com/document/d/1_mUBC7KEzIMYTI_Etqeddyx037XMJynN9lFZai2hP5M/edit#

KeymanConfigurationMockUp

# Introduction

The first section of this spec lists the changes required identified in the meeting. The second section  [Overall Requirements for Configuration](#overall-requirements-for-configuration-ui) that need to be referred to, making sure the changes still adhere to the requirements.

Refer to the mockup picture drawn on the whiteboard during the meeting. 
Goal: Reduce busyness on the screen, make the controls more consistent
## Changes 
#7206  Makes the following changes. 

- [x] - Remove checkboxes for enabling the keyboards. This will be moved to a button row within the drop-down for each keyboard.
- [x] - Remove the 3 controls on the right “?X^”.  
- [x] - Add a toggle for collapsing/showing information extended information
- [x]    - The drop-down shall show: keyboard version, Languages , ~~QR code~~, `down arrow` to show all the rest of the information (package info, copyright etc. Finally a row containg buttons `help` `[options]` `disable/enable` `uninstall`   `share` which is always visible when a keyboard is selected in the main list.
- [x] remove `keep in touch` tab and provide link under the `Support` tab. First do -  #7527
- [x] - The add language for the keyboard is a button that should rarely be used - make it an advanced function. add/remove button with a pop-up window.
- [ ] - Ensure mouse over is showing correct information.
- [x] - The right hand side column shall show the keyboard hotkey shortcut to select the keyboard - again see white board image.
- [x] -  If there is more than one language assigned to the keyboard then don't show the hotkey in the main list but only in the drop-down. As the hotkey is per-language. If there is only one language do the reverse only show it in the main keyboard layout list and not in drop-down list. 
- [x] - Only one keyboard can be expanded at a time. Therefore selecting a different keyboard row in the list will collapse the currently expanded keyboard as well as expand the newly selected keyboard.
- [x] - Instead of (none) when a keyboard has no hotkey, should show (no hotkey)  Remove the blue hyperlink and underline and show this on hover.
- [x] - Up and down arrow keys expand the list.
- [x] - Fix the selecting of list items when a new tab is selected for the Keyboard Layout, Options, and Hotkeys 
- [x] - assign left/right arrow key to the more details in the keyboard detail view.
- [x] Red Text in Languages if no languages have been added
- [x] Adjust spacing for the keyboard layout buttons to insure the overall flex box padding matches the gap between the buttons.
- [x] Remove blue hover from languages
- [x] Add selection colour for keyboard layout that has focus
- [x] Change Hover colour for buttons
- [x] Tweaks to the appearance of the Add/remove languages pop-up include a close button. 
- [x] Created to issue #7523
- [x] #7536

Note: Issue #3354 There is a mouse over for the current language and you can click an “x” to remove the language. will be addressed by this work. as it will remove the mouse over option `x`

## Overall Requirements for Configuration UI
This is a summary points for all of the Configuration wish items. Online updates for example will be done in a different PR.
- Display information about installed keyboards, install location (whether it’s user-specific or system-wide available)
- Online update check - indicator if new version is available
- Allows to install and remove keyboards. Installation from keyman.com or from local package file.
- Allows to add/remove language for keyboard
- Allows to update keyboard to newest version
- Allows to check for new versions and reload the list of keyboards
- Display detailed info about a keyboard. Show QR code for sharing keyboard.
- Display help about a specific keyboard
- Change options for specific keyboard
- Allow to select UI language
- Change options for Keyman in general
- Show Keyman version and links to help pages, community site etc
- Allows to see and define Hotkeys ? (defining hotkeys might depend on the OS. On Linux switching keyboards etc can be configured in the system; retrieving defined hotkeys from the system for display might be not so easy to do)

rc-swag:
Fixes #6296  See this issue for the main requirements for this PR.
Fixes #3354 

---

### Key Navigation
**Page Up/Down**  on the main menu toggles the Keyboard Layout, Keyboard Options, Hotkeys etc. It also focus the first item in each of the tabs. In the case of Keyboard layouts it will focus the last expanded keyboard layout, else the first keyboard layout.

**Up/Down Arrow** On the keyboard layout the up and down arrow will cycle through the keyboard layouts and expand the currently selected layout. 

**Left/Right**
When a Keyboard Layout is expanded pressing right arrow will show the further detail. Pressing left arrow will hide the extra detail

---

### The following TODO:

- [x]  All the additions to strings.xml will need to be localised
- [x] With Keyboard hotkeys if the keyboard_id is not defined in `//KeymanLanguage` in the XML file then we can not assign a hotkey. Further investigation I have worked out that `TKeymanLanguage.Serialize` in `keymanlanguage.pas` needs to be called when a language is uninstalled or installed. ~~It isn’t so then we do not get the <KeymanLanguage> tag update in the xml.~~ Closing and reopening configuration the xml is updated. It turns out it the DoRefresh for keyman language is using the TSF which is a cached value and not updated when this is called. The DoRefersh for the installed keyboards was accessing the Windows registry directly.  There is enough work for this to be a new  Issue  #7523

- [x]  Hotkey link doesn't display on some occasions. For hotkeys the bcp47code under the xml for `KeyamnKeyboardLanguageInstalled` does not match the bcp47code under `KeymanLanguage` (for the same keyboard) this means the hotkey can not be matched correctly. 

- [x]  Remove the hotkey configuration from the add/remove language popup, there are just too many places to configure it more of a confusion for users

- [x] ~~Turning on animations for the detailed lists breaks the qr and add/remove language popups~~ - it is a shame because the animations add made it easy for the eye to see what was expanding.
- [x]  Many test cases for every button, for hotkeys, for key and mouse navigation. 

# User Testing

If possible run the tests in the order they appear here as some tests rely on steps from the previous tests.

**TEST_GENERAL_LAYOUT**

This first test is general overview of the changes to the Configuration. 
  
- Open `Keyman Configuration`
- Using the `Download Keyboard` Button install the following keyboards: IPA (SIL), EuroLatin (SIL), Khmer Angkor and Yolngu
- Select the `Keyboard Layouts` tab  
- **Expected Result**
- All keyboards are listed
    
**TEST_DISABLE_BTN**
- Select the EuroLatin Keyboard Layout 
- Select the `Disable` button
- **Expected Result**
- The whole keyboard detail opacity is reduced to 60%
- The Keyboard title is greyed out
- Close the Configuration
- Click on the Keyman icon in the system tray and observe the disabled layout is no longer available

**TEST_ENABLE_BTN**
- Select the EuroLatin Keyboard Layout 
- Select the `Enable` button
- **Expected Result**
- The `Enable` button returns to light orange while the mouse hovers over it and grey when the mouse moves away.
- The Keyboard title is back to the default colour, no longer greyed out
- Close the Configuration
- Click on the Keyman icon in the system tray and observe the layout available
  
**TEST_KEYBOARD_OPTIONS_BTN**
- Install IPA (SIL) - if not already installed 
- Select the IPA (SIL) keyboard in the Keyboard layouts tab
- Select the `Keyboard Options` Button
- **Expected Result**
- IPA (SIL) Keyboard Options dialogue pops up
- Close the dialogue

**TEST_HELP_BTN**
- Install Hieroglyphic - if not already installed
- Select the Hieroglyphic keyboard in the Keyboard layouts tab.
- Select the `Help` Button
- **Expected Result**
- The help window shall open
  
**TEST_UNINSTALL_BTN**
- Install Hieroglyphic - if not already installed
- Select the Hieroglyphic keyboard in the Keyboard layouts tab.
- Select the `Uninstall` Button  
- **Expected Result**
-  Hieroglyphic Keyboard is uninstalled.
    
**TEST_ADDREMOVE_BTN**
- Select the EuroLatin Keyboard Layout 
- Press `Add/remove..` button
- **Expected Result**
- Add remove language dialogue appears
- Press the `Close` button and the dialogue closes.

**TEST_SHARE_KBD_BTN**
- Press the` Share keyboard` button
- **Expected Result**
- QR Code dialogue pops up
- Clicking outside the dialogue, on the greyed configuration area closes the dialogue

**TEST_ARROW_DOWN_BTN**
  - Select the  `Arrow Down` button, or press the right arrow key
  - **Expected Result**
- More detail is shown 
- Filename, Encodings, Layout Type, On-Screen Keyboard, Documentation
- Arrow changes to an up arrow. 
- Can continue to TEST_ARROW_UP_BTN

**TEST_ARROW_UP_BTN**
- First, follow the steps from TEST_ARROW_DOWN_BTN
- Select the `Arrow Up` button, or press the left arrow key
- **Expected Result**
- More detail is hidden 
- Arrow changes to a down arrow. 

**TEST_UP_DOWN_ARROW_KEYs**
- Open the configuration Keyboard Layout tab
- Use the `down arrow` to cycle through all the keyboard layouts 
- Repeat using the `up arrow key` 
- **Expected Result**
- As each layout is selected the summary detail for that layout with the buttons should be shown
- Only one layout can show the summary at a time so as you move down the list the previously selected Keyboard layout summary should be closed
- When the last keyboard layout is in the list pressing the down arrow should return to the start of the list. i.e. it is circular navigation.

 
**TEST_SET_HOTKEY_SINGLE_LANGUAGE**

Set the keyboard layout hotkey for a single language.
- Open the configuration and select the `Keyboard Layouts` tab
- Select the IPA SIL Keyboard - 
- On the left-hand side of the keyboard layout title click the `No Hotkey` text
- The Change hotkey dialogue box should appear.
- Set a hotkey and close the dialogue box. e.g. `Ctrl + 1`
- **Expected Result**
- The title should now have displayed the hotkey for this keyboard - `Ctrl + 1`

  
**TEST_ADD_MULTIPLE_LANGUAGE**  

- Open the configuration and select the `Keyboard Layouts` tab
- Install the EuroLatin (SIL) Keyboard - if not already installed
- Initially when there is only one language associated with the keyboard the `no hotkey` will appear on the same row as the keyboard layout name 
- Select the `Add/remove language..` button 
- From the `Add/remove language` dialogue select the add language button 
- Add 'fr' French
- **Expected Result**
- The `add/remove language` dialogue box now includes French
- Select the `Close` button
- The language list under the keyboard layout name will also include French
- [ It is a known issue that the hotkey to the right of the language will not show until the configuration is open and closed. ] Close the Keyman Configuration
- Open Keyman Configuration
- Select the Keyboard Layouts Tab
- Select the EuroLatin SIL Keyboard Layout
- Observe that there is no longer a `hotkey label` on the Keyboard Layout Title
- Observe that the summary view shows the hotkey label to the right of each language.

**TEST_REMOVE_LANGUAGE_MULTIPLE_LANGUAGE**  
- Follow the steps in TEST_ADD_MULTIPLE_LANGUAGE
- Ensure the EuroLatin (SIL) Keyboard layout is selected and that there is more than one language in the summary.
- Select the `Add/remove language` button
- Press the cross `x` beside the `French` 
- **Expected Result**
- The dialogue should be updated with the `French` now removed. Close the dialogue
- Note the `French` is also no longer showing in the summary.  
  - **Continuing**
- Select the `Add/remove language` button again
- Remove any languages still remaining using the cross `x` button.
- The dialogue box should now be empty. Select the close button 
- **Expected Result**
- The word `Languages:` is now the colour red.


windows/src/desktop/kmshell/xml/config.css
```diff
@@ -242,78 +287,83 @@ html
...
 {
-  float: right;
-  cursor: default;
-  background: url('help20.png') 0 0 no-repeat;
-  width: 20px;
-  height: 20px;
-  margin: 2px 4px 0 0;
+  display: block;
+ /* animation: growDown 500ms ease-in-out forwards;
+  transform-origin: top center;*/
```
rc-swag, line 358: If you uncomment these lines.  You well see nice animation on the keyboard layout expanding. However it then breaks the pop-up which use postion: fixed.

mcdurdin: I'm not all that keen on animations -- they slow down the user experience :grin:

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -130,15 +132,29 @@ document.addEventListener("DOMContentLoaded", windowResize);
...
-      var k = document.getElementById('list_detail_'+n);
-      if( !k ) break;
-      if(k.style.display == 'block') list_detail(event,n);
+      {
+        let keyboard_name = n.substring(9); // remove keyboard_ from element name
+        let kbd_grid = document.getElementById('keyboard_grid_'+keyboard_name);
+        if( !kbd_grid ) break;
+        let kbd_list = document.getElementById('list_'+n);
+        if( !kbd_list ) break;
+        if(!kbd_list.classList.contains('expanded')) break;
```
mcdurdin, line 141: Just wondering why these checks are necessary?

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -130,15 +132,29 @@ document.addEventListener("DOMContentLoaded", windowResize);
...
-      var k = document.getElementById('list_detail_'+n);
-      if( !k ) break;
-      if(!k.style.display || k.style.display == 'none') list_detail(event,n);
+      {
+        let keyboard_name = n.substring(9); // remove keyboard_ from element name
+        let kbd_grid = document.getElementById('keyboard_grid_'+keyboard_name);
+        if( !kbd_grid ) break;
+        let kbd_list = document.getElementById('list_'+n);
+        if( !kbd_list ) break;
+        if(!kbd_list.classList.contains('expanded')) break;
```
mcdurdin, line 154: ditto

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -130,15 +132,29 @@ document.addEventListener("DOMContentLoaded", windowResize);
...
+        ///  show more details for the expanded list
+        if (!$(kbd_grid).hasClass('grid_rows_hide')) more_detail(kbd_grid.id);
```
mcdurdin, line 156:
```suggestion
        if (!$(kbd_grid).hasClass('grid_rows_hide')) {
          more_detail(kbd_grid.id);
        }
```

Per style guide, trying to avoid one-liner ifs.

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -194,12 +210,37 @@ document.addEventListener("DOMContentLoaded", windowResize);
...
     save_state();
     return false;
   }
 
+  function list_expanded_selected (n){
+    if(loading_state) return;
+    var keyboards = $('div.list_item');
+    keyboards.each(function(index) {
+      if(this.id == n.id) {
+        $(this).toggleClass('expanded');
```
mcdurdin, line 224:
```suggestion
        $(this).addClass('expanded');
```

Is that safer than a toggle?

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -194,12 +210,37 @@ document.addEventListener("DOMContentLoaded", windowResize);
...
+    var k = document.getElementById(n);
+    var m = document.getElementById(n+'_more');
+    $(k).toggleClass('grid_rows_hide');
+    if ($(k).hasClass('grid_rows_hide')){
+      $(m).attr("src","/app/expand20.png");
+    }
+    else {
+      $(m).attr("src","/app/collapse20.png");
+    }
+  }
```
mcdurdin, line 242: 1. Use consts where possible:

```
    const k = document.getElementById(n);
    const m = document.getElementById(n+'_more');
```

2. Can we rename the variables so something meaningful, particularly the input param?

mcdurdin (approved):
Code changes look generally fine. xsl is never going to be elegant to read...

I would like to do a usability walkthrough with you when we have time, before merge. I see you have not yet defined user tests, also.

😸

windows/src/desktop/kmshell/xml/config.css
```diff
@@ -43,7 +44,8 @@
   bottom: 0;
   width: 100%;
   overflow: hidden;
-  background: #79C3DA;
+  /*TODO background: #79C3DA;*/
```
ermshiperete, line 47: Is this temporary code that you want to put back before merging? Or why different background colors from before?

windows/src/desktop/kmshell/xml/menu-frame.js
```diff
@@ -28,10 +28,17 @@
     save_state();
 
     if(!loading_state) {
-      var q = document.getElementById('content_'+c.menu_name).getElementsByTagName('span');
+      let q = null;
+      if (c.menu_name.includes("keyboardlist")) {
+        q = document.getElementById('content_'+c.menu_name).getElementsByClassName('list_item expanded');
+      }
+      // if none are expanded select the first item
```
ermshiperete, line 35: The comment is a bit misleading - we're potentially getting more than one item. Maybe this comment should be moved to just before the `for` loop and changed to `// select the first list item` ?

windows/src/desktop/kmshell/xml/menu-frame.js
```diff
@@ -28,10 +28,17 @@
...
+      if (q == null || q.length == 0) {
+        q = document.getElementById('content_'+c.menu_name).getElementsByClassName('list_item');
+      }
       for(var i = 0; i < q.length; i++) {
-        blah = q[i].id;
-        itemtype = blah.substr(0, 5);
+        list_array = q[i].id;
+        itemtype = list_array.substr(0, 5);
 
         if( itemtype == "list_" ) {
```
ermshiperete, line 43: Do we still need this check now that we get `list_item` items?

I guess we'd still need it if we could have items with class `list_item` that have an `id` not starting with `list_`.

windows/src/desktop/kmshell/xml/strings.xml
```diff
@@ -463,7 +482,7 @@
   <!-- Context: Configuration Dialog - Hotkeys tab -->
   <!-- String Type: PlainText -->
   <!-- Introduced: 7.0.230.0 -->
-  <string name="S_Hotkey_None" comment="Hotkey - text used when no hotkey set for an option">(none)</string>
+  <string name="S_Hotkey_None" comment="Hotkey - text used when no hotkey set for an option">(no hotkey)</string>
```
ermshiperete, line 485:
```suggestion
  <string name="S_Hotkey_None" comment="Hotkey - text used when no hotkey set">(no hotkey)</string>
```

I find mentioning of an option a bit confusing here - makes me wonder what other hotkeys there are. Since I assume this string is used for all hotkeys I think we can omit the option.

windows/src/desktop/kmshell/xml/config.css
rc-swag: When reviewing it was suggested to remove the blue background. In doing that it looked a bit too washed out. I then tried the grey defined in the `Keyman Design Approach Manual`. The reason I left the comment there was because I was still unsure which way to go. The grey looks ok in my opinion a bit dull but closer the Windows OS feel. Shawn had commented that UI which looks vastly different from the native feel can give the user are feeling of less trust in the application.  What are your thoughts?


blue
default
grey

windows/src/desktop/kmshell/xml/config.css
rc-swag: Ha true can be or flair and no function,  this wasn't that slow an animation it sort of made the flow less harsh and focused the eyesto the correct spot. However, it isn't worth the effort. If it breaks the pop-up

windows/src/desktop/kmshell/xml/config.js
rc-swag: I will but the kbd_grid id lower so it is checked just before use. 
First it checks to see if it is a keyboard list (item) and if it is expanded. Without the check you can in through select a keyboard with the mouse, press left arrow, set focus to another keyboard on the list then when, again focusing and expanding the original keyboard the more detail is already show even though the list has never been expanded.
Then show/hide more detail -  `more_detail` blindly does a toggle hence the need to check if it is already in the desired state.

windows/src/desktop/kmshell/xml/config.js
rc-swag: The desired behaviour here is to toggle the class for the matching id (not add) and remove it for all other ids. At most one keyboard item is to be expanded at a time.

windows/src/desktop/kmshell/xml/config.js
rc-swag: Yep good plan. I will go through and make these changes.

windows/src/desktop/kmshell/xml/menu-frame.js
rc-swag: I think I will remove it as it is no longer really needed with the class name look up and the xsl should make sure each id starts with a list_

ermshiperete (approved):

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -303,6 +356,28 @@ function keyboard_checkclick(n,toggle) {
...
 
+/* Disable/Enable Button replaces checkbox for now still use the "check" function in delphi webserver */
+function keyboard_toggle(n) {
+  var k = document.getElementById('keyboardcheck_'+n), li = document.getElementById('list_keyboard_'+n);
+  var btn = document.getElementById('button_'+n);
+
+  k.checked = !k.checked;
+  if (k.checked){
+    btn.value = "Disable";
+    btn.style.background = '#dadddc';
```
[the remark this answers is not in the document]
mcdurdin: Best way is to have two elements in the XML and toggle their visibility. Then you can avoid defining styles in the .js as well.

windows/src/desktop/kmshell/xml/config.css
mcdurdin: My pref would be either white bg or blue bg. Buttons look a little too 'disabled' however. May need to rethink the colour of those a little?

mcdurdin: We have trouble with matching native UI -- it changes so much that we can't keep up. So I think we don't try and keep up. AFAICT, most modern apps have their own UI themes anyway so we aren't losing much...

windows/src/desktop/kmshell/xml/menu-frame.js
```diff
@@ -28,15 +28,20 @@
...
+      let q = null;
+      if (c.menu_name.includes("keyboardlist")) {
+        q = document.getElementById('content_'+c.menu_name).getElementsByClassName('list_item expanded');
+      }
+      // if none are expanded get the full list
+      if (q == null || q.length == 0) {
+        q = document.getElementById('content_'+c.menu_name).getElementsByClassName('list_item');
+      }
+      // select the first list item
       for(var i = 0; i < q.length; i++) {
```
ermshiperete, line 40: Why do we need a `for` loop if we always break after the first item?

ermshiperete (approved):

windows/src/desktop/kmshell/xml/strings.xml
```diff
@@ -225,6 +239,11 @@
   <!-- Introduced: 9.0.420.0 -->
   <string name="S_Languages_Install" comment="Add language profile">Add language</string>
 
+  <!-- Context: Configuration Dialog - Captions -->
+  <!-- String Type: PlainText -->
+  <!-- Introduced: 16.0.47 -->
+  <string name="S_Languages_Modify" comment="Add or Remove languages for keyboard layout">Add/remove language</string>
```
ermshiperete, line 245: Is this a dialog title? Then it would be helpful to add that to the comment (assuming the comment is what's visible on Crowdin), e.g.

```suggestion
  <string name="S_Languages_Modify" comment="Title for dialog to add or remove languages for keyboard layout">Add/remove language</string>

```

mcdurdin: Believe this is the button label?

rc-swag: It is used for the button and the Title of the dialog the button opens. Yes, localisation context matters. Should I create one for the button and one for the Title? Note I also am adding the Ellipsis by concatenation. I am wondering if the translators would prefer to have that information also?

ermshiperete: Wondering - does adding the ellipsis by concatenation work for RTL languages, or do they need an extra whatever added so that it doesn't show "in front of" the text? If that's the case then it would make sense to have two separate strings, otherwise we can go with one, but I'd add the info that it's used for both button and title in the comment.

mcdurdin: The ellipsis is part of the translatable string and should not be concatenated. There should be two strings -- one for the button, and one for the title, because they may well translate differently in some languages.

rc-swag:
> We have trouble with matching native UI -- it changes so much that we can't keep up. So I think we don't try and keep up. AFAICT, most modern apps have their own UI themes anyway so we aren't losing much... 

I have stuck with the Blue then

ermshiperete (approved):

windows/src/desktop/kmshell/xml/config.css
rc-swag: Sticking with the blue

windows/src/desktop/kmshell/xml/config.js
```diff
@@ -297,6 +306,12 @@ document.addEventListener("DOMContentLoaded", windowResize);
             $('div.list_item[data-name="'+state.keyboards[i].name+'"]').addClass('expanded');
         }
       }
+      if(state.modify_languages) {
+        for(var i = 0; i < state.modify_languages.length; i++) {
+          if(state.modify_languages[i].modify)
+            $('div.modify[data-name="'+state.modify_languages[i].name+'"]').addClass('modify-visible');
```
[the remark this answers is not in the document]
rc-swag: No the xsl will not set the modify-visible class for any of the elements. It is added and removed by hideModifyLink/showModifyLink.   So when loading we only have set it based on the json.

rc-swag:
Commit 53ae7948772b6a79d91071a706477a23054ecf3c changes the class names to match the naming convention.

mcdurdin:
I've run through and tested this. I encountered no breaking issues. The new style is very clean; I like it. Some of the things I've noted here are probably not new, but just noting them here for completeness.

## Keyboard Layouts list

1. Clicking on `(no hotkey)` should not expand/contract the details -- need to cancel bubble on the click
2. `(no hotkey)` text should have margin-right about `8px`, not `21px`.
3. Don't need both Keyboard version and Package version above the fold -- Keyboard version is usually redundant, so put it below the fold.
4. Order of actions for keyboard, I think should be:
   * Keyboard options...
   * Add/remove language...
   * Share keyboard
   * Help
   * Disable
   * Uninstall
5. Enable button in keyboard details is red? Don't understand why
6. L/R Arrows and +/- seem backwards -- +/right should expand details, -/left should collapse
7. Tweak, optional: if Package name is same as keyboard name, perhaps move it below the fold.
8. If window is sized small horizontally, hotkey label overlaps borders:
   image

## Add/remove language popup dialog

1. Esc needs to cancel the popup
2. Space needed between items on the list
3. **Add language** and **Close** buttons should be bottom right like other dialogs, and styling should be more like Windows dialog, e.g.:
   image

## Share keyboard popup dialog

1. Esc needs to cancel the popup
2. Styling should also be more like Windows dialog

## Options tab

1. Keyman System Settings and Base Keyboard buttons don't highlight like other buttons (these are 'elevated' buttons
3. Elevated buttons have an obsolete icon, should be: 
   image
4. Border on selected elevated buttons is wonky
   image

## General notes

1. A few possible focus tweaks may be possible in another iteration - particularly around arrow keys and list item selection vs button selection. But I don't see this as urgent.
2. I can see us replicating the "boxed" styling of the keyboard list in the options and hotkeys lists, for consistency.

rc-swag:
I have added a check list to @mcdurdin  review. I have done the items marked off the rest I am having some trouble with. 
I wanted to check in what I have to get testing underway. Then I can add new tests for what is changed on the remaing items.

rc-swag:
@keymanapp-test-bot retest TEST_DISABLE_BTN, TEST_SET_HOTKEY_SINGLE_LANGUAGE

bharanidharanj:
* **TEST_DISABLE_BTN (PASSED):** Retested this as per the instructions and I noticed that the whole keyboard detail opacity is reduced to 60% out and the title bar was grayed out. 

 
   



* **TEST_SET_HOTKEY_SINGLE_LANGUAGE (PASSED):** Retested this as per the instructions and I verified that the title should now be  displayed with the hotkey for this keyboard - Ctrl + I

mcdurdin:
Going ahead and merging so we can launch beta. We will need to do another round of usability review during beta, but I expect any changes would be minor and cosmetic only :grin:

Good work!

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.94-beta](https://keyman.com/downloads/releases/beta/16.0.94)

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.1-alpha](https://keyman.com/downloads/releases/alpha/17.0.1)

PR #7211, opened 2022-09-06
refactor(developer): split kmxplus builder functions into separate files 🙀

mcdurdin:
This refactor work simply splits the existing functions into separate files to prevent kmx-plus-builder.ts getting too big.

The functions now have no side-effects. Some of them mutate input parameters, but they are fairly self-contained, which should make unit tests simpler. Note that `alloc_string` is used extensively, but works on the input parameter `sect_strs`.

Note: kmx-plus.ts has additional definitions for transform sections. These should have been in the next PR, but it's a bit of work to split them out for very little benefit.

@keymanapp-test-bot skip


developer/src/kmldmlc/src/keyman/kmx/kmx-plus.ts
```diff
@@ -10,6 +10,37 @@ export class Section {
 }
 
 // 'sect'
+
+export type Sect = Section;
+
+// 'bksp' -- see 'tran'
```
mcdurdin, line 16: Some of these transform-related definitions should have gone into #7212 but splitting them out is probably not a worthwhile exercise at this point.

PR #7212, opened 2022-09-06
feat(developer): kmldmlc compiler for transforms 🙀

mcdurdin:
Adds compilers, kmx builders, and minimal unit tests for transform sections `bksp`, `elem`, `finl`, `ordr`, and `tran`.

Adds a utility script for building test fixtures for manual analysis:

    ./build.sh build-fixtures

Refactors constants to give us design-time type safety once more.

Note that UnicodeSets are not yet supported.

@keymanapp-test-bot skip


developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder/build-elem.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  let str: BUILDER_ELEM_STRING = {
+    _value: value,
+    items: [],
+    length: value.length,
+    offset: 0 // will be filled in later
+  };
+
+  for(let v of value) {
```
srl295, line 82: is value implicitly a set? Spec says it is a "codepoint or a UnicodeSet" and a unicodeSet should have brackets `[]` around multiple characters.   This code seems like it woudl take `abc` and turn it into a, b, c

srl295 (approved):
Generally LGTM though a question about the unicodesets

developer/src/kmldmlc/src/keyman/kmx/kmx-plus-builder/build-elem.ts
mcdurdin: Currently, this code doesn't support UnicodeSets. See #7234.

PR #7219, opened 2022-09-06
fix(developer): suppress repeated warnings about unreachable code

### LINKED ISSUE #7216: bug(developer): kmcomp generates too many warnings for some rules
**Describe the bug**

Compiling lao_phonetic.kmn at https://github.com/keymanapp/keyboards/commit/7ca2bc96532de735053884aea7fbb479592e9e25 for web generates hundreds of warnings like:

```
lao_phonetic.kmn (237): Warning: 209A The rule will never be matched because its key code is never fired.
```

This relates to one letter in the store: `store(C)` has an 'ñ' to generate 'ຍ' -- but there is no 'ñ' key on the keyboard, so the web compiler generates a warning for it.

See also keymanapp/keyboards#1633

---

**Keyman Developer:**
 - Keyman Developer version: 15.0.269-stable 

---

**Keyboard**
 - Keyboard name:    lao_phonetic 
 - Keyboard version: 1.1.1 
 - Language name:    Lao

mcdurdin:
Fixes #7216.

The warning message 0x209A 'The rule will never be matched because its key code is never fired.' was being generated multiple times for a single line because the `JavaScript_Key` function it is generated by is used for various purposes.

This PR keeps a cache of reported key rules to ensure that the message is reported only once for a given key rule, and also improves the reporting to clarify which specific key is unreachable, which makes it easier to diagnose when using `any(k)` style messages, for example:

```
lao_phonetic.kmn (237): Hint: 109A The rule will never be matched for key 'ñ' because its key code is never fired.
```

This also reduces the warning to a hint, as this should not be a blocking issue for a keyboard, rather just a place the keyboard author can tidy up.

# User Testing

**TEST_LAO_PHONETIC:** Load the lao_phonetic keyboard from the keyboards repo, and verify that the message `109A` is generated at most once per line/key when compiling. 

1. you will need to edit the line `store(&TARGETS) 'windows macosx'` and change it to `store(&TARGETS) 'any'` in the keyboard before compiling in order to reproduce this and test it.

2. You may wish to contrast to the current alpha release of Keyman Developer, where hundreds of warnings would be generated.

### DIFF
--- a/developer/src/common/delphi/compiler/CompileErrorCodes.pas
+++ b/developer/src/common/delphi/compiler/CompileErrorCodes.pas
@@ -194,7 +194,7 @@ const
   CWARN_VisualKeyboardFileMissing                   = $2097;
   CWARN_ExtendedShiftFlagsNotSupportedInKeymanWeb   = $2098;   // I4118
   CWARN_TouchLayoutUnidentifiedKey                  = $2099;   // I4142
-  CWARN_UnreachableKeyCode                          = $209A;   // I4141
+  CHINT_UnreachableKeyCode                          = $109A;   // I4141
 
   CWARN_CouldNotCopyJsonFile                        = $209B;   // I4688
   CWARN_PlatformNotInTargets                        = $209C;
--- a/developer/src/kmcmpdll/comperr.h
+++ b/developer/src/kmcmpdll/comperr.h
@@ -191,7 +191,7 @@
 #define CWARN_VisualKeyboardFileMissing                    0x00002097
 #define CWARN_ExtendedShiftFlagsNotSupportedInKeymanWeb    0x00002098   // I4118
 #define CWARN_TouchLayoutUnidentifiedKey                   0x00002099
-#define CWARN_UnreachableKeyCode                           0x0000209A
+#define CHINT_UnreachableKeyCode                           0x0000109A
 
 #define CWARN_CouldNotCopyJsonFile                         0x0000209B
 #define CWARN_PlatformNotInTargets                         0x0000209C
--- a/developer/src/tike/compile/CompileKeymanWeb.pas
+++ b/developer/src/tike/compile/CompileKeymanWeb.pas
@@ -102,6 +102,7 @@ uses
   Winapi.Windows,
   System.Character,
   System.Classes,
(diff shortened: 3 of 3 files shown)


bharanidharanj:
* **TEST_LAO_PHONETIC (FAILED):** Tested this with Keyman Developer 16.0.57-alpha-test-7219 build along with lao_phonetic Keyboard and I noticed that 109A message is not generated when compiling the keyboard. I am seeing the same result when I tested with Keyman Developer 16.0.56-alpha build. Seems to be not working as expected. Or Did I missed out anything?

..in Keyman Developer 16.0.57 build

   

..in Keyman Developer 16.0.56 alpha build

mcdurdin:
Apologies -- I left out a crucial step in the user test -- editing the keyboard to target 'any' instead of just windows+macosx, which meant that the issue was not triggered. I have edited the user test steps accordingly now.

@keymanapp-test-bot retest

bharanidharanj:
> Apologies -- I left out a crucial step in the user test -- editing the keyboard to target 'any' instead of just windows+macosx, which meant that the issue was not triggered. I have edited the user test steps accordingly now.
> 
> @keymanapp-test-bot retest

Thanks, Marc for your comments. I will retest it with the new steps and will add my comments. :-)

MakaraSok:
> * Keyman Developer 15.0.56 current alpha release build

I guess you meant "Keyman Developer `16.0.56` current alpha".

mcdurdin:
> I guess you meant "Keyman Developer `16.0.56` current alpha".

It's possible that I meant version 16 but I refuse to confirm or deny! 🤣

bharanidharanj:
> > * Keyman Developer 15.0.56 current alpha release build
> 
> I guess you meant "Keyman Developer `16.0.56` current alpha".

@MakaraSok Ah..Yes, Makara. Sorry for the Version Confusion. It is 16.0.56 alpha from Keyman Site.

rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.61-alpha](https://keyman.com/downloads/releases/alpha/16.0.61)

Issue #7134, opened 2022-08-26
bug(core): ldml: improve utf-32 handling 🙀


srl295:
- use kmx_xstring and codecvt instead of hacks

#5015

srl295:
closed as both have landed

PR #7230, opened 2022-09-07
fix(core): ldml: improve utf-32 handling 🙀

srl295:
- remove hacky utf-32 processing
- add utf-32 test case to test_kmx_plus.cpp
- change kmx_plus to return std::u16string
- kmx_xstring: add Uni_IsBMP() macro

Fixes: #7134

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -161,20 +161,10 @@ COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
 
 std::u16string
 COMP_KMXPLUS_KEYS_ENTRY::get_string() const {
-  if (flags & LDML_KEYS_FLAGS_EXTEND) {
-    return std::u16string();  // error, extended
-  }
-  char16_t buf[2];
-  std::size_t len;
-  if (Uni_IsBMP(to)) {
-    buf[0] = (to & 0xFFFF);
-    len    = 1;
-  } else {
-    buf[0] = Uni_UTF32ToSurrogate1(to);
-    buf[1] = Uni_UTF32ToSurrogate2(to);
-    len    = 2;
-  }
-  return std::u16string(buf, len);
+  assert(!(flags & LDML_KEYS_FLAGS_EXTEND)); // should not be called.
+  char16_single buf;
+  const int len = Utf32CharToUtf16(to, buf);
+  return std::u16string(buf.ch, len);
 }
 
 bool
@@ -353,12 +343,11 @@ COMP_KMXPLUS_STRS::get(KMX_DWORD entry) const {
     if (entry >= count) {
         return std::u16string(); // Fallback: empty string
     }
-    KMX_DWORD offset = entries[entry].offset;
-    KMX_DWORD length = entries[entry].length;
+    const KMX_DWORD offset = entries[entry].offset;
+    const KMX_DWORD length = entries[entry].length;
     assert(offset+((length+1)*2) <= header.size); // assert not out of bounds
     const uint8_t* thisptr = reinterpret_cast<const uint8_t*>(this);
     const KMX_WCHAR* start = reinterpret_cast<const KMX_WCHAR*>(thisptr+offset);
-    assert(start[length] == 0); // assert null terminated
     return std::u16string(start, length);
 }
 
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -7,6 +7,7 @@
 #include <km_types.h>
 #include <kmx_file.h>
 #include <kmx/kmx_plus.h>
+#include <kmx/kmx_xstring.h>
 
 #if KMXPLUS_DEBUG
 #include <iostream>
@@ -158,6 +159,24 @@ COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
   return true;
 }
 
+std::u16string
+COMP_KMXPLUS_KEYS_ENTRY::get_string() const {
+  if (flags & LDML_KEYS_FLAGS_EXTEND) {
+    return std::u16string();  // error, extended
+  }
+  char16_t buf[2];
+  std::size_t len;
+  if (Uni_IsBMP(to)) {
+    buf[0] = (to & 0xFFFF);
+    len    = 1;
+  } else {
+    buf[0] = Uni_UTF32ToSurrogate1(to);
+    buf[1] = Uni_UTF32ToSurrogate2(to);
+    len    = 2;
+  }
+  return std::u16string(buf, len);
+}
+
 bool
 COMP_KMXPLUS_LOCA::valid(KMX_DWORD _kmn_unused(length)) const {
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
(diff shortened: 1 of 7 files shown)


core/src/kmx/kmx_plus.cpp
```diff
@@ -158,6 +159,24 @@ COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
   return true;
 }
 
+std::u16string
```
srl295, line 162: moved this here just so it's unit testable

core/src/kmx/kmx_plus.cpp
```diff
@@ -207,22 +226,36 @@ COMP_KMXPLUS_STRS::valid(KMX_DWORD _kmn_unused(length)) const {
...
-            KMXPLUS_PRINTF((" U+%04X ", str[j]));
+            KMXPLUS_PRINTF((" U+%04X ", start[j]));
         }
     }
     KMXPLUS_PRINTF(("\n"));
+
+    // // now print it as a string
+    // TODO-LDML
+    // std::u16string str = get(i);
+    // std::cerr << str << std::endl;
```
srl295, line 257: todo, figure out how to print u16strings

core/src/kmx/kmx_xstring.h
```diff
@@ -8,7 +8,9 @@ namespace kmx {
 
 #define Uni_IsSurrogate1(ch) ((ch) >= 0xD800 && (ch) <= 0xDBFF)
 #define Uni_IsSurrogate2(ch) ((ch) >= 0xDC00 && (ch) <= 0xDFFF)
+
 #define Uni_IsSMP(ch) ((ch) >= 0x10000)
```
srl295, line 12: Misnomer!   SMP is 0x010000…0x01FFFD inclusive.  0x020000 begins plane 2.

srl295: this macro was used elsewhere, so i didn't 'fix' it

srl295:
i realized there are no .kmx files anywhere which test the non-extend case 😭

core/src/kmx/kmx_plus.cpp
```diff
@@ -158,6 +159,24 @@ COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
   return true;
 }
 
+std::u16string
+COMP_KMXPLUS_KEYS_ENTRY::get_string() const {
+  if (flags & LDML_KEYS_FLAGS_EXTEND) {
+    return std::u16string();  // error, extended
+  }
```
mcdurdin, line 166: If this is an error, we need an assert() on it for debug builds. And I think the function should be named `unextended_to_as_string` or something like that?

Otherwise, this needs a TODO-LDML on it?

core/src/kmx/kmx_xstring.h
mcdurdin: Yep, longstanding misnomer. We should rename to `Uni_IsSupplementaryPlane()` which I think covers all planes 1-16?

core/src/kmx/kmx_plus.cpp
```diff
@@ -158,6 +159,24 @@ COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
...
+  char16_t buf[2];
+  std::size_t len;
+  if (Uni_IsBMP(to)) {
+    buf[0] = (to & 0xFFFF);
```
mcdurdin, line 170: You've just added `Uni_IsBMP()` so let's go ahead and add a macro `Uni_UTF32BMPToUTF16`? or something like that? Just so we are consistent with our bit fiddling?

core/src/kmx/kmx_plus.cpp
```diff
@@ -158,6 +159,24 @@ COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
...
+    len    = 1;
+  } else {
+    buf[0] = Uni_UTF32ToSurrogate1(to);
+    buf[1] = Uni_UTF32ToSurrogate2(to);
```
mcdurdin, line 174: I'd love to rename these macros to something clearer also. When these were written, we were so Windows-specific that it was obvious that we were talking UTF-16. And surrogate implies UTF-16. But they should be `Uni_UTF32toUTF16surrogate1()`? or something like that ... or get away from macros and put all this into an (inline?) function, something like:

```c++
typedef struct { char16_t c[3]; } char16_single;
void Utf32CharToUtf16(const KMX_DWORD ch32, char16_single& ch16) { ... }
std::u16string Utf32CharToU16String(const KMX_DWORD ch32) { ... }
```

And I would null terminate `buf` for c-safety (so `char16_t buf[3]`). I am keen to try and improve our style as we move forward... so now is the time to do this!

core/src/kmx/kmx_plus.cpp
mcdurdin: Yep, painful...

core/src/kmx/kmx_plus.cpp
```diff
@@ -314,23 +347,19 @@ KMX_DWORD COMP_KMXPLUS_SECT::find(KMX_DWORD ident) const {
...
+std::u16string
+COMP_KMXPLUS_STRS::get(KMX_DWORD entry) const {
     assert(entry < count);
     if (entry >= count) {
-        return nullptr;
+        return std::u16string(); // Fallback: empty string
     }
-    // TODO-LDML: will be improved in https://github.com/keymanapp/keyman/issues/7134
     KMX_DWORD offset = entries[entry].offset;
     KMX_DWORD length = entries[entry].length;
```
mcdurdin, line 357:
```suggestion
    const KMX_DWORD offset = entries[entry].offset;
    const KMX_DWORD length = entries[entry].length;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -314,23 +347,19 @@ KMX_DWORD COMP_KMXPLUS_SECT::find(KMX_DWORD ident) const {
...
-    assert(bufsiz > (length+1)); // assert bufsiz big enough
     assert(offset+((length+1)*2) <= header.size); // assert not out of bounds
     const uint8_t* thisptr = reinterpret_cast<const uint8_t*>(this);
     const KMX_WCHAR* start = reinterpret_cast<const KMX_WCHAR*>(thisptr+offset);
-    for(KMX_DWORD i=0;i<=length;i++) {
-        buf[i] = start[i];
-    }
-    return buf;
+    assert(start[length] == 0); // assert null terminated
```
mcdurdin, line 361: We need to be sure that assertions are around bugs in our code, not malformed inputs. Not sure that this is covering that off?

core/src/kmx/kmx_plus.h
```diff
@@ -142,6 +143,12 @@ struct COMP_KMXPLUS_KEYS_ENTRY {
     KMX_DWORD mod;
     KMX_DWORD to;     // to may be KMXPLUS_STR or UTF32 char
     KMX_DWORD flags;
+    /**
+     * @brief Get the 'to' as a string, if not EXTEND
```
mcdurdin, line 147:
```suggestion
     * @brief Get the 'to' as a string, if flags&LDML_KEYS_FLAGS_EXTEND is not set
```

mcdurdin (changes requested):
I'm being nitpicky here but I think I'd like to see a bit more done on get_string :grin:

Conceptually however, this is good.

core/src/kmx/kmx_plus.cpp
srl295: Zero length is unambiguous- no bit pattern would produce it.

core/src/kmx/kmx_plus.cpp
srl295: Ok

core/src/kmx/kmx_plus.cpp
mcdurdin: In my understanding, we shouldn't be calling this with a non-extend keys entry, so it should be fixed in caller -- that's why I wanted the assert. Empty string may cause all sorts of other issues on a release build where caller is expecting non-empty string?

core/src/kmx/kmx_xstring.h
srl295: You're right!  https://www.unicode.org/glossary/#supplementary_planes Practically, though, I have not heard this used as much.   Certainly 'the SMP' is only plane 1.  Anyway, I think it depends on what the callers are looking for.

srl295:
@mcdurdin PTAL.  Would like to get this improvement in, and tackle renaming of the macros (touching unrelated code) in a separate PR.

core/src/kmx/kmx_plus.cpp
srl295: removed

core/src/kmx/kmx_plus.cpp

core/src/kmx/kmx_plus.cpp
srl295: done, and will rename macros in a separate PR

core/src/kmx/kmx_plus.cpp

core/src/kmx/kmx_plus.cpp
srl295: changed to assert.

core/src/kmx/kmx_xstring.h
mcdurdin: Unicode Glossary does not mention SMP. But Supplementary Planes, Supplementary Characters all refer to 1-16. So I blame Unicode for the confusion! Doing a Google search for "supplementary multilingual planes" shows SMP is commonly used to refer to non-BMP. 

This macro was probably introduced around 2002... :grin:

mcdurdin (approved):
LGTM!

core/src/kmx/kmx_xstring.h
srl295: "the SMP" - https://www.unicode.org/roadmaps/smp/ this is where i most commonly see it. (UTC)

PR #7243, opened 2022-09-08
fix(windows): Delete both code units when deleting surrogate pairs in TSF-aware apps

rc-swag:
Fixes: #7231
When the windows platform receives the action for a backspace it will check to see if the character being deleted is a surrogate pair. If it is and the app is TSF aware app it will need to set the BK_SURROGATE flag in the platform action queue This will result in two backspaces to remove both parts of the surrogate pair. For legacy apps, this is not required.

Issue: #7251 will add some windows unit tests.
   
Using a public available keyboard.
# USER TESTING 
**Setup Text Services Framework(TSF) testing with Keyman for Windows**

1. Install Keyman Developer version 15 or greater. 
2. Open the Keyman Developers Test page in Firefox. **It must be [`Firefox Browser`](https://www.mozilla.org/en-US/firefox/new/)** 
  a. By default this is at localhost:8008 so you could type in to the firefox address bar
  b. You can also access it by pressing the `Test package on web` button found in the `Compile` tab for any keyboard package.
3. On the `Keymand Developer Keyboard Test Page` use the `Keyboard` dropdown to select `system keyboard`
4. Use the `Status Bar` to select the desired Keyman for the Windows keyboard. 
5. Typing text into the input field will now use the TSF with the keyboard selected in step 4.

* TEST_TSF_BACKSPACE_SURROGATE

1. Install [Hieroglyphic ](url)Keyboard from keyman.com
2. Open the `Keyman Developer Test Page`. Follow the steps from `Setup Text Services Framework(TSF) testing with Keyman for Windows` if needed.
3. Select the Hieroglyphic keyboard from the status bar
4. In the `Keyman Developer Test Page` Type: <kbd>d</kbd><kbd>i</kbd><kbd>spacebar</kbd>
5. Expected Output : 𓏙                 and the character bar just show the one character "0133d9" 
6. Type:<kbd>backspace</kbd>
7. Expected Output: Empty and the character bar above the input text should no have any character just showing `empty`.

* TEST_TSF_RULE_DELETES_SURROGATE

1. Install [Hieroglyphic ](url)Keyboard from keyman.com
2. Open the `Keyman Developer Test Page`. Follow the steps from `Setup Text Services Framework(TSF) testing with Keyman for Windows` if needed.
3. Select the Hieroglyphic keyboard from the status bar
4. In the `Keyman Developer Test Page` Type: <kbd>d</kbd><kbd>i</kbd><kbd>spacebar</kbd><kbd>spacebar</kbd>
[...]

PR #7244, opened 2022-09-08
chore(core): improve xstring macros 🙀

srl295:
- move Utf32CharToUtf16 to inline
- remove Is_SMP and call sites, use Utf32CharToUtf16 instead.

For: #7134

I did not modify call sites that are using the windows unicode.h

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_actions.cpp
+++ b/core/src/kmx/kmx_actions.cpp
@@ -45,11 +45,12 @@ KMX_BOOL KMX_Actions::QueueAction(int ItemType, KMX_DWORD dwData)
     break;
 
   case QIT_CHAR:
-    if(Uni_IsSMP(dwData)) {
-      m_context->Add(Uni_UTF32ToSurrogate1(dwData));
-      m_context->Add(Uni_UTF32ToSurrogate2(dwData));
-    } else {
-      m_context->Add((KMX_WORD) dwData);
+    {
+      char16_single buf;
+      int len = Utf32CharToUtf16(dwData, buf);
+      for(int i=0; i<len; i++) {
+        m_context->Add(buf.ch[i]);
+      }
     }
     break;
 
--- a/core/src/kmx/kmx_processor.cpp
+++ b/core/src/kmx/kmx_processor.cpp
@@ -242,11 +242,10 @@ kmx_processor::process_event(
   for (auto c = cp.begin(); c != cp.end(); c++) {
     switch (c->type) {
     case KM_KBP_CT_CHAR:
-      if (Uni_IsSMP(c->character)) {
-        ctxt += Uni_UTF32ToSurrogate1(c->character);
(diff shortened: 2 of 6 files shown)


core/src/kmx/kmx_actions.cpp
```diff
@@ -45,11 +45,12 @@ KMX_BOOL KMX_Actions::QueueAction(int ItemType, KMX_DWORD dwData)
...
-      m_context->Add(Uni_UTF32ToSurrogate1(dwData));
-      m_context->Add(Uni_UTF32ToSurrogate2(dwData));
-    } else {
-      m_context->Add((KMX_WORD) dwData);
+    {
+      char16_single buf;
+      int len = Utf32CharToUtf16(dwData, buf);
+      for(int i=0; i<len; i++) {
+        m_context->Add(buf.ch[i]);
+      }
```
mcdurdin, line 53: Hmm... it's a shame that the replacement requires a loop. But not seeing a clean, simple way to avoid that. (I can see ugly C++ ways of avoiding the loop. But NO, we won't go there.)

mcdurdin (approved):

core/src/kmx/kmx_actions.cpp
srl295: shouldn't the compiler optimize it?  'len' here is going to be 1 or 2 exactly.

srl295:
@mcdurdin merge if you need to, will be out for an errand

srl295:
@mcdurdin  a new usage of the windows version of  Is_SMP !

https://github.com/keymanapp/keyman/pull/7243/files#diff-d71f3c8ec6081ed84ecbc07ee305b2e81fa98637ce88a7d5be7f10f6270f6411R37

core/src/kmx/kmx_actions.cpp
mcdurdin: I would hope so. It's just a little verbose for code readers :grin:

srl295: The other way to do this is to update Add() to take a string and length. Like I did with the .append() call.

srl295: @mcdurdin  we could have a macro to collapse it like so:

```c
#define FOREACH_Utf32CharToUtf16(ch, var, action) { \
     char16_single buf;  \
     int len = Utf32CharToUtf16(ch, buf); \
     for(int i=0; i<len; i++) { \
        const char16_t &var = buf.ch[i]; /*alias*/ \
        (void)(action); \
     } \
  }
```

thus:

```c
  FOREACH_Utf32CharToUtf16(dwData, code_unit, m_context->add(code_unit));
```

… will turn into the same equivalent loop.

mcdurdin: yeah ... nah :grin: I'm also not a big fan of C macros burying logic..

PR #7250, opened 2022-09-08
chore(developer): kmc - use local reference for json schema 🙀

### LINKED ISSUE #7221: chore(developer/kmldmlc): use local ref for json schema 🙀
We should move this to the configure step and then use this file also in test/helpers/index.ts -- keeps things neat and tidy and avoids deeper cross-referencing. But will do this in a follow-up PR.

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/7184#discussion_r964223315_

mcdurdin:
Fixes #7221.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/build.sh
+++ b/developer/src/kmc/build.sh
@@ -31,13 +31,6 @@ builder_parse "$@"
 
 #-------------------------------------------------------------------------------------------------------------------
 
-if builder_has_action configure; then
-  verify_npm_setup
-  builder_report success configure
-fi
-
-#-------------------------------------------------------------------------------------------------------------------
-
 if builder_has_action clean; then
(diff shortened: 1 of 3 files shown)


developer/src/kmc/build.sh
```diff
@@ -31,24 +31,27 @@ builder_parse "$@"
 
 #-------------------------------------------------------------------------------------------------------------------
 
-if builder_has_action configure; then
-  verify_npm_setup
-  builder_report success configure
+if builder_has_action clean; then
+  rm -rf ./build/ ./tsconfig.tsbuildinfo
+  builder_report success clean
 fi
```
mcdurdin, line 37: `clean` needs to happen before `configure`...

srl295 (approved):

PR #7255, opened 2022-09-09
chore: make checksum optional 🙀

### LINKED ISSUE #7222: feat: make checksum optional in .kmx 🙀
Query: do we make checksum optional for 16.0 onwards? We'd still verify the checksum if non-zero but allow files with a zero field for checksum. Note that this would have ripple effects into other sources which also parse .kmx files -- but we are probably going to need to review those for kmxplus support at some point anyway.

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/issues/7220#issuecomment-1238708984_

mcdurdin:
Fixes #7222.

Includes a unit test for zero checksum file (not produced by the compiler, so we have to manually generate it)

@keymanapp-test-bot skip

### DIFF
--- a/core/tests/unit/kmx/fixtures/binary/meson.build
+++ b/core/tests/unit/kmx/fixtures/binary/meson.build
@@ -1,0 +1,42 @@
+# Copyright:    © 2018 SIL International.
+# Description:  Cross platform build script to compile libkmnkbp API unit tests.
+# Create Date:  19 Oct 2018
+# Authors:      Marc Durdin, Tim Eves (TSE)
+# History:      19  Oct 2018 - TSE - Added test for context API functions.
+#
+
+# binary file unit tests
+
+# Build all binary test keyboards
+
+if compiler.get_id() == 'emscripten'
+  binary_test_path = '/'
+else
+  binary_test_path = meson.current_build_dir()
+endif
+
+binary_tests = [
+  'b_000_zero_checksum'
+]
+
+foreach kbd : binary_tests
+  configure_file(
+    command: copy_cmd + ['@INPUT@', '@OUTPUT@'],
+    input: kbd + '.kmn',
+    output: kbd + '.kmn'
+  )
+
+  configure_file(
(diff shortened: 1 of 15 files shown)


core/tests/unit/kmx/fixtures/binary/meson.build
```diff
@@ -0,0 +1,42 @@
...
+# Build all binary test keyboards
+
+if compiler.get_id() == 'emscripten'
+  binary_test_path = '/'
+else
+  binary_test_path = meson.current_build_dir()
+endif
+
+binary_tests = [
+  'b_000_zero_checksum'
```
srl295, line 19: love the incremental improvement here

core/tests/unit/kmx/fixtures/meson.build
```diff
@@ -0,0 +1,6 @@
+if node.found()
+  # Note: if node is not available, we cannot build the keyboards; build.sh
+  # emits a warning that the 'ldml' keyboard tests will be skipped; that
```
srl295, line 3:
```suggestion
  # Note: if node is not available, we cannot build the binary test keyboards; build.sh
  # emits a warning that the 'kmx/binary' keyboard tests will be skipped; that
```

looks like a copy and paste…

srl295 (approved):
LGTM.

I assume kmx_plus should still validate the checksum, iff present?

core/tests/unit/kmx/fixtures/meson.build
mcdurdin: Actually, at present build.sh still reports that it can't build 'ldml' keyboards 😆... low priority message which we should hopefully never see.

mcdurdin:
> I assume kmx_plus should still validate the checksum, iff present?

Maybe? Or we could consider it irrelevant and deprecated and ignore it? I actually think that may be better. That suggests to me I should be adding a `// DEPRECATED in 16.0` comment to the header files.

PR #7265, opened 2022-09-12
fix(developer): suppress repeated warnings about unreachable code 🍒

### LINKED ISSUE #7216: bug(developer): kmcomp generates too many warnings for some rules

mcdurdin:
Fixes #7216.

Cherry-pick of #7219.

The warning message 0x209A 'The rule will never be matched because its key code is never fired.' was being generated multiple times for a single line because the `JavaScript_Key` function it is generated by is used for various purposes.

This PR keeps a cache of reported key rules to ensure that the message is reported only once for a given key rule, and also improves the reporting to clarify which specific key is unreachable, which makes it easier to diagnose when using `any(k)` style messages, for example:

```
lao_phonetic.kmn (237): Warning: 209A The rule will never be matched for key 'ñ' because its key code is never fired.
```

This also reduces the warning to a hint, as this should not be a blocking issue for a keyboard, rather just a place the keyboard author can tidy up.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/developer/TIKE/compile/CompileKeymanWeb.pas
+++ b/windows/src/developer/TIKE/compile/CompileKeymanWeb.pas
@@ -102,6 +102,7 @@ uses
   Winapi.Windows,
   System.Character,
   System.Classes,
+  System.Generics.Collections,
   System.UITypes,
 
   compile,
@@ -214,6 +215,7 @@ type
     FTouchLayoutFont: string;
     FFix183_LadderLength: Integer;
     FCloseBrace: Boolean;   // I4872
+    FUnreachableKeys: TList<PFILE_KEY>;
 
     function JavaScript_String(ch: DWord): string;  // I2242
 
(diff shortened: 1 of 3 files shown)


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.270](https://keyman.com/downloads/releases/stable/15.0.270)

PR #7267, opened 2022-09-12
fix(developer): show more useful error if out of space during Setup

### LINKED ISSUE #7261: bug(developer): installer fails with "Stream write error" if not enough disk space
**Describe the bug**

If there is not enough disk space to extract the installer, it fails with a small message box saying "Stream write error". No Sentry report is generated.

image

Error experienced on Peter's laptop.

---

**Keyman Developer:**
 - Keyman Developer version: 15.0.269

mcdurdin:
Fixes #7261.

Low severity; no user test required. This should work for both Developer and Desktop. Note that there are other scenarios where the installer may run out of space which are not covered by this, such as downloading additional media, running Windows Installer, etc. But this should handle the obscure 'Stream write error' messsage better!

@keymanapp-test-bot skip

### DIFF
--- a/common/windows/delphi/setup/SFX.pas
+++ b/common/windows/delphi/setup/SFX.pas
@@ -82,27 +82,38 @@ var
   fs: TFileStream;
   ms: TMemoryStream;
 begin
-  fs := TFileStream.Create(ParamStr(0), fmOpenRead or fmShareDenyWrite);
-  ms := TMemoryStream.Create;
   try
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.62-alpha](https://keyman.com/downloads/releases/alpha/16.0.62)

PR #7281, opened 2022-09-13
chore(linux): Update debian changelog 🍒

ermshiperete:
(🍒 picked from PR #7280)

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/changelog
+++ b/linux/debian/changelog
@@ -1,3 +1,10 @@
+keyman (15.0.270-1) unstable; urgency=medium
+
+  * New upstream release.
+  * Re-release to Debian
+
+ -- Eberhard Beilharz <eb1@sil.org>  Tue, 13 Sep 2022 11:20:25 +0200
+
 keyman (15.0.269-1) unstable; urgency=medium
 
   * New upstream release.


mcdurdin (approved):
LGTM.

Can this be automated in our release build trigger process?

ermshiperete:
> Can this be automated in our release build trigger process?

I'm not sure. There are several steps involved:

1. create the source package, sign it and upload to mentors.debian.net. I'll have to check if the signing could be automated and run on CI machine and if that would be acceptable with Debian policies.
2. Create draft PRs with updated changelog based on the uploaded source package. This needs to be a draft PR since it has to match what gets accepted by Debian.
3. Send an email to the Debian input method mailing list asking for a sponsor for the uploaded package.
4. Once the package got accepted by Debian, remove the draft status from the PRs and eventually merge them.

The first two steps are currently done by [a script](https://github.com/keymanapp/keyman/blob/master/linux/scripts/upload-to-debian.sh) (creating the PRs is only locally on my machine atm because I need to test it first).

We could think about adding the `auto` label to the PRs and enhance the GHA so that they get auto-merged.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.63-alpha](https://keyman.com/downloads/releases/alpha/16.0.63)

mcdurdin:
> I'm not sure. There are several steps involved:

OK, sounds like it's a bit tricky. This is good documentation in case someone else has to pick this up at some point, thanks!

PR #7302, opened 2022-09-16
chore(common): Update to Unicode 15.0

darcywong00:
Update to Unicode 15.0 characters from https://www.unicode.org/Public/15.0.0/ucd/

@keymanapp-test-bot skip


srl295 (approved):

keyman-server:

PR #7286, opened 2022-09-13
feat(developer): add kmx metadata to kmxplus compiles 🙀

mcdurdin:
Fixes #7273.

Adds 4 kmx system stores to allow existing configuration tools to load and process KMXPlus files without modification:

* `TSS_NAME` (first `<name>`)
* `TSS_COMPILEDVERSION` (only if `options.addCompilerVersion == true`)
* `TSS_KEYBOARDVERSION` (from `<version number>`)
* `TSS_TARGETS` (currently always 'desktop', will later include more targets)

In order for `TSS_COMPILEDVERSION` to be handled correctly, needed to plumb in the new `CompilerOptions` interface. This could be a separate PR, but the changes are not huge, so opted to append the changes here.

Updates related tests and fixtures. Refactored some of the e2e helpers into helpers/index.ts for use with the metadata-compiler tests.

@keymanapp-test-bot skip

PR #7311, opened 2022-09-18
chore(core): `kmc --outFile` to `kmc --out-file` (and `--` rider) 🙀

mcdurdin:
This fixes the failing Core tests in #7054 -- broke in #7286, but wasn't picked up
there because Core build didn't run on that PR.

## All about `./build.sh --`

This update also allows use of `./build.sh configure -- --param-one=foo ...`.

Updates Core build to support `--` (including build.bat wrapper for Windows builds). Fixes Linux debian rules to pass additional parameters in with this format again, and adds `:arch` to only build current architecture, not wasm as well...

Note that when using `--` you should typically avoid multiple actions in a single command, as each action may interpret the additional parameters differently!

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/rules
+++ b/linux/debian/rules
@@ -35,7 +35,7 @@ override_dh_auto_configure:
 
 override_dh_auto_build:
 	# keymankeyboardprocessor
-	cd core && ./build.sh build
+	cd core && ./build.sh build:arch
 	# ibus-keyman
 	cd linux/ibus-keyman && make
 	# keyman-config
@@ -48,7 +48,7 @@ override_dh_auto_build:
 override_dh_auto_test:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
 	# keymankeyboardprocessor
-	cd core && ./build.sh tests
+	cd core && ./build.sh test:arch
 	# ibus-keyman
 	cd linux/ibus-keyman && make check VERBOSE=1
 	# keyman-config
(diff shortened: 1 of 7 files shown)


ermshiperete (approved):
lgtm - although I don't particularly like putting these extra parameters in a `--configure` parameter. But that's a different thing...

linux/debian/rules
```diff
@@ -20,9 +20,9 @@ export LC_ALL=C.UTF-8
 override_dh_auto_configure:
 	# keymankeyboardprocessor
 	cd core && \
-		./build.sh configure -- --wrap-mode=nodownload --prefix=/usr --sysconfdir=/etc \
+		./build.sh configure --configure="--wrap-mode=nodownload --prefix=/usr --sysconfdir=/etc \
 			--localstatedir=/var --libdir=lib/$(DEB_TARGET_MULTIARCH) \
-			--libexecdir=lib/$(DEB_TARGET_MULTIARCH)
+			--libexecdir=lib/$(DEB_TARGET_MULTIARCH)"
```
srl295, line 25: I can't find why this was made… the `--` form seems cleaner, but there must be some reason…?

srl295 (approved):

mcdurdin:
> although I don't particularly like putting these extra parameters in a --configure parameter. But that's a different thing...

And it wasn't working anyway so fixed to use `--` again. But see PR description for caveats.

linux/debian/rules
mcdurdin: Back to using `--` form again now.

PR #7314, opened 2022-09-19
chore(developer): update multer dependency for server

mcdurdin:
Removes Keyman Developer Server's transitive dependency on dicer by updating multer to `1.4.5-lts.1`, which updates its dependency on busboy.

See https://github.com/expressjs/multer/pull/1097#issuecomment-1141286771 for reasoning behind use of  `-lts.1` rather than a full release version.

At some point in the future, multer will publish a full release with this fix, at which point we can move back to a full release version.

@keymanapp-test-bot skip


darcywong00 (approved):
packge-lock.json has merge conflicts, but lgtm after that's resolved :)

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.68-alpha](https://keyman.com/downloads/releases/alpha/16.0.68)

PR #7320, opened 2022-09-19
chore(common): remove unused ncc dependency

mcdurdin:
The `build:dist` script was never used; increment-version.sh uses only
`build:ts`.


mcdurdin:
wrong branch

PR #7328, opened 2022-09-19
chore(core): changes ldml tsconfig to es2020 🙀

mcdurdin:
Also splits the root tsconfig.json into tsconfig.cjs.json and tsconfig.esm.json. These are not typically used at present but in theory you could do a `tsc -b <tsconfig.x.json>` and build all modules. In practice, we still have some prebuild steps that make that a bit of a pain.

I have noted some inconsistency with `tsc -b` when referenced modules have different module modes, so splitting them seems wise. Also helpful for a simple index of which modules are in which format...

@keymanapp-test-bot skip

### DIFF
--- a/core/include/ldml/package.json
+++ b/core/include/ldml/package.json
@@ -8,6 +8,7 @@
     "unicode"
   ],
   "license": "MIT",
+  "type": "module",
   "main": "build/keyboardprocessor_ldml.js",
   "repository": {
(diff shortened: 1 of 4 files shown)


mcdurdin:
note: jenkins build failure addressed in #7311.

srl295 (commented):
pmlgtm

PR #7329, opened 2022-09-20
chore: patch update ansi-regex

mcdurdin:
@keymanapp-test-bot skip


mcdurdin:
Closing in favour of #7346 which does the same work.

Issue #7340, opened 2022-09-21
feat(developer/kmc): read .kmx file to retrieve `&keyboardversion` for package compiler 🙀


mcdurdin:
kmc-package will need to be able to parse a .kmx file to retrieve the keyboardversion value. This is necessary in order to meet the requirements of the `FollowKeyboardVersion` property of the source .kps file.

Note that we will *not* support parsing a generated .js file for the same property, as we will require keyboard packages to include the .kmx in the future anyway, for support for ldml keyboards.

As a temporary measure, I will read the keyboard version from the `Keyboards` array in the .kps, but this is not guaranteed to be up to date with the keyboard that's included -- because that value is informational only, and is updated only when the .kps is edited.

Part of #7277.

PR #7341, opened 2022-09-21
feat(developer): fixes to kmc-package for keyboard support 🙀

### LINKED ISSUE #7277: feat(developer): share kmlmc package compiler with kmc

mcdurdin:
Fixes #7277.

kmc-package had a number of gaps and errors. This fixes a bunch of problems:

* Wrong case on XML elements
* Wrong metadata in system element
* Missing kmp.json in `files` array (needs test in consumer apps later)
* Missing `displayFont`, `oskFont` in `keyboard` elements
* Including `copyLocation` when == 0
* Including an empty `startMenu` element
* Incorrect assumptions about location of source files in `source/` folder

Also fixes up test fixtures accordingly (mostly version numbers)

Note that support for `followKeyboardVersion` will come later; for now, it uses the possibly-stale value from the Keyboard element in the .kps file. See #7340 for tracking issue.

TODO: 
- [ ] Given the number of faults here, we will need to review our generation of lexical model packages at some point.
- [ ] Regression test lexical model packages built against Android & iOS

@keymanapp-test-bot skip


developer/src/kmc-package/src/kmp-compiler.ts
```diff
@@ -37,117 +40,227 @@ export default class KmpCompiler {
...
+
+  private kpsInfoToKmpInfo(info: KpsFileInfo): KmpJsonFileInfo {
+    let ni: KmpJsonFileInfo = {};
+
+    const keys: [(keyof KpsFileInfo), (keyof KmpJsonFileInfo)][] = [
+      ['author','author'],
+      ['copyright','copyright'],
+      ['name','name'],
+      ['version','version'],
+      ['webSite','website']
```
srl295, line 169: capital `S` answers 'why is this not a one dimensional list'

srl295 (approved):
LGTM, although are `developer/src/kmc-package/test/fixtures/khmer_angkor/build/khmer_angkor.js` and such (build directory) really supposed to be checked in?

developer/src/kmc-package/src/kmp-compiler.ts
mcdurdin: Yeah, gross right? It keeps me awake at night. Note that there is a camel-casify thing that I asked the xml2js wrapper to do when I wrote this code originally which I now regret and will back out in the future, which will mean more of these tables (but then at least it will say `WebSite` and `ID` just like the XML rather than this weird hybrid `webSite` and `iD` which matches nothing in any format!)

mcdurdin:
> are `developer/src/kmc-package/test/fixtures/khmer_angkor/build/khmer_angkor.js` and such (build directory) really supposed to be checked in?

In this case, yes. When building this we don't have access to the compiler, and we want to verify that the packaging phase will succeed when given files in various folders (../source, . (build/), etc).

developer/src/kmc-package/src/kmp-compiler.ts
mcdurdin: --> tracking in #7353

PR #7357, opened 2022-09-26
refactor(developer): move compiler callbacks out of readers 🙀

mcdurdin:
This is a minor refactor to prepare for moving the file readers into common/web/types, rather than keeping them as part of the compiler.

@keymanapp-test-bot skip


srl295 (approved):
makes sense

PR #7369, opened 2022-09-28
fix(developer): generate valid js when using unquoted digits in stores

### LINKED ISSUE #7005: bug(developer): compiler generates invalid javascript for unquoted numbers in `if()` statements
**Describe the bug**

```
if(option_key_t_is_tyun = 1) U+221A + [NCAPS K_T] > set(option_key_t_is_tyun = '0') save(option_key_t_is_tyun)
```

This generates invalid javascript: it generates a store with a value of `"\x1"`, which must be `"\x01"` to work. 

The source keyboard: armenian_mnemonic from https://community.software.sil.org/t/text-suggestion-not-working-in-some-cases/6490/4 ([armenian_mnemonic.zip](https://github.com/keymanapp/keyman/files/9204452/armenian_mnemonic.zip))

Proposed mitigations:

1. always generate minimum 2 char hexadecimal sequences. 
2. possibly, hint that an unquoted number might not be expected?
3. not really related, but compress the list of compared store values so we don't end up with hundreds of them in the compiled keyboard.

---

**Keyman Developer:**
 - OS: any 

mcdurdin:
Fixes #7005.

If a developer used single digit values in stores, unquoted (which is not really the usual approach, but should work just fine for non-zero values), the KMW compiler would emit invalid single-digit hexadecimal escapes for them, e.g. `"\x1"` instead of `"\x01"`.

Sample code:

```
store(option) 1

if(option = 1) + 'a' > 'one'
if(option = 2) + 'a' > 'two'
+ '1' > set(option = 1)
+ '2' > set(option = 2)
```

It is not immediately obvious from this code, but the value `1` is actually a character with value `'\x01'` or `U+0001`!

The more usual approach will not encounter this problem:

```
store(option) '1'

if(option = '1') + 'a' > 'one'
if(option = '2') + 'a' > 'two'
+ '1' > set(option = '1')
+ '2' > set(option = '2')
```

Note that the following is a compile error (due to internal use of null terminated strings).

```
store(option) 0
```

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tike/compile/CompileKeymanWeb.pas
+++ b/developer/src/tike/compile/CompileKeymanWeb.pas
@@ -1467,7 +1467,7 @@ begin
       9:  Result := '\t';
       10: Result := '\n';
       13: Result := '\r';
-      else Result := '\x'+IntToHex(ch,1);
+      else Result := '\x'+IntToHex(ch,2);
     end;
   end
   else


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.71-alpha](https://keyman.com/downloads/releases/alpha/16.0.71)

Issue #7238, opened 2022-09-08
feat(developer): emit .kvks and .keyman-touch-layout files and corresponding binaries in kmc 🙀


mcdurdin:
brain dump from discussion earlier this week.

`kmc` vs `kmldmlc`: see #7237

* kmldmlc must generate .kvks and .keyman-touch-layout.
* could also generate binary .kvk and 'compiled' touch layout JSON?
  * ideally kmldmlc generates these too
* questions on generating keyboard.js as well!
  * ideally kmldmlc generates this as well

Subquestion:
* do we include this data in kmxplus?
* Would mean separate .kvk  file wouldn't be needed, one day in the future.
* Would be equivalent to xml data, not an encapsulated .kvk file or .keyman-touch-layout
* Would look a lot like LDML layerMaps (with displayMap applied).
* Using synthesized vkeys for touch layouts (not ids)

mcdurdin:
.keyman-touch-layout is 18.0. .kvk is done.

darcywong00:
From A18S8 NA planning meeting, reassigning to @ermshiperete for A18S9

mcdurdin:
We probably won't be doing the touch layout generation this way but instead directly synthesizing it within Keyman Engine for Web from the KMX+ data. Pushing to 19.0 for consideration of touch layout support for kmn keyboards.

mcdurdin:
See #12620. May need to be in 18.0

PR #7380, opened 2022-09-29
feat(common): kvks writer for TypeScript 🙀

### LINKED ISSUE #7379: feat(common): write kvks file from Typescript
Extracted from #7355. We have support for read/write of KVK in that PR, and read of KVKS. Write of KVKS is not technically required for LDML project at this time, but for completeness, good to get it in.

mcdurdin:
Fixes #7379.

Relates to #7238.

Also splits attributes from sub-elements in kvks-file in order to make round-trip writing more pleasant.

Adds new round-trip test for kvks writer.

Removes incorrect property keybitmap, adds missing property layout.

@keymanapp-test-bot skip

### DIFF
--- a/common/schemas/kvks/kvks.schema.json
+++ b/common/schemas/kvks/kvks.schema.json
@@ -75,18 +75,27 @@
             "$ref": "#/definitions/kvk-layer"
           }
         },
-        "name": {
-          "$ref": "#/definitions/kvk-encoding-name"
-        },
-        "fontname": {
-          "type": "string"
-        },
-        "fontsize": {
-          "type": "string"
+        "$": {
+          "type": "object",
+          "properties": {
+            "name": {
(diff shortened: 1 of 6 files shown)


srl295 (approved):
LGTM. not clear why the element was renamed to `$` is that to save space?

mcdurdin:
> not clear why the element was renamed to `$` is that to save space?

Reason was: 

> Also splits attributes from sub-elements in kvks-file in order to make round-trip writing more pleasant.

That is, it also means a single intermediate in-memory `KVKSourceFile` structure, usable for both read/write, which makes the transforms clearer and DRYs out the code. The in-memory `KVKSourceFile` can mirror the file format fairly closely given we always do any real work with `VisualKeyboard`, not `KVKSourceFile`.

PR #7388, opened 2022-09-30
fix(android/engine): Dismiss subkeys on multi-touch

darcywong00:
Possibly addresses #6978

Keyman Engine for Android currently isn't set up to robustly handle multi-touch. If the user starts a long-press with one finger and switches layer with a second finger, it can cause KeymanWeb to process a subkey that doesn't match the base key anymore.

This PR causes subkeys to get dismissed when a multi-touch is detected.

## User Testing
**Setup**
1. Needs a physical Android device for multi-touch typing (two thumbs)
2. Install the PR build of Keyman for Android

* **TEST_MULTI_TOUCH_DISMISSES_LONGPRESS** - Verifies long-press keys aren't generated when multi-touch detected
1. Start Keyman for Android with the default sil_euro_latin keyboard.
2. Observe the OSK is on the Shift layer (upper-case)
3. With the right thumb, start long-press on G and hold. Immediately with the left thumb press and release the shift button to switch to Default layer
4. With the right thumb still held, observe the layer is default layer. Verify long-presses for uppercase-G **do not** appear.
5. Observe the base key that was held is output in the text.

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -302,6 +302,11 @@ final class KMKeyboard extends WebView {
       // if active, allowing for smooth, integrated gesture control.
       subKeysWindow.getContentView().findViewById(R.id.grid).dispatchTouchEvent(event);
     } else {
+      if (event.getPointerCount() > 1) {
+        // Multiple points touch the screen at the same time, so dismiss any pending subkeys
+        dismissKeyPreview(0);
+        dismissSubKeysWindow();
+      }
       gestureDetector.onTouchEvent(event);
     }
 


bharanidharanj:
* **TEST_MULTI_TOUCH_DISMISSES_LONGPRESS (FAILED):** Tested this with the attached PR build (16.0.72-alpha-test-7388) in my Android Mobile device (Ver 11.0) and noticed that even though the Keyboard layer was changed to default, the Capital letter 'G' appears on the text pane.

darcywong00:
@bharanidharanj 
The test steps say
*  Verify long-presses for uppercase-G do not appear.
* Observe the base key that was held is output in the text.

So, the long-press menu of "G" should not appear
but the base key that was held was "G" so the "G" should appear in the text.

@keymanapp-test-bot retest

bharanidharanj:
> @bharanidharanj The test steps say
> 
>     * Verify long-presses for uppercase-G do not appear.
> 
>     * Observe the base key that was held is output in the text.
> 
> 
> So, the long-press menu of "G" should not appear but the base key that was held was "G" so the "G" should appear in the text.
> 
> @keymanapp-test-bot retest

@darcywong00 Oh, Okay! Thank you so much for your clarification. Yes. It is working as expected.  :-)

bharanidharanj:
* **TEST_MULTI_TOUCH_DISMISSES_LONGPRESS (PASSED):** Tested this with the attached PR build (16.0.72-alpha-test-7388) in my Android Mobile device (Ver 11.0) and it is working as expected. (ie., the base key that was held is 'G' outputs in the text pane. Thanks for your clarification, Darcy!

mcdurdin (approved):
LGTM

I think this makes sense to me: it's probably the cleanest way to avoid the conflict. As you recognised, either starting a second touch or ending the first touch should always cancel the longpress gesture.

This may need revisiting with the new gestures work but that's a while away anyway.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.74-alpha](https://keyman.com/downloads/releases/alpha/16.0.74)

PR #7390, opened 2022-10-03
chore(android/engine): Display Toast notifications on Sentry errors

### LINKED ISSUE #7189: chore(android): Add Toast notifications to KMLog
Ironically, I would suggest adding toasts to LogError and LogException if they are not there already, so the user knows something has gone wrong and can report the error to us (assuming they remember the message at all!)

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/7178#discussion_r960186869_

darcywong00:
Fixes #7189
This updates the KMLog error handling to also display a Toast notification of the error message (provides user additional info about errors)

* Some classes aren't an AppCompatActivity, so they'll need BaseActivity.makeToast() to display the notification.
* Also updated the error string when downloads fail to use the destinationFilename, since filename could be blank.
* Found that api.keyman returns a valid empty array if no matching lexcal-model for a language is found. So I've removed that LogError.

Note: with this change, the localized strings used for Toast notifications are passed on to Sentry logging...

## User Testing
**Setup**
Install the PR build of Keyman for Android, and start the test with the device having network access

* **TEST_LOG_ERROR** - Verifies KMLog.LogError displays Toast notifications
1. Start Keyman for Android
2. In the Keyman app, go to Settings --> search for a Keyman keyboard (e.g. sil_cameroon_qwerty), but **do not** click the green button to install
3. Swipe down on the system notifications panel and set the device in airplane mode (disabling network access)
4. Back in the Keyman app, click on the green button to install the keyboard
5. Observe the keyboard fails to download
6. Verify several Toast notifications appear, one of which is "Could not download filename sil_cameroon_qwerty.kmp"

* **TEST_LOG_EXCEPTION** - Verifies KMLog.LogException displays Toast notifications
1. Start Keyman for Android
2. Launch a separate browser and download this invalid kmp package (the kmp.json file doesn't contain a valid keyboard version)
https://darcywong00.github.io/examples/invalid/khmer10/khmer10_noKbVer.kmp
3. When the kmp file downloads, click to open it (the file association should be handled in Keyman)
4. In the keyboard installation process, click "INSTALL"
5. Verify a Toast notification appears "No keyboards or predictive text to install"

* **TEST_KMP_INSTALL** - Verifies valid kmp package installs
1. Start Keyman for Android
2. Go to Keyman settings --> Install a keyboard from keyman.com (e.g. sil_cameroon_qwerty)
3. Continue with the keyboard package installation, picking "Akum" language, and install
4. Verify keyboard package installs and no Toast errors are displayed.

### DIFF
--- a/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/PackageActivity.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/PackageActivity.java
@@ -62,8 +62,7 @@ public class PackageActivity extends AppCompatActivity implements
     if (bundle != null) {
       kmpFile = new File(bundle.getString("kmpFile"));
       if (!kmpFile.exists()) {
-        KMLog.LogError(TAG, kmpFile.getAbsolutePath() + " not found. Unable to extract");
-        showErrorToast(context, getString(R.string.failed_to_extract));
+        showErrorToast(kmpFile.getAbsolutePath() + " not found. Unable to extract");
         return;
       }
       installMode = KmpInstallMode.fromString(bundle.getString("installMode"));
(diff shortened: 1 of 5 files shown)


jahorton:
Well... I guess this is one way to find out more information about some of our more persistent errors.  There are some Sentry events we've been receiving that haven't had any prior user feedback or concerns, so... this may cause an interesting reaction from some people.

mcdurdin:
I think we run this mode for alpha/beta -- more feedback is better in those versions. We can make an assessment as to whether we should run this for stable during the beta phase.

bharanidharanj:
* **TEST_LOG_ERROR (PASSED):** Tested this with the attached PR build (Keyman 16.0.73-alphat-test-7390) in my Android Mobile device (Ver 11) and it is showing the toast message 'Could not download filename sil_cameroon_qwerty.kmp' while trying to download the keyboard. 

   


* **TEST_LOG_EXCEPTION (PASSED):** Tested this with the attached PR build (Keyman 16.0.73-alphat-test-7390) in my Android Mobile device (Ver 11) and noticed that an error message appears after clicking the Install option.

bharanidharanj:
* **TEST_KMP_INSTALL (PASSED):** Tested this as per the instructions and I noticed that no toast notifications appear during the installation.

mcdurdin (approved):
Changes LGTM.

Let's review error toast frequency in use and if it is a problem, then obviously we need to be fixing the problems but also consider if we turn some of the (harmless?) toasts off for release builds.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.74-alpha](https://keyman.com/downloads/releases/alpha/16.0.74)

PR #7396, opened 2022-10-04
chore(linux): Cleanup unnecessary surrogate check

### LINKED ISSUE #6385: chore(linux): cleanup unnecessary surrogate check in engine.c
[`engine.c`](https://github.com/keymanapp/keyman/blob/93af2a3d84489e15f514c553546805c2fe90fd5e/linux/ibus-keyman/src/engine.c#L524) checks for `action_item->character` being a surrogate. This should never happen because `action_item->character` is a 32-bit character. We should verify that we don't hit this block in any test, and then remove this code block and remove the `keyman->firstsurrogate` field.
I suggest doing that in Keyman 16.

_Originally posted by @ermshiperete in https://github.com/keymanapp/keyman/issues/5615#issuecomment-1068360858_

ermshiperete:
Previously `process_unicode_char_action` checked for `action_item->character` being a surrogate. This should never happen because `action_item->character` is a 32-bit character. This change removes the unnecessary check and the `keyman->firstsurrogate` field.

Fixes #6385.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -64,7 +64,6 @@ struct _IBusKeymanEngine {
     gchar           *ldmlfile;
     gchar           *kb_name;
     gchar           *char_buffer;
-    gunichar         firstsurrogate;
     gboolean         lctrl_pressed;
     gboolean         rctrl_pressed;
     gboolean         lalt_pressed;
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.76-alpha](https://keyman.com/downloads/releases/alpha/16.0.76)

Issue #7404, opened 2022-10-04
bug(developer):  0-th elem needs zero offset 🙀 


srl295:
https://github.com/keymanapp/keyman/blob/feature-ldml/core/src/ldml/C7043_ldml.md#c704323-elemtransform-and-reorder-element-strings says about `elem`

> The first entry in the element string list MUST be the null element string, which has zero length and zero offset.

but on inspection, the first (0th) element string list has a zero length, but repeats the offset of the second (1st) entry.

-----
for #7042 
found during #7234

mcdurdin:
See also https://github.com/keymanapp/keyman/pull/7437/files#diff-95985cb44c34779804524113d19f84c84c0df3809790667a95bd3592e3a9e7cbR307

PR #7410, opened 2022-10-05
fix(developer): kmc: 0-th elem needs 0 offset #7404 

srl295:
Fixes: bug(developer):  0-th elem needs zero offset 🙀  #7404

@keymanapp-test-bot skip


srl295:
@mcdurdin wanted to jump into kmc a little.  Seems to work, but should ideally have a test/assertion

mcdurdin:
I note the build failed, probably because binary data no longer matches for the e2e test (haven't investigated). Additional test would be good.

srl295:
> I note the build failed, probably because binary data no longer matches for the e2e test (haven't investigated). Additional test would be good.

yes that's why it's a draft.  I did notice unit tests on the binary format also.

srl295:
> I note the build failed, probably because binary data no longer matches for the e2e test (haven't investigated). Additional test would be good.

it was the e2e test fixture, now updated.

mcdurdin (approved):

PR #7421, opened 2022-10-09
feat(common): add keyman-touch-layout format and readers/writers for typescript 🙀

mcdurdin:
Adds validating reader and conforming writer for .keyman-touch-layout files.

Does not yet write out to the 'compiled' format used by KeymanWeb.

@keymanapp-test-bot skip


common/web/types/src/keyman-touch-layout/keyman-touch-layout-file-reader.ts
```diff
@@ -0,0 +1,80 @@
...
+export class TouchLayoutFileReader {
+  public read(source: Uint8Array): TouchLayoutFile {
+    const decoder = new TextDecoder('utf-8', {fatal:true, ignoreBOM: true});
+    let sourceString: string;
+    try {
+      sourceString = decoder.decode(source);
+    /* c8 ignore next 7 */
+    } catch(e) {
+      if(e instanceof TypeError) {
+        // TODO: Do we want to do something else with this?
```
srl295, line 13: Maybe, it's an unrecoverable error in any event though

common/web/types/src/keyman-touch-layout/keyman-touch-layout-file.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export interface TouchLayoutFile {
+  tablet?: TouchLayoutPlatform;
+  phone?: TouchLayoutPlatform;
+  desktop?: TouchLayoutPlatform;
+};
+
+export type TouchLayoutFont = string;
+export type TouchLayoutFontSize = string;
+export type TouchLayoutDefaultHint = "none"|"dot"|"longpress"|"multitap"|"flick"|"flick-n"|"flick-ne"|"flick-e"|"flick-se"|"flick-s"|"flick-sw"|"flick-w"|"flick-nw";
```
srl295, line 18: This reminds me that the 'directions' schema should have a regex to validate this.

Would it be helpful in general to have supplemental metadata with a direction list?

srl295: Started a DTD PR https://github.com/unicode-org/cldr/pull/2437 - now back to this review

common/web/types/src/keyman-touch-layout/keyman-touch-layout-file.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export interface TouchLayoutFlick {
+  n?: TouchLayoutSubKey;
+  s?: TouchLayoutSubKey;
+  e?: TouchLayoutSubKey;
+  w?: TouchLayoutSubKey;
+  ne?: TouchLayoutSubKey;
+  nw?: TouchLayoutSubKey;
+  se?: TouchLayoutSubKey;
+  sw?: TouchLayoutSubKey;
```
srl295, line 86: a little surprising to see these enumerated, wouldn't it be more flexible to just have a map? Seems a little WET :)

srl295 (approved):
comments but LGTM

common/web/types/src/keyman-touch-layout/keyman-touch-layout-file-reader.ts
mcdurdin: Unrecoverable but we want to turn it into a friendly compiler error rather than a compiler crash

common/web/types/src/keyman-touch-layout/keyman-touch-layout-file.ts
mcdurdin: > Would it be helpful in general to have supplemental metadata with a direction list?

I think that given we are only supporting unidirectional flicks at this point, it doesn't help all that much.

common/web/types/src/keyman-touch-layout/keyman-touch-layout-file.ts
mcdurdin: This is a type, so it verifies that we have these and only these properties. I don't think we can do that in a map in Typescript types?

PR #7422, opened 2022-10-09
feat(developer): generate .js from ldml .xml 🙀

mcdurdin:
Relates to #7238.

Compiles a LDML keyboard .xml down to a KeymanWeb .js file. Currently only supports some metadata.

Next step is to generate touch layout data, then to embed the binary kmx data.

@keymanapp-test-bot skip


jahorton (commented):
Not that it has to be done yet, but it might be worth a unit test verifying that produced keyboards may be loaded safely by Web's `keyboard-processor` module.  I realize that checking against schemas does provide a similar role... but this would give us integration testing potential.

Of course, at this stage, there's not any _keystroke processing_, but even 'loading without an error' would be noteworthy.  And, technically, you should be able to retrieve the keyboard's layouts (once they're compiled - I see `compileTouchLayout` is a no-op at present) and verify that the expected layers, rows, and keys are present at that level.

Then, once we actually do start baking in the LDML core, it becomes possible to verify that Web is properly processing LDML-based keystroke 'rules.'

jahorton (approved):
That said... LGTM as is.

developer/src/kmc-keyboard/src/compiler/keymanweb-compiler.ts
```diff
@@ -0,0 +1,117 @@
...
+      // TODO-LDML: fill in KLS
+      `${tab}}`;
+
+    return result;
+  }
+
+  public compileTouchLayout(source: LDMLKeyboard.LDMLKeyboardXMLSourceFile) {
+    // const tlc = new TouchLayoutCompiler();
+    // const layout = tlc.compile(source);
+    // TODO-LDML
```
srl295, line 42: TODO: implement ! 😸

srl295 (approved):
i don't know all of the structure here but LGTM

mcdurdin:
> Not that it has to be done yet, but it might be worth a unit test verifying that produced keyboards may be loaded safely by Web's `keyboard-processor` module. I realize that checking against schemas does provide a similar role... but this would give us integration testing potential.

Yes, end-to-end integration testing would be a great future goal.

PR #7433, opened 2022-10-11
chore(ios): project changes for xcode 14 compatibility

sgschantz:
To upgrade to Xcode 14 (needed to build and debug apps withiOS 16), two changes generated warnings and caused the build to fail. These changes are described in the Xcode 14 [release notes](https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes):

1) increase the iOS target version from 9 to 11

>The Xcode 14 release supports on-device debugging in iOS 11 and later, tvOS 11 and later, and watchOS 4 and later. Xcode 14 requires a Mac running macOS Monterey 12.5 or later.

2) disable bit code

The xcode 14 release notes describe the deprecation of bit code:

>Deprecations
Starting with Xcode 14, bitcode is no longer required for watchOS and tvOS applications, and the App Store no longer accepts bitcode submissions from Xcode 14.
Xcode no longer builds bitcode by default and generates a warning message if a project explicitly enables bitcode: “Building with bitcode is deprecated. Please update your project and/or target settings to disable bitcode. The capability to build with bitcode will be removed in a future Xcode release. IPAs that contain bitcode will have the bitcode stripped before being submitted to the App Store. Debug symbols for past bitcode submissions remain available for download. (86118779)


mac/Keyman4MacIM/Keyman4MacIM/KMInputMethodAppDelegate.m
```diff
@@ -87,40 +88,50 @@ - (id)init {
...
-        }
-
-        self.runLoopEventSrc = CFMachPortCreateRunLoopSource(kCFAllocatorDefault, self.lowLevelEventTap, 0);
 
-        CFRunLoopRef runLoop = CFRunLoopGetCurrent();
-
-        if (self.runLoopEventSrc && runLoop) {
-            CFRunLoopAddSource(runLoop,  self.runLoopEventSrc, kCFRunLoopDefaultMode);
-        }
+      // first notify user and request access to Accessibility/PostEvent permissions
```
mcdurdin, line 92: These changes don't look like they belong in this PR.

mcdurdin (changes requested):
A few thoughts here. Version bumps are real busy-work, and seem to generate lots of tidy up. But it's important that we cover all the bases.

* It looks like your mac privacy fixes crept into this PR, so they should be moved into a separate PR.

* We need to make sure that we can continue to build earlier versions of Keyman targeting 9.0, so build agents won't update to xcode 14 at this time. So these projects need to continue to be able to build and deploy on older xcode, for now.

* Are there any impacts on macOS builds in upgrading to xcode 14? (e.g. I see this in the release notes: "Building for deployment to OS releases older than macOS 10.13, iOS 11, tvOS 11, and watchOS 4 is no longer supported. (92834476)")

* You'll also need to fixup the iOS sample projects (e.g. KMSample2 still has bitcode enabled, so it fails to build) and any other iOS projects such as KeyboardCalibrator.

* You'll need to write up some user tests (we'll do a regression test across supported iOS versions after this merges, though hopefully there won't be any issues).

* You should remove any existing code which is targeting iOS versions older than 11.0: search for `#available` in *.swift, looks like there are a few modules in KeyboardCalibrator.

* This will need at least four 'breaking change' release notes (add these to the sprint retrospective):
  * iOS app (because of min iOS version)
  * iOS engine (because of min iOS version and bitcode changes)
  * macOS app (because of min macOS version)
  * macOS engine (because of min macOS version)

* We should make sure we loop in the App Builder team (@chrisvire, @rpmargetts) so they have a heads-up on the changing requirements for Keyman 16. They'll probably already be dealing with it in their own move to iOS 16 but it's nice to be informed anyway! :grin:

chrisvire:
> * We should make sure we loop in the App Builder team (@chrisvire, @rpmargetts) so they have a heads-up on the changing requirements for Keyman 16. They'll probably already be dealing with it in their own move to iOS 16 but it's nice to be informed anyway! 😁

Please include @davidmoore1. He has done work in app-builders for Xcode 14 and iOS 16 (including removing bitcode). He has also done work in a fork of keyman to support Carthage.

mcdurdin:
> Please include @davidmoore1. He has done work in app-builders for Xcode 14 and iOS 16 (including removing bitcode). He has also done work in a fork of keyman to support Carthage.

Righto, will do. We'd also be interested at some point in understanding what is diverging in the fork and what we should be merging back upstream into the main Keyman project.

mcdurdin:
Please see #7483 for additional work we'll need to do in CI when we move to XCode 14.

chrisvire:
@mcdurdin: You can have multiple.  The question is whether `xcodebuild` will work correctly after each `xcode-select` call and if there is anything that resides outside of the Xcode install folder.

https://stackoverflow.com/questions/669367/can-i-have-multiple-xcode-versions-installed

mcdurdin:
Thanks @chrisvire, that's helpful info. Sad that we need to `sudo xcode-select` because that makes it hard to script in TeamCity... some exploration needed I guess.

darcywong00:
Is this replaced by #8015 ?

mcdurdin:
Replaced by #8015.

Issue #7397, opened 2022-10-04
bug(linux): debian build has man in wrong place


srl295:
**Keyman for Linux:**
 - Device:           VM
 - OS:                 Debian 11.5, x64
 - Keyman version: 04bde69ab6464c1e9a1db742a4fd3f6f3b05c6c6 16.0.72 alpha 2022-09-30

-----

linux/keyman-config/build-help.sh has:

```sh
    echo "Generating man pages..."
    mkdir -p ../../debian/man
```

But this generates man pages in `linux/keyman-config/../../debian`  However the installer (and `.gitignore`!) expect them in `linux/keyman-config/debian/man` 

This causes `make install` to fail.

Workaround:  `cd linux ; mv -v ../debian keyman-config/`

srl295:
version is recent `master` — 04bde69ab6464c1e9a1db742a4fd3f6f3b05c6c6 installed from git.

Issue #7400, opened 2022-10-04
bug(linux): km-config did not work on Debian 11.5


srl295:
**Keyman for Linux:**
 - Device:           VM
 - OS:                 Debian 11.5, x64
 - Keyman version: 04bde69ab6464c1e9a1db742a4fd3f6f3b05c6c6 16.0.72 alpha 2022-09-30

-----

km-config failed to start up with

```
Can not find lang_tags_map.py. Did you run "make"?
```

however, the file WAS inside the .egg file.

Fixed by (!!) setting `zip_safe=False` in the `setup.py` so that it installed a .egg dir instead of a .egg file.  Somehow, this worked.

mcdurdin:
Please include Keyman version info and method of installation for future reference :grin:

srl295:
version is recent `master` — 04bde69ab6464c1e9a1db742a4fd3f6f3b05c6c6 installed from git.

ermshiperete:
Updated `README.md` in #7434.

srl295:
@ermshiperete I'm still hitting on Ubuntu 22.04.1 LTS (happens to be WSL)

```shell
$ make reconf && make fullbuild && sudo make install
$ km-config
Can not find lang_tags_map.py. Did you run "make"?
```


and still 'fixable' with

```diff
diff --git a/linux/keyman-config/setup.py b/linux/keyman-config/setup.py
index c090be9fb..ba5b6f957 100644
--- a/linux/keyman-config/setup.py
+++ b/linux/keyman-config/setup.py
@@ -14,6 +14,8 @@ setup(
              'km-package-uninstall',
              'km-package-list-installed'],
 
+    zip_safe=False,  # SRL: don't install a zip
+
     install_requires=[
         'lxml', 'numpy', 'Pillow', 'requests', 'requests-cache',
         'python-magic', 'qrcode', 'sentry-sdk'
```

do you want a new ticket?

srl295:
@ermshiperete looks like this change didn't make it to feature-ldml yet (which is fine), but this patch does still seem necessary for me otherwise.

Issue #7398, opened 2022-10-04
bug(linux): mk-build-deps failed for Debian 11.5


srl295:
**Keyman for Linux:**
 - Device:           VM
 - OS:                 Debian 11.5, x64
 - Keyman version: 04bde69ab6464c1e9a1db742a4fd3f6f3b05c6c6 16.0.72 alpha 2022-09-30

-----

```shell
$ sudo mk-build-deps --install linux/debian/control
…
…
Investigating (0) keyman-build-deps:amd64 < 15.0.270-1 @iU mK Nb Ib >
Broken keyman-build-deps:amd64 Depends on python3-sentry-sdk:amd64 < 0.13.2-1 @ii mK > (>= 1.1)
  Removing keyman-build-deps:amd64 because I can't find python3-sentry-sdk:amd64
Broken keyman-build-deps:amd64 Depends on python3-raven:amd64 < none @un H >
  Removing keyman-build-deps:amd64 because I can't find python3-raven:amd64
  Or group remove for keyman-build-deps:amd64
```

srl295:
version is recent `master` — 04bde69ab6464c1e9a1db742a4fd3f6f3b05c6c6 installed from git.

ermshiperete:
Debian 11.5 Bullseye contains a version of `python3-sentry-sdk` that is too old. The workaround is to install a newer package version from the next Debian version, from packages.sil.org, or via pip. I updated the `README.md` with this information in #7434.

PR #7434, opened 2022-10-11
fix(linux): Fix make install

ermshiperete:
Several fixes for building and installing locally.

Fixes https://github.com/keymanapp/keyman/issues/7397, https://github.com/keymanapp/keyman/issues/7398, https://github.com/keymanapp/keyman/issues/7400

Also add separate READMEs for the legacy projects, and removes mentioning of the legacy projects in the current READMEs.
The motivation is to make it later easier to remove the legacy projects, as well as reducing the efforts to testing the README with
legacy projects. Since legacy projects aren't maintained anymore it doesn't make sense to still have them in the regular README,
suggesting that the instructions work for the legacy projects as well.

And finally ignore `.devcontainer` folder. [VSCode dev containers](https://code.visualstudio.com/docs/remote/containers)
allows to develop in a container and is configured through files in `.devcontainer`. However, that directory falls in the same
category as `.vscode` in that it might be different on each developer's machine, depending on preference and platform he's
working on. Therefore we exclude this directory similar to `.vscode`.

@keymanapp-test-bot skip


linux/LICENSE.md
```diff
@@ -1,17 +1,16 @@
 # LICENSE
 
-The [kmflcomp](./kmflcomp), [libkmfl](./libkmfl), and [keyman_config](./keyman_config) projects
-are covered by the [MIT license](./libkmfl/COPYING).
+The [keyman-config](./keyman-config) project iscovered by the
```
mcdurdin, line 3:
```suggestion
The [keyman-config](./keyman-config) project is covered by the
```

linux/README.md
```diff
@@ -56,37 +54,30 @@ See [license information](./LICENSE.md) about licensing.
 
 Used by TC for validating PRs
 
-Run `make tmpinstall` to build and install **keyboardprocessor**, **kmflcomp**, **libkmfl**, **ibus-kmfl** and **ibus-keyman** to `/tmp/kmfl`
+Run `make tmpinstall` to build and install **keyboardprocessor** and **ibus-keyman** to `/tmp/kmfl`
```
mcdurdin, line 57: Huh, it still installs to `/tmp/kmfl`?

linux/keyman-config/setup.py
```diff
@@ -20,8 +20,8 @@
     ],
 
     # metadata to display on PyPI
-    author="Daniel Glassey",
-    author_email="wdg@debian.org",
+    author="Keyman team",
+    author_email="support@keyman.com",
```
mcdurdin, line 24: :+1:

mcdurdin (approved):
LGTM, may be good to wait for @srl295 review also

linux/keyman-config/Makefile
```diff
@@ -8,17 +8,20 @@ langtags:
...
 	pip3 install qrcode sentry-sdk
+	# eventually change this to: pip3 install .
 	python3 setup.py install
 	# install icons
 	mkdir -p /usr/local/share/keyman/icons
 	cp keyman_config/icons/* /usr/local/share/keyman/icons
 	# install man pages
 	mkdir -p /usr/local/share/man/man1
-	cp debian/man/*.1 /usr/local/share/man/man1
+	cp ../../debian/man/*.1 /usr/local/share/man/man1
```
srl295, line 18: hrm … - okay, but ../../debian/man isn't in .gitignore, is it?

srl295 (approved):
Generally LGTM seems like it will fix the issues, think there may be a .gitignore issue still

ermshiperete:
> hrm … - okay, but ../../debian/man isn't in .gitignore, is it? 

Hmm, you're right. It's ignored on my machine, but why?

linux/keyman-config/Makefile
ermshiperete: > hrm … - okay, but ../../debian/man isn't in .gitignore, is it? 

Hmm, you're right. It's ignored on my machine, but why?

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.78-alpha](https://keyman.com/downloads/releases/alpha/16.0.78)

.gitignore
```diff
@@ -172,3 +173,14 @@ Thumbs.db
...
+# Linux packaging related
+/debian/
+results/
+*.deb
+*.ddeb
+*.dsc
+*.build
+*.buildinfo
+*.changes
+*.tar.?z
```
mcdurdin, line 186: These really should _not_ be in the top-level .gitignore. Can you move them to `linux/.gitignore`, or prefix as needed with `/linux/` please?

If it turns out that the Debian build is scattering files everywhere, let's try and fix that instead of using this big and dangerous hammer.

ermshiperete: Unfortunately these files do get generated in the root of the Keyman tree, so they'll have to live in this file. However, I limited the scope (PR #7451).
I also created an issue to try to refactor the Linux packaging (#7450).

Issue #7234, opened 2022-09-08
feat(developer): ldml - support UnicodeSets in LDML keyboards 🙀


mcdurdin:
UnicodeSets -- brain dump from discussion earlier this week

* Parser within compiler
* Appropriate binary form
  * Usage within keyboardprocessor: for now, will be a pure string of characters to choose from. later needs a performant binary format.
  * → UnicodeSet subset for ldml keyboard spec?
  * Allow UnicodeSets up to 1024 chars max for sanity for now. This will need to change because it doesn't cope with inverse sets.
  * Supported patterns
`\uxxxx`
`[abc]`
`[a-z123]`
`[\u1234-\u2468]`

Nopes for now:
* `[\u0001-\u10ffff]` ← binary format issues
* `[^a]` ← binary format issues
* `[]`  ← disallow, non-character

Refs:

* Intl  -> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
* https://github.com/tc39/ecma402
* https://caniuse.com/?search=intl
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes 
* https://github.com/node-unicode/node-unicode-data
* UnicodeSet in rust: https://unicode-org.github.io/icu4x-docs/doc/icu_collections/index.html

mcdurdin:
Will be using ICU components to transform to regex in compilation. Closing as discussion here is ancient history.

mcdurdin:
Turns out that we need to implement a UnicodeSet parser in the compiler. But the discussion here is still out of date.

srl295:
Just to consolidate this:

- linked PR pulls in ICU4C (but patched to not support properties)
- ICU4X TODO issue: https://github.com/unicode-org/icu4x/issues/168 
- Filed upstream ICU ticket https://unicode-org.atlassian.net/browse/ICU-22375 for changes

PR #7437, opened 2022-10-11
feat(core): kmx_plus: add accessors for elem 🙀

srl295:
- add accessors and tests for `elem` and related data structures
- other refactor and cleanup in kmx_plus, change to use kmx_debug debugging instead of printf/cerr

for #7234

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -17,26 +17,22 @@ namespace km {
 namespace kbp {
 namespace kmx {
 
+
 /**
- * @brief convert a section name to string, for debugging
- * @param ident the hex dword
- * @param buf buffer, must be size 5
- * @return const pointer to buffer
+ * \def _DEBUG_IDENT_SAFE for use by DEBUG_IDENT
+*/
+#define _DEBUG_IDENT_SAFE(x,b) (((x)>>b)&0x7F)<0x20?'?':(((x)>>b)&0x7F)
+
+/**
+ * \def DEBUG_IDENT
+ * Expands to four chars:  a,b,c,d
+ * for purposes of debug logging a hex identifier
  */
-inline const char *
-section_name_to_str(KMX_DWORD ident, char *buf) {
-  for (int i = 0; i < 4; i++) {
-    unsigned char ch = ident & 0xFF;
-    if (ch < 0x20 || ch > 0x7F) {
-      buf[i] = '?';
-    } else {
-      buf[i] = (char)ch;
-    }
-    ident >>= 8;
-  }
-  buf[4]=0;
-  return buf;
-}
+#define DEBUG_IDENT(x) \
+  _DEBUG_IDENT_SAFE(x,0), \
+  _DEBUG_IDENT_SAFE(x,8), \
+  _DEBUG_IDENT_SAFE(x,16), \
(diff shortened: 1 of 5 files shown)


srl295:
> Logging -- excellent, so much better. A separate PR to refactor the DebugLog function out of the kmx processor may be in our future.
> 
> I note that it isn't building yet, but code so far lgtm.

PTAL , was a problem with the debug messages

core/src/kmx/kmx_plus.cpp
```diff
@@ -9,24 +9,34 @@
...
-#define KMXPLUS_PRINTF(x)
-#endif
+/**
+ * @brief convert a section name to string, for debugging
+ * @param ident the hex dword
+ * @param buf buffer, must be size 5
+ * @return const pointer to buffer
+ */
+inline const char *
+section_name_to_str(KMX_DWORD ident, char *buf) {
```
mcdurdin, line 27: I'm just having a bit of fun here, but another way of skinning this would be:

```c++
#define _DEBUG_IDENT_SAFE(x,b) (((x)>>b)&0x7F)<0x20?'?':(((x)>>b)&0x7F)
#define DEBUG_IDENT(x) \
  _DEBUG_IDENT_SAFE(x,0), \
  _DEBUG_IDENT_SAFE(x,8), \
  _DEBUG_IDENT_SAFE(x,16), \
  _DEBUG_IDENT_SAFE(x,24)

  ...
  DebugLog("%c%c%c%c", DEBUG_IDENT(ident));
```

The primary advantage is it avoids loops (even if they are unwound), and reduces the debug statements to a single line, which reduces noise in the source.

core/src/kmx/kmx_plus.cpp
```diff
@@ -277,34 +281,109 @@ COMP_KMXPLUS_SECT::valid(KMX_DWORD length) const {
...
+      return false;
+    }
+  }
+  return true;
+}
+
+const COMP_KMXPLUS_ELEM_ELEMENT *
+COMP_KMXPLUS_ELEM::getElementList(KMX_DWORD elementNumber, KMX_DWORD &length) const {
+  if (elementNumber >= count) {
+    return nullptr;
```
mcdurdin, line 330: This should have a DebugLog statement and/or debug-build assertion.

core/src/kmx/kmx_plus.cpp
```diff
@@ -277,34 +281,109 @@ COMP_KMXPLUS_SECT::valid(KMX_DWORD length) const {
...
+  }
+  const COMP_KMXPLUS_ELEM_ENTRY &entry = entries[elementNumber];
+  length = entry.length;
+  if (length == 0) {
+    return nullptr;
+  }
+  if (entry.offset + (entry.length * sizeof(COMP_KMXPLUS_ELEM_ELEMENT)) > header.size) {
+    DebugLog("ERROR: !! COMP_KMXPLUS_ELEM::getElementList");
```
mcdurdin, line 338: This needs to return nullptr

mcdurdin (approved):
Minor fixes but otherwise LGTM

core/src/kmx/kmx_plus.cpp

core/src/kmx/kmx_plus.cpp
srl295: done

Issue #7418, opened 2022-10-07
bug(core): ldml_processor  needs to handle utf-32 items in context 🙀 


srl295:
ldml_processor is likely to slice UTF-32 codepoints into surrogate pairs, such as the character "🙀"

```c
      const std::u16string str = vkeys.lookup(vk, modifier_state);
…
      for(size_t i=0; i<str.length(); i++) {
        // TODO-LDML: needs to be per UTF-32 char? Seems this would push surrogates.
        state->context().push_character(str[i]);
        state->actions().push_character(str[i]);
      }
```

This should push one utf-32 character, not two surrogate pairs.

Context:  found during #7234 

#5015

srl295:
Fixed in #7443

PR #7443, opened 2022-10-13
feat(core): ldml: Use new UTF-16 to UTF-32 function 🙀

srl295:
- also a test case with a 🙀 key on the tiny keyboard

For #7418

Note this will currently conflict with #7437 and following

@keymanapp-test-bot skip

### DIFF
--- a/core/src/ldml/ldml_processor.cpp
+++ b/core/src/ldml/ldml_processor.cpp
@@ -168,8 +168,22 @@ ldml_processor::process_event(
 
     switch (vk) {
     case KM_KBP_VKEY_BKSP:
-      state->context().pop_back();
-      state->actions().push_backspace(KM_KBP_BT_UNKNOWN); // Assuming we don't know the character
+      {
+        KMX_DWORD last_char = 0UL; // Assuming we don't know the character
+        // attempt to get the last char
+        auto end = state->context().rbegin();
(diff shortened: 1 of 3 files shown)


core/src/ldml/ldml_processor.cpp
```diff
@@ -195,9 +195,10 @@ ldml_processor::process_event(
         return KM_KBP_STATUS_OK; // Nothing to do- no key
       }
       const std::u16string &str = key->second;
-      for(size_t i=0; i<str.length(); i++) {
-        state->context().push_character(str[i]);
-        state->actions().push_character(str[i]);
+      const std::u32string str32 = kmx::u16string_to_u32string(str);
+      for(size_t i=0; i<str32.length(); i++) {
+        state->context().push_character(str32[i]);
+        state->actions().push_character(str32[i]);
```
srl295, line 201: Surprisingly, this fix is actually _not needed_. It seems to have no effect on the output text.

srl295: in any event, …
aha! deletion event.  Backspace breaks it… testing

srl295: yup that was it.

mcdurdin (approved):

Issue #7337, opened 2022-09-21
bug(windows): [on behalf of user]  Something causes the left Alt key to be ‘stuck’


MakaraSok:
**Describe the bug**

> Quite frequently, when I switch between keyboards, something causes the left Alt key to be stuck down. This is at the software level: if I open the on-screen keyboard the left Alt key is purple instead of dark gray. Can anyone suggest what might be causing this? I’ve tried to pay attention to what key I might be hitting by mistake, but I can’t notice a pattern.
>
> This is Windows 11. I’ve disabled every setting I can find to do with sticky keys.
>
> My keyboard-switching shortcuts are Ctrl-Shift-1, Ctrl-Shift-2, etc.

**Expected behavior**

**Screenshots**

**Related issues**

---

**Keyman for Windows/macOS/Linux/iPhone/iPad/Android:**
   Device:             
 - OS:                  Windows 11
 - Keyman version:      15.0.268
 - Target Application:  Excel?

---

**Keyboard**
 - Keyboard name:     user's IPA keyboard?
 - Keyboard version:  ?
 - Language name:     IPA

**Additional context**

PR #7449 - Makes an improvement to by removing one possible cause for the ALT modifier becoming "stuck."
The issue still persists and therefore this issue cannot yet be closed.

MakaraSok:
originally posted at https://community.software.sil.org/t/something-causes-the-left-alt-key-to-be-stuck/6609/8

mcdurdin:
See also https://community.software.sil.org/t/shift-key-sticks-while-using-keyboard-in-paratext-9/6715

Related to #4884, #5190?

nilphite:
I’ve been having exactly the same issue for several weeks, even after updating to the latest version of Keyman. It affects all my Keyman keyboards. Tapping the alt key toggles it off for one character, and then it locks on again. It happens inconsistently; sometimes I can use the keyboards fine for extended periods and then it’ll just lock alt unpredictably. I’m running Windows 11 Developer Preview.

rc-swag:
PR https://github.com/keymanapp/keyman/pull/7449 - Makes an improvement to by removing one possible cause for the ALT modifier becoming "stuck."
The issue still persists and therefore this issue cannot yet be closed.

rc-swag:
Another user finding the same issue.
https://community.software.sil.org/t/typing-a-character-using-keyman-seems-to-activate-a-modifier-key/6819

mcdurdin:
[...]

PR #7449, opened 2022-10-13
fix(windows): ensure all modifier events go to serialized queue

rc-swag:
Looking at the logs it seems to indicate it occurs when switching between keyman and non-keyman keyboards. The use of <kbd>Alt</kbd> + <kbd>Tab</kbd> to switch applications. #6392 made a change that meant key events were only sent to the keyman serialized queue when a Keyman keyboard was active. This resolved the issue with EMOJI input pop-up. However, it is likely the cause as to why the modifier keys are sometimes being seen out of sequence.

This change makes it so that all modifier events are sent to the serialized server so that the modifier state does not lose synchronization.

# USER TESTING :

This PR changes code around modifier keys. Therefore the tests need to cover interactions around these keys being pressed.
Unfortunately, I was not able to find a way to reproduce the behaviour to test it is resolved. We need to make sure though that
there are no regressions.

**GROUP_WIN10**  - Run the tests in a Windows 10 OS
**GROUP_WIN11** - Run the tests in Windows 11 OS

**TEST_ALTGR_IN_RULES** 
This is to test the key rules which contain <kbd>AltGr</kbd> or <kbd>Shift</kbd> + <kbd>AltGr</kbd> still work correctly.

* Open LibreOffice or Notepad 
* Observe output of  ឱ​

* Observe output of ᧨

**TEST_HOTKEYS**
This is to test switching between non-keyman and keyman keyboards using hotkeys. To make sure this PR has not broken Hotkeys that use modifiers 
* Install two different keyman keyboards in addition to the default for the Windows installation.
* Use the Keyman Configuration to assign hotkeys to the default Keyboard as well as the two Keyman Keyboards that use modifiers for the hotkey for example <kbd>Ctrl</kbd> + <kbd>1</kbd>  or <kbd>Alt</kbd> + <kbd>K</kbd>
* Open Notepad 
* Use the hotkeys to switch between keyboards
* Verify that switching between the keyboards is working

**TEST_HOTKEYS_SWITCH_APPS** 
This extends the previous test to test switching between non-keyman and keyman keyboards using hotkeys and also using <kbd>Alt</kbd> + <kbd>Tab</kbd> to switch applications.
* Install two different keyman keyboards in addition to the default for the Windows installation.
* Use the Keyman Configuration to assign hotkeys to the default Keyboard as well as the two Keyman Keyboards
* Open two text input applications for example Libre Office and Notepad.
* Turn on the OSK keyboard
* Open Libre Office and select the default Keyboard with the Hotkeys type some text.
* Use <kbd>Alt</kbd> + <kbd>Tab</kbd> to switch to NotePad
* Use the Hotkeys to switch to a Keyman Keyboard and observe you can use that keyboard.
* Observe that on the <kbd>Alt</kbd> is not pressed down "stuck" on.
* Use the HotKeys to switch to the second Keyman keyboard
* Use <kbd>Alt</kbd> + <kbd>Tab</kbd> to switch to LibreOffice
* Type and observe the keyboard works correctly 
* Observe that the <kbd>Alt</kbd>does not get stuck
* Try a combination of switching between keyboards and applications. Pressing different modifier keys etc.
* We want to make use the ALT Key or any other modifier key does not stay down - become stuck.

**TEST_WIN_EMOJI_INPUT** 

* Start Keyman
* Select a non-keyman keyboard
* Open Notepad 
* Type some text on the physical keyboard. 
* Press the <kbd>&#x229E; Win</kbd> + <kbd>.</kbd> to bring up the emoji and gif window.
* In the search box type some characters.
* Does it allow input into the emoji panels text field? Verify that input still works and is not completely blocked.


windows/src/engine/keyman32/Keyman32.cpp
```diff
@@ -310,8 +311,10 @@ BOOL InitialiseProcess(HWND hwnd)
 	wm_keymanim_contextchanged = RegisterWindowMessage("WM_KEYMANIM_CONTEXTCHANGED");
 	wm_keymanshift = RegisterWindowMessage("WM_KEYMANSHIFT");
 
+
 	wm_keyman_control = RegisterWindowMessage(RWM_KEYMAN_CONTROL);
 	wm_keyman_control_internal = RegisterWindowMessage("WM_KEYMAN_CONTROL_INTERNAL");
+  wm_keyman_modifierevent = RegisterWindowMessage("WM_KEYMAN_MODIFIEREVENT");
```
mcdurdin, line 317: We should be able to do this and `wm_keyman_keyevent` as simple WM_USER messages, rather than globally registered messages, because they are only ever used internally to a window we own. But I think that change belongs in its own maintenance PR.

windows/src/engine/keyman32/k32_globals.cpp
```diff
@@ -73,6 +73,7 @@ UINT
   //TODO: consolidate these messages -- they are probably not all required now
   wm_keyman = 0,						// user message - ignore msg   // I3594
   wm_keyman_keyevent = 0,   // for serialized input
+  wm_keyman_modifierevent = 0,
```
mcdurdin, line 76:
```suggestion
  wm_keyman_keyevent = 0,      // for serialized input
  wm_keyman_modifierevent = 0, // for serialized input
```

windows/src/engine/keyman32/serialkeyeventserver.cpp
```diff
@@ -485,10 +485,13 @@ class SerialKeyEventServer: public ISerialKeyEventServer {
         input[1].ki.dwExtraInfo = EXTRAINFO_FLAG_SERIALIZED_USER_KEY_EVENT;
         input[1].ki.dwFlags = lParam & 0xFFFF;
 
-        if (!SendInput(2, input, sizeof(INPUT))) {
-          DebugLastError("SendInput");
+        if (msg == wm_keyman_keyevent){
```
mcdurdin, line 488:
```suggestion
        if (msg == wm_keyman_keyevent) {
          // We track changes to modifiers with wm_keyman_modifierevent, but only ever
          // pass them on to the app when we receive them with the wm_keyman_keyevent 
          // message.
```

windows/src/engine/keyman32/k32_lowlevelkeyboardhook.cpp
```diff
@@ -201,6 +201,10 @@ LRESULT _kmnLowLevelKeyboardProc(
   if(Flag != 0) {
     if(isUp) FHotkeyShiftState &= ~Flag;
     else FHotkeyShiftState |= Flag;
+    // #7337 Post the modifier state ensuring the serialized queue is in sync
```
mcdurdin, line 204:
```suggestion
    // #7337 Post the modifier state ensuring the serialized queue is in sync
    // Note that the modifier key may be posted again with wm_keyman_keyevent,
    // later in this function. This is intentional, as the wm_keyman_modifier 
    // event message only updates our internal modifier state, and does not do 
    // any additional processing or other serialization of the input queue.
```

mcdurdin (approved):

windows/src/engine/keyman32/Keyman32.cpp
rc-swag: created #7458

windows/src/engine/keyman32/k32_lowlevelkeyboardhook.cpp
```diff
@@ -201,6 +201,14 @@ LRESULT _kmnLowLevelKeyboardProc(
...
+    // Note that the modifier key may be posted again with wm_keyman_keyevent,
+    // later in this function. This is intentional, as the wm_keyman_modifier 
+    // event message only updates our internal modifier state, and does not do
```
ermshiperete, line 207:
```suggestion
    // later in this function. This is intentional, as the wm_keyman_modifierevent 
    // message only updates our internal modifier state, and does not do 
```

ermshiperete (approved):

ermshiperete (approved):

bharanidharanj:
* **TEST_ALTGR_IN_RULES (PASSED):** Tested this with the attached PR build (Keyman 16.082-alpha) in Windows 10 (Native Machine) and here is my observation: Verified that we are able to see the expected result. 

   

* **TEST_HOTKEYS (PASSED):** Verified that the switching between non-keyman and keyman keyboards using hotkeys are working as expected.

   


* **TEST_HOTKEYS_SWITCH_APPS (PASSED):** Verified that switching between non-keyman and keyman keyboards using hotkey and also using Alt + Tab to switch applications (working as expected.)

   


* **TEST_WIN_EMOJI_INPUT (FAILED):** After running this test, ALT key stuck at one point, and the keyboard is malfunctioning. I have to restart the system to sort out this problem. Seems to be an issue.

bharanidharanj:
### GROUP_WIN11: - Run the tests in Windows 11 OS

* **TEST_ALTGR_IN_RULES (PASSED):** notes
* **TEST_HOTKEYS (PASSED):** notes
* **TEST_HOTKEYS_SWITCH_APPS (PASSED):** notes
* **TEST_WIN_EMOJI_INPUT (PASSED):** notes

rc-swag:
> WIN_EMOJI_INPUT (FAILED): After running this test, ALT key stuck at one point, and the keyboard is malfunctioning. I have to restart the system to sort out this problem. Seems to be an issue.

Ok that is disappointing. 
* Can you remember the exact sequence? If so please provide steps to reproduce.
* Were you able to open the emoji input window at all? *
* Was it once you entered an emoji? 
* What part of the sequence do you press the ALT key,

rc-swag:
Question on the way forward for @mcdurdin?
@bharanidharanj  was unable to reproduce getting the ALT key stuck on the failed test. I would say that we can say that this change has not removed what is causing the issue. However, it seems like a logical improvement AND the other tests show it has caused a regression.  Should we 
1. Merge this PR but keep the issue open needing further investigation.
2. Close this PR without merging and start from scratch investigating this issue.

mcdurdin:
> I would say that we can say that this change has not removed what is causing the issue.

Keep in mind that there may be multiple causes for the issue. So this should have fixed one common source.

I am okay with merging this as I think it is a net positive. But perhaps we'll hold off back-porting to 15.0 for now?

rc-swag:
Yes so this PR address one of the causes for an ALT key becoming stuck. It has passed the regression tests, however, it has been shown the ALT key being stuck can still occur.
We have enough confidence to merge this PR. However, we will leave parent issue #7337 open for further investigation.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.86-alpha](https://keyman.com/downloads/releases/alpha/16.0.86)

MakaraSok:
## GROUP_WIN10 
Tested on Windows 10 Pro virtual machine with Keyman for Windows 16.0.88-alpha
* TEST_WIN_EMOJI_INPUT(PASSED): the emoji window shows up and the word typed would be counted as a query for the matching emoji (see the screenshot below).

image

PR #7451, opened 2022-10-13
chore(linux): Adjust Linux specific files in `.gitignore`

ermshiperete:
During Linux packaging some files get put in the root of the source directory. This change limits these files in `.gitignore` to the root directory. It would be better to not generate these files in the root directory in the first place, but that's not trivial and a separate task.

@keymanapp-test-bot skip

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,11 +176,9 @@ Thumbs.db
 
 # Linux packaging related
 /debian/
-results/
-*.deb
-*.ddeb
-*.dsc
-*.build
-*.buildinfo
-*.changes
-*.tar.?z
+/results/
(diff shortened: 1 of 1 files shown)


ermshiperete:
See discussion in https://github.com/keymanapp/keyman/pull/7434#discussion_r994055412

mcdurdin (approved):
I note *.dsc and *.build are not in the revised list, but otherwise LGTM, thanks for tightening this up.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.80-alpha](https://keyman.com/downloads/releases/alpha/16.0.80)

PR #7459, opened 2022-10-14
fix(ios):  Allow Full Access enabled to fix ios 16  invisible keyboard

### LINKED ISSUE #7391: bug (oem/fv/ios): FV keyboard not visible on iOS 16
**Describe the bug**
The FirstVoices app works normally on iOS 15, but after updating to iOS 16, the keyboard is completely blank.

**To Reproduce**
1. Install iOS 16
2. Open Pages
3. Open the keyboard
4. Tap the globe key and select FirstVoices 

**Screenshots**

image

**Related issues**

A similar issue [5831](https://github.com/keymanapp/keyman/issues/5831) appeared with the release of iOS 15.
The behavior was not identical, however, as the keyboard was visible initially and went blank after tapping a key.

**Reproduced with**
 - Device:            iPhone 13 mini
 - OS:                iOS 16.0.2
 - FirstVoices version:     15.0.270
 - Target Application: Pages

**Keyboard**
 - appears to happen with all FV keyboards

**Additional context**
rotating, locking and unlocking, killing and restarting the app have no effect

**First reported by Clarissa Forbes:**

>Hi Keyman team, we have a bug reported from a user, which sounds very much like the iOS 15 bug. I can't reproduce since I don't have an iPhone but here it is:

>>Since the recent iPhone update the Tuscarora keyboard doesn’t work via the FirstVoices keyboard on iPhone. Just a gray screen shows with no keyboard when selecting. I deleted the app and redownloaded it and I'm still experiencing the same issue. I'm using an iPhone 12 on iOS version 16.0 (20A362).

sgschantz:
fixes #7391 

Due to an undocumented change in iOS 16, the FirstVoices Keyboard was blank and unusable. This change adds the ability to set Allow Full Access under Settings > General > Keyboard > Keyboards > First Voices:

image

After tapping Allow Full Access, an alert will be displayed:

image

Select the Allow button to proceed.

# User Testing

* **TEST_ALLOW_FULL_ACCESS_IOS_16:**

1. On an iPhone running iOS 16, enable 'Allow Full Access' under Settings > General > Keyboard > Keyboards > First Voices
2. Switch to the Notes app and tap on a blank document page
3. Tap the globe key on the keyboard and then tap FirstVoices from the keyboard list.
4. Verify that the keyboard is displayed as normal rather than being blank

### DIFF
--- a/oem/firstvoices/ios/SWKeyboard/Info.plist
+++ b/oem/firstvoices/ios/SWKeyboard/Info.plist
@@ -38,7 +38,7 @@
 			<key>PrimaryLanguage</key>
 			<string>mul</string>
 			<key>RequestsOpenAccess</key>
-			<false/>
+			<true/>
 		</dict>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.keyboard-service</string>


bharanidharanj:
* **TEST_ALLOW_FULL_ACCESS_IOS_16 (OPEN):** Tested this with the attached FirstVoices Keyboard in iOS 16 / iPhone 14 Pro Max Simulator and here is my observation: I noticed that there is no Notes app installed in the Simulator. It asks some authentication to install it in the simulator. So, I tried with the "Reminders" app. I could able to see the First Voices Keyboard after I tap it using the globe key.

mcdurdin:
Note: the Simulator in the past had different behaviour with "Allow Full Access" toggle -- things would work even if that option was off. I don't know if this has been corrected in recent versions of the Simulator or not. But it may invalidate tests, if so!

mcdurdin (approved):
Code changes LGTM

mcdurdin:
Please create a cherry-pick for `stable-15.0`

darcywong00:
Would Keyman for iPhone and iPad need a similar change for iOS 16?

sgschantz:
@darcywong00 This setting is already available for Keyman, so this only impacts FV

jahorton:
I realize this may be a bit more involved, and can probably be done as a separate PR, but... a gentle reminder that we can make this a lot easier for users if we add a shortcut to the FV app similar to this shortcut in the Keyman app:

File (1)

We could also shorten the setup instructions on the FV app's main screen if doing so.

keyman-server:

mcdurdin:
> I realize this may be a bit more involved, and can probably be done as a separate PR

tracking @jahorton's suggestion in #7482.

PR #7465, opened 2022-10-17
fix(ios): Allow Full Access enabled to fix ios 16 invisible keyboard 🍒 

sgschantz:
🍒 -pick of #7459 to stable-15.0

relates to #7391 

Due to an undocumented change in iOS 16, the FirstVoices Keyboard was blank and unusable. This change adds the ability to set Allow Full Access under Settings > General > Keyboard > Keyboards > First Voices:

image

After tapping Allow Full Access, an alert will be displayed:

image

Select the Allow button to proceed.

# User Testing

* **TEST_ALLOW_FULL_ACCESS_IOS_16:**

1. On an iPhone running iOS 16, enable 'Allow Full Access' under Settings > General > Keyboard > Keyboards > First Voices
2. Switch to Safari and tap on the URL editor field to bring up the virtual keyboard
3. Tap the globe key on the keyboard and then tap FirstVoices from the keyboard list.
4. Verify that the keyboard is displayed as normal rather than being blank

### DIFF
--- a/oem/firstvoices/ios/SWKeyboard/Info.plist
+++ b/oem/firstvoices/ios/SWKeyboard/Info.plist
@@ -38,7 +38,7 @@
 			<key>PrimaryLanguage</key>
 			<string>mul</string>
 			<key>RequestsOpenAccess</key>
-			<false/>
+			<true/>
 		</dict>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.keyboard-service</string>


mcdurdin (approved):

bharanidharanj:
* **TEST_ALLOW_FULL_ACCESS_IOS_16 (PASSED):** Tested this with the attached PR build in iOS 16 / iPhone 14 Pro Max Simulator and verified that the On Screen Keyboard is displayed after I set the FirstVoices keyboard (using globe key).

keyman-server:

Issue #7411, opened 2022-10-05
chore(developer): update kmc to latest LDML keyboard DTD version 🙀 


srl295:
A number of things to fix. Need a new import of the [dtd](https://github.com/unicode-org/cldr/blob/67bcf4eef09a732306c54cf71bfd9111463348aa/keyboards/dtd/ldmlKeyboard.dtd)?

* `<backspace>` is no longer in the draft, replaced by `<transform type="backspace">`
* BkspCompiler should become more like the FinlCompiler 
* TODO: enumerate other changes here...

srl295:
and yes this is the moving target problem !

srl295:
> a bunch of other things to update


mhm! done

PR #7471, opened 2022-10-17
chore(resources): update to latest CLDR preview 🙀 

srl295:
Note: there's a commit that does nothing other than sort the existing schema, this is to allow comparison in the later commit.

Fixes: #7411

@keymanapp-test-bot skip


keyman-server:

mcdurdin:
Note: you're doing this from a fork, so you won't get trusted automatic builds. Suggest closing and reopening on a branch on keymanapp/keyman instead

srl295:
> Note: you're doing this from a fork, so you won't get trusted automatic builds. Suggest closing and reopening on a branch on keymanapp/keyman instead

accidental… just realized that now

PR #7402, opened 2022-10-04
feat(developer): implement import and implied 🙀

srl295:
- k_100, k_101, k_102 all failed for different reasons. 
- this PR adds support for import to common/web/types
- this also supports implied
- two of the three keyboards have had the following added, so that they now pass.

```xml
    <import base="cldr" path="techpreview/keys-Latn-implied.xml" />
```

For: 
- #7401
- #7534
- #7964

@keymanapp-test-bot skip

PR #7477, opened 2022-10-18
chore(resources): update to latest CLDR preview 🙀

srl295:
Note: there's a commit that does nothing other than sort the existing schema, this is to allow comparison in the later commit.

- moved dtd/xsd into the `dtd/` subdir - left the schema where they were
- brought in `3.0`, `test` subdirs. Now each subdir matches CLDR.
- Note, the CLDR xml files don't compile yet!  Will work on that in a subsequent PR

Fixes: #7411
reopen of #7471 in the head fork

@keymanapp-test-bot skip

PR #7479, opened 2022-10-18
fix(developer): support CLDR sample keyboards 🙀

srl295:
With this, kmc is at least able to load mt.xml and fr-t-k0-azerty.xml but runs into issues from #7402  and also lack of import support.

for #7411 

includes an update to CLDR data from PR https://github.com/unicode-org/cldr/pull/2469

@keymanapp-test-bot skip


mcdurdin (approved):
I see now this has the updated ldmlKeyboard.dtd for `displays?` :grin: that I suggested in #7477.

PR #6438, opened 2022-03-28
fix(android/engine): Clear WebView cache on package install

darcywong00:
Fixes #6292 

When overwriting an existing keyboard package with the same version, we need to clear the webView cache and reload for the updated keyboard to show.

## User Testing
**Setup**
On the Android emulator/device, save this modified keyboard package file [sil_cameroon_qwerty.kmp](https://darcywong00.github.io/examples/sil_cam_qwerty_azerty/sil_cameroon_qwerty.kmp) to the device's "Downloads" folder. This matches the current (at the time of creating this PR) version of [sil_cameroon_qwerty](https://downloads.keyman.com/keyboards/sil_cameroon_qwerty/6.0.6/sil_cameroon_qwerty.kmp) (version 6.0.6), but the keyboard file is replaced with a modified copy of sil_cameroon_azerty. The ID was also modified to remain `sil_cameroon_qwerty`, but the OSK will display *azerty* on the default layer.

* **TEST_CLEAR_CACHE_RELOAD**
1. Install and load the PR build of Keyman for Android
2. Close the "Getting Started" menu
3. From Keyman Settings --> Install Keyboard or Dictionary --> Install from keyman.com
4. Search for "sil_cameroon_qwerty" and install the keyboard package. (If the keyboard version is > 6.0.6, stop testing and comment on the keyboard version)
5. After the sil_cameroon_qwerty keyboard is installed, exit the menus and look at the sil_cameroon_qwerty OSK
6. Verify the top row of letters on the default layer is "qwerty" (below the top row of diacritics)
qwerty
7. From Keyman Settings --> Install Keyboard or Dictionary --> Install from local file 
8. Browse to the "Downloads" folder and install the modified "sail_cameroon_qwerty.kmp" file from "Setup"
9. After the modified sil_cameroon_qwerty keyboard is installed, exit the menus and look at the OSK
10. Verify the top row of letters on the default row is now "azerty" instead of "qwerty".
azerty

PR #7489, opened 2022-10-19
fix(android/engine): Verify keyboard file exists before switching

### LINKED ISSUE #7412: bug(android): internal engine unable to use keyboards
@jahorton here.

Sentry Issue: [KEYMAN-WEB-5K](https://sentry.io/organizations/keyman/issues/3457768509/?referrer=github_integration)

```
UnhandledRejection: Non-Error promise rejection captured with value: Cannot find the Warang Citi keyboard for Ho (Warang Citi) at /data/user/0/com.tavultesoft.kmapro/app_data/packages/warang_citi/warang_citi.js?v=1664969917797.
```

I've gone through and merged a number of related Sentry events that all resolve to the same point of failure - KMW's `installKeyboard` function, which loads a keyboard's script within the context of its host page, is unable to load the keyboard from the specified file.

https://github.com/keymanapp/keyman/blob/0191d9a3ac81c5da9a3fa44eeb617fdcdb4381e3/web/source/keyboards/kmwkeyboards.ts#L735-L749

The message actually just indicates that the loading `<script>` tag generated an `error`, not ` load`.  Technically, we aren't currently checking that the error's due to a 404 (file not found)... but the net result is the same.  Though... I suppose if we could verify exactly what sort of errors we're getting internally, that might help?

Just to make things extra explicit:  we've (naturally) been assuming that Android code can simply pass a file URL off to the internal KMW engine, which KMW can then load here via script tag.  For whatever reason, KMW is unable to load the provided file URL.  As KMW is confined within a WebView used by the Android Keyman engine, it's Android-code's responsibility to ensure that the file is ready and available for KMW.

A quick brainstorm of possible causes, some of which can probably be eliminated very quickly:
- Might there be file permissions at play?
- Cross-origin security issues, despite being file-system local?
- Have the installed files straight-up gone missing somehow?

It's worth noting that this happens across the full range of supported Android devices, so it _shouldn't_ be due to something specific to new versions of the OS:

image

darcywong00:
Fixes #7412

This change to Keyman Engine for Android verifies a Javascript keyboard file exists before having KeymanWeb switch to the keyboard.

For my testing, I installed sil_cameroon_qwerty package, then manually deleted the .js file before trying to switch to a second keyboard and verified the Toast notification of the [Sentry LogError](https://sentry.io/organizations/keyman/issues/3682438055/?alert_rule_id=8817330&alert_timestamp=1666168780714&alert_type=email&environment=local&project=5983520&referrer=alert_email) appeared.

@keymanapp-test-bot skip

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -572,6 +572,13 @@ final class KMKeyboard extends WebView {
 
     String keyboardPath = makeKeyboardPath(packageID, keyboardID, keyboardVersion);
 
+    // Verify keyboard file still exists
+    File kbFile = new File(keyboardPath);
+    if (kbFile != null && !kbFile.exists()) {
+      KMLog.LogError(TAG, keyboardID + " no longer installed.");
+      return false;
+    }
+
     JSONObject reg = new JSONObject();
     try {
       reg.put("KN", keyboardName);
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
@@ -417,6 +417,12 @@ public final class KeyboardPickerActivity extends BaseActivity {
   protected static boolean removeKeyboard(Context context, int position) {
     boolean result = false;
 
+    Keyboard info = KeyboardController.getInstance().getKeyboardInfo(position);
+    if (info != null) {
+      // Log keyboard uninstall to troubleshoot #7412
+      String msg = KMString.format("Uninstalling %s::%s, version: %s.", info.getPackageID(), info.getKeyboardID(), info.getVersion());
+      KMLog.LogInfo(TAG, msg);
+    }
     KeyboardController.getInstance().remove(position);
     result = KeyboardController.getInstance().save(context);
 
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudKeyboardPackageDownloadCallback.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/impl/CloudKeyboardPackageDownloadCallback.java
@@ -16,6 +16,7 @@ import com.tavultesoft.kmea.packages.LexicalModelPackageProcessor;
 import com.tavultesoft.kmea.packages.PackageProcessor;
 import com.tavultesoft.kmea.util.FileUtils;
 import com.tavultesoft.kmea.util.KMLog;
+import com.tavultesoft.kmea.util.KMString;
 
 import org.json.JSONException;
 
@@ -75,6 +76,11 @@ public class CloudKeyboardPackageDownloadCallback implements ICloudDownloadCallb
             if (packageID == null || packageID.isEmpty()) {
               KMLog.LogError(TAG, "Cloud URL " + _d.getCloudParams().url + " has null packageID");
             }
+
+            // Log keyboard package update to troubleshoot #7412 (languageID may be null)
+            String msg = KMString.format("Cloud keyboard package updated for: %s, language: %s.", packageID, languageID);
+            KMLog.LogInfo(TAG, msg);
+
             String kmpFilename = String.format("%s%s", packageID, FileUtils.KEYMANPACKAGE);
 
             // Extract the kmp.


jahorton:
> > I... wouldn't say this fixes the issue. It provides us somewhat better data _about_ the issue, though.
> > I'm pretty sure that users would say "it's an issue" if a keyboard appeared to have spontaneously uninstalled itself. If the file goes missing, but the keyboard is still in the language list, that doesn't sound like an intentional uninstallation.
> 
> What do you suggest for resolving the Sentry message?

I'll admit, I'm stumped on how to resolve it overall.

That said, this doesn't _resolve_ it either.  It just "passes the buck" to a different message.

> > Cannot find the Warang Citi keyboard for Ho (Warang Citi) at /data/user/0/com.tavultesoft.kmapro/app_data/packages/warang_citi/warang_citi.js?v=1664969917797.
> 
> I don't see how Keyman Engine gets in that state...

Yeah, like I said, I'm stumped on how to resolve it overall.  If I had a clear answer here, that'd be your resolution.

mcdurdin:
I concur with @jahorton on this one. My best guess is that I'd be looking at keyboard uninstallation and/or upgrade as a potential culprit here.

mcdurdin (changes requested):
Let's look at root causes rather than addressing the symptom

darcywong00:
I've added Sentry logging for keyboard uninstalls and updates via the cloud

mcdurdin:
Sadly, I'm not sure this logging will help us -- it does not provide enough detail on what the filesystem and metadata store looks like before + after the uninstalls and updates.

Given we are having trouble with diagnosing this and getting logging together, I wonder if we can try and get a repro on alpha and then go from there?

jahorton:
If only we had Android unit tests set up like the iOS unit tests defined here: https://github.com/keymanapp/keyman/blob/master/ios/engine/KMEI/KeymanEngineTests/MigrationTests.swift

A setup like that would allow loading from constructed "save states", facilitating before and after comparison of engine / install state, which could be useful for investigation here.  It could even be used to set up states that we've "patched out" to verify that they're properly transitioned.

darcywong00:
I'm currently wondering if there's some race condition from this #6438 change (added in 15.0 beta):

https://github.com/keymanapp/keyman/blob/30747f5600ccc0587eaa3f2b5ba300c43c3632d8/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/PackageActivity.java#L285

jahorton:
> I'm currently wondering if there's some race condition from this #6438 change (added in 15.0 beta):
> 
> https://github.com/keymanapp/keyman/blob/30747f5600ccc0587eaa3f2b5ba300c43c3632d8/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/PackageActivity.java#L285

I'm not seeing anything immediately on whether or not that function is async.

That said, one interesting tidbit I ran into, courtesy of https://stackoverflow.com/a/14393017 - https://developer.android.com/reference/android/webkit/WebSettings.html#setCacheMode%28int%29.  (The `LOAD_NO_CACHE` part, in particular.)

jahorton:
It's possible I messed up doing this, but could you do me a favor and try the following?

1. Download/prepare fv_all.kmp

```bash
KEYBOARDS_TARGET=fv_all.kmp KEYBOARDS_TARGET=temp.temp $KEYMAN_ROOT/oem/firstvoices/common/build_keyboards.sh -download-keyboards
```

2. Drag-copy the resulting downloaded `fv_all.kmp` into a live instance of the Keyman app.  (We're doing the package-based ad-hoc path here.)

3. Once the package installer appears, simply click "Install".

4. Check sentry.io and see if you get any error reports like these:
    - `keyman-android`: https://sentry.io/share/issue/80eee015a19047a5b81f56e5cc574248/
    - `keyman-web`: https://sentry.io/share/issue/12c8946f2d7748a1a232681542865716/
        - Note the `.local` on the correponding events:  I triggered these when following the above steps.

jahorton:
Following up on the previous comment:

>Non-Error promise rejection captured with value: Cannot find the Anicinapemi8in/Anishinàbemiwin keyboard for Algonquin (Latin) at /data/user/0/com.tavultesoft.kmapro.debug/app_data/packages/fv_all/fv_anicinapemi8in.js?v=1667206299329.

But:

image

(Within an unaltered un-zip of fv_all.kmp.)

jahorton:
Tried a second time after wiping the emulation instance; the error did not recur this time.  So there may be some sort of race condition... but it _can_ happen via ad-hoc installs, at least.  The linked Sentry reports are proof of that.

jahorton:
Also, for what it's worth...

The installed keyboards are all registered with lowercased language tags.  I inspected the app's filesystem, `app_userdata/keyboards_list.json` to verify this.  The base `fv_all.kmp` file's `kmp.json` does have mixed-case language tags, thus the app itself is confirmed to apply lowercasing on keyboards coming in via ad-hoc install.

jahorton:
Oh, and all of the above was done on `master`, not on this branch - just for clarification.

In retrospect, this may not have been the best place to leave these notes, but... at least they're available here.  Not sure where to move 'em at the moment, and it's the end of the workday.  "Somewhere" is at least better than "nowhere", so hopefully all of that proves useful.

jahorton:
While I remember:  I think the first run started with French Basic as the active keyboard, not SIL EuroLatin.  I know there were at least two keyboards total installed.  This may be relevant to repro attempts - that the prior active keyboard wasn't EuroLatin.

jahorton:
... yeah, starting to seem like I "lucked out" that one time; I haven't had any successful attempts so far today.

darcywong00:
> Probably of note: I was not prompted to select any languages to install.

@jahorton - Since fv_all.kmp contains multiple keyboards to install, we don't use a language picker during the package installation. It just picks the first language in each list. They can always add the desired language in the Settings menu.

darcywong00:
One thing I'll note. Keyman for Android starts with the default keyboard:
```json
{
  "packageID": "sil_euro_latin"
  "resourceID": "sil_euro_latin"
}
```

and after installing fv_all.kmp, the packageID in the keyboards_list.json gets updated to:
```json
{
  "packageID": "fv_all"
  "resourceID": "sil_euro_latin"
}
```

The sil_euro_latin folder is still intact though.


This is probably because the Keyboard controller sees a matching keyboardID: sil_eruo_latin and languageID: en
and updates the packageID....


This wouldn't explain other keyboards in the Sentry logs: arabic_izza, sil_cameroon_qwerty, ...

jahorton:
You know... a _lot_ of this feels kinda similar to issue #6862.  Granted, the errors logged by Sentry are different, but the end-user experience felt very similar.  I don't believe `fv_all.kmp` has a welcome.htm, so that aspect was (naturally) bypassed for the case reported here.

darcywong00:
Abandoning this as we chase down the root cause. (A17S13)

PR #7501, opened 2022-10-20
refactor(linux): Comment unused methods

### LINKED ISSUE #7297: bug(linux): wrong logic in ibus_keyman_engine_set_surrounding_text()

ermshiperete:
Closes #7297.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -115,34 +115,34 @@ static void ibus_keyman_engine_focus_out  (IBusEngine             *engine);
 static void ibus_keyman_engine_reset      (IBusEngine             *engine);
 static void ibus_keyman_engine_enable     (IBusEngine             *engine);
 static void ibus_keyman_engine_disable    (IBusEngine             *engine);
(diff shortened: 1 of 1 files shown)


jahorton (commented):
In my experience, we usually outright-delete code rather than comment it out.  Worst-case, it'll still live on in commit history.

ermshiperete:
In general I agree, however in this case I find it helpful to have them commented: the documentation of ibus isn't that great, and having them helps when debugging since it allows to quickly uncomment them.

mcdurdin (approved):
Generally I think we should delete unused code, but I think @ermshiperete's reasoning is solid here, so lgtm

keyman-server:

PR #7502, opened 2022-10-20
chore(linux): Rename internal method to `commit_string`

ermshiperete:
The previous method name `ibus_keyman_engine_commit_string` suggests that this method is overridden from ibus-engine class. However, while it is true that ibus-engine has a `ibus_engine_commit_string` method it can't be overridden. This change renames the method to make it clearer that this is an internal method that won't be called by ibus.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

mcdurdin (approved):

keyman-server:

PR #5687, opened 2021-09-14
chore(common): Add GitHub action to upload sources to crowdin

darcywong00:
This PR adds an action to upload sources per https://support.crowdin.com/github-crowdin-action/

The source files are organized at a top level folder named "master".

For now, I've set the action to trigger every two weeks (one sprint cycle) which should suffice.

@keymanapp-test-bot skip

PR #7509, opened 2022-10-21
fix(mac): load default privacy alert string

### LINKED ISSUE #7475: bug(mac): "privacy-alert-text" when trigger Keyman for the first time after installation
**Describe the bug**

Install Keyman build from this [PR](https://github.com/keymanapp/keyman/pull/7354) on a clean macOS while having the UI language of macOS set to "Hausa". Restart the machine after the installation completed, then trigger Keyman by clicking on its icon on the system's dropdown menu to see an error where this message is shown: "privacy-alert-text" (see the screenshot below).

image

**Expected behavior**

The expected message is either in EN or the localized version of the EN message that says:

image

**Related issues**

---

   Device:              MBP
 - OS:                  Monterey 12.6
 - Keyman version:      16.0.70
 - Target Application:  Popup dialog box

sgschantz:
Fixes #7475 

This fix loads the English string for privacy-alert-text and uses it as a default in case the string cannot be found for a localization. The problem occurred when:
1. Keyman was localized for a given language x
2. the string with the key privacy-alert-text was not found in Localizable.strings for language x
In that case, the key name, "privacy-alert-text", would be displayed instead of a localized string

It is best if we actually have an alert in the localized language, but when we don't, English is much better than an incomprehensible key name.

# User Testing

* **TEST_LOCALIZED_PRIVACY_ALERT:** - Confirm that a readable alert displays in English when Hausa is set as the preferred language.

1. Remove old version of Keyman **after** first removing its permissions from Security & Privacy > Privacy > Accessibility
2. Install the Keyman build for this PR. 
3. Go to Apple > System Preferences...
4. Choose Language and Region and select + to add Hausa, then drag Hausa to the top to be the primary preferred language
5. Restart the system
6. Select Keyman from the System Input Source menu
7. Confirm that the following dialog is displayed:
image

Note that after the Hausa localization is eventually updated, this test will not result in an alert in English.

### DIFF
--- a/mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyWindowController.m
+++ b/mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyWindowController.m
@@ -25,13 +25,31 @@
   NSString *bundleDisplayName = [[[NSBundle mainBundle] localizedInfoDictionary]
                            objectForKey:@"CFBundleDisplayName"];
   [self.window setTitle:bundleDisplayName];
-  [_alertText setStringValue:NSLocalizedString(@"privacy-alert-text", nil)];
+  
+  // need to provide a defaultValue for the possibility that:
+  // 1. the app is localized for a given language x
+  // 2. the string with the specified key is not found in Localizable.strings for language x
+  // if not, then UI will display the key instead of the string with that language selected
+
+  NSString *defaultValue = [self defaultLocalizableString: @"privacy-alert-text"];
+  [_alertText setStringValue:NSLocalizedStringWithDefaultValue(@"privacy-alert-text", nil, NSBundle.mainBundle, defaultValue, nil)];
   
   NSImage *keymanLogo = [NSImage imageNamed:NSImageNameApplicationIcon];
   if (keymanLogo) {
     [_appLogo setImage:keymanLogo];
   }
   [_okButton setEnabled:YES];
+}
+
+/**
+ * utility method to get the English string for the localizable key
+ * move to some generic place if it will be widely used
+ */
+- (NSString*) defaultLocalizableString:(NSString*) key {
+  NSBundle *main = [NSBundle mainBundle];
+  NSString *resourcePath = [main pathForResource:@"en" ofType:@"lproj"];
+  NSBundle *bundle = [NSBundle bundleWithPath:resourcePath];
+  return NSLocalizedStringFromTableInBundle(key, nil, bundle, nil);
 }
 
 - (IBAction)closeAction:(id)sender {


bharanidharanj:
* **TEST_LOCALIZED_PRIVACY_ALERT (PASSED):** Tested this in Monterey 12.6 macOS and here my observation: Followed the steps as they given in the PR and after restarting the system, I just selected the Keyman from the System Input Source men (6th step). Nothing happened. After that I clicked Keyman from macOS menu bar leads to open an alert message on the screen.

mcdurdin:
It seems this is a general issue and I imagine we have other untranslated strings elsewhere? Is this something we need to fix throughout?

sgschantz:
I don't think we need to make this change for strings that existed before our first internationalization effort. For those, we can expect the localization to include each string. It's only when new strings are added and we have an incomplete localization that the key would be missing. 

To help with this, @darcywong00 said that there might an option in crowdin that will include the default English strings during export for cases where a new localized string is not provided. That way, we won't end up with an undefined key unless we happen to add a new string to the project after localization has already been completed.

mcdurdin:
> I don't think we need to make this change for strings that existed before our first internationalization effort.

I don't know... We have less than 100% coverage on a lot of our languages.

> @darcywong00 said that there might an option in crowdin that will include the default English strings during export for cases where a new localized string is not provided.

That sounds like an appropriate solution which would also be less code.

darcywong00:
Ahh. I forgot we're using a GitHub action (triggers every two weeks) to keep crowdin strings up to date with the master branch. (#5687).

Currently, `privacy-alert-text` doesn't exist in crowdin, so maybe we should manually kick off the action.

re: fallback strings. Here's our project settings on what gets exported

image

The mac Localizable.strings is 80% for Hausa. And the untranslated strings are already populated with English by crowdin.

mcdurdin:
> Currently, `privacy-alert-text` doesn't exist in crowdin, so maybe we should manually kick off the action.

SGTM

> And the untranslated strings are already populated with English by crowdin.

🎆 So nothing to be done by @sgschantz then! Fantastic :grin: We can probably close this PR...

darcywong00:
> Fantastic 😁 We can probably close this PR...

In theory, yes... I triggered the GHA but the crowdin version  remained the same, so still investigating...

darcywong00:
After 🦆 with @mcdurdin we think this is how we'll handle the new 16.0 untranslated strings (including privacy-alert-text)
1. we can abandon this PR
2. I'll pull all the new crowdin strings in 16.0 into a new mega-PR for all the languages we've incorporated. It will be a ton of files but we can diff it

sgschantz:
@darcywong00, not sure if your conversation with @mcdurdin was before or after our conversation. Do you think that we have a reasonable procedure that we can repeat each time we add a new string to the app? Before we release the updated Keyman code, we would need to 
1. update crowdin with the key and English text for each new string or strings.
2. in crowdin, populate the corresponding localization file in all languages with the default Englsh text for the new string
3. export the localized files (each including one or more non-localized string) from crowdin back to the Keyman project

If I understand that correctly, then my questions are:
1. is this a fairly simple manual process? (assuming we can't automate it)
2. will it still be easy for the crowdin translators to identify the text that needs to be localized or does the default English text obscure that?

mcdurdin:
> Before we release the updated Keyman code

Just clarifying here -- before we go to 'beta' or 'stable'. It is acceptable for translations to be missing entirely in alpha and showing placeholder strings.

sgschantz:
> > Before we release the updated Keyman code
> 
> Just clarifying here -- before we go to 'beta' or 'stable'. It is acceptable for translations to be missing entirely in alpha and showing placeholder strings.

If our testers doing any testing with the preferred language not set to English, then they will probably want to open an issue when they see the key name substituted for the string, just as @MakaraSok did in #7475.

mcdurdin:
Yeah, I understand that, but we can document this as a known issue that can be encountered in Alpha, and our in-house testers will quickly become aware of that.

darcywong00:
> will it still be easy for the crowdin translators to identify the text that needs to be localized or does the default English text obscure that?

Crowdins translators can see untranslated strings for the languages they're working in.
When we manually pull down localizations via the Crowdin CLI, it will fill in missing strings with English.

sgschantz:
Closing this PR -- rather than programmatically setting an English string for the default, we will rely on the localizable.strings files for each language to be updated with English text until localization occurs.

PR #7520, opened 2022-10-25
fix(core): remove `KMN_API` from callback spec

### LINKED ISSUE #7487: bug(core): Compiling core on Linux outputs a warning
Compiling core outputs the following warning when building on Linux:

```
core/include/keyman/keyboardprocessor.h:130:1: warning: ‘visibility’ attribute ignored [-Wattributes]
  130 | typedef KMN_API uint8_t (*km_kbp_keyboard_imx_platform)(km_kbp_state*, uint32_t, void*);
      | ^~~~~~~
```

mcdurdin:
Fixes #7487.

`KMN_API` defines only import/export/static modifiers for functions, so it makes no sense to apply it to a typedef. (Previously, I thought it also included a calling convention, but it doesn't.)

@keymanapp-test-bot skip

### DIFF
--- a/core/include/keyman/keyboardprocessor.h
+++ b/core/include/keyman/keyboardprocessor.h
@@ -127,7 +127,7 @@ typedef struct km_kbp_option_item  km_kbp_option_item;
 // Callback function used to to access Input Method eXtension library functions
 // from Keyman Core
 //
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:

Issue #7529, opened 2022-10-25
bug(developer): debugger should use scan codes rather than vk codes to ensure consistent mapping


mcdurdin:
### Describe the bug

We need to always use kbdus.dll virtual key codes under the hood for the debugger, so that positional keyboards work correctly. The cleanest way to do this is to use our existing scan code mapping tables from Keyman Engine for Windows for kbdus virtual keys.

Don't forget the Brazilian ABNT2 103rd key.

### Related issues

This is precursor work for #1225, #1074.

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for iPhone and iPad
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [X] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

15.0.271

Issue #1225, opened 2018-09-29
[Developer] Select System Keyboard dialog is not working


mcdurdin:
The Select System Keyboard dialog in Keyman Developer 10 is not working. Also the help link is broken:

 /developer/10.0/context/select-system-keyboard

mcdurdin:
This is because the mnemonic recompiler support has not been integrated into the debugger. At present, interactive debugging must take place on a English US base keyboard.

Testing of other base keyboards can be accomplished by installing the keyboard into Keyman Desktop and testing within other apps with Keyman's normal Base Keyboard functionality.

We hope to be able to address this in Keyman 11 or 12 (it's a non-trivial fix).

mcdurdin:
See #1074. This is still non-trivial. Waiting on mnemonic layout recompiler to be rewritten.

To completely fix this, we need to also always use kbdus.dll under the hood (i.e. mapping via scan code instead of vk -- except for the Brazilian ABNT2 103rd key) -- see #7529

PR #7530, opened 2022-10-25
fix(developer): handle LControl being set by Windows when AltGr pressed

### LINKED ISSUE #7506: bug(developer): Right Alt treated as Left Ctrl + Right Alt on some base keyboards
**Describe the bug**

See https://community.software.sil.org/t/ctrl-key-latching-on-without-being-pressed/6803.

On some Windows hardware keyboards, Windows treats <kbd>AltGr</kbd> the same as <kbd>Left Ctrl</kbd> + <kbd>Right Alt</kbd>. The reason is historical, supporting users who had only one <kbd>Alt</kbd> key on their keyboards. Keyman for Windows should handle this case already, but I think when we rewrote the debugger integration in Keyman Developer 15, we missed this edge case. I have turned this into an issue to fix in the debugger.

This is probably a regression in Keyman Developer 15 with the rewrite of the debugger to use Keyman Core.

**Expected behavior**

The debugger should ignore the <kbd>Left Ctrl</kbd> when it is paired with a <kbd>Right Alt</kbd>.

---

**Keyman Developer:**
 - Keyman Developer version: 15.0.270

mcdurdin:
Fixes #7506.

In the keyboard debugger, we need to mask out <kbd>LControl</kbd> when <kbd>AltGr</kbd> is pressed by the user, because Windows will generate LControl + <kbd>AltGr</kbd> for European layouts.

Note: have also hidden the System Keyboard debug setting in this commit, given it is not currently implemented; see #1225.

A follow-up issue is #7529.

# User Testing

**TEST_ALTGR:** Verify that the <kbd>LControl</kbd> modifier is not set when <kbd>AltGr</kbd> is pressed.

1. Open a keyboard, and start debugging (<kbd>F5</kbd>).
2. Set the debugger into Single Step Mode.
3. In the **Windows language selector**, choose a European keyboard layout such as French or English (UK).
4. In the debugger, press <kbd>AltGr</kbd>+<kbd>X</kbd> (any other letter is okay).
5. The debug status window should show <kbd>RAlt</kbd>+<kbd>X</kbd>, not <kbd>Control</kbd>+<kbd>RAlt</kbd>+<kbd>X</kbd>.

### DIFF
--- a/developer/src/tike/actions/dmActionsKeyboardEditor.dfm
+++ b/developer/src/tike/actions/dmActionsKeyboardEditor.dfm
@@ -108,6 +108,7 @@ object modActionsKeyboardEditor: TmodActionsKeyboardEditor
       Category = 'Debug Control'
       Caption = 'Select System &Keyboard...'
       ImageIndex = 49
+      Visible = False
       OnExecute = actDebugSelectSystemKeyboardExecute
       OnUpdate = actDebugSelectSystemKeyboardUpdate
     end
--- a/developer/src/tike/actions/dmActionsKeyboardEditor.pas
+++ b/developer/src/tike/actions/dmActionsKeyboardEditor.pas
@@ -556,6 +556,7 @@ end;
 
 procedure TmodActionsKeyboardEditor.SelectDebugSystemKeyboard(k: TSystemKeyboardItem);
 begin
+//TODO: #1225, #1074 (note both menu item and toolbar combo are hidden while this is not working)
 //  if IsDebuggerVisible then   // I3655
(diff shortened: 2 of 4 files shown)


darcywong00 (approved):
lgtm

bharanidharanj:
* **TEST_ALTGR (PASSED):** Tested this PR with the attached Keyman Developer build 16.0.89-alpha-test-7530 in Windows 10 OS and noticed that after pressing AltGr+X keys, it is showing the same keys in the debugger status window.

keyman-server:

PR #7537, opened 2022-10-26
fix(developer): handle LControl being set by Windows when AltGr pressed 🍒

### LINKED ISSUE #7506: bug(developer): Right Alt treated as Left Ctrl + Right Alt on some base keyboards

mcdurdin:
Fixes #7506.

Cherry-pick of #7530.

In the keyboard debugger, we need to mask out LControl when AltGr is pressed by the user, because Windows will generate LControl + AltGr for European layouts.

Note: have also hidden the System Keyboard debug setting in this commit, given it is not currently implemented; see #1225.

A follow-up issue is #7529.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/developer/TIKE/actions/dmActionsKeyboardEditor.dfm
+++ b/windows/src/developer/TIKE/actions/dmActionsKeyboardEditor.dfm
@@ -108,6 +108,7 @@ object modActionsKeyboardEditor: TmodActionsKeyboardEditor
       Category = 'Debug Control'
       Caption = 'Select System &Keyboard...'
       ImageIndex = 49
+      Visible = False
       OnExecute = actDebugSelectSystemKeyboardExecute
       OnUpdate = actDebugSelectSystemKeyboardUpdate
     end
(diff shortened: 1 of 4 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.272](https://keyman.com/downloads/releases/stable/15.0.272)

PR #7553, opened 2022-10-26
fix(developer): exclude `nocaps` from layer id modifier components

### LINKED ISSUE #7551: bug(developer): `nocaps` should not be included as one of the modifier components in layer ids
### Describe the bug

We should remove the `nocaps` identifier from the extended list of modifier options in the touch layout editor. The presence of `nocaps` in the list  introduces a distinction that isn't real: including `nocaps` in the layout id has exactly the same meaning as excluding `caps` from the layout id. That also removes the really stupid last entry in the list: `caps-nocaps` :joy:

image

The `nocaps` component in a layer id has been used in only two keyboards in the keymanapp/keyboards repository: sil_cameroon_qwerty, and sil_cameroon_azerty. I propose that we patch these two keyboards rather than handling the `nocaps` modifier in the KeymanWeb code.

For reference, KeymanWeb currently does not currently handle `nocaps` in the layer id, and ends up treating `nocaps` as `caps` due to a naive substring match (mea culpa):

https://github.com/keymanapp/keyman/blob/65770747a3c71ceaec99758c1e2a64f158a69a7a/common/web/keyboard-processor/src/text/keyboardProcessor.ts#L390-L400

But I think the proper fix is as described above, rather than patching KeymanWeb.

### Reproduce the bug

_No response_

### Expected behavior

_No response_

### Related issues

https://community.software.sil.org/t/bug-introduced-in-cameroon-keyboard-by-km-15-web-android/6834/4

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [X] Keyman Developer
- [ ] KeymanWeb

### Keyman version

15.0

### Operating system

_No response_

### Device

_No response_

### Target application

_No response_

### Browser

_No response_

### Keyboard name

_No response_

### Keyboard version

_No response_

### Language name

_No response_

### Additional context

_No response_

mcdurdin:
Fixes #7551.

@keymanapp-test-bot skip I've already tested this to my satisfaction. I don't think we need a user test.

### DIFF
--- a/developer/src/tike/xml/layoutbuilder/constants.js
+++ b/developer/src/tike/xml/layoutbuilder/constants.js
@@ -672,14 +672,16 @@ $(function() {
 
   this.showAllModifierCombinations = false;
 
-  // Add CAPS and NO_CAPS variants for all of the above
-  (function(c, codes) {
-    var i, n = c.length;
-    for(i = 0; i < n; i++)
-      c.push(c[i] | codes.CAPS);
-    for(i = 0; i < n; i++)
-      c.push(c[i] | codes.NO_CAPS);
-  })(this.validModifierCombinations, this.modifierCodes);
+  // Add CAPS variants for all of the above
+  (function addCapsCombinations(validCombinations, CAPS) {
+    const n = validCombinations.length;
+    for(let i = 0; i < n; i++) {
+      const newCombination = validCombinations[i] | CAPS;
+      if(!validCombinations.includes(newCombination)) {
+        validCombinations.push(newCombination);
+      }
+    }
+  })(this.validModifierCombinations, this.modifierCodes.CAPS);
 
   this.modifierNames = [
     'leftctrl',   // 0x001


developer/src/tike/xml/layoutbuilder/constants.js
```diff
@@ -672,14 +672,16 @@ $(function() {
...
-      c.push(c[i] | codes.NO_CAPS);
-  })(this.validModifierCombinations, this.modifierCodes);
+  // Add CAPS variants for all of the above
+  (function addCapsCombinations(validCombinations, CAPS) {
+    const n = validCombinations.length;
+    for(let i = 0; i < n; i++) {
+      const newCombination = validCombinations[i] | CAPS;
+      if(!validCombinations.includes(newCombination)) {
+        validCombinations.push(newCombination);
+      }
```
mcdurdin, line 682: I tweaked this here because, for example, `caps` was in the list three times previously.

darcywong00: Is `nocaps` at l.696 still used?

mcdurdin: It's not, but I left it there for completeness because it maps to the definitions in kmn

darcywong00 (approved):
lgtm

keyman-server:

PR #7472, opened 2022-10-18
fix(android/engine): Dismiss key preview and subkeys on globe action

darcywong00:
Fixes #7424 

similar to #  , this dismisses key previews and subkeys when the Globe key is handled.

## User Testing
Setup - Install PR build of Keyman for Android on physical Android device. (Not an emulator because test involves multi-touches on the OSK)

**TEST_GLOBE_KEY_DISMISSES_KEYS** - Verifies key previews and subkeys are dismissed on Globe key action
1. Launch Keyman and dismiss the Get Started menu
2. With the default sil_euro_latin keyboard, long-press and hold a key with one finger 
3. While holding the long-press, use another finger to short-press the globe key
4. Verify nothing happens with th eglobe key and the long-press keys are still visible
5. Release the long-press key
6. With the default sil_euro_latin keyboard, first short-press a key with one finger, and then short-press the globe key with another finger and release both fingers (before a long-press triggers)
7. Verify the key preview briefly appears and disappears when the globe key action is handled:
  * If 1 keyboard installed, the Keyboard picker menu appears
  * If multiple keyboards installed, Keyman switches to the next keyboard

PR #7555, opened 2022-10-27
fix(android/engine): Dismiss subkeys when hiding keyboard

darcywong00:
Similar to #7388 and #7472
While trying to get a repro of #7412, I noticed subkeys can get stuck when dismissing/hiding the Keyman system keyboard

## User Testing
1. Install PR build of Keyman for Android on a physical device

* **TEST_LONGPRESS_DISAPPEARS_ON_HIDE** - Verifies longpress keys aren't stuck after hiding system keyboard
1. Launch Keyman for Android
2. On the "Get Started" menu, enable Keyman as the default system keyboard
3. Launch a separate app (e.g. Chrome) and touch the search bar
4. Switch system keyboard so Keyman is visible
5. With the right thumb, long-press and hold a key, select a long-press option, and continue holding
6. With the left thumb, press the down arrow on the system taskbar  to dismiss/hide the system keyboard
7. Verify the OSK disappears
8. Release all keyboard presses
9. Touch the Chrome search bar so the OSK reappears
10. Verify long-press keys are not visible

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KMKeyboard.java
@@ -269,6 +269,9 @@ final class KMKeyboard extends WebView {
   }
 
   public void hideKeyboard() {
+    dismissKeyPreview(0);
+    dismissSubKeysWindow();
+
     String jsString = "hideKeyboard()";
     loadJavascript(jsString);
   }


bharanidharanj:
* **TEST_LONGPRESS_DISAPPEARS_ON_HIDE (PASSED):** Tested this PR in my Android Mobile device (ver 11) and it seems to be working as expected. Verified that after I clicked the back button from my mobile the Keyman OSK disappeared from the Chrome browser. Then, I clicked the Chrome / Search bar. Now, the Keyman OSK reappears on the screen.

mcdurdin (approved):
This should also be cherry-picked

keyman-server:

Issue #7532, opened 2022-10-25
feat(developer): ldml: include all LDML data in kmx+ 🙀


srl295:
- KMX+ should include all data in the source
- Principle here: data representation, not prematurely optimizing for keyman build process, other than synthesized vkeys.

Scope:
- layouts
- may include another copy of the keybag also (and displays) due to gestures
- synthesized vkey ids need to match kvks

#7042 🙀

srl295:
should be closed with #7630

mcdurdin:
Closing with #7630

PR #7561, opened 2022-10-27
feat(core): ldml: add `layr`, `disp`, `key2`, `list` sections to C7043.1 spec 🙀

srl295:
- add layers, rows, display

This is a draft, feedback appreciated…

Note that I use `.x` in the numbering for the new sections, will renumber if these are accepted

for #7532 

@keymanapp-test-bot skip

### DIFF
--- a/core/src/ldml/C7043_ldml.md
+++ b/core/src/ldml/C7043_ldml.md
@@ -324,31 +324,30 @@ For each key:
 - `vkey`: Is the standard vkey, 0-255
 - `target`: Is the target (resolved) vkey, 0-255.
 
-### C7043.2.x `lyrs`—Layers list
+### C7043.2.13 `layr`—Layers list
 
-Represents layers on the keyboard. There may be multiple layer lists,
-corresponding to different form factors and hardware types.
+Represents layers on the keyboard.
+- Each key entry corresponds to a key on the row
 
 | ∆ | Bits | Name       | Description                              |
 |---|------|------------|------------------------------------------|
-| 0 |  32  | ident      | `lyrs`                                   |
+| 0 |  32  | ident      | `layr`                                   |
 | 4 |  32  | size       | int: Length of section                   |
-| 8 |  32  | count      | int: Number of layer lists               |
-|12 |  32  | layerstart | int: offset to first layer entry         |
-|16 |  32  | layercount | int: number of layer entries             |
-|20 |  96  | reserved   | padding                                  |
+| 8 |  32  | listCount  | int: Number of layer lists               |
+|12 |  32  | layerCount | int: number of layer entries             |
+|16 |  32  | rowCount   | int: number of row entries               |
+|20 |  32  | keyCount   | int: number of key entries               |
+|24 |  64  | reserved   | padding                                  |
 
-Note: `layerstart` must point after the end of the layer lists
+Each layer list corresponds to one `<layers>` element.
+There are `listCount` total lists.
 
-For each 'layers' - the layer lists
-
-| ∆ | Bits | Name       | Description                              |
-|---|------|------------|------------------------------------------|
-|16+|  32  | flags      | int: per-layers options                  |
-|24+|  32  | hardware   | str: layout (`us`,`iso`,`jis`,`abnt2`)   |
-|28+|  32  | layer      | int: index to first layer element        |
-|32+|  32  | layerCount | int: number of layer elements in list    |
-|36+|  96  | reserved   | short`[3]`: padding                      |
+| ∆ | Bits | Name       | Description                                |
+|---|------|------------|--------------------------------------------|
+| 0+|  32  | flags      | int: per-layers options                    |
+| 4+|  32  | hardware   | str: layout (`us`,`iso`,`jis`,`abnt2`)     |
+| 8+|  32  | layer      | int: index to first layer element          |
+|12+|  32  | count      | int: number of layer elements in this list |
 
 - `flags`: a 32-bit bitfield defined as below:
 
@@ -357,59 +356,32 @@ For each 'layers' - the layer lists
   |       0      | form     | 0: hardware          |
   |       0      | form     | 1: touch             |
 
-For each layer element
+Each layer entry corresponds to one `<layer>` element
+There are `layerCount` total layer entries.
 
 | ∆ | Bits | Name       | Description                              |
 |---|------|------------|------------------------------------------|
-|16+|  32  | id         | str: layer id such as `base` or `shift`  |
-|24+|  32  | modifier   | str: modifier string                     |
-|28+|  32  | row        | int: `rows` identifier of first row      |
-|32+|  32  | rowCount   | int: number of `rows` elements           |
+| 0+|  32  | id         | str: layer id such as `base` or `shift`  |
+| 4+|  32  | modifier   | str: modifier string                     |
+| 8+|  32  | row        | int: index into rows area (next section) |
+|12+|  32  | count      | int: number of `rows` elements for this layer  |
 
-### C7043.2.x `rows`—Row list
+Each row entry corresponds to one `<row>` element
+There are `rowCount` total row entries.
 
-Each list is a row of keys in the `key2` element
+| ∆ | Bits | Name       | Description                            |
+|---|------|------------|----------------------------------------|
+| 0+|  32  | key        | int: index into key element            |
+| 4+|  32  | key2count  | int: count of key elements in this row |
 
-| ∆ | Bits | Name    | Description                              |
-|---|------|---------|------------------------------------------|
-| 0 |  32  | ident   | `rows`                                   |
-| 4 |  32  | size    | int: Length of section                   |
-| 8 |  32  | count   | int: Total number of row list elements   |
-|12 |  32  | reserved| padding                                  |
-
-> TODO-LDML: Should there be a '0-key2count terminator'?
-
-For each 'row' list element
-
-| ∆ | Bits | Name       | Description                              |
-|---|------|------------|------------------------------------------|
-|16+|  32  | key2       | int: index into `key2` element           |
-|24+|  32  | key2count  | int: count of `key2` elements            |
-
-### C7043.2.x `key2`—Secondary key list
-
-This is a list of vkeys, but ordered by rows.  The `rows` elements point into the start and length of entries here.
-
-> TODO-LDML: Should there be a '0-vkey terminator'?
-> And shold the 0th element also be a 0 vkey so that key2=0 means null?
+Each key entry corresponds to a key in the row.
+There are `keyCount` total key entries.
 
 | ∆ | Bits | Name    | Description                              |
 |---|------|---------|------------------------------------------|
-| 0 |  32  | ident   | `key2`                                   |
-| 4 |  32  | size    | int: Length of section                   |
-| 8 |  32  | count   | int: Total number of key2 elements       |
-|12 |  32  | reserved| padding                                  |
+| 0+|  32  | vkey    | int: vkey ID                             |
 
-For each `key2` element:
-
-Note that the first two fields are the index into the `keys` element, except arranged according to the keyboard layout.
-
-| ∆ | Bits | Name    | Description                              |
-|---|------|---------|------------------------------------------|
-|16+|  32  | vkey    | int: vkey ID                             |
-|20+|  32  | mod     | int: modifier key flags                  |
-
-### C7043.2.x `disp`—Display list
+### C7043.2.14 `disp`—Display list
 
 | ∆ | Bits | Name          | Description                              |
 |---|------|---------------|------------------------------------------|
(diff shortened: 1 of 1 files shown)


core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,120 @@ For each key:
...
+|36+|  96  | reserved   | short`[3]`: padding                      |
+
+- `flags`: a 32-bit bitfield defined as below:
+
+  | Bit position | Meaning  |  Description         |
+  |--------------|----------|----------------------|
+  |       0      | form     | 0: hardware          |
+  |       0      | form     | 1: touch             |
+
+### C7043.2.x `layr`—Layer list
```
mcdurdin, line 354: I think I'd be inclined to fold this into the previous section

srl295: sure i can see that. just add 1 offset, 'start of layer list'

srl295: ptal

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+
+Note layerstart must point after the end of the layer lists
+
+For each 'layers' - the layer lists
+
+| ∆ | Bits | Name       | Description                              |
+|---|------|------------|------------------------------------------|
+|16+|  32  | flags      | int: per-layers options                  |
+|24+|  32  | hardware   | str: layout (`us`,`iso`,`jis`,`abnt2`)   |
+|20+|  32  | target     | int: target vkey ID (0…255)              |
```
mcdurdin, line 346: This doesn't seem to belong here?

```suggestion
```

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+|28+|  32  | layer      | int: index to first layer element        |
+|32+|  32  | layerCount | int: number of `layr` elements in list   |
```
mcdurdin, line 348:
```suggestion
|32+|  32  | layerCount | int: number of layer elements in list    |
```

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+  |       0      | form     | 1: touch             |
+
+For each layer element
+
+| ∆ | Bits | Name       | Description                              |
+|---|------|------------|------------------------------------------|
+|16+|  32  | id         | str: layer id such as `base` or `shift`  |
+|24+|  32  | modifier   | str: modifier string                     |
+|28+|  32  | row        | int: `rows` identifier of first row      |
+|32+|  32  | rowCount   | int: number of `rows` elements           |
```
mcdurdin, line 365: This implies a certain ordering in the `rows` section which I think is a little fragile. Either we pull `rows` into this `lyrs` section (my pref), or we rethink how we organise it.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+|12 |  32  | reserved| padding                                  |
+
+> TODO-LDML: Should there be a '0-key2count terminator'?
+
+For each 'row' list element
+
+| ∆ | Bits | Name       | Description                              |
+|---|------|------------|------------------------------------------|
+|16+|  32  | key2       | int: index into `key2` element           |
+|24+|  32  | key2count  | int: count of `key2` elements            |
```
mcdurdin, line 385: In the same way, we've built a dependency and fragility in the ordering of key2 elements. I would prefer each section to stand on its own.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+Each list is a row of keys in the `key2` element
+
+| ∆ | Bits | Name    | Description                              |
+|---|------|---------|------------------------------------------|
+| 0 |  32  | ident   | `rows`                                   |
+| 4 |  32  | size    | int: Length of section                   |
+| 8 |  32  | count   | int: Total number of row list elements   |
+|12 |  32  | reserved| padding                                  |
+
+> TODO-LDML: Should there be a '0-key2count terminator'?
```
mcdurdin, line 378: I don't think we need this. We have the count.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+|---|------|------------|------------------------------------------|
+|16+|  32  | key2       | int: index into `key2` element           |
+|24+|  32  | key2count  | int: count of `key2` elements            |
+
+### C7043.2.x `key2`—Secondary key list
+
+This is a list of vkeys, but ordered by rows.  The `rows` elements point into the start and length of entries here.
+
+> TODO-LDML: Should there be a '0-vkey terminator'?
+> And shold the 0th element also be a 0 vkey so that key2=0 means null?
```
mcdurdin, line 392: I think we don't need null term. But I think this should also be in the `lyrs` section.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+|12 |  32  | reserved| padding                                  |
+
+For each `key2` element:
+
+Note that the first two fields are the index into the `keys` element, except arranged according to the keyboard layout.
+
+| ∆ | Bits | Name    | Description                              |
+|---|------|---------|------------------------------------------|
+|16+|  32  | vkey    | int: vkey ID                             |
+|20+|  32  | mod     | int: modifier key flags                  |
```
mcdurdin, line 408: I don't think this is correct. This should be an array of `keys` indices, and have no reference either to `vkey` or to `mod`.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+
+| ∆ | Bits | Name          | Description                              |
+|---|------|---------------|------------------------------------------|
+| 0 |  32  | ident         | `disp`                                   |
+| 4 |  32  | size          | int: Length of section                   |
+| 8 |  32  | count         | int: Total number of disp elements       |
+|12 |  32  | baseCharacter | str: If non-null, default base.          |
+|16 | 128  | reserved      | padding (future displayOptions)          |
+
+The default baseCharacter is U+25CC, if baseCharacter is null.
```
mcdurdin, line 420:
```suggestion
The default `baseCharacter` is U+25CC, if `baseCharacter` is null.
```

mcdurdin (changes requested):
I feel like I am destroying your work here, so big apologies! I really think that structurally, the whole lyrs->layr->rows->row->keys thing should be one big section... I am willing to be convinced, but otherwise we end up with a lot of boilerplate handling for each section, and redundant data also.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,115 @@ For each key:
...
+|16 |  32  | layercount | int: number of layer entries             |
+|20 |  96  | reserved   | padding                                  |
+
+Note layerstart must point after the end of the layer lists
+
+For each 'layers' - the layer lists
+
+| ∆ | Bits | Name       | Description                              |
+|---|------|------------|------------------------------------------|
+|16+|  32  | flags      | int: per-layers options                  |
```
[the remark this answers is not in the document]
srl295: [ ] nope, thanks

core/src/ldml/C7043_ldml.md
srl295: makes some sense.

core/src/ldml/C7043_ldml.md
srl295: that's a better idea.

srl295: I'm a little unsure about how the modifier flags in here on the target `key` element.

mcdurdin: modifier flags are determined by the layer, aren't they?

srl295: for hardware-yes.  good point, each key inherits the modifier flags from that layer. For non hardware, it's more complex.  This gets back to the discussion about what `layer id="shift"` denotes for a non hardware layer.

mcdurdin: Is modifier even relevant for non-hardware? The output is defined by the `<key id>` association.

srl295: Agreed

srl295: More later

core/src/ldml/C7043_ldml.md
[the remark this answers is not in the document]
srl295: updated the deltas to be relative to 0 for each struct.

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,87 @@ For each key:
...
+|---|------|------------|----------------------------------------|
+| 0+|  32  | key        | int: index into key element            |
+| 4+|  32  | count      | int: count of key elements in this row |
+
+Each key entry corresponds to a key in the row.
+There are `keyCount` total key entries.
+
+| ∆ | Bits | Name    | Description                              |
+|---|------|---------|------------------------------------------|
+| 0+|  32  | vkey    | int: vkey ID                             |
```
mcdurdin, line 382: I'm not sure there is a 1-to-1 between vkey and key bag items? e.g. <kbd>s</kbd> vs <kbd>S</kbd> have the same vkey and differ only by modifier. We probably need a key bag index?

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,87 @@ For each key:
...
+| ∆ | Bits | Name       | Description                              |
+|---|------|------------|------------------------------------------|
+| 0 |  32  | ident      | `layr`                                   |
+| 4 |  32  | size       | int: Length of section                   |
+| 8 |  32  | listCount  | int: Number of layer lists               |
+|12 |  32  | layerCount | int: number of layer entries             |
+|16 |  32  | rowCount   | int: number of row entries               |
+|20 |  32  | keyCount   | int: number of key entries               |
+|24 |  64  | reserved   | padding                                  |
+
```
mcdurdin, line 341: Suggest listing the variable length data here in the table, and then adding subheadings for each sub-table, e.g. `#### layers sub-table`, `#### rows sub-table`, `#### keys sub-table`
```suggestion
|32 | var  | layers     | layers sub-table                         |
| - | var  | rows       | rows sub-table                           |
| - | var  | keys       | keys sub-table                           |

```

core/src/ldml/C7043_ldml.md
```diff
@@ -324,10 +324,87 @@ For each key:
...
+Each layer entry corresponds to one `<layer>` element
+There are `layerCount` total layer entries.
+
+| ∆ | Bits | Name       | Description                              |
+|---|------|------------|------------------------------------------|
+| 0+|  32  | id         | str: layer id such as `base` or `shift`  |
+| 4+|  32  | modifier   | str: modifier string                     |
+| 8+|  32  | row        | int: index into rows area (next section) |
+|12+|  32  | count      | int: number of `rows` elements for this layer  |
```
mcdurdin, line 367:
```suggestion
| ∆ | Bits | Name       | Description                                    |
|---|------|------------|------------------------------------------------|
| 0+|  32  | id         | str: layer id such as `base` or `shift`        |
| 4+|  32  | modifier   | str: modifier string                           |
| 8+|  32  | row        | int: index into rows area (next section)       |
|12+|  32  | count      | int: number of `rows` elements for this layer  |
```

mcdurdin (changes requested):
I think this is looking pretty close. Some more thoughts, one concern on `keyid`

core/src/ldml/C7043_ldml.md
srl295: this is a great suggestion that gives the overall picture in the main table

core/src/ldml/C7043_ldml.md
srl295: that would also future-proof us if there are other distinguishing features of a key, some future flag and so on.

srl295:
```suggestion
| 0+|  32  | key     | int: index into `keys` element           |
```

mcdurdin: Instead of `` `keys` element``, we want `` `keys` section``? But I don't think it currently exists as we want it -- the current `keys` section is designed around vkey+modifier and is optimized for hardware layout.

srl295: OK. May need a new keys section then (`key2`) that is designed around the entire data set- flicks, longpress and so on. This is where the duplication comes in which we expected.

mcdurdin: I wonder if renaming the original `keys` section to `hkey` (aka "hardware key") is worth the hassle?

srl295: The difference is still going to need to be explained. So I'm not sure it really buys much. In any event, I would rename after key2 is already in place.

It also might be that if key2 has vkey information, we can eventually drop the original keys array. Not a primary goal.

srl295:
@mcdurdin PTAL

Added a list of strings.

mcdurdin (approved):

PR #7570, opened 2022-10-29
fix(developer): ldml: fix for unescaping Unicode values 🙀

### LINKED ISSUE #7569: bug(developer): ldml does not support escaping 🙀
```xml
    <display to="\u{0300}" display="`" />
```

should go to U+0300  etc.

Note:
- [ ] see and resolve https://github.com/keymanapp/keyman/pull/7570#issuecomment-1372602027 before closing this once #7630 is merged. 

#7054

srl295:
fix(developer): ldml: add escaping 🙀

- common util unescapeString with its own tests
- add a test case using the disp element

Fixes #7569

@keymanapp-test-bot skip


developer/src/kmc-keyboard/test/fixtures/basic.xml
```diff
@@ -18,6 +18,11 @@
     <vkey from="Q" to="A" />
   </vkeys>
 
+  <displays>
```
srl295, line 21: Hm, maybe don't need this in basic also.

srl295:
test case is

```xml
    <display to="\u{0300}" display="`" />
    <displayOptions baseCharacter="x"/>
  </displays>
```

but shouldn't be in basic.xml.. yet

common/web/types/src/util/util.ts
```diff
@@ -15,3 +15,13 @@ export function boxXmlArray(o: any, x: string): void {
     }
   }
 }
+
+const MATCH_HEX_ESCAPE = /\\u{([0-9a-fA-F]{1,5})}/g;
```
mcdurdin, line 19:
```suggestion
const MATCH_HEX_ESCAPE = /\\u{([0-9a-fA-F]{1,6})}/g;
```

`\u{10FFFF}` has 6 digits.

Query: `\u{a}` doesn't seem like it should be acceptable? Do we want to allow control characters (below 0x20) in these strings? This perhaps should be handled outside this function though.

common/web/types/src/util/util.ts
```diff
@@ -15,3 +15,13 @@ export function boxXmlArray(o: any, x: string): void {
...
+
+export function unescapeString(s: string): string {
+  if(!s) return s;
+
+  s = s.replaceAll(MATCH_HEX_ESCAPE, (str,hex) => String.fromCodePoint(Number.parseInt(hex, 16)));
+
+  return s; // null imp
```
mcdurdin, line 26: What does this comment mean?

common/web/types/test/util/test-unescape.ts
```diff
@@ -0,0 +1,28 @@
...
+    assert.equal(unescapeString('\\u{0127}==\\u{0127}'), 'ħ==ħ');
+    assert.equal(unescapeString('Q\\u{0127}'), 'Qħ');
+  });
+
+  it("should correctly handle 1..5 char strings", function() {
+    assert.equal(unescapeString('\\u{9}'),     '\t');
+    assert.equal(unescapeString('\\u{4a}'),    'J');
+    assert.equal(unescapeString('\\u{3c8}'),   'ψ');
+    assert.equal(unescapeString('\\u{304B}'),  'か');
+    assert.equal(unescapeString('\\u{1e109}'), '𞄉');
```
mcdurdin, line 26:
```suggestion
    assert.equal(unescapeString('\\u{1e109}'), '\u{1e109}');
    assert.equal(unescapeString('\\u{10fff0}'), '\u{10fff0}');
```

Suggest using the ES6 escaped strings on the expecteds for clarity.

common/web/types/test/util/test-unescape.ts
```diff
@@ -0,0 +1,28 @@
...
+    assert.equal(unescapeString('abc'), 'abc');
+  });
+
+  it("should correctly unescape strings", function() {
+    assert.equal(unescapeString('\\u{0127}'), 'ħ');
+    assert.equal(unescapeString('\\u{0127}==\\u{0127}'), 'ħ==ħ');
+    assert.equal(unescapeString('Q\\u{0127}'), 'Qħ');
+  });
+
+  it("should correctly handle 1..5 char strings", function() {
```
mcdurdin, line 21:
```suggestion
  it("should correctly handle 1..6 char escapes", function() {
```

common/web/types/src/util/util.ts
```diff
@@ -15,3 +15,13 @@ export function boxXmlArray(o: any, x: string): void {
     }
   }
 }
+
+const MATCH_HEX_ESCAPE = /\\u{([0-9a-fA-F]{1,5})}/g;
+
+export function unescapeString(s: string): string {
+  if(!s) return s;
+
+  s = s.replaceAll(MATCH_HEX_ESCAPE, (str,hex) => String.fromCodePoint(Number.parseInt(hex, 16)));
```
mcdurdin, line 24: This will throw `RangeError` on invalid values such as `\u{110000}`. How do we want to handle that?

common/web/types/src/util/util.ts
```diff
@@ -15,3 +15,13 @@ export function boxXmlArray(o: any, x: string): void {
     }
   }
 }
+
+const MATCH_HEX_ESCAPE = /\\u{([0-9a-fA-F]{1,5})}/g;
+
+export function unescapeString(s: string): string {
+  if(!s) return s;
```
mcdurdin, line 22:
```suggestion
  if(!s) {
    return s;
  }
```

mcdurdin (changes requested):
Looking good; a couple of changes requested.

common/web/types/src/util/util.ts
srl295:
```suggestion
  return s;
```

it was a leftover!

common/web/types/src/util/util.ts
srl295: Perhaps it should be caught here and thrown as its own `CompilerError` specifically stating an invalid escape.

mcdurdin: That would be good. Then the compiler will handle `CompilerError`s and reframe them through the compiler logging. A real `RangeError` elsewhere will then be unhandled, as it should be.

We also want to block surrogate pairs and illegal chars somewhere, but not here.

common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -129,6 +133,8 @@ export class Strs extends Section {
       throw new Error('alloc_string: s must be a string, undefined, or null.');
     }
 
+    s = unescapeString(s);
```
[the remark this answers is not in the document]
srl295: That is a good question. I will double check.

Actually I can think of one class of cases:  Where the NMTOKEN(S) are being included, such as where the literal key ID is pulled in.

srl295:
@mcdurdin I think I have addressed the main code issues, except for:

- [ ] verify what should be unescaped and what shouldn't.

When I do that I'll move this out of draft.  The regex I wanted to write didn't work, yet.  So I am using a simpler one.

srl295:
One issue I see is that, for example, `longPress` needs to be escaped, but this PR is at the bottom of the pile right now and the `layr` PR is closer to the top.  So this may take a couple iterations until escaping is in the right place.  I will see if I can add a DTD annotation of some sort to the spec to make it easier to sort these out.

srl295:
Over on the DTD side, I'm proposing a new annotation for escaping: https://github.com/unicode-org/cldr/pull/2642 It's already in the text of the spec, but this makes it easier to see at a glance which attributes allow unescaping.

srl295:
this is now based on #7630  and so I can bring in the rest of the escaped attributes

srl295:
OK.  Pulled in the last  few attributes that needed escaping.  With unit tests.

common/web/types/src/kmx/kmx-plus.ts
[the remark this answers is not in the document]
srl295: I think this is now addressed.

mcdurdin:
I might wait until you have resolved merge conflicts before reviewing?

common/web/types/src/util/util.ts
srl295: it's thrown as an `UnescapeError` now.

srl295:
@mcdurdin as i noted i'm throwing an `EscapingError` if escaping goes awry.  Should that be caught in `Compiler.compile()` perhaps? Since it's cross module I didn't see a good way to get at `callbacks.reportMessage()` from util down in common/web/types.  Maybe a future PR, leaving this issue open?

developer/src/kmc-keyboard/test/test-keys.ts
```diff
@@ -17,6 +17,14 @@ describe('keys', function () {
     assert.equal(keys.keys.length, 1);
   });
 
+  it('should compile escaped keys data', function() {
+    let keys = loadSectionFixture(KeysCompiler, 'sections/keys/escaped.xml', compilerTestCallbacks) as Keys;
+    assert.isNotNull(keys);
+    assert.equal(compilerTestCallbacks.messages.length, 0);
+    assert.equal(keys.keys.length, 1);
+    assert.equal(keys.keys[0].to.value, '🪦');
```
mcdurdin, line 25: Perhaps should use escaped value here as it has notdef glyph for me.

mcdurdin (approved):
LGTM tiny nit only

PR #7584, opened 2022-11-01
fix(developer): generate a default icon for basic keyboard projects

### LINKED ISSUE #2803: bug(developer): New basic project doesn't add icon feature
Unlike when creating a new project based on a Windows keyboard, creating a new basic project doesn't add the icon feature and it should.

mcdurdin:
Fixes #2803.

The new basic project template will now include a default icon, generated either from the first BCP 47 tag, or else from the first 3 letters of the filename of the keyboard.

The icon will not be included if the keyboard does not target desktop platforms, as web/touch do not currently support the icon.

# User Testing

**TEST_BASIC:** Test that creating a new basic keyboard project includes an icon by default, with the first 3 letters of the BCP 47 tag or the filename. Once the keyboard is created, you can verify the icon in the keyboard editor, Icon tab. The text should usually fit within the icon square (`WWW` may be
difficult!)

**TEST_NO_ICON:** Test that creating a new basic keyboard project that targets only 'web' does not include an icon.

### DIFF
--- a/common/windows/delphi/ui/Keyman.System.Util.RenderLanguageIcon.pas
+++ b/common/windows/delphi/ui/Keyman.System.Util.RenderLanguageIcon.pas
@@ -7,11 +7,19 @@ uses
   System.UITypes,
   Vcl.Graphics;
 
-procedure DrawLanguageIcon(ACanvas: TCanvas; x, y: Integer; code: string);   // I3933
+{
+  Renders a two or three character language icon into the canvas
+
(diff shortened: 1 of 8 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.95-beta](https://keyman.com/downloads/releases/beta/16.0.95)

PR #7608, opened 2022-11-02
feat(web): Custom layer targets for Shift double-tap

mcdurdin:
Applies proposed feature https://community.software.sil.org/t/testing-caps-on-touch/6849/5:

1. A `caps` layer must still exist in the keyboard.
2. If there is at least one multitap subkey defined on the Shift key on the active layer, then the `nextlayer` property of the first multitap subkey for the Shift key will be used instead of the default `caps`.
3. This does not change the behavior of the Shift key otherwise.

For example, the following defines a valid multitap key on Shift on the `rightalt` layer to switch to a custom extended `rightalt-shift-caps` layer.

```
  {
    "id": "K_SHIFT",
    "text": "*Shift*",
    "width": 150,
    "sp": 1,
    "nextlayer": "rightalt-shift",
    "layer": "default",
    "multitap": [
      {
        "text": "",
        "id": "T_new_694",
        "sp": "1",
        "nextlayer": "rightalt-shift-caps"
      }
    ]
  },
```

This does not implement any other multitap functionality. It just allows the keyboard developer to specify a target layer for a given double-tap.

This change should be forward-compatible with the multitap gesture support.

@keymanapp-test-bot skip

We'll do some testing on this with @MattGyverLee once this hits beta.

### DIFF
--- a/web/source/osk/browser/pendingMultiTap.ts
+++ b/web/source/osk/browser/pendingMultiTap.ts
@@ -18,6 +18,7 @@ namespace com.keyman.osk {
     private _state: PendingMultiTapState = PendingMultiTapState.Waiting;
     private _timeout: Promise<void>;
     private cancelDelayFactor = 125; // 125msec * count
+    private _destinationLayerId;
 
     public get timeout() {
       return this._timeout;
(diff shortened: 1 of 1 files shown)


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.96-beta](https://keyman.com/downloads/releases/beta/16.0.96)

MattGyverLee:
I've just tested the Cameroon Azerty on my fork (Master branch of https://github.com/MattGyverLee/keyboards/tree/master/release/sil/sil_cameroon_azerty) on 16.0.96, and it works completely as expected! So far, I have only tested in the debugger and not on native Android/iOS.

Now I just need to decide what version of Keyman should be stable before I release it on keymanapp/keyman.

mcdurdin:
> It's interesting that your pseudocode above uses `rightalt-shift-caps`

That's because my pseudocode is dumb. It should be `rightalt-caps` ... 😆

PR #7620, opened 2022-11-04
fix(android/engine): Check temporary kmp file is valid

darcywong00:
Fixes some of the Sentry crashes in #6862 where the temporary kmp file during a package installation suddenly isn't available.

We still haven't found what can cause the temporary kmp file to be removed...

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM, seems like it is masking the real problem though? Perhaps we should be logging all those instances where we are `return specs;` without actually adding anything to `specs`?

darcywong00:
>  Perhaps we should be logging all those instances where we are return specs; without actually adding anything to specs?

I have mixed feelings about this. It would be helpful to know which "fail" caused this to return empty specs, but we still wouldn't have insight/control on the root cause...

PR #7650, opened 2022-11-08
fix(core): emit keystroke was writing to wrong queue

### LINKED ISSUE #7649: bug(core): emit keystroke action was queued out of order
The emit keystroke action is being queued out of order -- the `internal_process_queued_actions()` call adds everything from the kmx processor internal queue to the `state->actions` queue:

https://github.com/keymanapp/keyman/blob/68dc7e8cb9523dca441b5b5772ec8cc02782cbfb/core/src/kmx/kmx_processor.cpp#L264-L269

This causes the debugger to process an "unknown" type backspace event first, deleting any markers in the context prior to the insertion point, before processing the "marker" backspace event(s) that ended up later in the queue, which triggers this assertion.

Keyman Engine on Linux is ignoring marker-type events:

https://github.com/keymanapp/keyman/blob/badf5e6d4b30e3bc5e4a55d6c5d82b11bf6aaab4/linux/ibus-keyman/src/engine.c#L577-L579

Keyman Engine on Windows has ... layers so it's a little harder to piece together, but basically it  replicates the behaviour that Keyman Core does for us on a normal backspace and ignores the marker-type events (this could be improved in a future version):

https://github.com/keymanapp/keyman/blob/ee14b17b12dafc2b99149dfc67c93da12e19946a/windows/src/engine/keyman32/appint/aiWin2000Unicode.cpp#L156-L163

---

Sentry Issue: [KEYMAN-DEVELOPER-C1](https://sentry.io/organizations/keyman/issues/3355188386/?referrer=github_integration)

```
EAssertionFailed: Exception EAssertionFailed in module tike.exe at 009A25CC.
Assertion failure (C:\BuildAgent\work\82cac59564031c74\keyman\windows\src\developer\tike\child\UfrmDebug.pas, line 744).

  File "UfrmDebug.pas", line 745, in UfrmDebug.DoBackspace
    memo.SelText := '';
  File "System.pas", line 24855, in System.@Assert
  File "UfrmDebug.pas", line 744, in UfrmDebug.DoBackspace
    Assert(BackspaceType = KM_KBP_BT_UNKNOWN);
  File "UfrmDebug.pas", line 920, in UfrmDebug.TfrmDebug.ExecuteEventAction
    KM_KBP_IT_BACK:           DoBackspace(km_kbp_backspace_type(dwData));
  File "UfrmDebug.pas", line 621, in UfrmDebug.TfrmDebug.ExecuteEvent
    then ExecuteEventAction(n)
...
(12 additional frame(s) were not displayed)
```

mcdurdin:
Fixes #7649.

emit_keystroke code path in kmx processor was writing directly to the core queue instead of to the internal kmx processor queue. This caused it to be out-of-order in the actions sent to the engine/debugger.

Engines didn't really care but it broke the debugger, for example if only a deadkey was in the buffer and backspace was pressed, Developer would assert as it would get unexpected context for the deletions.

# User Testing

**TEST_REGRESSION:** I would like to ensure that this is covered in the Linux, Developer, and Windows regression tests in Beta. I don't know if it needs a separate test.

The test we should aim for is to ensure that backspace is working correctly in all contexts, including with rule processing, particularly when deadkeys/markers are involved.

### DIFF
--- a/developer/src/tike/main/Keyman.System.KeymanCoreDebug.pas
+++ b/developer/src/tike/main/Keyman.System.KeymanCoreDebug.pas
@@ -106,6 +106,11 @@ pkm_kbp_state_debug_item = ^km_kbp_state_debug_item;
 
 km_kbp_debug_type = type uint32_t;
 
+// These types are used only for debugging convenience
+type
+  km_kbp_state_debug_item_array = array[0..100] of km_kbp_state_debug_item;
+  pkm_kbp_state_debug_item_array = ^km_kbp_state_debug_item_array;
+
 const
   KM_KBP_DEBUG_BEGIN = 0;
   //KM_KBP_DEBUG_BEGIN_ANSI: km_kbp_debug_type = 1, // not supported; instead rewrite ansi keyboards to Unicode with mcompile
--- a/core/src/kmx/kmx_actions.h
+++ b/core/src/kmx/kmx_actions.h
@@ -31,8 +31,9 @@ typedef struct
 #define QIT_CAPSLOCK  8
 #define QIT_INVALIDATECONTEXT 9
 #define QIT_SAVEOPT 10
+#define QIT_EMIT_KEYSTROKE 11
 
-#define QIT_MAX     10
+#define QIT_MAX     11
 
 #define QVK_EXTENDED 0x00010000 // Flag for QIT_VKEYDOWN to indicate an extended key
 #define QVK_KEYMASK  0x0000FFFF
--- a/core/src/kmx/kmx_processor.cpp
+++ b/core/src/kmx/kmx_processor.cpp
@@ -178,6 +178,9 @@ kmx_processor::internal_process_queued_actions(km_kbp_state *state) {
           option{KM_KBP_OPT_KEYBOARD, rStoreToSave.dpName, rStoreToSave.dpString});
       }
       break;
+    case QIT_EMIT_KEYSTROKE:
+      state->actions().push_emit_keystroke();
+      break;
     case QIT_VKEYDOWN:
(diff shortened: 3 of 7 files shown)


developer/src/tike/main/Keyman.System.KeymanCoreDebug.pas
```diff
@@ -106,6 +106,11 @@ km_kbp_state_debug_item = record
 
 km_kbp_debug_type = type uint32_t;
 
+// These types are used only for debugging convenience
+type
+  km_kbp_state_debug_item_array = array[0..100] of km_kbp_state_debug_item;
+  pkm_kbp_state_debug_item_array = ^km_kbp_state_debug_item_array;
+
```
rc-swag, line 113: Where is this array and the one in `KeymanCore.pas` used?

mcdurdin: These are used only in the debugger. If you cast the relevant variable to `pkm_kbp_state_debug_item_array` type in the watch window, you can see the whole array rather than just the current item. It's annoying that the Delphi debugger doesn't do pointers to arrays well otherwise, in any way that I've been able to find!

mcdurdin:
TEST_REGRESSION PASSED. @bharanidharanj can you make sure that use of backspace is thoroughly tested in the regression test for Windows, Linux, Developer?

rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.99-beta](https://keyman.com/downloads/releases/beta/16.0.99)

PR #7661, opened 2022-11-09
fix(windows): Cleanup CEF more correctly in shutdown

mcdurdin:
Hopefully fixes KEYMAN-WINDOWS-1MR, and probably many related crash reports. Will monitor on alpha before porting to stable/beta.

This changes the form destruction sequence to make sure that it doesn't close the form until CEF has been shutdown correctly.

@keymanapp-test-bot skip

We are going to need to do a full regression on this, but that can wait until we exit beta.


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.3-alpha](https://keyman.com/downloads/releases/alpha/17.0.3)

mcdurdin:
This PR should be ported to stable-16.0 (or beta pre-16.0)

PR #7662, opened 2022-11-09
fix(web): Add additional checks on className.indexOf

darcywong00:
Fixes Sentry [report](https://sentry.io/organizations/keyman/issues/3728694505/?project=5983520&query=is%3Aunresolved&referrer=issue-stream) which goes back at least to 14.0.287

> Uncaught TypeError: Cannot read property 'indexOf' of undefined

@keymanapp-test-bot skip

### DIFF
--- a/web/source/dom/domEventHandlers.ts
+++ b/web/source/dom/domEventHandlers.ts
@@ -352,7 +352,7 @@ namespace com.keyman.dom {
         if(el && typeof el.kmwInput != 'undefined' && el.kmwInput == false) {
           return true;
         }
-      } else if(el && el.className.indexOf('kmw-disabled') >= 0) {
+      } else if(el?.className?.indexOf('kmw-disabled') >= 0) {
         return true;
       }
 


mcdurdin (approved):

web/source/dom/domEventHandlers.ts
```diff
@@ -352,7 +352,7 @@ namespace com.keyman.dom {
         if(el && typeof el.kmwInput != 'undefined' && el.kmwInput == false) {
           return true;
         }
-      } else if(el && el.className.indexOf('kmw-disabled') >= 0) {
+      } else if(el?.className?.indexOf('kmw-disabled') >= 0) {
```
mcdurdin, line 355:
```suggestion
      } else if(el?.classList?.indexOf('kmw-disabled') >= 0) {
```

Tidier because it's not a substring match and `classList` is [supported](https://caniuse.com/?search=classlist) in all our environments (`includes()` is not, although probably okay because of es6-shim?)

darcywong00: Would that also apply to domManager? It's got several uses (I only highlighted the first two)

https://github.com/keymanapp/keyman/blob/d1ab61eac11f9601931752d56336af707b97c2e0/web/source/dom/domManager.ts#L818-L826

Maybe a future PR?

mcdurdin: > Maybe a future PR?

OK

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.100-beta](https://keyman.com/downloads/releases/beta/16.0.100)

PR #7667, opened 2022-11-10
fix(windows): base layout support in Keyman Core

mcdurdin:
While Keyman Core already supported the base layout flags, they were not being passed in from Keyman Engine for Windows. This meant that any rules that used `baselayout` would fail, such as in sil_ipa with <kbd>AltGr</kbd>+<kbd>0</kbd> for `}`: https://github.com/keymanapp/keyboards/blob/0bf72bed74f563f823b1af002f6a8ccef8510ebf/release/sil/sil_ipa/source/sil_ipa.kmn#L348-L350

This patch adds support for the following properties:

* `KM_KBP_KMX_ENV_BASELAYOUT`: the legacy base layout name, e.g. kbdus.dll
* `KM_KBP_KMX_ENV_BASELAYOUTALT`: the BCP 47 base layout name, e.g. en-US
* `KM_KBP_KMX_ENV_SIMULATEALTGR`: whether the user has selected to emulate AltGr with Ctrl+Alt, for example on laptops without a right alt key.
* `KM_KBP_KMX_ENV_BASELAYOUTGIVESCTRLRALTFORRALT`: whether the system base layout in use generates Ctrl+RAlt when RAlt is pressed, which can be true for many European system layouts. This is necessary when associating the keyboard with a European language because the base keyboard becomes the language's default keyboard, rather than kbdus.

# User Testing

**TEST_BASELINE:** We need to verify that SIL IPA continues to work correctly on US English base keyboard. Verify that key sequences from the SIL IPA keyboard documentation work correctly in Word and Notepad.

**TEST_GERMAN:** Switch the 'Base Keyboard' in Keyman Configuration / Options tab to German. With the SIL IPA keyboard, type <kbd>a</kbd> <kbd>AltGr</kbd>+<kbd>0</kbd>. The output should be `aˈ`.

**TEST_SIMULATE_ALTGR_OFF:** We need to test the Simulate AltGr with Ctrl+Alt option in Keyman Configuration. First, turn the option off. Then with Khmer Angkor, press <kbd>AltGr</kbd>+<kbd>B</kbd>, and verify that it emits `ឞ` (U+179E). Then, press<kbd>LCtrl</kbd>+<kbd>LAlt</kbd>+<kbd>B</kbd>. You should get no output.

**TEST_SIMULATE_ALTGR_ON:** We need to test the Simulate AltGr with Ctrl+Alt option in Keyman Configuration. Now, turn the option on. Then with Khmer Angkor, press <kbd>AltGr</kbd>+<kbd>B</kbd>, and verify that it emits `ឞ` (U+179E). Then, press<kbd>LCtrl</kbd>+<kbd>LAlt</kbd>+<kbd>B</kbd>. It should also emit `ឞ` (U+179E).

---

Ref: https://community.software.sil.org/t/bug-report-on-keyman-ipa-sil-for-windows/4811

### DIFF
--- a/windows/src/engine/keyman32/CoreEnvironment.cpp
+++ b/windows/src/engine/keyman32/CoreEnvironment.cpp
@@ -1,0 +1,32 @@
+#include "pch.h"
+
+BOOL SetupCoreEnvironment(km_kbp_option_item **core_environment) {
+  km_kbp_option_item *items = new km_kbp_option_item[5];
+
+  items[0].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[0].key = KM_KBP_KMX_ENV_BASELAYOUT;
+  items[0].value = reinterpret_cast<km_kbp_cp*>(Globals::get_BaseKeyboardName());
+
+  items[1].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[1].key = KM_KBP_KMX_ENV_BASELAYOUTALT;
+  items[1].value = reinterpret_cast<km_kbp_cp*>(Globals::get_BaseKeyboardNameAlt());
+
+  items[2].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[2].key = KM_KBP_KMX_ENV_SIMULATEALTGR;
+  items[2].value = Globals::get_SimulateAltGr() ? u"1" : u"0";
+
+  items[3].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[3].key = KM_KBP_KMX_ENV_BASELAYOUTGIVESCTRLRALTFORRALT;
+  items[3].value = KeyboardGivesCtrlRAltForRAlt() ? u"1" : u"0";
+
+  items[4] = KM_KBP_OPTIONS_END;
+
+  *core_environment = items;
+  return TRUE;
+}
+
+void DeleteCoreEnvironment(km_kbp_option_item *core_environment) {
+  // All keys and values are pointers to owned memory, so only need to delete
+  // the array of options
+  delete [] core_environment;
+}
--- a/windows/src/engine/keyman32/K32_load.cpp
+++ b/windows/src/engine/keyman32/K32_load.cpp
@@ -101,12 +101,17 @@ BOOL LoadlpKeyboardCore(int i)
   }
   delete keyboardPath;
 
-  const km_kbp_option_item test_env_opts[] =
-  {
-    KM_KBP_OPTIONS_END
-  };
+  km_kbp_option_item *core_environment = nullptr;
 
-  err_status = km_kbp_state_create(_td->lpKeyboards[i].lpCoreKeyboard, test_env_opts, &_td->lpKeyboards[i].lpCoreKeyboardState);
+  if(!SetupCoreEnvironment(&core_environment)) {
+    SendDebugMessageFormat(0, sdmLoad, 0, "LoadlpKeyboardCore: Unable to set environment options for keyboard %ls", keyboardPath);
+    return FALSE;
+  }
+
+  err_status = km_kbp_state_create(_td->lpKeyboards[i].lpCoreKeyboard, core_environment, &_td->lpKeyboards[i].lpCoreKeyboardState);
+
+  DeleteCoreEnvironment(core_environment);
+
   if (err_status != KM_KBP_STATUS_OK) {
     SendDebugMessageFormat(
         0, sdmLoad, 0, "LoadlpKeyboardCore: km_kbp_state_create failed with error status [%d]", err_status);
--- a/windows/src/engine/keyman32/Keyman32.cpp
+++ b/windows/src/engine/keyman32/Keyman32.cpp
@@ -575,9 +575,18 @@ extern "C" BOOL  _declspec(dllexport) WINAPI Keyman_ForceKeyboard(PCSTR FileName
     delete keyboardPath;
     SendDebugMessageFormat(0, sdmGlobal, 0, "Keyman_ForceKeyboard Core: %s OK", FileName); // TODO: 5442 - remove word Core
 
-    const km_kbp_option_item test_env_opts[] = {KM_KBP_OPTIONS_END};
+    km_kbp_option_item *core_environment = nullptr;
+
+    if(!SetupCoreEnvironment(&core_environment)) {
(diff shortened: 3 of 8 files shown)


mcdurdin:
@rc-swag are the last two options required? How would we test these? I see that the legacy `IsEquivalentShift` is calling `KeyboardGivesCtrlRAltForRAlt` for states 2 and 6:

https://github.com/keymanapp/keyman/blob/54ee26f7a0253640c9ef143143d1a1d4bef045c4/windows/src/engine/keyman32/keystate.cpp#L149-L155

And we have the same thing in Core:

https://github.com/keymanapp/keyman/blob/be7c9c58a9e81f596a64065cbbb0a58b786d9446/core/src/kmx/kmx_modifiers.cpp#L130-L136

So presumably this was also getting wrong behaviour given we were never passing those flags into Core and just leaving them as their defaults (false and false):

https://github.com/keymanapp/keyman/blob/be7c9c58a9e81f596a64065cbbb0a58b786d9446/core/src/kmx/kmx_environment.cpp#L18-L19

Note that I have also not set the value for `KM_KBP_KMX_ENV_PLATFORM` given it matches our Windows env, but we should probably do that, as a future PR.

mcdurdin:
> How would we test these?

I've added a test for `SIMULATEALTGR`, but need to think through how to properly test the final option `BASELAYOUTGIVESCTRLRALTFORRALT`

rc-swag (approved):
lgtm

bharanidharanj:
* **TEST_BASELINE (PASSED):** Tested with Keyman 16.0.100-beta-test-7667 build in Windows 10 OS and here is my observation: Verified that the SIL IPA continues to work correctly on US English base keyboard. ie., typing on the hardware keyboard produces the expected output in the LibreOffice Writer.

bharanidharanj:
* **TEST_GERMAN (PASSED):** Verified after switching the Base Keyboard into German, then typing a AltGr+0 shows the expected output a'.

bharanidharanj:
* **TEST_SIMULATE_ALTGR_OFF (PASSED):** Turned OFF the Simulate AltGr option in the Keyman Configuration. Pressing AltGr+B shows the expected output in Khmer Angkor. Pressing LCtrl+LAlt+B gives no output. 

   


* **TEST_SIMULATE_ALTGR_ON (PASSED):** Turned ON the Simulate AltGr option in the Keyman Configuration. Pressing AltGr+B shows the expected output in Khmer Angkor. Pressing LCtrl+LAlt+B gives the same output.

windows/src/engine/keyman32/CoreEnvironment.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+  items[2].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[2].key = KM_KBP_KMX_ENV_SIMULATEALTGR;
+  items[2].value = Globals::get_SimulateAltGr() ? u"1" : u"0";
+
+  items[3].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[3].key = KM_KBP_KMX_ENV_BASELAYOUTGIVESCTRLRALTFORRALT;
+  items[3].value = KeyboardGivesCtrlRAltForRAlt() ? u"1" : u"0";
+
+  items[4] = KM_KBP_OPTIONS_END;
```
ermshiperete, line 22: On Linux we also set `KM_KBP_KMX_ENV_PLATFORM` - I assume we don't need that on Windows?

windows/src/engine/keyman32/CoreEnvironment.cpp
```diff
@@ -0,0 +1,32 @@
...
+  items[0].key = KM_KBP_KMX_ENV_BASELAYOUT;
+  items[0].value = reinterpret_cast<km_kbp_cp*>(Globals::get_BaseKeyboardName());
+
+  items[1].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[1].key = KM_KBP_KMX_ENV_BASELAYOUTALT;
+  items[1].value = reinterpret_cast<km_kbp_cp*>(Globals::get_BaseKeyboardNameAlt());
+
+  items[2].scope = KM_KBP_OPT_ENVIRONMENT;
+  items[2].key = KM_KBP_KMX_ENV_SIMULATEALTGR;
+  items[2].value = Globals::get_SimulateAltGr() ? u"1" : u"0";
```
ermshiperete, line 16: Do we have to set `KM_KBP_KMX_ENV_SIMULATEALTGR` and `KM_KBP_KMX_ENV_BASELAYOUTGIVESCTRLRALTFORRALT` on Linux as well?

Maybe not since on Linux we add options from dconf, so if it's stored there we're fine.

windows/src/engine/keyman32/CoreEnvironment.cpp
ermshiperete: ok, I see. It uses the default which happens to be "windows hardware desktop native". Seems to me it would be cleaner and clearer to explicitly set it...

ermshiperete (approved):

windows/src/engine/keyman32/CoreEnvironment.cpp
mcdurdin: I agree, we should add that as well. I might do it as a separate PR?

windows/src/engine/keyman32/CoreEnvironment.cpp
mcdurdin: The `KM_KBP_KMX_ENV_SIMULATEALTGR` option is useful on Linux, but I am not sure if we surface it for the user?

`KM_KBP_KMX_ENV_BASELAYOUTGIVESCTRLRALTFORRALT` is probably only helpful on Windows where the Windows base keyboard may emit LCtrl+RAlt when RAlt is pressed, for some keyboards.

ermshiperete: >The KM_KBP_KMX_ENV_SIMULATEALTGR option is useful on Linux, but I am not sure if we surface it for the user?

Yes, that doesn't currently surface for the user - unless he does the equivalent of a registry hack and sets the dconf setting manually...

keyman-server:

windows/src/engine/keyman32/CoreEnvironment.cpp
mcdurdin: Future feature for Keyman for Linux: make the Simulate AltGr option available (and then we need to test it thoroughly I guess!)

PR #7668, opened 2022-11-10
fix(android/engine): Use IME package name if query permission denied

### LINKED ISSUE #7664: bug(android/app): PackageManager$NameNotFoundException
Sentry Issue: [KEYMAN-ANDROID-ZE](https://sentry.io/organizations/keyman/issues/3228497252/?referrer=github_integration)

```
PackageManager$NameNotFoundException: eu.dominicweb.africankeyboard
    at android.app.ApplicationPackageManager.getApplicationInfoAsUser(ApplicationPackageManager.java:428)
    at android.app.ApplicationPackageManager.getApplicationInfo(ApplicationPack
TageManager.java:417)
    at com.tavultesoft.kmea.KeyboardPickerActivity.getIMEList(KeyboardPickerActivity.java:548)
    at com.tavultesoft.kmea.KeyboardPickerActivity.onResume(KeyboardPickerActivity.java:254)
    at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1571)
...
(13 additional frame(s) were not displayed)
```

As a system keyboard, Keyman will query for other IME's to append to the keyboard picker menu. Starting with Android 11, it seems apps must declare in their Manifest each specific package name to query. Otherwise, get a bunch of NameNotFoundExceptions in the logs.

While we could include common IME's to query in the list, there's also a flag `QUERY_ALL_PACKAGES`. But that would involve a "Permission Declaration" form with the Google Play store review team.

See: https://support.google.com/googleplay/android-developer/answer/9214102#

darcywong00:
Fixes #7664 

Rather than applying for the QUERY_ALL_PACKAGES permission, we'll handle NameNotFoundExceptions as follows

> Quick fix seems like would be we handle NameNotFoundExceptions and just use the package name in that case?

So ~on Android 11+ devices,~ we'll fallback to display the packageName (not the descriptive keyboard name) on the keyboard picker.

## User Testing
Setup
1. Install the PR build of Keyman for Android on an Android 11+ device/emulator. You'll need to be signed into the Google Play Store for the other setup step
2. From the Google Play Store, install another IME (e.g "dominicweb african keyboard" is the one from some Sentry logs)
3. Setup the IME as an enabled system keyboard

* **TEST_PACKAGE_NAME_LISTED** - Verifies NameNotFound exception isn't thrown and other IME's appear on the keyboard picker
1. Launch Keyman for Android
2. On the "Get Started" menu,  enable Keyman as a system keyboard and set as the default keyboard
3. Open Chrome browser and click on the search box to display the keybaord
4. If Keyman keyboard doesn't appear, change keyboard to Keyman
5. Long-press on the globe key to display the Keyman keyboard picker
6. Verify the following:
    * Keyboard picker displays other IMEs (Samsung and Gboard should appear fine)
    * No Toast errors displayed about "NameNotFound exceptions"
    * The Dominic African keyboard appears as "eu.dominicweb.africankeyboard"
keyboard picker
7. Select "eu.dominicweb.africankeyboard"
8. Verify the Domonic African keyboard appears

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
@@ -538,25 +538,23 @@ public final class KeyboardPickerActivity extends BaseActivity {
     PackageManager packageManager = context.getPackageManager();
     for (InputMethodInfo imeInfo : imeList) {
       String id = imeInfo.getId();
-      String imeName = "";
       if (!id.startsWith(selfPackageName)) {
         // Attempt to get a readable name
         // Reference: https://stackoverflow.com/questions/62512501/get-human-readable-name-of-default-keyboard-not-package-name
         ComponentName componentName = ComponentName.unflattenFromString(id);
         if (componentName != null) {
           String packageName = componentName.getPackageName();
+          String imeName = "";
           try {
             ApplicationInfo info = packageManager.getApplicationInfo(packageName, PackageManager.GET_META_DATA);
             imeName = (String)packageManager.getApplicationLabel(info);
           } catch (PackageManager.NameNotFoundException e) {
             // For Android 11+, this exception is thrown because we don't have QUERY_ALL_PACKAGES permission.
             // We'll just display the package name instead.
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-              imeName = packageName;
-            } else {
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
               KMLog.LogException(TAG, "Name not found", e);
-              continue;
             }
+            imeName = packageName;
           }
 
           if (!imeName.isEmpty()) {
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
@@ -29,6 +29,7 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.graphics.Typeface;
 import android.inputmethodservice.InputMethodService;
+import android.os.Build;
 import android.os.Bundle;
 import android.view.MenuItem;
 import android.view.View;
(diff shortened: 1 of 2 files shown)


darcywong00:
I contemplated :cherries: picking this to stable-15.0, but then this will become stable in ~2 weeks

mcdurdin:
> I contemplated 🍒 picking this to stable-15.0, but then this will become stable in ~2 weeks

Happy to have it cherry-picked to stable-15.0, as we don't have a guaranteed release date for 16.0.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
```diff
@@ -537,23 +538,32 @@ private static ArrayList<HashMap<String, String>> getIMEList(Context context) {
...
+            imeName = (String)packageManager.getApplicationLabel(info);
+          } catch (PackageManager.NameNotFoundException e) {
+            // For Android 11+, this exception is thrown because we don't have QUERY_ALL_PACKAGES permission.
+            // We'll just display the package name instead.
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+              imeName = packageName;
+            } else {
+              KMLog.LogException(TAG, "Name not found", e);
+              continue;
+            }
```
mcdurdin, line 559: So, for earlier versions, you just won't list the input method? Why not use the same fallback?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
```diff
@@ -537,23 +538,32 @@ private static ArrayList<HashMap<String, String>> getIMEList(Context context) {
     PackageManager packageManager = context.getPackageManager();
     for (InputMethodInfo imeInfo : imeList) {
       String id = imeInfo.getId();
+      String imeName = "";
```
mcdurdin, line 541: This could probably go after line 547?

mcdurdin (approved):

android/KMEA/app/src/main/java/com/tavultesoft/kmea/KeyboardPickerActivity.java
darcywong00: k, I can strike the `continue` and use the same fallback

darcywong00:
UX: Would it look better to replace `.` with spaces and title case the package name?

mcdurdin:
> UX: Would it look better to replace `.` with spaces and title case the package name?

I wouldn't make any changes to the id.

bharanidharanj:
* **TEST_PACKAGE_NAME_LISTED (PASSED):** Tested this PR in my Android Mobile device (Redmi Note 8 Pro - Ver 11) and here is my observation:  Verified that the 1. Keyboard picker displays other IMEs. 2. No Toast errors appeared on the screen. 3. The Dominican African keyboard appears as "eu.dominicweb.africankeyboard". 4. The Dominic African keyboard appeared on the screen without showing any error message. 

..Keyboard picker menu displays other IMEs

   



..The Dominic African Keyboard

keyman-server:

Issue #7364, opened 2022-09-27
bug(android): Version Conflict happens in APKs page after installing Keyman


bharanidharanj:
**Describe the bug**
I noticed that there is a version conflict happens after installing different versions of Keyman. 

**To Reproduce**
1. Install the pre-release version of (here, it is 16.0.61-alpha) Keyman. 
2. Open Keyman In-App.
3. Click the 'More Info' option from the 'Get Started' menu. 
4. We may notice the pre-release version of Keyman (16.0.61-alpha) at the bottom of the page. 
5. Close Keyman. 
6. Install current release version of (15.0.270) without uninstalling the pre-release version. 
7. Verify that Keyman 15.0.270-beta has been installed in the device. 
8. Open Keyman / More Info page. 
9. Here it is showing 'Keyman 16.0.61-alpha' build.
10. Open File Manager / APK files in the Android Mobile device. 
11. Here, I noticed that 'Keyman 15.0.27.apk' Installed message. 

**Expected behavior**
A warning message should appear with a message 'the Keyman has been already installed' on the screen. 

**Screenshots**

..Info page shows pre-release version

   <img width="300" src="https://user-images.githubusercontent.com/19683143/192501484-357d3c86-c4fc-4f22-b7ff-8fd691c033d3.jpg">

..APKs page shows current-release version

   <img width="300" src="https://user-images.githubusercontent.com/19683143/192501519-a54596b3-b6a3-48a8-b440-2a9376ed5875.jpg">

..APKs page showing both versions are installed..

   <img width="300" src="https://user-images.githubusercontent.com/19683143/192502309-15197218-a9ce-4d16-a0be-581899a11c05.jpg">

---

**Keyman for Android:**
   Device:             Android API 26 emulator, Redmi Note 8 Pro 
 - OS:                 Android 8.0, Android 11
 - Keyman version:     15.0.270, 16.0.70
 - Target Application: Keyman App

---

**Keyboard**
 - Keyboard name:    
 - Keyboard version: 
 - Language name:    

**Additional context**
I am seeing different behavior If I install the stable release (15.0.270) over the pre-release version (16.0.70) build. ie., the APKs page shows both the versions are installed in the device. But, I noticed that (16.0.70) version was installed in the device. I have attached the Screenshot (both versions installed) for your reference. Thanks.

mcdurdin:
[...]

PR #7674, opened 2022-11-10
fix(android/app): Append tier to app name

darcywong00:
Addresses #7364 by appending the tier to the app name. For stable, we'll keep the default 'Keyman'.

Developers likely aren't going to have `TIER` defined as an environment variable though so I'm updating version.gradle to include VERSION_TIER. (CI only sets it for Release- and not Test- builds anyway)

## User Testing
Setup - Install the PR build of Keyman for Android 

* **TEST_APP_NAME** - Verifies the app name shows the tier
1. From the device launcher, show all apps
2. Browse to Keyman and verify the name is "Keyman Beta-test-7674"
3. Launch Keyman 
4. On the "Get Started" menu, enable Keyman as a system keyboard
5. Verify the Android system keyboard list has "Keyman Beta-test-7674"
6. On the Keyman "Get Started" menu, set Keyman as default keyboard
7. Verify the Android "Choose input method" list shows "Keyman Beta-test-7674"

### DIFF
--- a/android/KMAPro/kMAPro/build.gradle
+++ b/android/KMAPro/kMAPro/build.gradle
@@ -73,6 +73,8 @@ android {
     applicationVariants.all { variant ->
         // Append tier and environment to app name
         // Beware: Gradle sync will nullify these values
+        executeBuildUtils()
+
         String appSuffix = ''
         String environment = VERSION_ENVIRONMENT
         String tag = VERSION_TAG
@@ -152,6 +154,25 @@ dependencies {
     // we "lose" it in the dependency management)
     implementation ('com.github.kenglxn.QRGen:android:2.6.0') {
         transitive = true
+    }
+}
+
+// Use build-utils.sh to get common build version info.
+// Beware: Android Studio IDE Gradle sync will invalidate these values
+def executeBuildUtils() {
+    try {
+        exec {
+            commandLine 'bash', '-c', 'pwd; THIS_SCRIPT=build.gradle; . ../../../resources/build/build-utils.sh'
+        }
+    }
+    catch(Exception e) {
+        GradleException("Exception: $e")
+    }
+    // Debugging
+    println "version.gradle TAG: " + System.getenv("VERSION_TAG") + ", ENVIRONMENT: " + System.getenv("VERSION_ENVIRONMENT");
+    ext {
+        VERSION_TAG=System.getenv("VERSION_TAG")
+        VERSION_ENVIRONMENT=System.getenv("VERSION_ENVIRONMENT")
     }
 }
 
--- a/android/KMAPro/kMAPro/build.gradle
+++ b/android/KMAPro/kMAPro/build.gradle
@@ -71,17 +71,21 @@ android {
     }
 
     applicationVariants.all { variant ->
-        String tier = ''
-        switch (VERSION_TIER) {
-            case 'beta':
-                tier = ' Beta'
-                break
-            case 'stable':
-                break
-            default:
-                tier = ' Alpha'
+        // Append tier and environment to app name
+        // Beware: Gradle sync will nullify these values
+        String appSuffix = ''
+        String environment = VERSION_ENVIRONMENT
+        String tag = VERSION_TAG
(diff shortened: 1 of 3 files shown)


mcdurdin:
Is there any way to use the build-utils.sh environment status for -local, -test to determine local and PR builds? That would be really helpful!

android/KMAPro/kMAPro/build.gradle
```diff
@@ -138,6 +157,25 @@ dependencies {
     }
 }
 
+// Use build-utils.sh to get common build version info.
+// Beware: Android Studio IDE Gradle sync will invalidate these values
+def executeBuildUtils() {
+    try {
+        exec {
+            commandLine 'bash', '-c', 'pwd; THIS_SCRIPT=build.gradle; . ../../../resources/build/build-utils.sh'
```
darcywong00, line 165: > Is there any way to use the build-utils.sh environment status for -local, -test to determine local and PR builds? That would be really helpful!

I'm trying to go down that path but am running into some pain points:
1. Gradle calling a shell script will vary on WIndows vs Linux build agent (gosh!). atm this fails on Windows
2. The exported variables from build-utils.sh don't seem to persist when the Android Studio IDE does a Gradle sync (happens when installing the apk through the IDE).

darcywong00: I also had tried putting this in version.gradle but that breaks the FV Android build (uses the exact same `android/version.gradle` file)

darcywong00: Ah, since the parent shell has already called build-utils, hopefully all the vars are in the env.

The Gradle sync issue is still unresolved

mcdurdin: When there is a Gradle sync via IDE we can definitely just say 'local'.

darcywong00: OK - I'll update so local / IDE build will say "Keyman Beta-local"
And for a PR build it would be "Keyman Beta-test-1234"

android/KMAPro/kMAPro/build.gradle
```diff
@@ -70,6 +70,29 @@ android {
...
+        if (match.find()) {
+            // -test-PR_NUMBER
+            appSuffix = appSuffix + match.group(1)
+        } else {
+            String environment = VERSION_ENVIRONMENT ? VERSION_ENVIRONMENT : 'local';
+            if (environment.matches("local")) {
+                appSuffix = appSuffix + '-' + environment;
+            }
+        }
+        println "tag: " + tag + ", appSuffix: " + appSuffix;
```
darcywong00, line 93: TODO: Remove before merging

bharanidharanj:
# Test Results

* **TEST_APP_NAME (PASSED):** Tested with the attached PR build in my Android Mobile device (ver 11) and verified that it is showing the Keyman app name with tier in the following views.

mcdurdin (approved):

android/KMAPro/kMAPro/build.gradle
```diff
@@ -70,6 +70,29 @@ android {
     productFlavors {
     }
 
+    applicationVariants.all { variant ->
+        // Append tag to app name
+        // Gradle sync will nullify these values, so have a fallback
```
mcdurdin, line 75:
```suggestion
        // Append tag to app name to help differentiate versions for app testers:
        //  - stable builds have no suffix
        //  - beta and alpha release builds have "Beta" and "Alpha" suffixes
        //  - local builds (incl. from IDE) have "-local" suffix
        //  - PR builds have "-test-1234" suffix (where 1234 is PR#)
        // Gradle sync will nullify these values, so have a fallback
```

android/KMAPro/kMAPro/build.gradle
mcdurdin: Did you do this?

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.102-beta](https://keyman.com/downloads/releases/beta/16.0.102)

PR #7678, opened 2022-11-10
fix(linux): Fix keyboard icon in system tray

### LINKED ISSUE #7547: bug (linux): keyboard icon doesn't display in system tray
### Describe the bug

In Linux, distros using Gnome (such as Ubuntu) as their desktop environments integrate ibus into their internal keyboard layout solution, so in their panel ("system tray") they do not support using the ibus keyboard icons, instead they show a 2 character text-based short name for the language (or "en1" and "en2" if there are different keyboard layouts for english, for example).

However, other desktop environments that do not refactor ibus but instead use the ibus tray icon directly do show the keyboard icons in the panel ("system tray"). This is much preferred. Unfortunately, Keyman keyboards only show the generic Keyman icon instead of the keyboard icon. This means that the user is not sure from a visual glance what keyboard is active. Screenshots below are using Cinnamon (default DE for Wasta-Linux, for example), but the same behavior exists in other DEs such as XFCE, Mate, etc. Note that ibus-kmfl keyboards do correctly show the keyboard icon in the ibus tray icon.

---
Default ibus icon: default ibus tray icon

---

Keyman generic keyboard ibus tray icon: keyman ibus icon

---

ibus-kmfl keyboard ibus tray icon: ibus-kmfl ibus icon

---

### Reproduce the bug

_No response_

### Expected behavior

_No response_

### Related issues

#1182 seems like it may be related, but this was supposedly fixed some years ago? It is about icon format, there may be something to this.

### Keyman apps

- [ ] Keyman for Android
- [X] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb

### Keyman version

15.0.271-1+focal1

### Operating system

Ubuntu / Wasta-Linux - Cinnamon DE

### Device

Linux laptop

### Target application

_No response_

### Browser

_No response_

### Keyboard name

All

### Keyboard version

_No response_

### Language name

_No response_

### Additional context

_No response_

ermshiperete:
Show the keyboard specific icon in the system tray on Wasta instead of the generic keyman one.

Fixes #7547.

# User Testing

## Preparations

- The test should be run on Wasta 20.04 with Cinnamon

- Install the Khmer Angkor keyboard

- Reboot

## Tests

**TEST_ICON**: keyboard specific icon shows

- switch to Khmer Angkor keyboard by clicking on the icon in the system tray:

  image

- verify that the keyboard specific icon shows instead of the generic Keyman icon:

  image

### DIFF
--- a/linux/ibus-keyman/src/keymanutil.c
+++ b/linux/ibus-keyman/src/keymanutil.c
@@ -104,7 +104,7 @@ gchar * keyman_get_icon_file(const gchar *kmx_file)
 
     p=rindex(kmx_file,'.');
     filename = g_strndup(kmx_file, p-kmx_file);
-    full_path_to_icon_file=g_strdup_printf("%s.ico.png", filename);
+    full_path_to_icon_file=g_strdup_printf("%s.bmp.png", filename);
     g_free(filename);
 
     if (!g_file_test(full_path_to_icon_file, G_FILE_TEST_EXISTS)) {


mcdurdin (approved):
Well that is an easier fix than I was expecting! Are the image files always named to .bmp.png? (because Keyman keyboards can have .ico or .bmp format images embedded)

mcdurdin:
@rik-shaw FYI

bharanidharanj:
* **TEST_ICON (PASSED):** Tested with the attached PR build in Wasta 20.04 Linux OS (VM) and here is my observation: Able to see the Khmer Angkor Icon in the system tray. (after switching to Khmer Angkor Keyboard).

ermshiperete:
> Well that is an easier fix than I was expecting! Are the image files always named to .bmp.png? (because Keyman keyboards can have .ico or .bmp format images embedded)

I think it's always .bmp.png. During keyboard installation, if the keyboard has a .ico file we convert it first to a .bmp and then do the regular .bmp to .png conversion.

rik-shaw:
Thanks again to the whole team for this very helpful fix. I regret not reporting it earlier. Honestly, I think my lack of awareness of `km-package install --shared` has had me sticking to older `kmfl` keyboards since having system-wide pre-installed keyboards in Wasta-Linux is a big benefit.

Question: do beta / daily builds of keyman exist maybe in a separate PPA? Just curious how easy it is to test for non-developers like me. Of course it is understood that such a PPA wouldn't be for regular end users.

ermshiperete:
Yes, there's a PPA for the beta versions: https://launchpad.net/~keymanapp/+archive/ubuntu/keyman-beta :smile: 

They also get uploaded to the `-proposed` section on llso (e.g. `jammy-proposed`).

keyman-server:

mcdurdin:
> Yes, there's a PPA for the beta versions: https://launchpad.net/~keymanapp/+archive/ubuntu/keyman-beta 😄

And for alpha versions: https://launchpad.net/~keymanapp/+archive/ubuntu/keyman-alpha (which are built almost every night)

You can also see latest version numbers quickly at https://status.keyman.com/

rik-shaw:
Thanks again, everyone! I haven't been able to be very active the past few years with updates to the SIL Ethiopia keyboards, but am certainly quite motivated especially with this updates. Mobile layouts are something I never tackled with them, they could benefit from more standardized keycharts, etc.

rik-shaw:
Thank you again team, I can confirm that with `16.0.108-beta` icons now show in the tray, in the keyboard lists, in ibus-preferences, and in alt-tab. Nice work!

Question (or this may be a new bug report?): resolution is low (esp in alt-tab where it will show larger icons than in the typical taskbar size), is there any discussion or plan to increase icon size in this age of hi-res screens?

mcdurdin:
> Question (or this may be a new bug report?): resolution is low (esp in alt-tab where it will show larger icons than in the typical taskbar size), is there any discussion or plan to increase icon size in this age of hi-res screens?

This is supported in Keyman for Windows and multi-resolution icons can be loaded in Keyman Developer with the .ico format (Keyman Developer won't allow you to edit multi-resolution icons at this time, but you can import from other tools). However, Keyman for Linux AFAIK currently only loads the one resolution. So yes, please open this as a separate feature request.

rik-shaw:
> This is supported in Keyman for Windows and multi-resolution icons can be loaded in Keyman Developer with the .ico format (Keyman Developer won't allow you to edit multi-resolution icons at this time, but you can import from other tools). However, Keyman for Linux AFAIK currently only loads the one resolution. So yes, please open this as a separate feature request.

@mcdurdin thanks again for your help, as you can see I am not keeping up with my todo list :-) I noticed the `gff_amharic.ico` file is 64x64 and looks great in Linux. It must be a "multi-res icon file" as you reference above? If I could just use a 64x64 single res icon I would be happy, not sure if this would work or not? I'll try to get keyman developer (Windows) installed in a VM to look at the GFF icon, maybe I will learn something, or find some possible multi-res icon editor that may work to inspect it if it isn't possible to just include a single-res 64x64 icon.

mcdurdin:
If the icons are going to be identical, just automatically scaled with a tool, then there's not a lot of value in different sizes -- the OS does that anyway.

Generally, smaller icons should be thematically simpler, e.g. losing shadows, 3D effects, perspective, as well as removing detail, whereas larger icons can include more of all that. But auto scaling doesn't give you that.

Don't include up-scaled images -- the OS will do that anyway.

If you want a higher-res one for hidpi monitors, then 16 and 64 should be sufficient. If your original is 256 or higher, then include a 256 icon as well.

If you were going to create a full set, tweaking each size for clarity and style, then, include at least, 16, 32, and 64. May also include 256, and possibly 24, 48.

rik-shaw:
@mcdurdin I had misunderstood your prior comments, I had been led to believe that a 16x16 icon was required by Keyman, so the only way to have anything higher would be to use a multi-res icon that includes a 16x16 resolution. I am not sure where back in history I had been led to believe 16x16  was required, but as you note I don't have a need to refactor the icons, and I intentionally keep them without an alpha channel so that light / dark variants aren't needed.

The original graphic is .svg so can scale to any size. For hidpi alt-tab switching 64 is a bit minimal and is slightly blurry: 128 works better, I will go with that for now and drop the multi-res.

mcdurdin:
> I had been led to believe 16x16 was required

I don't believe it is required, but we are not testing that, so that may not be a safe assumption. I would still recommend always including 16x16.

rik-shaw:
> I don't believe it is required, but we are not testing that, so that may not be a safe assumption. I would still recommend always including 16x16.

OK sorry replying back and forth in comments leads to misunderstanding. I thought you were advising me above to not worry about using a multi-res icon since I am not refactoring it for smaller / higher resolutions but am simply scaling the same source image. So I thought just including the single higher res icon was recommended.

But from your most recent comment, it seems I should use a multi-res icon (even though automatically scaled with image magick)?

mcdurdin:
> But from your most recent comment, it seems I should use a multi-res icon (even though automatically scaled with image magick)?

Yes, 128 + 16 at least, perhaps also 48 if you like. Mostly, I think we should always include 16 for safety as we haven't tested without it in all contexts (particularly Windows which is finicky about icons).

PR #7679, opened 2022-11-10
fix(android/engine): Use IME package name if query permission denied 🍒

darcywong00:
:cherries: pick of #7664 to stable-15.0

Rather than applying for the QUERY_ALL_PACKAGES permission, we'll handle NameNotFoundExceptions as follows

> Quick fix seems like would be we handle NameNotFoundExceptions and just use the package name in that case?

So we'll fallback to display the packageName (not the descriptive keyboard name) on the keyboard picker.

## User Testing
Setup
1. Install the PR build of Keyman for Android on an Android 11+ device/emulator. You'll need to be signed into the Google Play Store for the other setup step
2. From the Google Play Store, install another IME (e.g "dominicweb african keyboard" is the one from some Sentry logs)
3. Setup the IME as an enabled system keyboard

* **TEST_PACKAGE_NAME_LISTED** - Verifies NameNotFound exception isn't thrown and other IME's appear on the keyboard picker
1. Launch Keyman for Android
2. On the "Get Started" menu,  enable Keyman as a system keyboard and set as the default keyboard
3. Open Chrome browser and click on the search box to display the keybaord
4. If Keyman keyboard doesn't appear, change keyboard to Keyman
5. Long-press on the globe key to display the Keyman keyboard picker
6. Verify the following:
    * Keyboard picker displays other IMEs (Samsung and Gboard should appear fine)
    * No Toast errors displayed about "NameNotFound exceptions"
    * The Dominic African keyboard appears as "eu.dominicweb.africankeyboard"
keyboard picker
7. Select "eu.dominicweb.africankeyboard"
8. Verify the Domonic African keyboard appears

### DIFF
--- a/android/KMAPro/kMAPro/src/main/AndroidManifest.xml
+++ b/android/KMAPro/kMAPro/src/main/AndroidManifest.xml
@@ -9,6 +9,10 @@
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
   <uses-permission android:name="android.permission.VIBRATE" />
 
+  <!-- Ideally we would declare this permission so we can get other installed keyboard names.
+  But would involve a long review process with the Play Store:
+  https://support.google.com/googleplay/android-developer/answer/9214102#
+  <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" /> -->
   <application
     android:allowBackup="true"
     android:hardwareAccelerated="true"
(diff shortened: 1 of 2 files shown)


bharanidharanj:
* **TEST_PACKAGE_NAME_LISTED (PASSED):** Tested this PR with Keyman 15.0.273-test-7679 build in my Android Mobile (Redmi Note 8 Pro - Ver 11) and here is my observation: Verified that the 1. Keyboard picker displays other IMEs. 2. No Toast errors appeared on the screen. 3. The Dominican African keyboard appears as "eu.dominicweb.africankeyboard". 4. The Dominic African keyboard appeared on the screen without showing any error message.

..Keyman Picker Menu:

   


..Dominic African Keyboard:

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.273](https://keyman.com/downloads/releases/stable/15.0.273)

PR #7686, opened 2022-11-10
refactor(linux): Use consts instead of strings

ermshiperete:
This refactoring makes use of the already defined strings. It also simplifies setting the values by directly specifying an UTF-16 string.

Motivated by the changes in PR #7667.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -38,6 +38,7 @@
 #endif
 
 #include <keyman/keyboardprocessor.h>
+#include <keyman/keyboardprocessor_consts.h>
 
 #include "keymanutil.h"
 #include "keyman-service.h"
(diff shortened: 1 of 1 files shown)


mcdurdin:
You are still freeing those strings later in the function‼

mcdurdin (changes requested):
This will surely crash seeing as you are `g_free`ing the const strings :grin:

ermshiperete:
ahh, you're right! Thanks for noticing!

mcdurdin (approved):
LGTM.

I notice that Windows sets the keyboard opts [in a separate call](https://github.com/keymanapp/keyman/blob/ee14b17b12dafc2b99149dfc67c93da12e19946a/windows/src/engine/keyman32/keyboardoptions.cpp#L236-L285) to `km_kbp_state_options_update` which may be worth considering to avoid array arithmetic.

Another future update may put `km_kbp_status` values into the `g_warning` content (e.g. L452) so that we can see _why_ a given call fails.

And it'd be better to use a _single_ status variable rather than multiple in the same function (`status_state`, `status_keyboard`) -- multiple variables with same purpose lead to them being accidentally used in the wrong place.

Also, if `km_kbp_keyboard_load` fails, then `keyman->keyboard` is invalid and `km_kbp_state_create` certainly should not be called and the function should probably cleanup and `return NULL`. (This is a definite bug.)

Similarly, if `km_kbp_state_create` fails, then we're going to be in a broken state so we should clean up and `return NULL` here too.

It's also a very long function, with heavy lifting such as getting the .ldml file for the keyboard all mixed into the one function. It'd be cleaner to refactor into multiple functions.

Okay, so once I started looking at the whole function I had a lot of ideas. I don't think those changes should be applied to this PR but I think it would be worth considering doing them as code quality updates in a future PR.

ermshiperete:
Created #7717 for additional refactoring task.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.103-beta](https://keyman.com/downloads/releases/beta/16.0.103)

PR #7689, opened 2022-11-10
chore(windows): set platform core env explicitly

### LINKED ISSUE #7687: chore(windows): set `KM_KBP_KMX_ENV_PLATFORM ` explicitly
The `KM_KBP_KMX_ENV_PLATFORM` option is not set by Windows Engine, because it happens to match the default in Core. We should set it, to be explicit and clear.

---

ok, I see. It uses the default which happens to be "windows hardware desktop native". Seems to me it would be cleaner and clearer to explicitly set it...

_Originally posted by @ermshiperete in https://github.com/keymanapp/keyman/pull/7667#discussion_r1019484488_

mcdurdin:
Fixes #7687.

# User Testing

## Preparation

We'll want to build a keyboard to test that the platform is set correctly. The keyboard should have a single rule:

```
platform('windows hardware desktop native') + 'a' > 'Success!'
```

## Test

* **TEST_PLATFORM:** Install and activate the keyboard that you just built above. Press <kbd>a</kbd> and the result should be `Success!`.

### DIFF
--- a/windows/src/engine/keyman32/CoreEnvironment.cpp
+++ b/windows/src/engine/keyman32/CoreEnvironment.cpp
@@ -1,7 +1,9 @@
 #include "pch.h"
 
+#define WINDOWS_PLATFORM_ENV u"windows hardware desktop native"
+
 BOOL SetupCoreEnvironment(km_kbp_option_item **core_environment) {
-  km_kbp_option_item *items = new km_kbp_option_item[5];
+  km_kbp_option_item *items = new km_kbp_option_item[6];
 
   items[0].scope = KM_KBP_OPT_ENVIRONMENT;
   items[0].key = KM_KBP_KMX_ENV_BASELAYOUT;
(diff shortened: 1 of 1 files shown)


bharanidharanj:
* **TEST_PLATFORM (PASSED):** Tested with Keyman Developer 16.0.100-beta (?) build (from Keyman com) in Windows 10 OS and here is my observation: 1. Installed the Keyman 16.0.101-beta-test-7689 build in the system. 2. Created a Tamil Keyboard in Keyman Developer then compiled it. 3. Set the rule as it mentioned in the above testing instructions. 4. In the debugger window, Pressed the a button. Noticed that the there is a message 'success' appeared on the pane.  Did I follow the correct steps or Did I missed out anything? Please, let me know.

ermshiperete (approved):

mcdurdin:
> * Did I follow the correct steps or Did I missed out anything? Please, let me know.

Yeah, one thing missing, sorry! You need to install the keyboard into Keyman for Windows and test it, e.g. in Notepad. Apart from that, perfect :)

@keymanapp-test-bot retest

rc-swag (approved):
lgtm

bharanidharanj:
* **TEST_PLATFORM (FAILED):** Retested this as per Marc's instructions and here is my observation: After installing the custom keyboard in the system, I noticed that after pressing 'a' from the hardware keyboard or pressing Tamil 'ya' letter from the Virtual Keyboard shows 'Success!' on the Notepad. Seems to be an issue. Note: Pressing any other keys on the Virtual or Hardware keyboard outputs the corresponding Tamil letter on the Notepad.

mcdurdin:
@bharanidharanj That looks like what should be happening -- that custom rule you added to the keyboard should be generating the “Success!” message. So that's all good!

TEST_PLATFORM PASS

keyman-server:

PR #7707, opened 2022-11-13
chore(windows): Add safecall for internal CustomisationStorage API interface

mcdurdin:
This makes it easier to track what is happening for errors where otherwise we found that exception messages were being cropped; for example:

* [KEYMAN-WINDOWS-X](https://sentry.io/organizations/keyman/issues/2700324563/?project=5983518&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d)
* [KEYMAN-WINDOWS-BR](https://sentry.io/organizations/keyman/issues/2830726485/?project=5983518&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d)

The error messages were being cropped because the exception handler was in the exe module, but the exception was being raised in kmcomapi module. When the exception was handled, the handler tested to see if the exception object inherited from exemodule.Exception, but it was inheriting from kmcomapi.Exception instead, and so it did not attempt to load the message detail.

By adding `safecall` calling convention to `KeymanCustomisation`, we move responsibility for handling the exception to the `TKeymanAutoObject.SafeCallException` function, and we get full detail for the exception.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/engine/kmcomapi/com/system/keymancontrol.pas
+++ b/windows/src/engine/kmcomapi/com/system/keymancontrol.pas
@@ -168,7 +168,7 @@ type
     procedure Refresh;
 
     { IKeymanCustomisationAccess }
-    function KeymanCustomisation: IKeymanCustomisation;
+    function KeymanCustomisation: IKeymanCustomisation; safecall;
   public
     constructor Create(AContext: TKeymanContext);
     destructor Destroy; override;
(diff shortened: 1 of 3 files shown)


rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.103-beta](https://keyman.com/downloads/releases/beta/16.0.103)

PR #7711, opened 2022-11-14
fix(windows): base layout support in Keyman Core 🍒

mcdurdin:
Cherry-pick of #7667.

While Keyman Core already supported the base layout flags, they were not being passed in from Keyman Engine for Windows. This meant that any rules that used `baselayout` would fail, such as in sil_ipa with AltGr+0 for }

https://github.com/keymanapp/keyboards/blob/0bf72bed74f563f823b1af002f6a8ccef8510ebf/release/sil/sil_ipa/source/sil_ipa.kmn#L348-L350

This patch adds support for the following properties:

* KM_KBP_KMX_ENV_BASELAYOUT: the legacy base layout name, e.g. kbdus.dll
* KM_KBP_KMX_ENV_BASELAYOUTALT: the BCP 47 base layout name, e.g. en-US
* KM_KBP_KMX_ENV_SIMULATEALTGR: whether the user has selected to emulate AltGr with Ctrl+Alt, for example on laptops without a right alt key.
* KM_KBP_KMX_ENV_BASELAYOUTGIVESCTRLRALTFORRALT: whether the system base layout in use generates Ctrl+RAlt when RAlt is pressed, which can be true for many European system layouts. This is necessary when associating the keyboard with a European language because the base keyboard becomes the language's default keyboard, rather than kbdus.

@keymanapp-test-bot skip


darcywong00:
For the PR description, is this a cherry-pick of #7667?

darcywong00 (approved):
lgtm

rc-swag (approved):
lgtm

windows/src/global/inc/keymanengine.h
```diff
@@ -277,5 +278,7 @@ BOOL SelectKeyboardTSF(DWORD KeymanID, BOOL foreground);   // I3933   // I3949
 BOOL ReportKeyboardChanged(WORD wCommand, DWORD dwProfileType, UINT langid, HKL hkl, GUID clsid, GUID guidProfile);
 void ProcessModifierChange(UINT key, BOOL isUp, BOOL isExtended);   // I4793
 
-#endif  // _KEYMAN64_LIGHT
+BOOL SetupCoreEnvironment(km_kbp_option_item **test_env_opts);
+void DeleteCoreEnvironment(km_kbp_option_item *test_env_opts);
+
```
mcdurdin, line 283:
```suggestion

#endif  // _KEYMAN64_LIGHT
```

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.273](https://keyman.com/downloads/releases/stable/15.0.273)

PR #7720, opened 2022-11-15
fix(android/engine): Remove duplicate file suffix when installing .kmp

darcywong00:
While investigating this [Sentry issue](https://sentry.io/organizations/keyman/issues/3738919593/?project=5983520) I saw the user was attempting to install a keyboard package of packageID `obolo_chwerty-1`.

Since packageID is determined by the filename of the .kmp file, it seems to be a case where a .kmp file is downloaded multiple times, and the system appended `-1` to make the filename `obolo_chwerty-1.kmp` (On my Samsung device, it appends ` (1)` so it would be `obolo_chwerty (1).kmp`).

This updates PackageProcessor to strip out the extra numbers in the filename (packageID should never contain spaces or hyphens).

In addition to adding a unit test, I verified with Android Studio that installing "khmer_angkor-1.kmp" and "khmer_angkor (1).kmp" results in the correct packageID "khmer_angkor".

## User Testing
Setup
2. From the Chrome browser, download the [khmer_angkor.kmp](https://downloads.keyman.com/keyboards/khmer_angkor/1.3/khmer_angkor.kmp) file multiple times without overwriting filenames. This should result in Downloads/ containing:
    * khmer_angkor.kmp
    * khmer_angkor (1).kmp or khmer_angkor-1.kmp (depending on the vendor)

* **TEST_KHMER_ANGKOR_1** - Verifies khmer_angkor (1).kmp installs correctly
2. From Keyman Settings --> Install Keyboard or Dictionary --> Install from local file --> Browse to the Downloads/ folder and select "khmer_angkor (1).kmp" or khmer_angkor-1.kmp (depending on how your device handled the multiple files
3. Install the keyboard package
4. When the welcome.htm file is displayed, verify the title is "Welcome to khmer_angkor" 
5. Click OK and finish the keyboard package installation

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
@@ -72,7 +72,7 @@ public class PackageProcessor {
   public String getPackageID(File path) {
     // Sometimes, downloading a kmp multiple times results in a filename with a suffix:
     // " (#).kmp" or "-#.kmp" in the filename, so strip out the number
-    String filename = path.getName().toLowerCase().replaceAll("(\\s+\\(|-)\\d+\\)?", "");
+    String filename = path.getName().toLowerCase().replaceAll("(\\s+\\(|-)\\d+\\)?.kmp$", ".kmp");
 
     if(resourceRoot == null) {
       throw new IllegalStateException("The PackageProcessor has not been initialized!");
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
@@ -70,7 +70,9 @@ public class PackageProcessor {
    * @throws IllegalArgumentException
    */
   public String getPackageID(File path) {
-    String filename = path.getName().toLowerCase();
+    // Sometimes, downloading a kmp multiple times results in a filename with a suffix:
+    // " (#).kmp" or "-#.kmp" in the filename, so strip out the number
+    String filename = path.getName().toLowerCase().replaceAll("(\\s+\\(|-)\\d+\\)?", "");
 
     if(resourceRoot == null) {
       throw new IllegalStateException("The PackageProcessor has not been initialized!");
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
@@ -72,7 +72,8 @@ public class PackageProcessor {
   public String getPackageID(File path) {
     // Sometimes, downloading a kmp multiple times results in a filename with a suffix:
     // " (#).kmp" or "-#.kmp" in the filename, so strip out the number
-    String filename = path.getName().toLowerCase().replaceAll("(\\s+\\(|-)\\d+\\)?.kmp$", ".kmp");
+    String filename = path.getName().toLowerCase().replaceAll(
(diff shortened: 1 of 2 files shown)


mcdurdin (commented):
It'd be good to restrict the change to the end of the string?

bharanidharanj:
* **TEST_KHMER_ANGKOR_1 (PASSED):** Tested with the attached PR build in API 31 / Android 12 emulator and here is my observation: 1. Downloaded the Khmer Angkor.kmp file 3 times in the Download folder.  2. From Keyman app, install Khmer Angkor(1).kmp file from the local folder. 3. Able to see the Welcome to khmer_angkor Screen during the installation.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
```diff
@@ -70,7 +70,9 @@ public PackageProcessor(File resourceRoot) {
    * @throws IllegalArgumentException
    */
   public String getPackageID(File path) {
-    String filename = path.getName().toLowerCase();
+    // Sometimes, downloading a kmp multiple times results in a filename with a suffix:
+    // " (#).kmp" or "-#.kmp" in the filename, so strip out the number
+    String filename = path.getName().toLowerCase().replaceAll("(\\s+\\(|-)\\d+\\)?.kmp$", ".kmp");
```
mcdurdin, line 75: I guess this is fine but technically it'd match `khmer_angkor-10).kmp` as well 😆 Also `.kmp` should really be `\\.kmp`, otherwise it matches `xkmp` etc.

You could use the regex `/((\s+\(\d+\))|(-\d+))\.kmp$/` (omitted Java string escapes for readability here!)

mcdurdin (approved):

android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
darcywong00: ok - I tweaked the regex again (and pulled the spaces out of the groupings)

android/KMEA/app/src/main/java/com/tavultesoft/kmea/packages/PackageProcessor.java
```diff
@@ -72,7 +72,8 @@ public PackageProcessor(File resourceRoot) {
   public String getPackageID(File path) {
     // Sometimes, downloading a kmp multiple times results in a filename with a suffix:
     // " (#).kmp" or "-#.kmp" in the filename, so strip out the number
-    String filename = path.getName().toLowerCase().replaceAll("(\\s+\\(|-)\\d+\\)?.kmp$", ".kmp");
+    String filename = path.getName().toLowerCase().replaceAll(
+      "\\s*((\\(\\d+\\))|(-\\d+))\\.kmp$", ".kmp");
```
mcdurdin, line 76: I don't think it's quite right yet. This will match `khmer_angkor -1.kmp`?

darcywong00: I didn't know if other Android devices might have whitespace before `-1.kmp` so I wanted to strip that out.
"khmer_angkor  1.kmp" would then become "khmer_angkor.kmp"

mcdurdin: OK, but `khmer_angkor 1.kmp` wouldn't match, would it?

darcywong00: True. I think we just update regex to handle "khmer_angkor 1.kmp" if we start seeing those in Sentry.

keyman-server:

PR #7726, opened 2022-11-15
docs(linux): Add example for keyman:// URL format

ermshiperete:
The help text for `km-config` mentioned a `keyman://` URL but we nowhere explained how that URL should look like. This change adds an example keyman:// URL with a bcp47 tag which should make things clearer.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

Can this also get added to the debian man page for km-config?
I forget where that's generated

https://help.keyman.com/products/linux/16.0/reference/km-config

ermshiperete:
> Can this also get added to the debian man page for km-config? I forget where that's generated

I think it gets generated automatically as the last build step on a TC release build.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.105-beta](https://keyman.com/downloads/releases/beta/16.0.105)

PR #7727, opened 2022-11-15
fix(linux): set DEB_VERSION

### LINKED ISSUE #7724: bug(linux): python3-keyman-config package is missing `pkgversion` number
The `pkgversion` variable in `version.py` should contain the package version but doesn't get set during a package build.

ermshiperete:
DEB_VERSION is used to set `pkgversion` in the `override_dh_auto_build` build step. When we moved to the consolidated source package and thus could no longer use the default debhelper build tools we lost the automatic setting of DEB_VERSION, so this change adds the relevant definitions.

Fixes #7724.

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/rules
+++ b/linux/debian/rules
@@ -12,6 +12,10 @@ export KEYMAN_PKG_BUILD=1
 # xenial needs this to be explicit
 export LC_ALL=C.UTF-8
 
+# Definitions from /usr/share/dpkg/pkg-info.mk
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.105-beta](https://keyman.com/downloads/releases/beta/16.0.105)

Issue #7705, opened 2022-11-13
chore: add safety checks on release build triggers


mcdurdin:
alpha build:
* branch must be `master`
* TIER.md must be `alpha`

beta build:
* branch must be `beta`
* TIER.md must be `beta`

stable build:
* branch must be `stable-*`
* TIER.md must be `stable`
* VERSION.md major.minor must match branch suffix

mcdurdin:
# stable check

```bat
rem Specify expected branch / tier
rem First, extract major.minor from VERSION.md
for /f "delims=. tokens=1,2" %%%%a in (VERSION.md) do (set branchversion=%%%%a.%%%%b)
echo stable > temp-tier.txt
echo stable-%%branchversion%% > temp-branch.txt

rem check that we are on the correct branch for this build type
git branch --show-current > temp-git-branch.txt
fc /w temp-git-branch.txt temp-branch.txt
if errorlevel 1 exit /b 98

rem check that we are on the correct tier for this build type
fc /w TIER.md temp-tier.txt
if errorlevel 1 exit /b 99

rem cleanup
del temp-git-branch.txt
del temp-branch.txt
del temp-tier.txt
```

# beta check

```bat
rem Specify expected branch / tier
echo beta > temp-tier.txt
echo beta > temp-branch.txt

rem check that we are on the correct branch for this build type
git branch --show-current > temp-git-branch.txt
fc /w temp-git-branch.txt temp-branch.txt
if errorlevel 1 exit /b 98

rem check that we are on the correct tier for this build type
fc /w TIER.md temp-tier.txt
if errorlevel 1 exit /b 99

rem cleanup
del temp-git-branch.txt
del temp-branch.txt
del temp-tier.txt
```

# alpha check

```bat
rem Specify expected branch / tier
echo alpha > temp-tier.txt
echo master > temp-branch.txt

rem check that we are on the correct branch for this build type
git branch --show-current > temp-git-branch.txt
fc /w temp-git-branch.txt temp-branch.txt
if errorlevel 1 exit /b 98

rem check that we are on the correct tier for this build type
fc /w TIER.md temp-tier.txt
if errorlevel 1 exit /b 99

rem cleanup
del temp-git-branch.txt
del temp-branch.txt
del temp-tier.txt
```

rc-swag:
lgtm

mcdurdin:
Resolved in CI scripts as above. We will probably transfer to in-repo scripts in due course.

PR #7735, opened 2022-11-16
fix(android): bump version_code major calculation

mcdurdin:
See #7736 for same change applied to master/alpha. This PR and https://github.com/keymanapp/keyman/pull/7736 should be merged at the same time.

Accidental publish of beta 17.0.6 during 16.0 beta cycle has forced a version_major bump, which is a permanent change. Relates to issue #7705.

---

Version code is, for 17.1.33: 1810033. This supports up to 9999 builds for a given major.minor version. We only support up to 9 minor versions for a given major version but at present we are not using minor versions at all.

Despite this calculation, the version_code should be treated as an opaque integer, used only for Play Store publishing.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #7742, opened 2022-11-17
feat(linux): Add symbols file

ermshiperete:
Debian/lintian recommends to add a symbols file which allows to check for API changes. The package build on Jenkins now has an additional build step that will compare the API of the newly built package with the committed symbols file and will fail the build if the API changed and we didn't update the symbols file yet. The current symbols file gets stored as build artifact, so it can easily be copied from Jenkins.

During the build of the binary package a similar check is run but won't fail the build.

@keymanapp-test-bot skip


linux/debian/libkmnkbp0-0.symbols
```diff
@@ -0,0 +1,43 @@
+libkmnkbp0.so.0 ibkmnkbp0-0 #MINVER#
+* Build-Depends-Package: libkmnkbp-dev
+
+ kbp_state_get_intermediate_context@Base 15.0
```
mcdurdin, line 4: This seems wrongly named 😢

mcdurdin (approved):

keyman-server:

PR #7754, opened 2022-11-17
fix(windows): tweak elevate icon in Config dialog

mcdurdin:
Fixes #7737.

@keymanapp-test-bot skip

image

Issue #7750, opened 2022-11-17
bug(windows): arrow keys don't work consistently in Options tab


mcdurdin:
### Describe the bug

Arrow keys don't allow you to move through all available options in the Options tab

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for iPhone and iPad
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [x] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

16.0.106-beta

mcdurdin:
Also, rearrange buttons to make them in-list rather than all-over :grin:

PR #7757, opened 2022-11-17
fix(windows): buttons should be in list in Options tab

mcdurdin:
Relates to #7750.

Makes the buttons in the Options tab part of the list and can be accessed via arrow keys.

@keymanapp-test-bot skip

image

Note that the image has the old 'elevation' icons as these are fixed in #7754.


rc-swag (approved):
lgtm

ermshiperete (approved):

keyman-server:

PR #7762, opened 2022-11-18
fix(mac): Enable right-alt/option key mapping

### LINKED ISSUE #7476: bug(mac): Keys on right alt and shift + right alt layers of Khmer Angkor don't output expected characters
**Describe the bug**

While testing this [PR](https://github.com/keymanapp/keyman/pull/7354), I've run into an issue where keys on right alt and shift + right alt do not output the expected characters. None Khmer characters get output instead of the Khmer ones which have been defined and proved to work without any issue on Windows.

**To Reproduce**

1. On a fresh macOS install the build from this PR (https://github.com/keymanapp/keyman/pull/7354)
2. Allow necessary permission for Keyman in System Preferences > Security & Privacy
3. Open Notes app and to see error
4. type right alt + `K` to get "˚" instead of the expected "ឝ"; type right alt + shift + `K` to get "" instead of the expected "᧳".

**Expected behavior**

The Khmer characters assigned to the keys being pressed should be output instead of the none Khmer ones.

**Related issues**

* https://github.com/keymanapp/keyman/issues/5721
* right alt and shift + right alt on OSK also erroneously output the unexpected characters

---

   Device:              MBP
 - OS:                  Big Sur 11.0.1
 - Keyman version:      16.0.70
 - Target Application:  any texture

---

**Keyboard**
 - Keyboard name:     khmer_angkor
 - Keyboard version:  1.3
 - Language name:     Khmer

**Additional context**

https://github.com/keymanapp/keyman/pull/7354#issuecomment-1273075472

sgschantz:
Fixes #7476 

When typing the right option key (Mac equivalent for the Alt key) ensure that it produces the correct character for the active Keyman keyboard.

# User Testing

* **TEST_RIGHT_ALT:** - Verify that the right option key with Keyman for Mac produces the correct character

1. Remove old version of Keyman **after** first removing its permissions from Security & Privacy > Privacy > Accessibility
2. Install the Keyman build for this PR
3. Restart the system
4. Select Keyman from the system input source menu
5. Enable privacy settings as prompted
6. Restart the system (again)
7. Install the Khmer Angkor keyboard 
8. Switch to a text editor like TextEdit or Pages
9. With Keyman running, hold down the right option key and type the K key
10. Verify that the character ឝ is emitted

* **TEST_RIGHT_ALT_SHIFT:** - Verify that the right option key + shift with Keyman for Mac produces the correct character

1. Switch to a text editor like TextEdit or Pages
2. With Keyman running with the Khmer Angkor keyboard, hold down right option and shift keys and type the K key
3. Verify that the character ᧳ is emitted

### DIFF
--- a/mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyConsent.m
+++ b/mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyConsent.m
@@ -78,7 +78,7 @@
   } else {
     // set completionHandler to be called after privacyDialog is dismissed
     _completionHandler = withCompletionHandler;
-   NSLog(@"do not have Accessibility, present Privacy Dialog");
+    NSLog(@"do not have Accessibility, present Privacy Dialog");
     [self showPrivacyDialog];
   }
 }
--- a/mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyConsent.m
+++ b/mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyConsent.m
@@ -62,7 +62,6 @@
  */
 - (void)requestPrivacyAccess:(void (^)(void))withCompletionHandler
 {
-  _completionHandler = withCompletionHandler;
   BOOL hasAccessibility = NO;
   
   // check if we already have accessibility
@@ -74,8 +73,12 @@
   
   if (hasAccessibility) {
     NSLog(@"already have Accessibility: no need to make request.");
+    // call completionHandler immediately -> privacyDialog will not be presented
+    withCompletionHandler();
   } else {
-    NSLog(@"do not have Accessibility, present Privacy Dialog");
+    // set completionHandler to be called after privacyDialog is dismissed
+    _completionHandler = withCompletionHandler;
+   NSLog(@"do not have Accessibility, present Privacy Dialog");
     [self showPrivacyDialog];
   }
 }


mac/Keyman4MacIM/Keyman4MacIM/Privacy/PrivacyConsent.m
```diff
@@ -74,8 +73,12 @@ - (void)requestPrivacyAccess:(void (^)(void))withCompletionHandler
   
   if (hasAccessibility) {
     NSLog(@"already have Accessibility: no need to make request.");
+    // call completionHandler immediately -> privacyDialog will not be presented
+    withCompletionHandler();
   } else {
-    NSLog(@"do not have Accessibility, present Privacy Dialog");
+    // set completionHandler to be called after privacyDialog is dismissed
+    _completionHandler = withCompletionHandler;
+   NSLog(@"do not have Accessibility, present Privacy Dialog");
```
mcdurdin, line 81:
```suggestion
    NSLog(@"do not have Accessibility, present Privacy Dialog");
```

mcdurdin (commented):
I'm confused as to how this fix addresses the Right Alt issue?

mcdurdin:
> I'm confused as to how this fix addresses the Right Alt issue?

@sgschantz still waiting on the updated explanation of the connection here so I can complete review on this :grin:

sgschantz:
In KMInputMethodAppDelegate (unchanged in this commit) -- the method initCompletion is passes to requestPrivacyAccess. That method is a completion handler that does some initialization of Keyman that is better done after the privacy setting changes have been requested first. In the case that the privacy dialog did not need to be presented, this completion handler was not being invoked which meant that the event tap that provides awareness of the alt key being pressed was never created.

For the privacy settings code, there are two cascading completion handlers that are executed after the dialog is presented: one created in our privacy code to request the setting changes from macOS and a second that can be passed in from the caller to perform some work after all the privacy settings code is executed. It is this second one that was being neglected when we did not need to execute the first, internal completion handler.

I hope that makes sense. It is absolutely not obvious that this was related to the alt key issue, and I thought that it must be pre-existing until I found I could only reproduce with the beta. So glad @MakaraSok found this!

bharanidharanj:
* **TEST_RIGHT_ALT (PASSED):** Tested with the attached PR build in macMini 12.6 Monterey OS and here is my observation: 1. After pressing Alt+K key, shows the expected output. 

   


* **TEST_RIGHT_ALT_SHIFT (PASSED):** Pressing Alt+Shift+K key, shows the expected output.

mcdurdin (approved):
LGTM. Now I understand :grin: Thank you for the explanation!

keyman-server:

PR #5230, opened 2021-06-06
feat(android): Automatically install keyboard through Play Store

mcdurdin:
Resolves #5219.

Uses the Google Play Install Referrer API to automatically discover the keyboard to install together with Keyman; this keyboard is passed in a `referrer` parameter to the Play Store page for the Keyman app from the Keyman website.

When the app launches for the first time, the Get Started screen appears and then a moment later, the Install Keyboard process starts for the associated keyboard. Once that completes, the first checkbox in Get Started is automatically checked.

This can be a little tricky to test because it only runs first-time after installing the app from the Play Store (by design). But the test version is not yet in the Play Store. However, you can test on an Android device by:

1. Uninstall the Keyman app
2. Open a link to the play store with an embedded referrer value -- but don't click Install. See below for some example links. The Play Store will record the referrer for later use by the Keyman app.
3. Click the Home button.
4. Install the test .apk associated with this PR, or load it from adb.

Example install links:
* [Install Keyman + GFF Amharic](https://play.google.com/store/apps/details?id=com.tavultesoft.kmapro&referrer=source%3Dkeyman%26package%3Dgff_amharic%26bcp47%3Dam)
* [Install Keyman + Khmer Angkor](https://play.google.com/store/apps/details?id=com.tavultesoft.kmapro&referrer=source%3Dkeyman%26package%3Dkhmer_angkor)
* [Install Keyman + Cameroon QWERTY, Adamawa Fulfulde](https://play.google.com/store/apps/details?id=com.tavultesoft.kmapro&referrer=source%3Dkeyman%26package%3Dsil_cameroon_qwerty%26bcp47%3Dfub-latn)

I have done a first round of cleanup of the `downloadKMP` pathway, eliminating the class variable `data` which was (a) obscure, and (b) only required for a single path in the download process (requesting permissions), replacing it with local variables and parameters as required. However, the pathways are still overly convoluted and there are mixed responsibilities for URL parsing which should ideally be done *only* in the `KMPLink` class. I didn't want to go overboard here, as it was not relevant to this PR, so the changes I have made are really to support the new download pattern that this PR needs. I'd love to see further cleanup of keyboard downloads.

An associated PR for the Keyman website is on its way.

PR #7766, opened 2022-11-18
fix(android/app): Only check Install Referrer API with Play Store installs

darcywong00:
Addresses a significant proportion of the Sentry errors [onInstallReferrerSetupFinished:FEATURE_NOT_SUPPORTED](https://sentry.io/organizations/keyman/issues/2700298447/?project=5983520&query=is%3Aunresolved+FEATURE_NOT_SUPPORTED&referrer=issue-stream&statsPeriod=90d)

#5230 added the feature to automatically install keyboard through Google Play Store. The API is only supported when Keyman is installed directly from the Google Play Store.

This updates CheckInstallReferrer to return before the Install Referrer API is called when Keyman is installed via:
* side loaded install (which includes nearly all local and test builds)
* package manager other than Google Play Store ("com.android.vending")

Unfortunately, we can't really do user-testing until the app is available on the pre-release Play Store...

@keymanapp-test-bot skip

### DIFF
--- a/android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
@@ -88,7 +88,9 @@ public class CheckInstallReferrer {
     if (installerInfo == null) {
       // Side loaded so just return
       return;
-    } else if (!installerInfo.equalsIgnoreCase(GOOGLE_PLAY)) {
+    }
+    
+    if (!installerInfo.equalsIgnoreCase(GOOGLE_PLAY)) {
       KMLog.LogInfo(TAG, "Skipping install referrer. Installed from " + installerInfo);
       return;
     }
--- a/android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
+++ b/android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
@@ -20,13 +20,16 @@ package com.keyman.android;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.os.Build;
 import android.os.RemoteException;
 import android.util.Log;
 
 import com.android.installreferrer.api.InstallReferrerClient;
 import com.android.installreferrer.api.InstallReferrerStateListener;
 import com.android.installreferrer.api.ReferrerDetails;
+import com.tavultesoft.kmapro.BuildConfig;
 import com.tavultesoft.kmapro.MainActivity;
 import com.tavultesoft.kmapro.R;
 import com.tavultesoft.kmea.KmpInstallMode;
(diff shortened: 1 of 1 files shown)


android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
```diff
@@ -59,6 +64,35 @@ public static void checkGooglePlayInstallReferrer(MainActivity mainActivity, Con
...
     editor.commit();
 
+    // local environment or test builds are nearly always side loaded
+    if (BuildConfig.VERSION_ENVIRONMENT.equalsIgnoreCase("local") ||
+        BuildConfig.VERSION_NAME.matches("^.*(-test-\\d+)$")) {
+      return;
+    }
+
+    // Determine if Keyman installed from Google Play store
+    PackageManager packageManager = context.getPackageManager();
```
mcdurdin, line 74: Suggest moving this inside the `try` as we only use it within that scope.

android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
```diff
@@ -59,6 +64,35 @@ public static void checkGooglePlayInstallReferrer(MainActivity mainActivity, Con
...
+      }
+    } catch (PackageManager.NameNotFoundException e) {
+      KMLog.LogException(TAG, "Error determining packageManager", e);
+      return;
+    }
+
+    if (installerInfo == null) {
+      // Side loaded so just return
+      return;
+    } else if (!installerInfo.equalsIgnoreCase(GOOGLE_PLAY)) {
```
mcdurdin, line 91:
```suggestion
    }
    
    if (!installerInfo.equalsIgnoreCase(GOOGLE_PLAY)) {
```

Given we are `return`ing above there's no need for the `else`.

android/KMAPro/kMAPro/src/main/java/com/keyman/android/CheckInstallReferrer.java
```diff
@@ -59,6 +64,35 @@ public static void checkGooglePlayInstallReferrer(MainActivity mainActivity, Con
...
+
+    // Determine if Keyman installed from Google Play store
+    PackageManager packageManager = context.getPackageManager();
+    String installerInfo = null;
+    try {
+      String packageName = context.getPackageName();
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+        installerInfo = packageManager.getInstallSourceInfo(packageName).getInstallingPackageName();
+      } else {
+        installerInfo = packageManager.getInstallerPackageName(packageName);
```
mcdurdin, line 81: Note that in theory `getInstallerPackageName` can throw [`IllegalArgumentException`](https://developer.android.com/reference/java/lang/IllegalArgumentException) if the given package name is not installed ([ref](https://developer.android.com/reference/android/content/pm/PackageManager#getInstallerPackageName(java.lang.String)))

mcdurdin (approved):
I think this LGTM. It will be important to watch for new unexpected exceptions as we are introducing calls to `PackageManager` that we haven't used in the past.

PR #7779, opened 2022-11-21
fix(windows):debug info modifier state

rc-swag:
This branch will not be checked in but is just used to create a build for a user whose machine can reliably reproduce the ALT modifier key being stuck.


rc-swag:
This was only ever for creating a signed debug build. A fix has now been implemented this PR will be closed.

PR #7788, opened 2022-11-22
fix(android/engine): Check if kmp.json is null

darcywong00:
This fixes PackageProcessor.getPackageTarget from throwing exceptions if kmp.json is null.

I'm still tracing the Sentry logs upstream for why the extracted kmp file would have null kmp.json. (maybe IOException if no space left on device?)

Fixes https://sentry.io/organizations/keyman/issues/2706237233/?project=5983520


mcdurdin (approved):

keyman-server:

darcywong00:
Currently not 🍒 -picked to stable-15.0, so Sentry flagged as "regression" in 15.0.273
https://sentry.io/organizations/keyman/issues/2700298447

PR #7792, opened 2022-11-22
fix(linux): Allow to build without patched ibus version

### LINKED ISSUE #7774: bug(linux): ibus-keyman not buildable with any ibus stable release.
### Describe the bug

ibus-keyman integrate ibus pr to early and make it impossible to be built without patching ibus. This is not a good practice and will be troublesome for distro packager to do the upgrade. (not to mention it's still unknown whether the PR to ibus itself will be accepted or not).

Whether it contains the required change should at least be checked at build time instead of blindly apply it.

### Reproduce the bug

_No response_

### Expected behavior

_No response_

### Related issues

_No response_

### Keyman apps

- [ ] Keyman for Android
- [X] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb

### Keyman version

16.0.107

### Operating system

_No response_

### Device

_No response_

### Target application

_No response_

### Browser

_No response_

### Keyboard name

_No response_

### Keyboard version

_No response_

### Language name

_No response_

### Additional context

_No response_

ermshiperete:
This change makes it possible to compile even when the updated ibus version is not installed. Of course ordered output won't work in that case, but at least it will compile and the rest of Keyman will work.

Fixes #7774.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -40,10 +40,17 @@
 #include <keyman/keyboardprocessor.h>
 #include <keyman/keyboardprocessor_consts.h>
 
+#include "config.h"
 #include "keymanutil.h"
 #include "keyman-service.h"
 #include "engine.h"
 #include "keycodes.h"
+
+// Fallback for older ibus versions that don't define IBUS_PREFILTER_MASK
+#ifndef IBUS_HAS_PREFILTER
+#warning Updated ibus version is not installed. Ordered output will not be available.
+#define IBUS_PREFILTER_MASK (1 << 23)
+#endif
 
 #define MAXCONTEXT_ITEMS 128
 #define KEYMAN_BACKSPACE 14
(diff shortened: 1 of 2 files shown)


mcdurdin:
> Of course ordered output won't work in that case

That is, the guaranteed ordered output. Keyman will still 'mostly work', right?

linux/ibus-keyman/src/engine.c
```diff
@@ -40,11 +40,18 @@
...
 
+#include "config.h"
 #include "keymanutil.h"
 #include "keyman-service.h"
 #include "engine.h"
 #include "keycodes.h"
 
+// Fallback for older ibus versions that don't define IBUS_PREFILTER_MASK
+#ifndef IBUS_HAS_PREFILTER
+#warning Updated ibus version is not installed. Ordered output will not be available.
```
mcdurdin, line 51: The warning as it stands is not very helpful to a dev encountering it. Perhaps we should reference a specific PR in the warning?

```suggestion
#warning Compiling against ibus version that does not include prefilter mask patch (https://github.com/ibus/ibus/pull/2440). Output ordering guarantees will be disabled.
```

mcdurdin (approved):

ermshiperete:
> > Of course ordered output won't work in that case
> 
> That is, the guaranteed ordered output. Keyman will still 'mostly work', right?

Correct.

keyman-server:

PR #7804, opened 2022-11-23
refactor(linux): Refactor setting keyboard options ⚙️

### LINKED ISSUE #7717: refactor(linux): Refactor `ibus_keyman_engine_constructor`
@mcdurdin wrote in a [comment](https://github.com/keymanapp/keyman/pull/7686#pullrequestreview-1177762194) in #7686 regarding [`ibus_keyman_engine_constructor`](https://github.com/keymanapp/keyman/blob/b4ea9ce7ef91b39c84e5514e7c1658a369b35b06/linux/ibus-keyman/src/engine.c#L321-L475):

> I notice that Windows sets the keyboard opts [in a separate call](https://github.com/keymanapp/keyman/blob/ee14b17b12dafc2b99149dfc67c93da12e19946a/windows/src/engine/keyman32/keyboardoptions.cpp#L236-L285) to km_kbp_state_options_update which may be worth considering to avoid array arithmetic.
> 
> Another future update may put km_kbp_status values into the g_warning content (e.g. L452) so that we can see why a given call fails.
> 
> And it'd be better to use a single status variable rather than multiple in the same function (status_state, status_keyboard) -- multiple variables with same purpose lead to them being accidentally used in the wrong place.
> 
> Also, if km_kbp_keyboard_load fails, then keyman->keyboard is invalid and km_kbp_state_create certainly should not be called and the function should probably cleanup and return NULL. (This is a definite bug.)
> 
> Similarly, if km_kbp_state_create fails, then we're going to be in a broken state so we should clean up and return NULL here too.
> 
> It's also a very long function, with heavy lifting such as getting the .ldml file for the keyboard all mixed into the one function. It'd be cleaner to refactor into multiple functions.

ermshiperete:
- set keyboard options in a call to `km_kbp_state_options_update` separate from setting the environment options
- output status value if method fails
- if any of the methods fails we now return NULL
- split setting the options in multiple methods

Fixes #7717.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -319,6 +319,114 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
   }
 }
 
+static km_kbp_cp* get_base_layout()
+{
+  return u"en-US";
+
+#if 0  // in the future when mnemonic layouts are to be supported
+  const gchar *lang_env = g_getenv("LANG");
+  gchar *lang;
+  if (lang_env != NULL) {
+    g_message("LANG=%s", lang_env);
+    gchar **splitlang = g_strsplit(lang_env, ".", 2);
+    g_message("before . is %s", splitlang[0]);
+    if (g_strrstr(splitlang[0], "_")) {
+      g_message("splitting %s", splitlang[0]);
+      gchar **taglang = g_strsplit(splitlang[0], "_", 2);
+      g_message("lang of tag is %s", taglang[0]);
+      g_message("country of tag is %s", taglang[1]);
+      lang = g_strjoin("-", taglang[0], taglang[1], NULL);
+      g_strfreev(taglang);
+    }
+    else {
+      lang = g_strdup(splitlang[0]);
+    }
+    g_strfreev(splitlang);
+  }
+  else {
+    lang = strdup("en-US");
+  }
+  g_message("lang is %s", lang);
+  km_kbp_cp *cp = g_utf8_to_utf16(lang, -1, NULL, NULL, NULL);
+  return cp;
+  // g_free(lang);
+#endif
+}
+
+static km_kbp_status
+setup_environment(IBusKeymanEngine *keyman)
+{
+  g_assert(keyman);
+  g_message("%s: setting up environment", __FUNCTION__);
+
+  km_kbp_option_item *keyboard_opts = g_new0(km_kbp_option_item, 4);
+
+  keyboard_opts[0].scope = KM_KBP_OPT_ENVIRONMENT;
+  keyboard_opts[0].key   = KM_KBP_KMX_ENV_PLATFORM;
+  keyboard_opts[0].value = u"linux desktop hardware native";
+
+  keyboard_opts[1].scope = KM_KBP_OPT_ENVIRONMENT;
+  keyboard_opts[1].key   = KM_KBP_KMX_ENV_BASELAYOUT;
+  keyboard_opts[1].value = u"kbdus.dll";
+
+  keyboard_opts[2].scope = KM_KBP_OPT_ENVIRONMENT;
+  keyboard_opts[2].key   = KM_KBP_KMX_ENV_BASELAYOUTALT;
+  keyboard_opts[2].value = get_base_layout();  // TODO: free when mnemonic layouts are to be supported
+
+  keyboard_opts[3].scope = 0;
+  keyboard_opts[3].key   = 0;
+  keyboard_opts[3].value = NULL;
+
+  km_kbp_status status = km_kbp_state_create(keyman->keyboard, keyboard_opts, &(keyman->state));
+  if (status != KM_KBP_STATUS_OK) {
+    g_warning("%s: problem creating km_kbp_state. Status is %u.", __FUNCTION__, status);
+  }
+  g_free(keyboard_opts);
+  return status;
+}
+
+static km_kbp_status
+load_keyboard_options(IBusKeymanEngine *keyman)
+{
+  g_assert(keyman);
+
+  // Retrieve keyboard options from DConf
+  // TODO: May need unique packageID and keyboard ID
+  g_message("%s: Loading options for kb_name: %s", __FUNCTION__, keyman->kb_name);
+  GQueue *queue_options = keyman_get_options_queue_fromdconf(keyman->kb_name, keyman->kb_name);
+  int num_options       = g_queue_get_length(queue_options);
+  if (num_options < 1)
+    return KM_KBP_STATUS_OK;
+
+  // Allocate enough options for: num_options plus 1 pad struct of 0's
+  km_kbp_option_item *keyboard_opts = g_new0(km_kbp_option_item, num_options + 1);
+
+  for (int i = 0; i < num_options; i++) {
+    km_kbp_option_item *item = g_queue_pop_head(queue_options);
+    keyboard_opts[i].scope = item->scope;
+    keyboard_opts[i].key   = item->key;
+    keyboard_opts[i].value = item->value;
+  }
+
+  keyboard_opts[num_options].scope = 0;
+  keyboard_opts[num_options].key   = 0;
+  keyboard_opts[num_options].value = NULL;
+
+  // once we have the option list we can then update the options using the public api call
+  km_kbp_status status = km_kbp_state_options_update(keyman->state, keyboard_opts);
+
+  if (status != KM_KBP_STATUS_OK) {
+    g_warning("%s: problem creating km_kbp_state. Status is %u.", __FUNCTION__, status);
+  }
+  for (int i = 0; i < num_options; i++) {
+    g_free((km_kbp_cp *)keyboard_opts[i].key);
+    g_free((km_kbp_cp *)keyboard_opts[i].value);
+  }
+  g_queue_free_full(queue_options, NULL);
+  g_free(keyboard_opts);
+  return status;
+}
+
 static GObject*
 ibus_keyman_engine_constructor(
   GType type,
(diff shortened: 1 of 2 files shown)


linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,114 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+#endif
+}
+
+static km_kbp_status
+setup_environment(IBusKeymanEngine *keyman)
+{
+  g_assert(keyman);
+  g_message("%s: setting up environment", __FUNCTION__);
+
+  km_kbp_option_item *keyboard_opts = g_new0(km_kbp_option_item, 4);
```
darcywong00, line 362: Should this be called `environment_opts` now?

darcywong00, line 362: Maybe comment 
```
// Allocate enough options for: 3 environments plus 1 pad struct of 0's
```

ermshiperete: done

ermshiperete: done

linux/ibus-keyman/src/engine.c
```diff
@@ -387,84 +496,26 @@ ibus_keyman_engine_constructor(
...
 
-    km_kbp_status status_state = km_kbp_state_create(keyman->keyboard,
-                                  keyboard_opts,
-                                  &(keyman->state));
-    if (status_state != KM_KBP_STATUS_OK)
-    {
-        g_warning("%s: problem creating km_kbp_state", __FUNCTION__);
+    status = setup_environment(keyman);
+    if (status != KM_KBP_STATUS_OK) {
+      g_free(abs_kmx_path);
```
darcywong00, line 511: Is this already freed at l. 502?

darcywong00:
Confirming this isn't needed for 16.0 beta?

linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,115 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+  keyboard_opts[num_options].value = NULL;
+
+  // once we have the option list we can then update the options using the public api call
+  km_kbp_status status = km_kbp_state_options_update(keyman->state, keyboard_opts);
+
+  if (status != KM_KBP_STATUS_OK) {
+    g_warning("%s: problem creating km_kbp_state. Status is %u.", __FUNCTION__, status);
+  }
+  for (int i = 0; i < num_options; i++) {
+    g_free((km_kbp_cp *)keyboard_opts[i].key);
```
darcywong00, line 423: Should `scope` also be freed? (I notice the original code doesn't free it either)

linux/ibus-keyman/src/engine.c
ermshiperete: yep. Thanks!

linux/ibus-keyman/src/engine.c
ermshiperete: no, `scope` is a simple `int`, so we don't have to free it. Whereas `key` and `value` are pointers.

ermshiperete:
> Confirming this isn't needed for 16.0 beta?

No, it's not needed. And since it does quite a bit of shuffling of code I'd rather not put it in beta this late - rather not introduce new unnoticed bugs :smile:

linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,115 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+}
+
+static km_kbp_status
+setup_environment(IBusKeymanEngine *keyman)
+{
+  g_assert(keyman);
+  g_message("%s: setting up environment", __FUNCTION__);
+
+  // Allocate enough options for: 3 environments plus 1 pad struct of 0's
+  km_kbp_option_item *environment_opts = g_new0(km_kbp_option_item, 4);
```
mcdurdin, line 363: Might be better to allocate these on the stack:
```suggestion
  km_kbp_option_item environment_opts[4] = {0};
```

linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,115 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+  environment_opts[1].key   = KM_KBP_KMX_ENV_BASELAYOUT;
+  environment_opts[1].value = u"kbdus.dll";
+
+  environment_opts[2].scope = KM_KBP_OPT_ENVIRONMENT;
+  environment_opts[2].key   = KM_KBP_KMX_ENV_BASELAYOUTALT;
+  environment_opts[2].value = get_base_layout();  // TODO: free when mnemonic layouts are to be supported
+
+  environment_opts[3].scope = 0;
+  environment_opts[3].key   = 0;
+  environment_opts[3].value = NULL;
```
mcdurdin, line 379:
```suggestion
```

Because allocating on stack and zeroing above

linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,115 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+
+  km_kbp_status status = km_kbp_state_create(keyman->keyboard, environment_opts, &(keyman->state));
+  if (status != KM_KBP_STATUS_OK) {
+    g_warning("%s: problem creating km_kbp_state. Status is %u.", __FUNCTION__, status);
+  }
+  g_free(environment_opts);
```
mcdurdin, line 385:
```suggestion
```

Not needed if stack allocated

linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,115 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+{
+  g_assert(keyman);
+
+  // Retrieve keyboard options from DConf
+  // TODO: May need unique packageID and keyboard ID
+  g_message("%s: Loading options for kb_name: %s", __FUNCTION__, keyman->kb_name);
+  GQueue *queue_options = keyman_get_options_queue_fromdconf(keyman->kb_name, keyman->kb_name);
+  int num_options       = g_queue_get_length(queue_options);
+  if (num_options < 1)
+    return KM_KBP_STATUS_OK;
```
mcdurdin, line 400:
```suggestion
  if (num_options < 1) {
    g_queue_free_full(queue_options, NULL);
    return KM_KBP_STATUS_OK;
  }
```

linux/ibus-keyman/src/engine.c
```diff
@@ -319,6 +319,115 @@ ibus_keyman_engine_init(IBusKeymanEngine *keyman) {
...
+  for (int i = 0; i < num_options; i++) {
+    km_kbp_option_item *item = g_queue_pop_head(queue_options);
+    keyboard_opts[i].scope = item->scope;
+    keyboard_opts[i].key   = item->key;
+    keyboard_opts[i].value = item->value;
+  }
+
+  keyboard_opts[num_options].scope = 0;
+  keyboard_opts[num_options].key   = 0;
+  keyboard_opts[num_options].value = NULL;
```
mcdurdin, line 414:
```suggestion
```

`g_new0` already zeroes memory doesn't it?

linux/ibus-keyman/src/engine.c
```diff
@@ -387,84 +496,26 @@ ibus_keyman_engine_constructor(
...
-    km_kbp_status status_keyboard = km_kbp_keyboard_load(abs_kmx_path, &(keyman->keyboard));
+    status = km_kbp_keyboard_load(abs_kmx_path, &(keyman->keyboard));
     g_free(abs_kmx_path);
 
-    if (status_keyboard != KM_KBP_STATUS_OK)
-    {
-        g_warning("%s: problem creating km_kbp_keyboard", __FUNCTION__);
+    if (status != KM_KBP_STATUS_OK) {
+      g_warning("%s: problem creating km_kbp_keyboard. Status is %u.", __FUNCTION__, status);
+      return NULL;
```
mcdurdin, line 506:
```suggestion
      IBUS_OBJECT_CLASS (parent_class)->destroy ((IBusObject *)keyman);
      return NULL;
```

linux/ibus-keyman/src/engine.c
```diff
@@ -387,84 +496,26 @@ ibus_keyman_engine_constructor(
...
-    km_kbp_status status_state = km_kbp_state_create(keyman->keyboard,
-                                  keyboard_opts,
-                                  &(keyman->state));
-    if (status_state != KM_KBP_STATUS_OK)
-    {
-        g_warning("%s: problem creating km_kbp_state", __FUNCTION__);
+    status = setup_environment(keyman);
+    if (status != KM_KBP_STATUS_OK) {
+      g_free(abs_kmx_path);
+      return NULL;
```
mcdurdin, line 512:
```suggestion
      IBUS_OBJECT_CLASS (parent_class)->destroy ((IBusObject *)keyman);
      return NULL;
```

linux/ibus-keyman/src/engine.c
```diff
@@ -387,84 +496,26 @@ ibus_keyman_engine_constructor(
...
     }
-    for (int i = KEYMAN_ENVIRONMENT_OPTIONS; i < KEYMAN_ENVIRONMENT_OPTIONS + num_options + 1; i++) {
-      g_free((km_kbp_cp *)keyboard_opts[i].key);
-      g_free((km_kbp_cp *)keyboard_opts[i].value);
+
+    status = load_keyboard_options(keyman);
+    if (status != KM_KBP_STATUS_OK) {
+      return NULL;
```
mcdurdin, line 517:
```suggestion
      IBUS_OBJECT_CLASS (parent_class)->destroy ((IBusObject *)keyman);
      return NULL;
```

mcdurdin (changes requested):
There's a few leak-on-failure that I think I saw here. Not critical but useful just tidy. However, the repetition smells a bit sorry 😢

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.13-alpha](https://keyman.com/downloads/releases/alpha/17.0.13)

PR #7806, opened 2022-11-23
fix: test builds should always run latest changes

mcdurdin:
While release builds need to build a specific commit hash, test builds should always build the HEAD of the branch. This stops multiple merges and/or pushes generating multiple test builds in the queue.

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/run-required-test-builds.sh
+++ b/resources/build/run-required-test-builds.sh
@@ -60,7 +60,7 @@ function triggerTestBuilds() {
         triggerJenkinsBuild "$job" "$branch" "$force"
       else
         echo "  -- Triggering build configuration $test_build on teamcity"
-        triggerTeamCityBuild "$test_build" "$vcs_test" "$branch"
+        triggerTeamCityBuild true "$test_build" "$vcs_test" "$branch"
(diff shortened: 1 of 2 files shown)


mcdurdin:
I tried triggering this build multiple times and looks like it did exactly what we wanted -- only adding the most recent changes to the queue. Should improve our build queues significantly!

darcywong00 (approved):
lgtm

jahorton (approved):
LGTM?  I'm not exactly versed in TC command-fu, but the script changes and implied intent of the commands looks like it'd do what we want.

Hopefully I'm drawing the right inferences from the parts I'm not familiar with.

mcdurdin:
> I'm not exactly versed in TC command-fu

Neither am I, and it's terribly complicated and very minimally documented...

mcdurdin:
Note iOS and macOS builds failed due to unrelated agent/CI system failure.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.112-beta](https://keyman.com/downloads/releases/beta/16.0.112)

PR #7781, opened 2022-11-21
fix(android/engine): Make it easier to test keyboard updates

darcywong00:
Currently there's a lot of hoops involved for testing Keyman keyboard package updates (see https://github.com/keymanapp/keyman/pull/7740#issuecomment-1319640048)

This PR adds overrides for local and PR test builds to ignore calendar comparisons for checking keyboard updates. (Note: this won't apply to release builds used in Acceptance Testing)

I also separated the updateKMP comparison so that
* keyboard info is updated if the cloud has a newer version
* if updateKMP has a value, trigger notification for update available (this was causing updates to only appear on the 2nd boot)

Additional changes:
* Refactors version environment checks to new `VersionUtils()`

Commit https://github.com/keymanapp/keyman/commit/2d0008e2e38e1ffc0481af5f84a5f88bd8a1e3b7 is temporary test code for installing an old version 6.0.4 of sil_cameroon_qwerty.
- [x] DO NOT MERGE until reverting test code

## User Testing
Setup - Use Android device/emulator (phone in portrait orientation) with network access enabled

* **TEST_KEYBOARD_UPATES** 
1. Install the PR build of Keyman for Android
2. Launch Keyman for Android a first time.  This installs the old version 6.0.4 of sil_cameroon_qwerty
3. Completely close Keyman
4. Launch Keyman for Android a second time. This performs a check for keyboard updates
5. Dismiss the "Get Started" menu
6. Click the Keyman overflow menu
7. Verify icon for "Install Updates" is visible

PR #7807, opened 2022-11-24
fix(android/engine): Append languageID for keyboard updates

### LINKED ISSUE #7801: bug(android/engine): Downloading keyboard updates in the background don't get preferred language
Reported during testing of #7740 for 16.0 beta

When installing a keyboard package update (uses the background download manager), the installation process doesn't set the preferred language. So in the sil_cameroon_qwerty example, it ends up creating a new entry for the first language "Afade" in addition to the existing "Akum" language.

We should propagate the preferred language when installing keyboard updates.

darcywong00:
Fixes #7801 and follow-on to #7781

This PR appends the language ID to the kmp links returned from api.keyman.com keyboard package update queries. This way, the installer applies the correct language tag in the package.

Note, if a keyboard package changes languageIDs (e.g. sil_cameroon_qwerty between version 6.0.4 and 6.0.8 changes from aku-Latn to aku), the "old" languageID isn't found in the new kmp.json so the first listed language will get installed.

## User Testing
**Setup** - Use any Android emulator/device
2. On the device, download the old version 1.0 of [sil_nko](https://downloads.keyman.com/keyboards/sil_nko/1.0/sil_nko.kmp) keyboard.

* **TEST_UPDATE_ASSOCIATES_LANGUAGE** - Verifies keyboard package update uses existing language (if found)
1. Launch Keyman for Android and dismiss "Get Started"
2. From Keyman settings --> Install the downloaded local keyboard package: sil_nko (version 1.0)
3. During the sil_nko keyboard package installation, install for the language "Mandingo" (third languageID in kmp.json, though the language list is sorted alphabetically)
4. After sil_nko is installed, long-press on the globe key to display the Keyboard picker menu
5. On the Keyboard picker menu, press the "i" icon to display keyboard info of sil_nko
6. Verify the sil_nko version is 1.0
7. Completely exit Keyman
8. Relaunch Keyman and wait for notifications of keyboard package updates.
9. If no keyboard updates appear, exit Keyman and re-launch Keyman
10. From Keyman Settings, click "Install Updates" and accept the download for sil_nko
11. Wait for sil_nko to update in the background. This may take a few minutes depending on the network
12. After the keyboard is updated, a Toast notification will give a status about resources being updated.
13. Long-press on the globe key to display the Keyboard Picker menu
14. Verify the only listed languages are "English" and "Mandingo"
15. Click the "i" icon to display keyboard info of sil_nko
16. Verify the sil_nko version is now 1.1


bharanidharanj:
* **TEST_UPDATE_ASSOCIATES_LANGUAGE (FAILED):** Tested with the attached PR build in API 31 / Android 12.0 emulator and here is my observation: 1. Installed the PR build in the emulator. 2. Installed the sil_nko.kmp file from local folder. 3. Completely closed Keyman app. Relaunched once again. 4. Noticed that no 'Install Updates' option available in the Keyman menu list. 5. Opened Installed languages menu. Opened sil_nko language. Noticed that there is an updated option available against the language name. 6. Clicking the on the Download option leads to crash the Keyman app. 

..sil_nko update

   


..Keyman Error message

darcywong00:
> Opened Installed languages menu. Opened sil_nko language. Noticed that there is an updated option available against the language name. 6. Clicking the on the Download option leads to crash the Keyman app

I had to merge beta in to get the updates from #7781. When I clicked Installed Languages --> update option available --> Clicking on Download ended up downloading the keyboard package for me (without crashing)

@keymanapp-test-bot retest TEST_UPDATE_ASSOCIATES_LANGUAGE

bharanidharanj:
* **TEST_UPDATE_ASSOCIATES_LANGUAGE (PASSED):** Retested this PR with my Android Mobile device (Redmi Note 8 Pro / Ver 11.0) as per the instructions and it is working as expected. ie., I was able to upgrade version of the sil_nko keyboard. 

..before update






..after update

mcdurdin:
> Note, if a keyboard package changes languageIDs (e.g. sil_cameroon_qwerty between version 6.0.4 and 6.0.8 changes from aku-Latn to aku), the "old" languageID isn't found in the new kmp.json so the first listed language will get installed.

Obviously this is not ideal, but it'll be much less widespread, so I am somewhat more comfortable with delaying a fix to this specific part of the problem.

However, given the original language code still stays installed, can we just avoid the problem altogether by never updating language associations during an update?

darcywong00:
> However, given the original language code still stays installed, can we just avoid the problem altogether by never updating language associations during an update?

Currently, a keyboard update is treated the same as a new kmp installation (the workflow is the same after the kmp files are downloaded). We can prefer a language, and if it's not found, the default is the first language.

Are you asking during a keyboard update, if the preferred language isn't found, to not add the default language? Would have to investigate how to implement that...

mcdurdin:
> Currently, a keyboard update is treated the same as a new kmp installation (the workflow is the same after the kmp files are downloaded). We can prefer a language, and if it's not found, the default is the first language.
> 
> Are you asking during a keyboard update, if the preferred language isn't found, to not add the default language? Would have to investigate how to implement that...

In my way of thinking, a keyboard update should be preserving the existing keyboard language settings, not be treated the same as a fresh install.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDataJsonUtil.java
```diff
@@ -158,9 +158,10 @@ public static void processKeyboardPackageUpdateJSON(Context aContext, JSONObject
                 // Cloud catalog has newer KMP version available
                 String updateKMP = kbd.getUpdateKMP();
                 if (updateKMP != null) {
-                  if (!updateKMP.equalsIgnoreCase(cloudKMP)) {
-                    // Update keyboard info with the latest KMP link
-                    kbd.setUpdateKMP(cloudKMP);
+                  if (!updateKMP.startsWith(cloudKMP)) {
+                    // Update keyboard info with the latest KMP link after appending languageID
+                    String languageID = kbd.getLanguageID();
+                    kbd.setUpdateKMP(String.format("%s&bcp47=%s", cloudKMP, languageID));
```
mcdurdin, line 164: Technically, we should be encoding these urls. However, languageID is not likely to cause trouble as it uses only safechars, and cloudKMP also should be using only safechars. So I'm almost okay with it (except for defense-in-depth and pure principles!)

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDataJsonUtil.java
```diff
@@ -158,9 +158,10 @@ public static void processKeyboardPackageUpdateJSON(Context aContext, JSONObject
                 // Cloud catalog has newer KMP version available
                 String updateKMP = kbd.getUpdateKMP();
                 if (updateKMP != null) {
-                  if (!updateKMP.equalsIgnoreCase(cloudKMP)) {
-                    // Update keyboard info with the latest KMP link
-                    kbd.setUpdateKMP(cloudKMP);
+                  if (!updateKMP.startsWith(cloudKMP)) {
```
mcdurdin, line 161: I am not entirely sure what the content of `updateKMP` and `cloudKMP` should be. But I can see two potential issues:

1. `startsWith` is case sensitive, so that's different to previously.
2. If I am understanding this correctly, this will now match `updateKMP == "my_keyboard_ex"` for `cloudKMP == "my_keyboard"`.

mcdurdin (changes requested):
NQLGTM (Not Quite Look Good To Me); am slightly concerned by the change in line 161. Also not entirely sure why that change is necessary?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDataJsonUtil.java
darcywong00: > I am not entirely sure what the content of updateKMP and cloudKMP should be

They contain the "kmp" property returned from the api.keyman.com [package-version](https://api.keyman.com/package-version?platform=android&keyboard=sil_cameroon_qwerty) query. updateKMP is empty if there's no updates to make.

I changed the comparison to `startsWith()` to handle the scenario where previous checks had stored the appended languageID:
updateKMP = `https://keyman.com/go/package/download/keyboard/sil_cameroon_qwerty?version=6.0.8&update=1&bcp47=aku` and
cloudKMP = `https://keyman.com/go/package/download/keyboard/sil_cameroon_qwerty?version=6.0.8&update=1`

mcdurdin: I'm not very comfortable with making these assumptions on the structure of the URL. I think we should use a URL parser and do a more precise match rather than assume that there isn't already a bcp47 code in it, etc.

darcywong00: I've refactored the URI validation to CloudDataJSONUtils where it checks if the cloud KMP link is a newer version.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDataJsonUtil.java
```diff
@@ -350,4 +352,38 @@ public static CloudApiTypes.CloudApiReturns retrieveJsonFromDownload(
...
+
+      if (updateKMP == null || updateKMP.isEmpty()) {
+        return true;
+      }
+
+      Uri localLink = Uri.parse(updateKMP);
+
+      boolean pathsMatch = localLink.getLastPathSegment().equalsIgnoreCase(cloudLink.getLastPathSegment());
+      boolean cloudVersionNewer = FileUtils.compareVersions(
+        localLink.getQueryParameter(CDKey_Version), cloudLink.getQueryParameter(CDKey_Version)) == FileUtils.VERSION_LOWER;
```
mcdurdin, line 381: 👏🏼👏🏼👏🏼

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDataJsonUtil.java
```diff
@@ -350,4 +352,38 @@ public static CloudApiTypes.CloudApiReturns retrieveJsonFromDownload(
...
+    try {
+      Uri cloudLink = Uri.parse(cloudKMP);
+
+      if (updateKMP == null || updateKMP.isEmpty()) {
+        return true;
+      }
+
+      Uri localLink = Uri.parse(updateKMP);
+
+      boolean pathsMatch = localLink.getLastPathSegment().equalsIgnoreCase(cloudLink.getLastPathSegment());
```
mcdurdin, line 379: If we rename a keyboard using the Cloud deprecation structures, is that handled here? I can't remember...

mcdurdin (approved):
Just one question on deprecated keyboards (e.g. vai1.1 is deprecated by athinkra_vai)?

android/KMEA/app/src/main/java/com/tavultesoft/kmea/cloud/CloudDataJsonUtil.java
darcywong00: Possibly a source of sadness?  The vai1.1 query returns
```
{
    "keyboards": {
        "vai1.1": {
            "version": "1.1",
            "error": "not available for platform",
            "kmp": "https://keyman.com/go/package/download/keyboard/vai1.1?version=1.1&update=1",
            "deprecatedBy": "athinkra_vai"
        }
    }
}
```

Currently, since the return has "error" field, we'll skip processing the update.

I'm not sure what happens if we were able to update vai1.1

mcdurdin: Probably not a good choice then -- given it was not an Android-capable one. Which one could we use instead? How about tamil99m?

darcywong00:
```json
{
    "keyboards": {
        "tamil99m": {
            "version": "1.1",
            "error": "not available as package",
            "deprecatedBy": "basic_kbdtam99"
        }
    }
}
```

and european2 returns
```json
{
    "keyboards": {
        "european2": {
            "version": "1.6",
            "error": "not available as package",
            "deprecatedBy": "sil_euro_latin"
        }
    }
}
```

darcywong00: https://api.keyman.com/package-version?platform=android&keyboard=basic_kbdughr
returns
```json
{
    "keyboards": {
        "basic_kbdughr": {
            "version": "1.0",
            "kmp": "https://keyman.com/go/package/download/keyboard/basic_kbdughr?version=1.0&update=1",
            "deprecatedBy": "basic_kbdughr1"
        }
    }
}
```

darcywong00: I'll go ahead and merge this. If we need to handle package updates for deprecated keyboards, I'll address in a separate issue.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.116-beta](https://keyman.com/downloads/releases/beta/16.0.116)

PR #7833, opened 2022-11-28
chore(linux): Fix upload-to-debian script

ermshiperete:
While we're in beta phase we need to modify `changelog` in beta branch instead of `master`.

@keymanapp-test-bot skip

### DIFF
--- a/linux/scripts/upload-to-debian.sh
+++ b/linux/scripts/upload-to-debian.sh
@@ -81,6 +81,12 @@ stable_branch=${stable_branch##* }
 git checkout "${stable_branch#origin/}"
 git pull origin "${stable_branch#origin/}"
 
+if git branch -r | grep origin/beta; then
+    ISBETA=true
+else
+    ISBETA=false
+fi
+
 cd "$KEYMAN_ROOT/linux"
 echo_heading "Building source package"
 DIST=unstable DEBREVISION=$REVISION scripts/debian.sh
@@ -101,7 +107,7 @@ if [ -n "$PUSH" ]; then
     $NOOP git push origin chore/linux/changelog
 fi
 
-if git branch -r | grep origin/beta; then
+if $ISBETA; then
     CLBRANCH=origin/beta
 else
     CLBRANCH=origin/master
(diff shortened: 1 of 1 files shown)


linux/scripts/upload-to-debian.sh
```diff
@@ -101,11 +101,21 @@ if [ -n "$PUSH" ]; then
     $NOOP git push origin chore/linux/changelog
 fi
 
-git checkout -B chore/linux/cherry-pick/changelog origin/master
+if git branch -r | grep origin/beta; then
```
darcywong00, line 104: Since l. 34 sources `build-utils.sh`, does this script have access to `$TIER`?

mcdurdin: Yep, just wondering what branch this script runs on? Is it running on a variety of branches? If so, then yep, we might need to check for existence of origin/beta branch as @ermshiperete has done here.

linux/scripts/upload-to-debian.sh
```diff
@@ -101,11 +101,21 @@ if [ -n "$PUSH" ]; then
...
+
+git checkout -B chore/linux/cherry-pick/changelog ${CLBRANCH}
 git cherry-pick -x chore/linux/changelog
 if [ -n "$PUSH" ]; then
     $NOOP git push origin chore/linux/cherry-pick/changelog
 fi
 
 echo_heading "Finishing"
-git checkout master
+if git branch -r | grep origin/beta; then
```
mcdurdin, line 117: I'd prefer to store this as a flag earlier in the script for consistency and reuse it here.

linux/scripts/upload-to-debian.sh
ermshiperete: Yes, the intention is that it doesn't matter what branch you're on, it will always upload the latest stable release to Debian.

linux/scripts/upload-to-debian.sh
ermshiperete: Done

mcdurdin (approved):

PR #7836, opened 2022-11-28
fix(windows): lower case filenames for projects

### LINKED ISSUE #7828: chore(developer): debug symbols are not being uploaded
Having a look at downloads.keyman.com/windows/symbols/, it appears that tike.exe has not been uploaded for months. tike.dbg is available all the way through.

* 2022-06-03 00:52 is the latest timestamp for https://downloads.keyman.com/windows/symbols/tike.exe/629956C1e82000/

* 2022-11-26 18:41 is the latest timestamp for tike.dbg: (https://downloads.keyman.com/windows/symbols/tike.dbg/6382594Dfc6000/

mcdurdin:
Fixes #7828.

Moves all projects to lower case filenames so that we don't end up with filename conflicts in symbol servers.

@keymanapp-test-bot skip


developer/src/tike/tike.dproj
```diff
@@ -47,9 +47,9 @@
...
-        <Icon_MainIcon>Tike_Icon.ico</Icon_MainIcon>
+        <Icon_MainIcon>tike_Icon.ico</Icon_MainIcon>
```
darcywong00, line 52: Could this be tike_icon.ico? Or it doesn't matter?

mcdurdin: Yeah, this one doesn't matter -- I meant to only change the tike.dpr ones but did this one accidentally. Won't hurt anything. :grin:

rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.115-beta](https://keyman.com/downloads/releases/beta/16.0.115)

PR #7844, opened 2022-11-29
fix(android/engine): Use FLAG_IMMUTABLE for PendingIntent for Android S+

darcywong00:
Fixes #7843 
Starting with Android S+ (API 31), we need to use `FLAG_IMMUTABLE` for the PendingIntent used to download package updates.

From the Sentry crash report
```
IllegalArgumentException: com.tavultesoft.kmapro.debug: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
    at android.app.PendingIntent.checkFlags(PendingIntent.java:375)
    at android.app.PendingIntent.getActivitiesAsUser(PendingIntent.java:593)
    at android.app.PendingIntent.getActivities(PendingIntent.java:575)
    at android.app.TaskStackBuilder.getPendingIntent(TaskStackBuilder.java:275)
    at android.app.TaskStackBuilder.getPendingIntent(TaskStackBuilder.java:249)
...
(19 additional frame(s) were not displayed)

RuntimeException: Error receiving broadcast Intent { act=android.intent.action.DOWNLOAD_COMPLETE flg=0x10 pkg=com.tavultesoft.kmapro.debug (has extras) } in com.tavultesoft.kmea.cloud.CloudDownloadMgr$1@562b474
    at android.app.LoadedApk$ReceiverDispatcher$Args.lambda$getRunnable$0$LoadedApk$ReceiverDispatcher$Args(LoadedApk.java:1689)
    at android.app.LoadedApk$ReceiverDispatcher$Args$$ExternalSyntheticLambda0.run
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:201)
...
(5 additional frame(s) were not displayed)
```

## User Testing - Based on tests for #7832
Setup - For each group, use the Android device, orientation, and Android OS as specified in the group.
1. On the device, download the old version [1.0 of sil_nko Keyman keyboard](https://downloads.keyman.com/keyboards/sil_nko/1.0/sil_nko.kmp)
2. Install the PR build of Keyman for Android

[...]

PR #7814, opened 2022-11-25
chore(android): Update Target SDK Version to 31 🍒

darcywong00:
🍒 pick of #7761 to stable-15.0

Updates Gradle files to target SDK version 31 to comply with Google Play Store

## User Testing
Setup - Install PR build of Keyman for Android on an Android emulator/device of Android 12 (SDK version 31)

* **TEST_KEYMAN_FUNCTIONALITY** - Verifies Keyman can download keyboards and be used as system keyboard
1. Launch Keyman
2. On the "Get Started" menu, add a keyboard for your language from keyman.com
3. Search for and install the "khmer_angkor" keyboard package
4. Verify the khmer_angkor keyboard successfully installs (along with the dictionary)
5. Go to Keyman Settings --> "Get Started" --> enable Keyman as system-wide keyboard and set Keyman as default keyboard
6. Dismiss the "Get Started" menu
7. Launch the Chrome browser app and click on the search box
8. Set the Keyboard to Keyman
9. Verify the Keyman OSK appears and that you can type with it.

PR #7846, opened 2022-11-29
fix(android/engine): Use FLAG_IMMUTABLE for PendingIntent for Android S+ 🍒

darcywong00:
🍒 pick of #7844 to stable-15.0 
It may be another week till 16.0 gets released, so this prevents keyboard updates from crashing on Android S+ devices.

## User Testing - Modified from #7844 since 15.0 doesn't contain updates to make it easier to test keyboard updates
Setup - For each group, use the Android device, orientation, and Android OS as specified in the group. Do **not** install Keyman for Android until instructed in Setup step
1. On the device, download the old version [1.0 of sil_nko Keyman keyboard](https://downloads.keyman.com/keyboards/sil_nko/1.0/sil_nko.kmp)
2. On the device, turn on airplane mode (sets device offline)
3. On the device Settings, manually set the date/time to 1 year  past (~Oct 2021)
4. Install the PR build of Keyman for Android
5. Launch Keyman for Android
6. From Keyman Settings, install the old version of sil_nko from local file. (Install for "Mandingo" language)
7. Completely exit Keyman
8. On the device Settings, set the date/time to automatically update from the network
9. On the device, turn off airplane mode (sets device back online)

**GROUP_PHONE_PORTRAIT_API_21** - Verifies update installs for phone in portrait orientation on Android 5.0 (API 21)
**GROUP_PHONE_PORTRAIT_API_31** - Verifies update installs for phone in portrait orientation on Android 12.0 (API 31)

**TEST_KEYBOARD_UPDATE**
2. After a while, see if notifications appear about keyboard updates being available. If they don't appear, completely exit Keyman and relaunch.
3. When there's notifications about Keyman updates being available, click the Keyman overflow menu to select "Install Updates"
4. Verify Keyman prompts confirmation to download a keyboard update for sil_nko
5. Accept the dialog
6. Verify after a while, the keyboard updates in the background.
7. After a while, the keyboard updates. Note, 15.0 doesn't contain some keyboard package updates so it likely will install for the first listed language in sil_nko

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/logic/ResourcesUpdateTool.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/logic/ResourcesUpdateTool.java
@@ -421,7 +421,7 @@ public class ResourcesUpdateTool implements KeyboardEventHandler.OnKeyboardDownl
     stackBuilder.addNextIntentWithParentStack(intent);
     // Get the PendingIntent containing the entire back stack
     PendingIntent startUpdateIntent =
-      stackBuilder.getPendingIntent(notification_id, PendingIntent.FLAG_UPDATE_CURRENT);
+      stackBuilder.getPendingIntent(notification_id, PendingIntent.FLAG_IMMUTABLE | PendingIntent.FLAG_UPDATE_CURRENT);
 
     Builder builder = new Builder(currentContext, getClass().getName())
       .setSmallIcon(R.drawable.ic_keyboard)


mcdurdin (approved):
LGTM if needed

darcywong00:
> This would only be needed if we are targeting Android S+ right? Is that API level 31?

Right, which we recently had to do in #7814

darcywong00:
> This would only be needed if we are targeting Android S+ right? Is that API level 31?

Hmm, this even causes crashes on my Android 11.0 device (API 30) trying to install keyboard from the QR code link.
Should be resolved when the next 16.0 beta build goes out.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 15.0.274](https://keyman.com/downloads/releases/stable/15.0.274)

PR #7875, opened 2022-12-07
chore(linux): Improve package build

ermshiperete:
During package builds on Debian servers the build often failed with an error related to X11 when running the tests. This change skips the tests in `linux/ibus-keyman/src/test` when building packages, similar to the way we already do it for the tests in `linux/ibus-keyman/tests`.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.120-beta](https://keyman.com/downloads/releases/beta/16.0.120)

PR #7882, opened 2022-12-08
chore(core): Update meson version

ermshiperete:
We no longer support Ubuntu 18.04 Bionic, so we can update to a newer meson version. The oldest supported platform is now Ubuntu 20.04 Focal which comes with Meson 0.53, so we make that the minimal required version.

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM. That will also remove an annoying warning from the meson build :)

ermshiperete:
This needs updated build agents on both TC (currently running on ba-bionic64ta) and Jenkins (currently running on autopackager-4) before the build will succeed.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.42-alpha](https://keyman.com/downloads/releases/alpha/17.0.42)

PR #7887, opened 2022-12-09
fix(developer): invalid charmap cell selection when updating to empty search result

### LINKED ISSUE #7879: bug(developer): Typing a search in the character map produced an error and developer exited
### Describe the bug

Details for error report for Keyman app:
  Application Identifier: tike-15.0.274
  Error Identifier: 4164B7C919F540AFD4394FC8419060E2

### Reproduce the bug

_No response_

### Expected behavior

_No response_

### Related issues

_No response_

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [X] Keyman Developer
- [ ] KeymanWeb

### Keyman version

15.0.274

### Operating system

Windows 11 22H2

### Device

Windows laptop

### Target application

_No response_

### Browser

_No response_

### Keyboard name

_No response_

### Keyboard version

_No response_

### Language name

_No response_

### Additional context

_No response_

mcdurdin:
Fixes #7879.

If we enter a search string that results in zero results, the grid resize to one row attempts to select row zero. However, we prevent this. This leaves the current row outside the range of valid rows, which means the grid is in an invalid state.

This later causes a crash if the user attempts to mousewheel up. There may be other similar crashes in this situation, but I haven't found any yet.

Fix is to allow selection of row zero when there is only one row. I have not found any problems caused by allowing this.

@keymanapp-test-bot skip

### DIFF
--- a/common/windows/delphi/charmap/UfrmCharacterMapNew.pas
+++ b/common/windows/delphi/charmap/UfrmCharacterMapNew.pas
@@ -904,7 +904,7 @@ end;
 
 procedure TfrmCharacterMapNew.gridSelectCell(Sender: TObject; ACol, ARow: Integer; var CanSelect: Boolean);
 begin
-  CanSelect := ARow > 0;
+  CanSelect := (ARow > 0) or (grid.RowCount = 1);
 end;
 
 function TfrmCharacterMapNew.GetDragObject: TCharacterDragObject;


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.122-beta](https://keyman.com/downloads/releases/beta/16.0.122)

PR #7884, opened 2022-12-09
chore: merge master into feature-ldml 🙀

srl295:
for #7042

@keymanapp-test-bot skip

PR #7888, opened 2022-12-10
fix(developer): kmc-model esbuild node_modules path 🙀

mcdurdin:
The node_modules path is relative to the build output, not package.json. This was causing the web tests to fail due to dependence on the lexical-model-compiler.cjs.

@keymanapp-test-bot skip


mcdurdin:
@srl295 this should fix the web test with #7884.

srl295 (approved):

Issue #7896, opened 2022-12-13
chore(android/samples): Remove Sentry dependency


darcywong00:
While updating Android dependencies for the start of our 17.0 development cycle, I was wondering if we can remove the Sentry gradle dependencies for KMSample1 and KMSample2 (and test apps).

KMManager may need some logic to determine if the Sentry library is included (similar to how `QRCodeUtil.libraryExists` checks for the library).

This should be doable in A17S2

Issue #2293, opened 2019-11-05
feat(android) Update to Java 11


darcywong00:
in #2279 @safolit had started the process of updating the Android build to use Java 11.

Though we're staying with Java 8 for now, I'm documenting the steps for when this happens.

Note, in order to run robolectric for Android SDK 29, we'll need to update Java
(see robolectric.properties)

Java SE Development Kit 11 (Java 9 or newer)
## Install Java
It is recommended to use openJDK because of oracle license issues.
Tested with latest release for openJDK 11 from
https://github.com/ojdkbuild/ojdkbuild

1. Download and unpack the zip archive
2. On Windows: use the default java path C:\Program Files\Java to avoid error message 
> "Error 0x80010135 Path Too Long"
3. Aso set an environment variable for JAVA_HOME e.g C:\Program Files\Java\openjdk-11.0.5.10-1

darcywong00:
Per https://android-developers.googleblog.com/2020/12/announcing-android-gradle-plugin.html

Java 11 is a pre-req for using Gradle plugin 7.0 (#4603)

darcywong00:
We don't need Java 11 for Keyman 15.0. Deferring to Future

rc-swag:
Or the appropriate version of Java at the time.

darcywong00:
I've done preliminary [work](https://github.com/keymanapp/keyman/tree/chore/android/java-11) to update the Android projects to build with MS [OpenJDK 11](https://learn.microsoft.com/en-us/java/openjdk/download#openjdk-11).

One dilemma with CI is the builds depend on the environment variable `JAVA_HOME` to specify the Java version which is
currently set to Java 8 for stable-16.0

darcywong00:
Discussing w/ @mcdurdin , instead of sdkman we can set the environment vars
JAVA_HOME_8
JAVA_HOME_11
and have the builder script export JAVA_HOME accordingly...

> But a more consolidated look at dependences will probably be needed (whole sprint effort from the team I think)

PR #7897, opened 2022-12-13
chore(android): Update Gradle dependencies; targetSDKVersion to 33

darcywong00:
This PR is updating our Android Gradle dependencies for the start of 17.0 development.

Since we're remaining on Java 8, this limits what version of Android Gradle Plugin (AGP) we can upgrade to, which then limits some of the other dependencies.

The targetSDKVersion is also bumped to 33, which necessitates adding the `POST_NOTIFICATIONS` permission to the Android Manifest.

Updating junit impacted `FileUtilsTest.java`, where the expected log now contains 3 items instead of 4.

We'll also need to see if the nightly build successfully publishes (after PR merge) to verify the triplet.play plugin.

Some follow-on work could be done in #7896 to remove Sentry dependencies in the Sample/test apps.

## User Testing
**Setup** - Install the PR build on an Android emulator/device (minimum SDK version of 21)

* **TEST_KEYMAN_APP** - Verifies the Keyman app runs
1. Launch Keyman for Android
3. From Keyman settings, search for the "khmer_angkor" keyboard and install the keyboard package
4. Verify Keyman installs the khmer_angkor keyboard
5. Verify the touch keyboard displays khmer_angkor keyboard.
6. Long-press on the globe-key to display the "Keyboard Picker" menu
7. Click the "i" info icon for khmer_angkor
8. Verify the keyboard info for khmer_angkor is displayed
9. Verify a QR code for sharing the khmer_angkor keyboard is displayed

### DIFF
--- a/android/KMAPro/build.gradle
+++ b/android/KMAPro/build.gradle
@@ -6,9 +6,10 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.2.1'
-        classpath 'io.sentry:sentry-android-gradle-plugin:2.1.0'
-        classpath 'name.remal:gradle-plugins:1.3.1'
+        classpath 'com.android.tools.build:gradle:4.2.2'
+        // sentry-android-gradle-plugin 2.1.5+ requires AGP 7.0
+        classpath 'io.sentry:sentry-android-gradle-plugin:2.1.2'
+        classpath 'name.remal:gradle-plugins:1.5.0'
 
         // From jcenter() which could be sunset in future
         // https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/
--- a/android/KMAPro/gradle/wrapper/gradle-wrapper.properties
+++ b/android/KMAPro/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8-bin.zip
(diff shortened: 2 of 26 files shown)


bharanidharanj:
# Test Results

* **TEST_KEYMAN_APP (PASSED):** Tested with the attached PR build (Keyman 16.0.121-beta) in Android 5.0 / API 21 emulator and here is my observation: 1. Verified the Khmer touch keyboard appears on the screen after installing the Khmer Angkor keyboard. 2. Able to see the added Khmer Keyboard in the Keyman list menu. 3. Clicking on the 'i' info icon opens the the QR code page.

mcdurdin:
> Since we're remaining on Java 8, this limits what version of Android Gradle Plugin (AGP) we can upgrade to, which then limits some of the other dependencies.

Shall we upgrade Java or does that introduce other issues?

darcywong00:
> Shall we upgrade Java or does that introduce other issues?

From what I can tell, [production-ready binaries](https://openjdk.org/projects/jdk/11/) for Java 11 are currently only available through Oracle. We can probably wait till the next release cycle to tackle the Java upgrade. (See #2293)

mcdurdin:
https://developer.android.com/studio/intro/studio-config#jdk suggests using the embedded JDK included with Android Studio, which is 11.0.13 on my version of Android Studio. So I don't think we need to be limited to Java 8. One minor concern is compatibility with the Web build which uses Closure, as we don't want to require two different versions of Java to be installed.

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.17-alpha](https://keyman.com/downloads/releases/alpha/17.0.17)

PR #7901, opened 2022-12-14
chore(android/samples): Remove Sentry dependencies

### LINKED ISSUE #7896: chore(android/samples): Remove Sentry dependency

darcywong00:
Follow-on to #7897 and fixes #7896

The Android sample/test apps don't use Sentry, so this removes the Sentry dependency. This means consumers of Keyman Engine for Android also won't need to include Sentry in their build.

## User Testing
Setup - Install the PR builds of the specified apps on an Android device/emulator

* **TEST_KEYMAN** - 
1.  Install Keyman for Android
2. Type on the in-app keyboard
3. Verify text is output in the app
4. Long-press on the globe key to display the keyboard picker menu
5. Click the "i" icon to get info on sil_euro_latin keyboard
6. Verify the keyboard info appears **with** a QR code for sharing
 
* **TEST_KMSAMPLE1** - Verifies KMSample1 functions
1. Install KMSample1
2. Type on the in-app keyboard
3. Verify text is output in the app
4. Click a suggestion from the suggestion banner
5. Verify the suggestion is inserted into the app.
6. Long-press on the globe key to display the keyboard picker menu
7. Click the "i" icon to get info on the tamil keyboard
8. Verify the keyboard info appears **without** a QR code for sharing

* **TEST_FV** - Verifies FirstVoices app functions
1. Install FirstVoices for Android
2. On the starting menu, click "Select keyboards" and enable a FirstVoices keyboard (any region & language)
3. On the starting menu, click "Setup" and enable FirstVoices and set as the current input method
4. Launch a separate app (e.g. Chrome) and select a text field
6. Set the keyboard to FirstVoices
7. Type with the FirstVoices keyboard
8. Verify text is output
9. Long-press on the globe key to display the keyboard picker menu
10. Click the "i" to get info on the FV keyboard
11. Verify the keyboard info appears **without** a QR code for sharing.


android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/KMLog.java
```diff
@@ -17,6 +17,21 @@
...
+  public static boolean sentryLibraryExists() {
+    boolean result = false;
+    try {
+      Class.forName("io.sentry.Sentry");
+      return true;
+    } catch (ClassNotFoundException e) {
+      // Intentionally not sending to Sentry because 3rd party apps may not include Sentry library
+      return false;
+    }
+  }
```
mcdurdin, line 33: Can we set this up to run only once and store the result for future executions? Exceptions in Java are [expensive](https://stackoverflow.com/questions/299068/what-are-the-effects-of-exceptions-on-performance-in-java).

mcdurdin (changes requested):
The change looks good, just would like a perf optimisation.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/KMLog.java
darcywong00: OK, I've refactored this to DependencyUtil. But KMLog won't have context available to store the result in a SharedPreference so it'll still go through the Exception once before storing it statically.

bharanidharanj:
# Test Results

* **TEST_KEYMAN (PASSED):** Tested with the attached PR build in API 31 / Android 12.0 emulator and here is my observation: 1. Verified the keyboard info appears with a QR code after clicking the "i" icon. 

   


* **TEST_KMSAMPLE1 (PASSED):** Installed the KMSample app in API 31 / Android 12.0 emulator and here is my observation: 1. Able to type in the Tamil text. 2. Verified the suggestion are inserted into the app. 3. Verified the Keyboard info appears without a QR code for sharing after clicking the 'i' icon. 

   


* **TEST_FV (PASSED):** Installed the FirstVoices app in API 31 / Android 12.0 emulator and here is my observation: 1. Able to type in the FirstVoices keyboard in Chrome browser. 2. Verified the keyboard infor appears without a QR code for sharing after clicking the 'i' icon. FYI, there is some special characters appears near the version label.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/KMLog.java
mcdurdin: hmm, I think just an in-memory value is fine. Don't think we should store it long-term. One exception on first use is acceptable.

android/KMEA/app/src/main/java/com/tavultesoft/kmea/util/DependencyUtil.java
```diff
@@ -0,0 +1,52 @@
...
+
+    try {
+      Class.forName(library.toString());
+      return true;
+    } catch (ClassNotFoundException e) {
+      // Intentionally not sending Exception to Sentry because 3rd party apps may not include the dependency library
+      libraryExists[library.ordinal()] = false;
+      return false;
+    }
+  }
```
mcdurdin, line 50: I think it'd be better to have an initialization flag pattern (note capital-B `Boolean` means it is a boxed type which allows `null`):

```suggestion
  private static Boolean libraryExists[] = {null, null};

  /**
   * Utility to determine if a dependency library is available in an app
   * @param library - LibraryType enum of the library to check: SENTRY or QRCODE
   * @return boolean - true if the library exists
   */
  public static boolean libraryExists(LibraryType library) {
    int ord = library.ordinal();
    if (libraryExists[ord] == null) {
      try {
        Class.forName(library.toString());
        libraryExists[ord] = true;
      } catch (ClassNotFoundException e) {
        // Intentionally not sending Exception to Sentry 
        // because 3rd party apps may not include the dependency library
        libraryExists[ord] = false;
      }
    }
    return libraryExists[ord];
  }
```

mcdurdin (changes requested):
Looking good. I like the consolidation with QRCode stuff too. One suggestion around storing detection values. Query: have you tested this with an app that doesn't include Sentry?

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.20-alpha](https://keyman.com/downloads/releases/alpha/17.0.20)

PR #7917, opened 2022-12-16
fix(linux): Fix dependency of ibus-keyman 🚑

ermshiperete:
When we added the API check we introduced a mis-spelled dependency which lead to Keyman being uninstallable. This change fixes things.

# User Testing

**TEST_INSTALLATION**: 
- It doesn't matter on which Linux platform these tests are run.
- Verify that there are no errors shown in the terminal window


darcywong00 (approved):
Lgtm

mcdurdin:
I'll go ahead and merge this before user test but we'll review after user testing in case of other issues.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.123-beta](https://keyman.com/downloads/releases/beta/16.0.123)

PR #7918, opened 2022-12-16
fix(linux): Fix dependency of ibus-keyman 🚑 🍒

ermshiperete:
When we added the API check we introduced a mis-spelled dependency which lead to Keyman being uninstallable. This change fixes things.

(🍒-picked from PR #7917)

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.19-alpha](https://keyman.com/downloads/releases/alpha/17.0.19)

PR #7825, opened 2022-11-27
fix(windows): Cleanup CEF more correctly in shutdown 🍒

mcdurdin:
Cherry-pick of #7661. Fixes #7824.

Hopefully fixes KEYMAN-WINDOWS-1MR, and probably many related crash reports.

This changes the form destruction sequence to make sure that it doesn't close the form until CEF has been shutdown correctly.

Will monitor this on beta before release and roll it back if other issues arise.

# User Testing

Please do a regression test on Keyman Configuration (just the user interface portions).

## SUITE_BASELINE: Keyman for Windows Base Line Tests

- **TEST_KEYBOARD_INSTALLATION_REMOTE**:
  - Start Keyman Configuration
  - Click Download Keyboard, select a keyboard from the download dialog and install it
  - If the package includes a readme file, verify that it appears in the install window.
  - If the package includes a welcome file, verify that it appears after installation.
  - Click OK to close Keyman Configuration
  - Start Notepad
  - Verify that the newly installed keyboard appears in the Keyman menu and can be selected

- **TEST_INSTALL_PKG_DISK**:
  - Install a package from disk ##
  - Locate a .kmp file on disk and double-click to install it
  - If the package includes a readme file, verify that it appears in the install window.
  - If the package includes a welcome file, verify that it appears after installation.
  - Start Notepad
  - Verify that the keyboard appears in the Keyman menu

## SUITE_CONFIGURATION: Testing of Keyman Configuration shutdown process

- **TEST_STRESS_CONFIGURATION**:
  - The purpose of this test is to stress test closing Keyman Configuration.
  - Open, and then close Keyman Configuration in as many ways as you can find.
  - For example Alt+F4, right-click Close on taskbar, etc.
  - Try opening and closing child dialogs such as 'Download Keyboard'
  - Open Keyman Configuration from the startup screen of Keyman, or from the Keyman icon, or from the Start Menu.
  - We are looking to see if we get any unexpected application behaviour or crashes.

- **TEST_CONFIGURATION_WINDOWS_SHUTDOWN**:
  - The purpose of this test is to verify that Keyman Configuration behaves correctly when restarting Windows.
  - Open Keyman Configuration, and leaving it open, restart Windows.
  - Try this a few times. Try also when you have a dialog such as 'Download Keyboard' open.
  - You may not see evidence of crash reports, but if any occur, we will see them in Sentry.

PR #7933, opened 2022-12-20
fix(windows): shutdown fix masked modal result 🔥 🍒

mcdurdin:
Cherry-pick of #7932.

Regression introduced in #7661 (17.0 alpha), #7825 (16.0 beta).

When shutting down CEF cleanly, we needed to pause the form destruction sequence in order for CEF to complete its shutdown tasks. However, for a dialog form, it will commonly be closed by setting `ModalResult` to a non-zero value. In our fix, we inadvertently lost any `ModalResult` value by calling `Close` after CEF was finished, which sets `ModalResult` to `mrCancel` (value of 2).

This fix caches the `ModalResult` value, and if non-zero, uses it instead after CEF shuts down, to achieve the same result.

# User Testing

**TEST_START_KEYMAN:** Do a clean install of Keyman for Windows, and verify that the Start Keyman button in the Splash screen actually starts Keyman.
**TEST_STOP_KEYMAN:** Select Exit Keyman from the Keyman tray icon, and verify that the OK button in the "Exit Keyman" dialog that appears actually does exit Keyman.

### DIFF
--- a/windows/src/desktop/kmshell/main/UfrmWebContainer.pas
+++ b/windows/src/desktop/kmshell/main/UfrmWebContainer.pas
@@ -60,6 +60,7 @@ type
     FDialogName: WideString;
     FCanClose: Boolean;
     FIsClosing: Boolean;
+    FLastModalResult: Integer;
     procedure WMUser_FormShown(var Message: TMessage); message WM_USER_FormShown;
     procedure WMUser_ContentRender(var Message: TMessage); message WM_USER_ContentRender;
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

rc-swag:
I have run the test with the build linked to this PR.
I can verify both tests pass

# Test Results

* **TEST_START_KEYMAN PASSED:**  Keyman now opens from clicking "start keyman on the splash screen.
* **TEST_STOP_KEYMAN PASSED:** Keyman Exists after clicking the ok on the "are you sure" pop up

PR #7958, opened 2022-12-30
fix(linux): add IBUS_HAS_PREFILTER ifdef to linux/ibus-keyman/tests

srl295:
see #7774 

apply the same change to engine.c as to linux/ibus-keyman/tests/ibusimcontext.c so that we can build without IBUS_HAS_PREFILTER

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/tests/ibusimcontext.c
+++ b/linux/ibus-keyman/tests/ibusimcontext.c
@@ -32,6 +32,11 @@
 #include <config.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.23-alpha](https://keyman.com/downloads/releases/alpha/17.0.23)

Issue #7965, opened 2023-01-02
feat(developer): support 'space bar' row, support other hardware layouts 🙀 


srl295:
followon to PRs #7402 and #7630, when trying to build the mt.xml example file:

- different hardware forms weren't supported, so the 'iso' style with extra keys in the row did not work
- the 'space bar' row was not supported

Two other issues are already covered elsewhere:
- #7533
- #7964

And two followon issues cover TODOs from this:
- #8160 
- #8161

for #7054 🙀

PR #7969, opened 2023-01-02
fix(common): update USVirtualKeyMap to actually be US🙀

srl295:
comon/web/types:
- the USVirtualKeyMap was actually somewhat ISO-ish. Correct it.
- also, add the spacebar row.
- create a new ISOVirtualKeyMap, but not linked in anywhere.

core/tests:
- add a new test keyboard, k_010_mt.xml which is Maltese 47-key
- update k_003 which was impacted by this

For: #7965

@keymanapp-test-bot skip

### DIFF
--- a/common/web/types/src/consts/virtual-key-constants.ts
+++ b/common/web/types/src/consts/virtual-key-constants.ts
@@ -136,8 +136,27 @@ export const USVirtualKeyMap: number[][] = [
   [ k.K_Q, k.K_W, k.K_E, k.K_R, k.K_T, k.K_Y, k.K_U, k.K_I, k.K_O, k.K_P, k.K_LBRKT, k.K_RBRKT, k.K_BKSLASH ],
   // [caps] A S D F G H J K L ; ' [enter]
   [ k.K_A, k.K_S, k.K_D, k.K_F, k.K_G, k.K_H, k.K_J, k.K_K, k.K_L, k.K_COLON, k.K_QUOTE ],
+  // [shift] Z X C V B N M , . / [shift] *=oE2
+  [ k.K_Z, k.K_X, k.K_C, k.K_V, k.K_B, k.K_N, k.K_M, k.K_COMMA, k.K_PERIOD, k.K_SLASH ],
+  // space
+  [ k.K_SPACE ],
+];
+
+/**
+ * TODO-LDML: WIP ISO layout for #7965
+ * May not be correct to use US codes?
+ */
+export const ISOVirtualKeyMap: number[][] = [
+  // ` 1 2 3 4 5 6 7 8 9 0 - = [bksp]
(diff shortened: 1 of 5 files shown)


common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -136,8 +136,27 @@ export const USVirtualKeyMap: number[][] = [
   [ k.K_Q, k.K_W, k.K_E, k.K_R, k.K_T, k.K_Y, k.K_U, k.K_I, k.K_O, k.K_P, k.K_LBRKT, k.K_RBRKT, k.K_BKSLASH ],
   // [caps] A S D F G H J K L ; ' [enter]
   [ k.K_A, k.K_S, k.K_D, k.K_F, k.K_G, k.K_H, k.K_J, k.K_K, k.K_L, k.K_COLON, k.K_QUOTE ],
+  // [shift] Z X C V B N M , . / [shift] *=oE2
+  [ k.K_Z, k.K_X, k.K_C, k.K_V, k.K_B, k.K_N, k.K_M, k.K_COMMA, k.K_PERIOD, k.K_SLASH ],
```
mcdurdin, line 140: Note that some "US" style mongrel keyboards have a 102nd key...

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -136,8 +136,27 @@ export const USVirtualKeyMap: number[][] = [
...
+  // space
+  [ k.K_SPACE ],
+];
+
+/**
+ * TODO-LDML: WIP ISO layout for #7965
+ * May not be correct to use US codes?
```
mcdurdin, line 147: AFAIK the US codes are still correct.

mcdurdin (approved):
I *think* this is okay as it was introduced specifically for LDML keyboards. However, we do need to be a little careful as there are US-layout keyboards that have a 102nd key (that is, they are neither ISO nor US, really!)

common/web/types/src/consts/virtual-key-constants.ts
srl295: is the key on the right hand side? we may need to handle it as a separate hardware layout then

srl295:
Demo of typing with the included mt layout 
image

common/web/types/src/consts/virtual-key-constants.ts
mcdurdin: I don't think we need to handle it as part of the LDML keyboard spec. In Core, we should make the 102nd key inherit the behaviour of the Windows US keyboard (emit `\` and `|`) when used with a `"us"` hardware layout ldml keyboard, so it will still work okay.

srl295:
a test case broke

srl295:
@mcdurdin I'm rebasing this because the import and implied PRs need to be merged together.  It's the same as you saw, except the maltese keyboard is disabled for now 🇲🇹

rc-swag (approved):
lgtm

PR #4588, opened 2021-03-04
fix(android): Localize Toast notifications

darcywong00:
Follow-on to #4261 and fixes #4475

Currently, overriding string resources to use the chosen display language only work for classes that extend `BaseActivity` (`AppCompatActivity`). For the remaining classes, Toast notifications end up using the device's default locale.

This PR adds a helper method `BaseActivity.makeToast()` so the Toast messages will be consistent with the display language. ~I had to simplify some Toast strings to avoid `makeToast` needing to `String.format` with variable parameters.~

(updated): The Toast strings optionally handle formatted strings. Note, messages to Sentry aren't localized.

### Other cleanup
* Remove unused string `failed_to_access_downloaded_file`

## User Testing
@MakaraSok , please test with the scenarios you reported in the issue

PR #7976, opened 2023-01-04
fix(android/engine): Add utility for localized strings

### LINKED ISSUE #7975: bug(android/engine,web):  Globe help string always in English
Seen while testing #7970 (16.0 beta)
On a fresh install of Keyman for Android, whenever I change the app Display Language, the globe help tip remains in English: "Tap here to change keyboard"
globe

The globe key help bubble was re-implemented in #7612 so as a starting investigation point, we can check if the localized string gets passed to KMW

darcywong00:
Fixes #7975 where the globe help tip was stuck in English and didn't match the UI locale.

Similar to the fix in #4588, this adds a BaseActivity utility so non AppCompatActivities can access localized strings.

## User Testing
Setup - Do a clean Install PR build of Keyman for Android on default English locale.

* **TEST_GLOBE_KEY_LOCALIZED** - Verifies the globe key text is localized and matches the current display language
1. Launch Keyman for Android
2. Verify the globe key text is in English 
3. From Keyman, go to Settings --> Display Language --> Khmer
4. Exit the Keyman settings menu
5. In the app, verify the UI is in Khmer, along with the globe key text
khmer-globe

6. From Keyman, go to Settings --> Display Language --> German
7. Exit the Keyman settings menu
8. In the app, verify the UI is in German, along with the globe key text
 
german-globe

### DIFF
--- a/android/KMEA/app/src/main/java/com/tavultesoft/kmea/BaseActivity.java
+++ b/android/KMEA/app/src/main/java/com/tavultesoft/kmea/BaseActivity.java
@@ -49,6 +49,21 @@ public class BaseActivity extends AppCompatActivity {
     }
   }
 
+  /**
+   * Some classes aren't an AppCompatActivity and need this helper to retrieve localized Strings
+   * in the updated locale.
+   * @param defaultContext - the context to fallback if localUpdatedContext is null
+   * @param resID - the resource ID of the string
+   * @return String - localized string
+   */
+  public static String getString(Context defaultContext, int resID) {
+    Context context = (localeUpdatedContext != null) ? localeUpdatedContext : defaultContext;
+    if (context != null) {
+      return context.getString(resID);
+    };
+    return "";
+  }
+
   @Override
   protected void attachBaseContext(Context newBase) {
     // Override the app locale using the BCP 47 tag from shared preferences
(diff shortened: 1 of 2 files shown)


bharanidharanj:
# Test Results

* **TEST_GLOBE_KEY_LOCALIZED (PASSED):** Tested with the attached PR build (16.0.129-beta-test-7976) in API 31 / Android 12.0 emulator and here is my observation: 1. Able to see the globe key text in Khmer language. 2. Changed the UI language into German. Able to see the globe key text is changed to German. Seems to be working as expected.

srl295 (approved):
FWIW LGTM and makes sense

android/KMEA/app/src/main/java/com/tavultesoft/kmea/BaseActivity.java
```diff
@@ -49,6 +49,21 @@ public static void makeToast(Context defaultContext, String msg, int duration) {
...
+   * @param defaultContext - the context to fallback if localUpdatedContext is null
+   * @param resID - the resource ID of the string
+   * @return String - localized string
+   */
+  public static String getString(Context defaultContext, int resID) {
+    Context context = (localeUpdatedContext != null) ? localeUpdatedContext : defaultContext;
+    if (context != null) {
+      return context.getString(resID);
+    };
+    return "";
```
mcdurdin, line 64: We should log this as it is a failure case

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.129-beta](https://keyman.com/downloads/releases/beta/16.0.129)

PR #7988, opened 2023-01-06
chore(linux): Remove unnecessary variable

ermshiperete:
We can directly use `logging.error` instead of `logging.log`. Then we don't need the `severity` variable.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.25-alpha](https://keyman.com/downloads/releases/alpha/17.0.25)

PR #7992, opened 2023-01-06
fix(core): ldml: add asserts to C++ 🙀

### LINKED ISSUE #7984: feat(core): ldml assert on invalid read 🙀
https://github.com/keymanapp/keyman/pull/7630/files#r1063295825

in kmx_plus.cpp:

> We should consider assert()ing these _(when invalid section data is encountered)_ as they indicate an error which it'd be good to catch in a debug build.

#7042

srl295:
- found a logic bug in kmx_plus.cpp

Fixes #7984

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -65,25 +65,25 @@ validate_section_name(KMX_DWORD ident) {
  */
 static const kmx::COMP_KMXPLUS_HEADER *
 header_from_bytes(const uint8_t *data, KMX_DWORD length, uint32_t ident) {
-  assert(data);
   if (!data) {
     DebugLog("!data");
+    assert(false);
     return nullptr;
   }
-  assert(length >= LDML_LENGTH_HEADER);
   if (length < LDML_LENGTH_HEADER) {
     DebugLog("length < LDML_LENGTH_HEADER");
+    assert(false);
     return nullptr;
   }
   const COMP_KMXPLUS_HEADER *all = reinterpret_cast<const COMP_KMXPLUS_HEADER *>(data);
   if (!all->valid(length)) {
-    assert(false);
     DebugLog("header failed validation");
+    assert(false);
     return nullptr;
   }
-  assert(all->ident == ident);
   if (all->ident != ident) {
     DebugLog("header had wrong section id");
+    assert(false);
     return nullptr;
   }
   return all;
@@ -99,9 +99,9 @@ header_from_bytes(const uint8_t *data, KMX_DWORD length, uint32_t ident) {
  */
 template <class T>
 const T *section_from_bytes(const uint8_t *data, KMX_DWORD length) {
-  assert(length >= sizeof(T));
   if (length < sizeof(T)) { // Does not include dynamic data. First check.
     DebugLog("length < sizeof(section)");
+    assert(false);
     return nullptr;
   }
   const COMP_KMXPLUS_HEADER *header = header_from_bytes(data, length, T::IDENT);
@@ -118,9 +118,9 @@ template <class T>
 const T *section_from_sect(const COMP_KMXPLUS_SECT* sect) {
   DebugLog("----"); // separate a new section's validation
   const uint8_t *rawbytes = reinterpret_cast<const uint8_t *>(sect);
-  assert(rawbytes != nullptr);
   if (rawbytes == nullptr) {
     DebugLog("section_from_sect(nullptr) == nullptr");
+    assert(false);
     return nullptr;
   }
   KMX_DWORD offset = sect->find(T::IDENT);
@@ -135,14 +135,14 @@ const T *section_from_sect(const COMP_KMXPLUS_SECT* sect) {
 bool
 COMP_KMXPLUS_HEADER::valid(KMX_DWORD length) const {
   DebugLog("%c%c%c%c: (%X) size 0x%X\n", DEBUG_IDENT(ident), ident, size);
-  assert(size >= LDML_LENGTH_HEADER);
   if (size < LDML_LENGTH_HEADER) {
     DebugLog("size < LDML_LENGTH_HEADER");
+    assert(false);
     return false;
   }
-  assert(size <= length);
   if (size > length) {
     DebugLog("size > length");
+    assert(false);
     return false;
   }
   if (!validate_section_name(ident)) {
@@ -156,8 +156,8 @@ bool
 COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
   DebugLog(" count: #0x%X\n", count);
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
-    assert(false);
     DebugLog("header.size < expected size");
+    assert(false);
     return false;
   }
   for (KMX_DWORD i = 0; i<count; i++) {
@@ -217,8 +217,8 @@ bool
 COMP_KMXPLUS_VKEY::valid(KMX_DWORD _kmn_unused(length)) const {
   DebugLog("vkey: count 0x%X\n", count);
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
-    assert(false);
     DebugLog("header.size < expected size");
+    assert(false);
     return false;
   }
   // TODO-LDML
@@ -230,8 +230,8 @@ bool
 COMP_KMXPLUS_DISP::valid(KMX_DWORD _kmn_unused(length)) const {
   DebugLog("disp: count 0x%X\n", count);
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
-    assert(false);
     DebugLog("header.size < expected size");
+    assert(false);
     return false;
   }
   // TODO-LDML
@@ -258,9 +258,9 @@ COMP_KMXPLUS_STRS::valid(KMX_DWORD _kmn_unused(length)) const {
     }
     const uint8_t* thisptr = reinterpret_cast<const uint8_t*>(this);
     const KMX_WCHAR* start = reinterpret_cast<const KMX_WCHAR*>(thisptr+offset);
-    assert(start[length] == 0);
     if(start[length] != 0) {
       DebugLog("#0x%X: String of length 0x%x not null terminated", i, length);
+      assert(start[length] == 0);
       return false;
     }
     // TODO-LDML: validate valid UTF-16LE?
@@ -335,9 +335,9 @@ COMP_KMXPLUS_ELEM::valid(KMX_DWORD _kmn_unused(length)) const {
 
 const COMP_KMXPLUS_ELEM_ELEMENT *
 COMP_KMXPLUS_ELEM::getElementList(KMX_DWORD elementNumber, KMX_DWORD &length) const {
-  assert(elementNumber < count);
   if (elementNumber >= count) {
     DebugLog("ERROR: COMP_KMXPLUS_ELEM::getElementList(%d) >= count %d", elementNumber, count);
+    assert(false);
     return nullptr;
   }
   const COMP_KMXPLUS_ELEM_ENTRY &entry = entries[elementNumber];
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -47,6 +47,7 @@ validate_section_name(KMX_DWORD ident) {
     unsigned char ch = ident & 0xFF;
     if (ch < 0x20 || ch > 0x7F) {
       DebugLog("Invalid section name %c%c%c%c (0x%X)", DEBUG_IDENT(ident), ident);
+      assert(false);
       return false;
     }
     ident >>= 8;
@@ -64,18 +65,23 @@ validate_section_name(KMX_DWORD ident) {
  */
 static const kmx::COMP_KMXPLUS_HEADER *
 header_from_bytes(const uint8_t *data, KMX_DWORD length, uint32_t ident) {
+  assert(data);
   if (!data) {
     DebugLog("!data");
     return nullptr;
   }
+  assert(length >= LDML_LENGTH_HEADER);
   if (length < LDML_LENGTH_HEADER) {
     DebugLog("length < LDML_LENGTH_HEADER");
+    return nullptr;
   }
   const COMP_KMXPLUS_HEADER *all = reinterpret_cast<const COMP_KMXPLUS_HEADER *>(data);
   if (!all->valid(length)) {
+    assert(false);
     DebugLog("header failed validation");
     return nullptr;
   }
+  assert(all->ident == ident);
   if (all->ident != ident) {
     DebugLog("header had wrong section id");
     return nullptr;
@@ -93,6 +99,7 @@ header_from_bytes(const uint8_t *data, KMX_DWORD length, uint32_t ident) {
  */
 template <class T>
 const T *section_from_bytes(const uint8_t *data, KMX_DWORD length) {
+  assert(length >= sizeof(T));
   if (length < sizeof(T)) { // Does not include dynamic data. First check.
     DebugLog("length < sizeof(section)");
     return nullptr;
@@ -102,6 +109,7 @@ const T *section_from_bytes(const uint8_t *data, KMX_DWORD length) {
   if (section != nullptr && section->valid(length)) {
     return section;
   } else {
+    assert(false);
     return nullptr;
   }
 }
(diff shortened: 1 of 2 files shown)


core/src/kmx/kmx_plus.cpp
```diff
@@ -376,17 +401,19 @@ COMP_KMXPLUS_LAYR_Helper::setLayr(const COMP_KMXPLUS_LAYR *newLayr) {
   if (newLayr == nullptr) {
     // null = invalid
     is_valid = false;
+    // No assert here: just a missing layer
     return false;
   }
   layr = newLayr;
-  const uint8_t *rawdata = reinterpret_cast<const uint8_t *>(this);
+  const uint8_t *rawdata = reinterpret_cast<const uint8_t *>(newLayr);
```
srl295, line 408: Woops! This was the error.

srl295:
not sure about 2 copies of the logic

```c
assert (a <= b);
if(a > b) {
   return nullptr;
}
```

mcdurdin:
> not sure about 2 copies of the logic

Agreed, this turns into a classic copypasta problem. Simple always-false assert is probably safer. (I wish we could do messages in C asserts!)

```c
  if(a > b) {
    assert(false);  // throw in debug mode
    return nullptr;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -64,18 +65,23 @@ validate_section_name(KMX_DWORD ident) {
  */
 static const kmx::COMP_KMXPLUS_HEADER *
 header_from_bytes(const uint8_t *data, KMX_DWORD length, uint32_t ident) {
+  assert(data);
   if (!data) {
     DebugLog("!data");
     return nullptr;
```
mcdurdin, line 71:
```suggestion
  if (!data) {
    DebugLog("!data");
    assert(false);
    return nullptr;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -64,18 +65,23 @@ validate_section_name(KMX_DWORD ident) {
...
   }
+  assert(length >= LDML_LENGTH_HEADER);
   if (length < LDML_LENGTH_HEADER) {
     DebugLog("length < LDML_LENGTH_HEADER");
+    return nullptr;
   }
```
mcdurdin, line 77:
```suggestion
  if (length < LDML_LENGTH_HEADER) {
    DebugLog("length < LDML_LENGTH_HEADER");
    assert(false);
    return nullptr;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -64,18 +65,23 @@ validate_section_name(KMX_DWORD ident) {
...
   const COMP_KMXPLUS_HEADER *all = reinterpret_cast<const COMP_KMXPLUS_HEADER *>(data);
   if (!all->valid(length)) {
+    assert(false);
     DebugLog("header failed validation");
     return nullptr;
   }
```
mcdurdin, line 83:
```suggestion
  if (!all->valid(length)) {
    DebugLog("header failed validation");
    assert(false);
    return nullptr;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -64,18 +65,23 @@ validate_section_name(KMX_DWORD ident) {
...
+  assert(all->ident == ident);
   if (all->ident != ident) {
     DebugLog("header had wrong section id");
     return nullptr;
```
mcdurdin, line 87:
```suggestion
  if (all->ident != ident) {
    DebugLog("header had wrong section id");
    assert(false);
    return nullptr;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -93,6 +99,7 @@ header_from_bytes(const uint8_t *data, KMX_DWORD length, uint32_t ident) {
  */
 template <class T>
 const T *section_from_bytes(const uint8_t *data, KMX_DWORD length) {
+  assert(length >= sizeof(T));
   if (length < sizeof(T)) { // Does not include dynamic data. First check.
     DebugLog("length < sizeof(section)");
     return nullptr;
```
mcdurdin, line 105:
```suggestion
  if (length < sizeof(T)) { // Does not include dynamic data. First check.
    DebugLog("length < sizeof(section)");
    assert(false);
    return nullptr;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -110,6 +118,7 @@ template <class T>
 const T *section_from_sect(const COMP_KMXPLUS_SECT* sect) {
   DebugLog("----"); // separate a new section's validation
   const uint8_t *rawbytes = reinterpret_cast<const uint8_t *>(sect);
+  assert(rawbytes != nullptr);
   if (rawbytes == nullptr) {
     DebugLog("section_from_sect(nullptr) == nullptr");
     return nullptr;
```
mcdurdin, line 124:
```suggestion
  if (rawbytes == nullptr) {
    DebugLog("section_from_sect(nullptr) == nullptr");
    assert(false);
    return nullptr;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -126,10 +135,12 @@ const T *section_from_sect(const COMP_KMXPLUS_SECT* sect) {
 bool
 COMP_KMXPLUS_HEADER::valid(KMX_DWORD length) const {
   DebugLog("%c%c%c%c: (%X) size 0x%X\n", DEBUG_IDENT(ident), ident, size);
+  assert(size >= LDML_LENGTH_HEADER);
   if (size < LDML_LENGTH_HEADER) {
     DebugLog("size < LDML_LENGTH_HEADER");
     return false;
   }
```
mcdurdin, line 142:
```suggestion
  if (size < LDML_LENGTH_HEADER) {
    DebugLog("size < LDML_LENGTH_HEADER");
    assert(false);
    return false;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -126,10 +135,12 @@ const T *section_from_sect(const COMP_KMXPLUS_SECT* sect) {
...
+  assert(size <= length);
   if (size > length) {
     DebugLog("size > length");
     return false;
```
mcdurdin, line 146:
```suggestion
  if (size > length) {
    DebugLog("size > length");
    assert(false);
    return false;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -145,6 +156,7 @@ bool
 COMP_KMXPLUS_KEYS::valid(KMX_DWORD _kmn_unused(length)) const {
   DebugLog(" count: #0x%X\n", count);
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
+    assert(false);
     DebugLog("header.size < expected size");
     return false;
   }
```
mcdurdin, line 162:
```suggestion
  if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
    DebugLog("header.size < expected size");
    assert(false);
    return false;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -203,6 +217,7 @@ bool
 COMP_KMXPLUS_VKEY::valid(KMX_DWORD _kmn_unused(length)) const {
   DebugLog("vkey: count 0x%X\n", count);
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
+    assert(false);
     DebugLog("header.size < expected size");
     return false;
   }
```
mcdurdin, line 223:
```suggestion
  if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
    DebugLog("header.size < expected size");
    assert(false);
    return false;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -215,6 +230,7 @@ bool
 COMP_KMXPLUS_DISP::valid(KMX_DWORD _kmn_unused(length)) const {
   DebugLog("disp: count 0x%X\n", count);
   if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
+    assert(false);
     DebugLog("header.size < expected size");
     return false;
   }
```
mcdurdin, line 236:
```suggestion
  if (header.size < sizeof(*this)+(sizeof(entries[0])*count)) {
    DebugLog("header.size < expected size");
    assert(false);
    return false;
  }
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -236,10 +253,12 @@ COMP_KMXPLUS_STRS::valid(KMX_DWORD _kmn_unused(length)) const {
...
       DebugLog("#0x%X: expected end of string past header.size", i);
+      assert(false);
       return false;
     }
     const uint8_t* thisptr = reinterpret_cast<const uint8_t*>(this);
     const KMX_WCHAR* start = reinterpret_cast<const KMX_WCHAR*>(thisptr+offset);
+    assert(start[length] == 0);
     if(start[length] != 0) {
       DebugLog("#0x%X: String of length 0x%x not null terminated", i, length);
       return false;
```
mcdurdin, line 264:
```suggestion
    if(start[length] != 0) {
      DebugLog("#0x%X: String of length 0x%x not null terminated", i, length);
      assert(start[length] == 0);
      return false;
```

core/src/kmx/kmx_plus.cpp
```diff
@@ -314,6 +335,7 @@ COMP_KMXPLUS_ELEM::valid(KMX_DWORD _kmn_unused(length)) const {
 
 const COMP_KMXPLUS_ELEM_ELEMENT *
 COMP_KMXPLUS_ELEM::getElementList(KMX_DWORD elementNumber, KMX_DWORD &length) const {
+  assert(elementNumber < count);
   if (elementNumber >= count) {
     DebugLog("ERROR: COMP_KMXPLUS_ELEM::getElementList(%d) >= count %d", elementNumber, count);
     return nullptr;
```
mcdurdin, line 341:
```suggestion
  if (elementNumber >= count) {
    DebugLog("ERROR: COMP_KMXPLUS_ELEM::getElementList(%d) >= count %d", elementNumber, count);
    assert(false);
    return nullptr;
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -66,6 +67,7 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
       std::u16string str;
       const kmx::COMP_KMXPLUS_KEYS_ENTRY &entry = kplus.keys->entries[i];
       if (entry.flags && LDML_KEYS_FLAGS_EXTEND) {
+        assert(nullptr != kplus.strs);
         if (nullptr == kplus.strs) {
           DebugLog("for keys: kplus.strs == nullptr"); // need a string table to get strings
           return;
```
mcdurdin, line 73:
```suggestion
        if (nullptr == kplus.strs) {
          DebugLog("for keys: kplus.strs == nullptr"); // need a string table to get strings
          assert(false);
          return;
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -78,10 +80,12 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
     }
   } // else: no keys! but still valid. Just, no keys.
   if (kplus.tran != nullptr) {
+    assert(nullptr != kplus.elem);
     if (nullptr == kplus.elem) {
       DebugLog("for tran: kplus.elem == nullptr");
       return;
     }
```
mcdurdin, line 87:
```suggestion
    if (nullptr == kplus.elem) {
      DebugLog("for tran: kplus.elem == nullptr");
      assert(false);
      return;
    }
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -78,10 +80,12 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
...
+    assert(nullptr != kplus.strs);
     if (nullptr == kplus.strs) {
       DebugLog("for tran: kplus.strs == nullptr"); // need a string table to get strings
       return;
```
mcdurdin, line 91:
```suggestion
    if (nullptr == kplus.strs) {
      DebugLog("for tran: kplus.strs == nullptr"); // need a string table to get strings
      assert(false);
      return;
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -98,6 +102,7 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
 
       // process From
       const KMX_DWORD elemNo = entry.from;
+      assert(elemNo < kplus.elem->count);
       if (elemNo >= kplus.elem->count) {
         DebugLog("tran[%d].from = %d, out of range for elem->count=%d", i, elemNo, kplus.elem->count);
         return;
```
mcdurdin, line 108:
```suggestion
      if (elemNo >= kplus.elem->count) {
        DebugLog("tran[%d].from = %d, out of range for elem->count=%d", i, elemNo, kplus.elem->count);
        assert(false);
        return;
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -106,10 +111,12 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
       // TODO-LDML: refactor this
       KMX_DWORD elemListLength                       = 0;
       const kmx::COMP_KMXPLUS_ELEM_ELEMENT *elemList = kplus.elem->getElementList(elemNo, elemListLength);
+      assert(elemList != nullptr);
       if (elemList == nullptr) {
         DebugLog("tran[%d].from = %d, could not load element list", i, elemNo);
         return;
       }
```
mcdurdin, line 118:
```suggestion
      if (elemList == nullptr) {
        DebugLog("tran[%d].from = %d, could not load element list", i, elemNo);
        assert(false);
        return;
      }
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -106,10 +111,12 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
...
+      assert(elemListLength != 0);
       if (elemListLength == 0) {
         DebugLog("tran[%d].from = %d, element list has 0 length", i, elemNo);
         return;
```
mcdurdin, line 122:
```suggestion
      if (elemListLength == 0) {
        DebugLog("tran[%d].from = %d, element list has 0 length", i, elemNo);
        assert(false);
        return;
```

mcdurdin (changes requested):
I'm fiddling here...

One could also define an `assert_false_and_return(retval, logmessage, ...)` macro which would bundle that DebugLog(), assert(), return pattern. (May need two versions, one which does a plain `return;`.) I think this could be a good strategy, and enforce a consistent pattern of use.

In lieu of that, I went through and moved the assertion in each case _after_ the `DebugLog` so that the last debug log message relates directly to the assertion.

I'm not 100% _sure_ that all the assertions here are appropriate -- assertions shouldn't typically be used for input data validation, but for program logic validation, and some of these assertions are being thrown for invalid kmx_plus structures. That said, we are also responsible for generating said structures in kmc, so perhaps this is still appropriate (and certainly early-fail makes it easier to track down _why_ a file is invalid!)

srl295:
> I'm fiddling here...

🎻 

> One could also define an `assert_false_and_return(retval, logmessage, ...)` macro which would bundle that DebugLog(), assert(), return pattern. (May need two versions, one which does a plain `return;`.) I think this could be a good strategy, and enforce a consistent pattern of use.

it would actually be something like the `test_assert…` calls, but would end up with assert(false) instead of test failure.

> In lieu of that, I went through and moved the assertion in each case _after_ the `DebugLog` so that the last debug log message relates directly to the assertion.

yeah that was the other way i was considering.

> I'm not 100% _sure_ that all the assertions here are appropriate -- assertions shouldn't typically be used for input data validation, but for program logic validation, and some of these assertions are being thrown for invalid kmx_plus structures. That said, we are also responsible for generating said structures in kmc, so perhaps this is still appropriate (and certainly early-fail makes it easier to track down _why_ a file is invalid!)

i hear you, but actually *all* of these are for invalid kmx_plus structures. To quote @mcdurdin (emphasis mine) 
> We should consider assert()ing these (_when invalid section data is encountered_) as they indicate an error which it'd be good to catch in a debug build.

The asserts are noops in release mode (right? right??) so no harm is done there.   I'd think unless/until kmx_plus.cpp is expected to have some sort of `validate_with_results()` call (that gives internal state about why validation failed),  assert(false) at least gives you an abort / debugger trap at the failing source line.

mcdurdin:
> i hear you, but actually _all_ of these are for invalid kmx_plus structures.

Gotcha. That's what I get for trying to do reviews when I am in 'moving' mode... Yes, all looks good to me then.

mcdurdin:
> asserts are noops in release mode (right? right??)

Yep

srl295:
> > i hear you, but actually _all_ of these are for invalid kmx_plus structures.
> 
> Gotcha. That's what I get for trying to do reviews when I am in 'moving' mode... Yes, all looks good to me then.

can i get a stamp then? thanks

mcdurdin (approved):

srl295:
that's _the_ stamp

PR #8011, opened 2023-01-12
feat(linux): Improve uninstallation

ermshiperete:
Even if the directory doesn't exist, we can still remove the keyboard from the keyboard list.

# User Testing

## Preparations
- start km-config
- install "Khmer Angkor" keyboard
- verify that it shows up in the language/keyboard dropdown in the taskbar

## Tests

**TEST_UNINSTALL**: Uninstallation with unavailable keyboard directory still removes keyboard from language dropdown

- start km-config
- in a terminal window, run the command `rm -rf ~/.local/share/keyman/khmer_angkor` (don't close km-config!)
- in km-config, select installed "khmer angkor" keyboard
- click uninstall button
- click "Yes" on the verification dialog
- verify a message dialog is shown that the keyboard directory doesn't exist
- verify that the Khmer Angkor keyboard no longer shows up in the language/keyboard dropdown in the taskbar

### DIFF
--- a/linux/keyman-config/keyman_config/uninstall_kmp.py
+++ b/linux/keyman-config/keyman_config/uninstall_kmp.py
@@ -17,7 +17,10 @@ from keyman_config.ibus_util import (get_ibus_bus, restart_ibus,
 from keyman_config.kmpmetadata import get_metadata
 
 
-def delete_dir(dir: str) -> bool:
+error_dirs = set()
+
+
+def _delete_dir(dir: str) -> bool:
     if not os.path.isdir(dir):
         logging.error("%s is not a directory", dir)
         return False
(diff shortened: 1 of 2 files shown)


linux/keyman-config/keyman_config/uninstall_kmp.py
```diff
@@ -146,14 +144,18 @@ def uninstall_kmp_user(packageID):
             uninstall_keyboards_from_ibus(keyboards, kbdir)
     else:
         logging.warning("could not uninstall keyboards")
-    delete_dir(kbdir)
-    logging.info("Removed user keyman directory: %s", kbdir)
+    if not os.path.isdir(kbdir):
+        msg = _("Keyboard directory for %s does not exist." % packageID)
```
ermshiperete, line 148: I'm debating whether it is useful to show this error to the user (in a dialog). Since he can't do anything about it, and the directory is already gone it seem we don't need to show it. Thoughts?

bharanidharanj:
# Test Results

* **TEST_UNINSTALL (FAILED):** Tested with the attached PR build (keyman 17.0.28-alpha (package version 17.0.28-1~PR-8011+focal2)) in Ubuntu 20.04 Focal Linux OS (VM) and here is my observation: 1. Installed Khmer Angkor Keyboard. 2. Opened the terminal window and run the above mentioned (rm)command. 3. run km-config command in the terminal. Here, I noticed that there is no Khmer Angkor keyboard appear in the list. But, the same keyboard appears on the task bar. Seems to be an issue.

ermshiperete:
@bharanidharanj I had the order of steps wrong, sorry. `km-config` needs to be started before and kept open while the `rm` command gets runs. Can you test again, please?

@keymanapp-test-bot retest

linux/keyman-config/keyman_config/uninstall_kmp.py
mcdurdin: Don't show it. Just log it. Generally, if there's nothing the user can do about it, then probably better to avoid showing it (otherwise it just causes distress for nothing).

linux/keyman-config/keyman_config/uninstall_kmp.py
```diff
@@ -36,15 +36,26 @@ def uninstall_kmp_shared(packageID):
...
+            uninstall_keyboards_from_gnome(keyboards, kbdir)
+        else:
+            uninstall_keyboards_from_ibus(keyboards, kbdir)
+    else:
+        logging.warning("could not uninstall keyboards")
+
     if not os.path.isdir(kbdir):
         msg = _("Keyboard directory for %s does not exist." % packageID)
         logging.error(msg)
         return msg
```
darcywong00, line 57: I see even if `kbdir` doesn't exist, the keyboards get uninstalled from `uninstall_keyboard_from_gnome` or `uninstall_keyboards_from_ibus`.

But if kbdir doesn't exist, won't l 57. exit? Then, how does the kbddocdir at l.68 or kbfontdir at l. 72 get removed?

bharanidharanj:
> @bharanidharanj I had the order of steps wrong, sorry. `km-config` needs to be started before and kept open while the `rm` command gets runs. Can you test again, please?
> 
> @keymanapp-test-bot retest

@ermshiperete Sure. I will retest it.

bharanidharanj:
# Test Results

* **TEST_UNINSTALL (PASSED):** Retested with the updated PR build (Keyman 17.0.29-alpha (package version 17.0.29-1~PR-8011+focal5) in Ubuntu 20.04 Focal Linux OS (VM) and it is showing the expected error message in the terminal window.

linux/keyman-config/keyman_config/uninstall_kmp.py
ermshiperete: Good point. Did some more refactoring how the uninstallation is done.

linux/keyman-config/keyman_config/uninstall_kmp.py
```diff
@@ -28,63 +31,55 @@ def delete_dir(dir: str) -> bool:
...
 
-    Args:
-        packageID (str): Keyboard package ID
-    """
-    kbdir = get_keyboard_dir(InstallLocation.Shared, packageID)
-    if not os.path.isdir(kbdir):
-        msg = _("Keyboard directory for %s does not exist." % packageID)
-        logging.error(msg)
-        return msg
+        _delete_dir(dir)
```
darcywong00, line 40: Would it be useful to check the status from _delete_dir?

The original code didn't use it though

darcywong00 (approved):
lgtm

ermshiperete:
@bharanidharanj Would you please re-test because this PR had major surgery. Thanks!

@keymanapp-test-bot retest

bharanidharanj:
# Test Results

* **TEST_UNINSTALL (FAILED):** Retested with the attached PR build (keyman 17.0.31-alpha (package version 17.0.31-1~PR-8011+focal13)) in Ubuntu 20.04 Focal Linux and here is my observation: 1. Noticed that the Khmer Angkor keyboard still appears on the language/keyboard dropdown in the taskbar after uninstalling it. Seems to be an issue.

ermshiperete:
TEST_UNINSTALL (SKIPPED): My bad. The test I suggested is invalid (at least with/after this PR). km-config needs the installed files to enumerate over the installed keyboards. If we delete the directory before that (as in this test), we don't know what keyboards to remove and so don't do anything.

PR #8021, opened 2023-01-13
chore(linux): Don't report KeyboardInterrupt to Sentry

ermshiperete:
When the user presses Ctrl-C we don't want to create a Sentry issue. This change ignores that exception.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.29-alpha](https://keyman.com/downloads/releases/alpha/17.0.29)

PR #8022, opened 2023-01-13
chore(linux): Don't report KeyboardInterrupt to Sentry 🍒

ermshiperete:
When the user presses Ctrl-C we don't want to create a Sentry issue. This change ignores that exception.

(Should have cherry-picked from `beta` to `master`, but thought too late that it would be good to have this change in 16.0)

(🍒-picked from #8021)

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/__init__.py
+++ b/linux/keyman-config/keyman_config/__init__.py
@@ -37,6 +37,15 @@ def secure_lookup(data, key1, key2 = None):
     return None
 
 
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.131-beta](https://keyman.com/downloads/releases/beta/16.0.131)

PR #8025, opened 2023-01-16
fix(linux): Properly set context after changing IP

ermshiperete:
This change fixes a bug where we the context was wrong after putting the IP in a new location by mouse-click.

Fixes #8024.

# User Testing

## Preparations

- [Install build artifacts of this PR](https://github.com/keymanapp/keyman/wiki/How-to-test-artifacts-from-pull-requests-for-Keyman-for-Linux)
- Reboot
- Install **IPA (SIL)** keyboard

## Tests

**TEST_SETS_CONTEXT**: Correctly gets surrounding text of current cursor location

- open gedit
- switch to IPA (SIL) keyboard
- Type
  ```
  an
  m
  ```
- click after `n`
- type `>`
- verify that the output is
  ```
  aŋ
  m
  ```

PR #8026, opened 2023-01-16
fix(linux): Properly set context after changing IP 🍒

### LINKED ISSUE #8024: bug(linux): Wrong context after setting IP to different line
To reproduce with **IPA (SIL)** keyboard:
- open gedit
- Type
  ```
  an
  m
  ```
- click after `n`
- type `>`

## Expected output
```
aŋ
m
```

## Actual output
```
aɱ
m
```

ermshiperete:
This change fixes a bug where we the context was wrong after putting the IP in a new location by mouse-click.

Fixes #8024.

(🍒-picked from PR #8025)

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #8036, opened 2023-01-18
chore(linux): log failures to `km_kbp_context_clear(context)`

### LINKED ISSUE #7540: chore(linux): log failures to `km_kbp_context_clear(context)`
If this fails, I think we should `km_kbp_context_clear(context)` and log.
```suggestion
      if ((status = km_kbp_context_items_from_utf8(surrounding_text, &context_items)) == KM_KBP_STATUS_OK) {
        km_kbp_context_set(context, context_items);
        km_kbp_context_items_dispose(context_items);
      } else {
        km_kbp_context_clear(context);
        g_message("%s: setting context failed with %d", __FUNCTION__, status);
      }
```

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/7084#discussion_r1001038333_

TheSchlunz:
This change adds behavior when setting the context fails. It then clears the context and logs the failure. 

Fixes #7540

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -277,12 +277,16 @@ reset_context(IBusEngine *engine) {
       if (km_kbp_context_items_from_utf8(surrounding_text, &context_items) == KM_KBP_STATUS_OK) {
         km_kbp_context_set(context, context_items);
         km_kbp_context_items_dispose(context_items);
+      } else {
+        km_kbp_context_clear(context);
+        g_message("%s: setting context failed", __FUNCTION__);
       }
     }
     g_free(surrounding_text);
     g_free(current_context_utf8);
   } else {
     km_kbp_context_clear(context);
+    g_message("%s: client does not support surrounding text", __FUNCTION__);
   }
 }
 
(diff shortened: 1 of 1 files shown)


linux/ibus-keyman/src/engine.c
```diff
@@ -277,12 +277,16 @@ reset_context(IBusEngine *engine) {
       if (km_kbp_context_items_from_utf8(surrounding_text, &context_items) == KM_KBP_STATUS_OK) {
         km_kbp_context_set(context, context_items);
         km_kbp_context_items_dispose(context_items);
+      } else {
+        km_kbp_context_clear(context);
+        g_message("%s: setting context failed", __FUNCTION__);
```
ermshiperete, line 282: It would be good to also output the status that `km_kbp_context_items_from_utf8` returned.

ermshiperete (changes requested):

linux/ibus-keyman/src/engine.c
TheSchlunz: I've now added the status code to the output.

ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.31-alpha](https://keyman.com/downloads/releases/alpha/17.0.31)

PR #8037, opened 2023-01-18
chore(linux): use faster zero-length string check

### LINKED ISSUE #7541: chore(linux): use faster zero-length string check
```suggestion
    if (!g_str_has_suffix(surrounding_text, current_context_utf8) || g_utf8_strlen(current_context_utf8, -1) == 0) {
```

I know it has the same effect, but `g_utf8_strlen` is returning a length value, wheres `g_str_has_suffix` is returning a boolean, so the semantic meaning is different.

Note: `g_utf8_strlen(x,-1)` is costly for non-zero length strings. Is this better?

```suggestion
    if (*current_context_utf8 == 0 || !g_str_has_suffix(surrounding_text, current_context_utf8)) {
```

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/7084#discussion_r1001034330_

TheSchlunz:
This changes a zero-length string check to a more efficient way of checking.

Fixes #7541

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -272,7 +272,7 @@ reset_context(IBusEngine *engine) {
       surrounding_text, context_end - context_start, cursor_pos, anchor_pos);
 
     current_context_utf8 = get_current_context_text(context);
-    if (!g_str_has_suffix(surrounding_text, current_context_utf8) || !g_utf8_strlen(current_context_utf8, -1)) {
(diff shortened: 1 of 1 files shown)


keyman-server:

ermshiperete (approved):

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.31-alpha](https://keyman.com/downloads/releases/alpha/17.0.31)

PR #8057, opened 2023-01-20
fix(linux): html tags in package description in keyboard's About page

### LINKED ISSUE #4769: bug(linux): html tags in package description in keyboard's About page
**Describe the bug**

image

The description is shown with these html tags `<p>...</p>` when triggered it from the <kbd>About</kbd> button on the Keyman Configuration.

**Expected behavior**

The tags should not be shown explicitly.

---

**Keyman for Linux:**
 - OS: Ubuntu 20.04.2 LTS -->
 - Keyman version: 14.0.270
 - Target Application: Keyman

---

**Keyboard**
 - Keyboard name: khmer_angkor
 - Keyboard version: 1.0.7
 - Language name: Khmer

TheSchlunz:
This change fixes a bug where html tags were shown in the package description in a keyboard's About page.

Fixes #4769

# User Testing
## Preparations
  - install "khmer_angkor" keyboard 
 
## Tests

**TEST_NO_HTML_TAGS**: no html tags are shown in the package description in keyboard's About page
  - start `km-config`
  - select "Khmer Angkor" keyboard
  - click "About" button
  - verify that no html tags are shown in the package description

### DIFF
--- a/docs/linux/keyman-config.md
+++ b/docs/linux/keyman-config.md
@@ -10,7 +10,7 @@ then you will need to:
 ```bash
 sudo apt install python3-lxml python3-magic python3-numpy python3-qrcode python3-pil \
     python3-requests python3-requests-cache python3 python3-gi gir1.2-webkit2-4.0 dconf-cli \
-    python3-setuptools python3-pip python3-dbus ibus libglib2.0-bin liblocale-gettext-perl
+    python3-setuptools python3-pip python3-dbus ibus libglib2.0-bin liblocale-gettext-perl python3-bs4
 ```
 
 Either `python3-raven` or `python3-sentry-sdk` (>= 1.4) is required as well. On Ubuntu 22.04 and later run:
--- a/linux/build/agent/dependencies.config
+++ b/linux/build/agent/dependencies.config
@@ -20,4 +20,4 @@
 # any=@precise.any
 
 [common]
-any=git autotools-dev build-essential dh-autoreconf libibus-1.0-dev flex bison meson python3 python3-pip cargo xserver-xephyr xvfb metacity libx11-dev gawk libgtk-3-dev libjson-glib-dev
+any=git autotools-dev build-essential dh-autoreconf libibus-1.0-dev flex bison meson python3 python3-bs4 python3-pip cargo xserver-xephyr xvfb metacity libx11-dev gawk libgtk-3-dev libjson-glib-dev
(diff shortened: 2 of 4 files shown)


keyman-server:

ermshiperete (changes requested):
LGTM. The only question I have is if this isn't overkill. Another approach would be to remove everything between `<` and `>`, but I'm torn which way is better.

Anyways, this change introduces a new dependency, so we'll have to update the documentation (`docs/linux/keyman-config.md`) and the files that list dependencies (`linux/build/agent/dependencies.config` and `linux/debian/control`).

BeautifulSoup is available as Ubuntu package python3-bs4.

TheSchlunz:
Removing everything between `<` and `>` is actually what I did first but stuff like this still looked bad because some text was just connected afterwards:
image

ermshiperete (approved):

mcdurdin:
I don't think we can fix the inclusion of HTML tags like this. There is also potential for hyperlinks, images, and other content there.

I think it would be better to consider updates to the .keyboard_info spec to either exclude rich text, or move to Markdown.

ermshiperete:
I just noticed that on Linux we're displaying different text from Windows, so the solution might be to simply change what we display.

mcdurdin:
> I just noticed that on Linux we're displaying different text from Windows, so the solution might be to simply change what we display.

Yes, on Windows, the details from the .keyboard_info / api.keyman.com response are never shown in the keyboard info; only the package description from the .kps is shown. This would be a better solution IMO.

mcdurdin (changes requested):
Change package description to reading data from kmp.json file. (In fact, we should avoid storing the .keyboard_info data, given that is coming from api.keyman.com, which is not available for side-loaded keyboards.)

ermshiperete:
Closing this PR since we're displaying the wrong data in the first place.

ermshiperete:
> Change package description to reading data from kmp.json file. (In fact, we should avoid storing the .keyboard_info data, given that is coming from api.keyman.com, which is not available for side-loaded keyboards.)

@mcdurdin  I'm confused. `kmp.json` doesn't contain anything useful to display for the package description. `khmer_angkor.json` contains a description field, which is what we're already displaying. I can't find the text that Windows displays in the installed keyboard folder. Or are you saying that Windows should do it differently?

mcdurdin:
True, the only thing that Windows shows right now is the "Message" from the .kmx. We plan to merge the .keyboard_info and .kps metadata in 17.0, so let's hold off on these changes for now.

Issue #7895, opened 2022-12-13
chore(common): move to release@semver versioning for sentry reporting


mcdurdin:
Per sentry:

The releases aren't in the SemVer format we are expecting. For example, `release-16.0.117-beta`  isn't something we can parse but if the format had been `release@16.0.117-beta` it would've worked. If you change the formatting for the release, it should work according to your current workflow. For more information on how our releases and SemVer work, click [this link](https://docs.sentry.io/platforms/node/guides/serverless-cloud/configuration/releases/#bind-the-version).

darcywong00:
Is this why Sentry kept reopening issues as regressions?

mcdurdin:
> Is this why Sentry kept reopening issues as regressions?

yep

PR #8063, opened 2023-01-21
chore(common): revise tag variable name to VERSION_GIT_TAG

mcdurdin:
Introduce new $VERSION_GIT_TAG environment variable, and update all references to release@version in all projects to use the new variable.

@keymanapp-test-bot skip

We will need to manually verify that all projects have the correct values. Could I ask the platform owners to verify their platforms please?

- [x] Android: @darcywong00 
- [x] Developer: @mcdurdin 
- [ ] iOS: @sgschantz 
- [x] Linux: @ermshiperete 
- [ ] mac: @sgschantz 
- [x] web: @jahorton 
- [x] Windows: @rc-swag


darcywong00:
I'm investigating why Android fails to build locally.

Can we also echo `$VERSION_GET_TAG` in `build-utils.printVersionUtilsDebug`?

oem/firstvoices/android/app/build.gradle
```diff
@@ -24,6 +24,7 @@ android {
         versionCode VERSION_CODE as Integer
         versionName VERSION_NAME
         buildConfigField "String", "VERSION_ENVIRONMENT", "\""+VERSION_ENVIRONMENT+"\""
+        buildConfigField "String", "VERSION_GIT_TAG", "\""+VERSION_GIT_TAG+"\""
```
darcywong00, line 27: This needs to be done in android/KMAPro/kMAPro/build.gradle too

jahorton (commented):
So, I didn't ask before, but especially given this... I take it that Sentry expects the '@' tagging not only in what the apps themselves use when reporting events, but also in the repo?

I assume there's no interest in maintaining both the older `-`-based tags and the newer `@` tags?

oem/firstvoices/android/app/build.gradle
darcywong00: Though in my debugger, BuildConfig.VERSION_GIT_TAG is coming up null...

android/version.gradle
```diff
@@ -70,6 +70,7 @@ ext {
     VERSION_NAME=getVersionName()
     VERSION_ENVIRONMENT=getVersionEnvironment()
     VERSION_TAG=System.getenv("VERSION_TAG")
+    VERSION_GIT_TAG=System.getenv("VERSION_GIT_TAG")
```
darcywong00, line 73: I think the IDE gradle sync keeps nulling this.

Following the getVersionEnvironment pattern, I had to add
```gradle
def getVersionGitTag = { ->
    String env_git_tag = System.getenv("VERSION_GIT_TAG")
    if (env_git_tag != null) {
        println "Using $env_git_tag from VERSION_GIT_TAG"
        return "$env_git_tag"
    } else {
        // Building probably from IDE
        println "Using release@$VERSION_NAME"
        return "release@$VERSION_NAME"
    }
}
```

and then set
```
VERSION_GIT_TAG=getVersionGitTag()
```

darcywong00:
> I assume there's no interest in maintaining both the older --based tags and the newer @ tags?

I thought we're sticking with @ tags for #7895

ermshiperete (commented):
The Linux part LGTM

jahorton:
> > I assume there's no interest in maintaining both the older --based tags and the newer @ tags?
> 
> I thought we're sticking with @ tags for #7895

Note that before this...

image

We have an incredible amount of older tagged releases using the old, `-` based tags; just check out https://github.com/keymanapp/keyman/tags.  I'm asking if there are any issues worth addressing that may arise from swapping out the character.

oem/firstvoices/android/app/build.gradle
mcdurdin: OK

android/version.gradle
mcdurdin: Gotcha

mcdurdin:
> So, I didn't ask before, but especially given this... I take it that Sentry expects the `@`-tagging not only in what the apps themselves use when reporting events, but also in the repo?

Yes, Sentry links its releases using the git tag `release@[version]`, per the release data sent in the client setup.

> I assume there's no interest in maintaining both the older `-`-based tags and the newer `@` tags?

I don't think so. I couldn't find anywhere where we used the older `-` tags that was particularly significant. It *might* be worthwhile to go through and re-tag all the existing `release-` releases with the corresponding `release@` tags. But I am not convinced of the need to do that.

darcywong00:
Android's working on both apps

android/KMAPro/kMAPro/src/main/java/com/keyman/android/SystemKeyboard.java
```diff
@@ -55,7 +55,7 @@ public void onCreate() {
     if (DependencyUtil.libraryExists(LibraryType.SENTRY) && !Sentry.isEnabled()) {
       Log.d(TAG, "Initializing Sentry");
       SentryAndroid.init(getApplicationContext(), options -> {
-        options.setRelease("release@"+com.tavultesoft.kmapro.BuildConfig.VERSION_NAME);
+        options.setRelease(com.tavultesoft.kmapro.BuildConfig.VERSION_GIT_TAG);
```
darcywong00, line 58: Maybe for another time, but do we want to also pass in these options for oem/firstvoices/android SystemKeyboard?

I see MainActivity already has it.

mcdurdin: yep good catch

developer/src/server/src/version-data.ts
```diff
@@ -42,6 +44,7 @@ export function extractVersionData(version: string): Environment {
     versionTag: versionData[4],
     versionWithTag: version,
     versionEnvironment: versionData[6] || versionData[5] || 'stable',
+    versionGitTag: 'release@'+version,
```
darcywong00, line 47: From the comment in l.15, should this be `+version` or `+versionWithTag`?

mcdurdin: See line 45 -- `version` is equivalent to `versionWithTag` :grin:

ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.35-alpha](https://keyman.com/downloads/releases/alpha/17.0.35)

PR #8067, opened 2023-01-23
chore(windows,developer): Use VsDevCmd.bat as replacement for vcvarsall.bat

SabineSIL:
In `developer/src/kmcmplib/build.bat`  (l.110) we have a comment about `vcvarsall.bat` stating we would need to find a different solution for this.:
```
rem Standalone build, so we'll make the environment available to the caller
rem Also setup
rem Note: Visual Studio 2022 doesn't provide vcvarsall.bat, so we'll have to find a different solution

```
There there is **VsDevCmd.bat** available in VS2022. This might be used instead. 

See comment **https://github.com/keymanapp/keyman/pull/6980#discussion_r1060401824   in feat(developer): cross-platform kmx compiler 🚑 https://github.com/keymanapp/keyman/pull/6980**


mcdurdin:
This was branched from your kmcompx PR I think, which means it has pulled in all the commits on that branch. Can you rebase on master so we get only the vsdevcmd changes?

SabineSIL:
moved to issue

Issue #6189, opened 2022-02-01
chore(linux): refactor linux scripts and extract common parts


ermshiperete:
There are a lot of places where the same logic [i.e. setting `legacy_projects`] is being applied in script. I wonder if it makes sense to have a projects.inc.sh script (this could be left for another day)?

Something along the lines of:

```shell
projects=keyman
legacy_projects=""
if [ -n "$BUILD_LEGACY" ]; then
    legacy_projects="kmflcomp libkmfl ibus-kmfl"
fi
all_projects="$projects $legacy_projects"
```

(but I think it'd be a bit more complex than that)

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/6184#pullrequestreview-868380976_

ermshiperete:
This should also include the list of distributions that should appear only in one place...

ermshiperete:
Superseded by #8102 which removes the legacy projects.

PR #8102, opened 2023-01-26
chore(linux): Remove legacy projects

ermshiperete:
This change removes the KMFL legacy projects. They are still available in the stable-16 branch if we need them.

Also removed some shellcheck warnings.

@keymanapp-test-bot skip


linux/ibus-keyman/src/keymanutil.h
```diff
@@ -3,9 +3,9 @@
 /*
  * Keyman Input Method for IBUS (The Input Bus)
  *
- * Copyright (C) 2018, 2020 SIL International
+ * Copyright (C) 2018-2023 SIL International
  *
- * kmflutil is dual licensed under the MIT or GPL licenses as described below.
+ * keymanutil is dual licensed under the MIT or GPL licenses as described below.
```
darcywong00, line 8: If KMFL is removed, does this mean we only  use the MIT license now?

darcywong00:
Does this address part of #6189?

ermshiperete:
> Does this address part of #6189?

Yes, this makes #6189 obsolete. Thanks for the reminder, I had intended to link that one but then forgot.

mcdurdin (approved):

linux/ibus-keyman/src/keymanutil.h
ermshiperete: No, there are still files under GPL.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.37-alpha](https://keyman.com/downloads/releases/alpha/17.0.37)

PR #8113, opened 2023-01-27
chore(linux): Refactor scripts and packaging 🏗️

ermshiperete:
This change modifies the scripts and the Debian packaging to use the meson build / build.sh script for building instead of autotools.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

mcdurdin:
LGTM but why are the builds failing? :grin:

ermshiperete:
> LGTM but why are the builds failing? grin

There's still a dependency problem. core needs to be built before ibus-keyman can be configured.

linux/debian/rules
```diff
@@ -62,7 +59,7 @@ override_dh_auto_install:
 	install -d $(CURDIR)/debian/tmp
 	cd core && DESTDIR=$(CURDIR)/debian/tmp ./build.sh install
 	# ibus-keyman
-	cd linux/ibus-keyman && make install DESTDIR=$(CURDIR)/debian/tmp
+	cd linux/ibus-keyman && DESTDIR=$(CURDIR)/debian/tmp ./build.sh install
```
mcdurdin, line 62: Is it still appropriate to be setting `DESTDIR` here?

linux/scripts/reconf.sh
```diff
@@ -43,20 +33,17 @@ fi
...
-    cd "$BASEDIR"
-done
-
 for proj in ${extra_projects}; do
     if [ "${proj}" == "keyboardprocessor" ] || [ "${proj}" == "keyman" ]; then
         rm -rf keyboardprocessor
         cp ../VERSION.md ../core/
         ../core/build.sh -t keyboardprocessor configure
+        # we need to build core first before we can configure ibus-keyman
+        ../core/build.sh -t keyboardprocessor build
```
mcdurdin, line 42: Does this mean that we could build keyboardprocessor twice? Looking above it seems that `keyboardprocessor` and `keyman` never end up specified together, so perhaps this is okay, but it seems kinda funny.

mcdurdin (approved):
LGTM

linux/debian/rules
ermshiperete: yes, that's needed so that the files we install end up in a dedicated directory so that the packaging system knows what to put in the package.

ermshiperete:
Don't merge until #8112 is approved!

PR #8115, opened 2023-01-27
chore(linux): Remove po files from ibus-keyman 🏗️

ermshiperete:
We currently don't have any localizable strings in ibus-keyman, so we don't need the po directory.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

mcdurdin (approved):

PR #8116, opened 2023-01-27
chore(linux): Remove obsolete files 🏗️

ermshiperete:
We no longer use these files, so we can remove them. Instead we can use `checkAndInstallRequirements`.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

linux/scripts/jenkins.sh
```diff
@@ -13,6 +13,8 @@ THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BA
 
 . "$KEYMAN_ROOT/resources/shellHelperFunctions.sh"
 
+. "$(dirname "$THIS_SCRIPT")/package-build.inc.sh"
```
mcdurdin, line 16:
```suggestion
. "$THIS_SCRIPT_PATH/package-build.inc.sh"
```

mcdurdin (approved):

PR #8121, opened 2023-01-30
fix(linux): Fix CI build 🩹

ermshiperete:
This fixes the build on Jenkins and TC that got broken by including the wildcard in the quoted string (which causes it to be interpreted literally), introduced in #8102.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.38-alpha](https://keyman.com/downloads/releases/alpha/17.0.38)

PR #8154, opened 2023-02-02
chore(linux): Fix lintian warnings 🍒

ermshiperete:
Also improve upload script to force push changelog branches.

Lintian v2.115 (which is run on Debian servers) still produces pedantic warnings, but there seems to be a bug - no matter what pattern I add to `lintian-overrides`, I can't get it to ignore the files. The following version v2.116 has these rules built in, so we don't need `lintian-overrides` anymore.

(🍒 picked from PR #8153)

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.42-alpha](https://keyman.com/downloads/releases/alpha/17.0.42)

Issue #7533, opened 2022-10-25
feat(core): ldml: support modifier keys 🙀


srl295:
Need to support modifier keys in key lookup

- Spec side: C7043.md still lists `keys.key.mod` as TODO

https://github.com/keymanapp/keyman/issues/7042 🙀

srl295:
There's a developer side to this as well, in supporting modifiers in lookups.

srl295:
in testing https://github.com/keymanapp/keyman/issues/7533 this was found to be already working

srl295:
ok now its done :)

srl295:
- [ ] @mcdurdin  reopening to address https://github.com/keymanapp/keyman/pull/8205#discussion_r1106649499

PR #8159, opened 2023-02-02
feat(common): ldml modifier support 🙀

srl295:
- spec: update spec for layr.mod field
- core/include/ldm: update constants
- bitfield modelled after s_modifier_names[]  / LCTRLFLAG etc.
- new test case k_004_tinyshift, also update k_010_mt
- update const builder to ignore all *_map properties
- split utility out to utils, add test

For: #7533

And not to forget:

@keymanapp-test-bot skip


core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   * Constant for no modifiers
+   */
+  readonly keys_mod_none = 0;
+  /**
+   * bitmask for modifier key
+   */
+  readonly keys_mod_altL = 1 << 2;
+  /**
+   * bitmask for modifier key
+   */
```
mcdurdin, line 258:
```suggestion
  /**
   * bitmask for Left Alt modifier key
   */
  readonly keys_mod_altL = 1 << 2;
  /**
   * bitmask for Right Alt (AltGr) modifier key
   */
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+  readonly keys_mod_altR = 1 << 3;
+  /**
+   * bitmask for either alt
```
mcdurdin, line 261:
```suggestion
   * bitmask for either Alt (Windows) or Option (Apple) modifier key
```

Do we need to have a code for a chirality-agnostic version of this, independently? Haven't thought through all the possibilities here yet.

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_alt  = this.keys_mod_altL | this.keys_mod_altR;
+  /**
+   * bitmask for modifier key
```
mcdurdin, line 265:
```suggestion
   * bitmask for Caps modifier key
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_caps = 1 << 7;
+  /**
+   * bitmask for modifier key
+   */
+  readonly keys_mod_cmd =  1 << 15;
```
mcdurdin, line 271:
```suggestion
   * bitmask for Command (Apple) / Windows (Microsoft) modifier key
   */
  readonly keys_mod_cmd =  1 << 15;
```

Do we need a left/right version of this? (Apple have L/R command keys). Note that we do not propose allowing LDML keyboards to use this key, so perhaps we should be excluding it?

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+  /**
+   * bitmask for modifier key
```
mcdurdin, line 273:
```suggestion
   * bitmask for Left Control modifier key
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_ctrlL = 1 << 0;
+  /**
+   * bitmask for modifier key
```
mcdurdin, line 277:
```suggestion
   * bitmask for Right Control modifier key
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_ctrlR = 1 << 1;
+  /**
+   * bitmask for either ctrl
```
mcdurdin, line 281:
```suggestion
   * bitmask for either Control modifier key
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_ctrl = this.keys_mod_ctrlL | this.keys_mod_ctrlR;
+  /**
+   * bitmask for modifier key
```
mcdurdin, line 285:
```suggestion
   * bitmask for Option modifier key
```

Do we need this? Alt and Option are synonymous.

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_opt = 1 << 14;
+  /**
+   * bitmask for either shift
```
mcdurdin, line 289:
```suggestion
   * bitmask for either Shift modifier key
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_shift = 1 << 4;
+  /**
+   * bitmask for modifier key
```
mcdurdin, line 293:
```suggestion
   * bitmask for Left Shift modifier key
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,80 @@ class Constants {
...
+   */
+  readonly keys_mod_shiftL = this.keys_mod_shift;
+  /**
+   * bitmask for modifier key
```
mcdurdin, line 297:
```suggestion
   * bitmask for Right Shift modifier key
```

core/src/ldml/C7043_ldml.md
```diff
@@ -130,7 +130,25 @@ For each key:
...
+|      0       | `ctrlL`  | `ctrl = ctrlL + ctrlR`                      |
+|      1       | `ctrlR`  |                                             |
+|      2       | `altL`   |                                             |
+|      3       | `altR`   | `alt` (both) = `altL + altR`                |
+|      4       | `shift`  | Either shift                                |
+|      7       | `caps`   |                                             |
+|     15       | `cmd`    | Nonstandard for keyman                      |
+|     14       | `opt`    | Nonstandard for keyman                      |
+|      4       | `shiftL` | alias for `shift`                           |
+|      4       | `shiftR` | alais for `shift`                           |
```
mcdurdin, line 150: Should `shiftL` and `shiftR` ever be permitted? I would argue not -- these keys are traditionally viewed as synonymous, unlike `alt`. By listing them, we encourage keyboard authors to try and use them, and then run into frustration when they are unsupported.

core/src/ldml/C7043_ldml.md
```diff
@@ -130,7 +130,25 @@ For each key:
...
+|--------------|----------|---------------------------------------------|
+| `0x00000000` | `none`   | All zeros = no modifiers                    |
+|      0       | `ctrlL`  | `ctrl = ctrlL + ctrlR`                      |
+|      1       | `ctrlR`  |                                             |
+|      2       | `altL`   |                                             |
+|      3       | `altR`   | `alt` (both) = `altL + altR`                |
+|      4       | `shift`  | Either shift                                |
+|      7       | `caps`   |                                             |
+|     15       | `cmd`    | Nonstandard for keyman                      |
+|     14       | `opt`    | Nonstandard for keyman                      |
```
mcdurdin, line 148: `cmd` should not be included because it is system-reserved on all platforms.
`opt` is synonymous with `alt`, afaik, so should not be a separate bit.

developer/src/kmc-keyboard/src/util/util.ts
```diff
@@ -54,3 +54,25 @@ export function allUsedKeyIdsInLayers(layersList : LKLayers[] | null): Set<strin
...
+export function translateLayerAttrToModifier(layer: LDMLKeyboard.LKLayer) : number {
+  const {id, modifier} = layer;
+  if(id == 'base') {
+    return constants.keys_mod_none;
+  }
+  if (modifier) {
+    let mod = constants.keys_mod_none;
+    for (let str of modifier.split(' ')) {
+      const submod = constants.keys_mod_map.get(str);
+      mod |= submod;
```
mcdurdin, line 72: Can we assume that the input string contains only valid modifier strings?

developer/src/kmc-keyboard/src/util/util.ts
```diff
@@ -54,3 +54,25 @@ export function allUsedKeyIdsInLayers(layersList : LKLayers[] | null): Set<strin
...
+/**
+ * Determine modifier from layer info
+ * @param layer layer obj
+ * @returns modifier
+ */
+export function translateLayerAttrToModifier(layer: LDMLKeyboard.LKLayer) : number {
+  const {id, modifier} = layer;
+  if(id == 'base') {
+    return constants.keys_mod_none;
+  }
```
mcdurdin, line 67: Is `base` a reserved id? Is this documented? Or should we be using only `modifier="none"` (or omitted `modifier` attribute)?

mcdurdin (changes requested):
Looking good; a few changes requested

developer/src/kmc-keyboard/src/util/util.ts
srl295: if not, validation would have failed.

developer/src/kmc-keyboard/src/util/util.ts
srl295: i think it is reserved, but we should probably be using modifier only.

core/src/ldml/C7043_ldml.md
srl295: LDML: https://github.com/unicode-org/cldr/blob/keyboard-preview/docs/ldml/tr35-keyboards.md#element-layer 
has `cmd`, `opt`, and `alt` separately

core/src/ldml/C7043_ldml.md
mcdurdin: image

core/src/ldml/C7043_ldml.md
srl295: proposing to drop these two in https://github.com/unicode-org/cldr/pull/2661#issuecomment-1419772903

core/include/ldml/keyboardprocessor_ldml.ts
srl295: trying to match the CLDR spec where `alt` can be a keyword meaning both.

developer/src/kmc-keyboard/src/util/util.ts
srl295: will drop this code

PR #8171, opened 2023-02-04
feat(common): ldml modifier support 🙀

srl295:
### REDO of #8159 - comments which see

- spec: update spec for layr.mod field
- core/include/ldm: update constants
- bitfield modelled after s_modifier_names[]  / LCTRLFLAG etc.
- new test case k_004_tinyshift, also update k_010_mt
- update const builder to ignore all *_map properties
- split utility out to utils, add test

For: #7533
(cherry picked from commit 0346bea48813314bb9d6163fedf9e7f76250f3b2)

@keymanapp-test-bot skip


srl295:
@mcdurdin PTAL

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -245,6 +245,60 @@ class Constants {
    */
   readonly keys_flags_extend = 1;
 
+  /**
```
mcdurdin, line 248: Note that the `keys_mod_alt` and `keys_mod_ctrl` masks do not currently correspond to the KMX modifier bitmasks in kmx_file.h. Perhaps we should make them match? KMX distinguishes between Alt and L/R Alt -- each gets a separate bit in the mask. This is partly historical (originally kmx did not support L/R Alt).

https://github.com/keymanapp/keyman/blob/ebff42e657fcf8bd9f32a36b1f1d7114903a4cd3/common/include/kmx_file.h#L268-L289

```suggestion
  /* Modifier key bitmasks. These correspond to the modifier bitmasks from kmx_file.h */
  
  /**
```

mcdurdin (approved):

core/include/ldml/keyboardprocessor_ldml.ts
srl295: @mcdurdin  `alt` and `ctrl` as such are needed to support the spec. `alt` should be `0x4 | 0x8 === 0xC`

mcdurdin: Yeah, we can diverge from kmx on the `alt` and `ctrl` modifiers -- just should note that we are doing that.

PR #8180, opened 2023-02-07
fix(linux): Fix autopkgtests

ermshiperete:
Autopkgtests couldn't find pkg-config, so this change adds that as test dependency.

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/control
+++ b/linux/debian/control
@@ -37,8 +37,8 @@ Build-Depends:
  metacity,
  gawk,
 Standards-Version: 4.6.1
-Vcs-Git: https://github.com/keymanapp/keyman.git [linux/debian]
-Vcs-Browser: https://github.com/keymanapp/keyman/tree/master/linux/debian
+Vcs-Git: https://github.com/keymanapp/keyman.git -b stable-16.0 [linux/debian]
+Vcs-Browser: https://github.com/keymanapp/keyman/tree/stable-16.0/linux/debian
 Homepage: https://www.keyman.com
 Rules-Requires-Root: binary-targets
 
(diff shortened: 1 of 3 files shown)


linux/debian/control
```diff
@@ -37,8 +37,8 @@ Build-Depends:
  metacity,
  gawk,
 Standards-Version: 4.6.1
-Vcs-Git: https://github.com/keymanapp/keyman.git [linux/debian]
-Vcs-Browser: https://github.com/keymanapp/keyman/tree/master/linux/debian
+Vcs-Git: https://github.com/keymanapp/keyman.git -b stable-16.0 [linux/debian]
+Vcs-Browser: https://github.com/keymanapp/keyman/tree/stable-16.0/linux/debian
```
darcywong00, line 41: So will this have to change with every release?

ermshiperete: Yes. (That's one reason why I'd prefer if it'd be named `stable` and we would use tags to distinguish the different versions)

mcdurdin: We could consider this change for branch names -- but it would require a fair bit of work on tooling, so I'm not in a hurry to do it!

mcdurdin (approved):

linux/debian/control
mcdurdin: Can you add instructions on this to the release wiki at https://github.com/keymanapp/keyman/wiki/Release-workflow?

ermshiperete: ok, done.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.139](https://keyman.com/downloads/releases/stable/16.0.139)

PR #8181, opened 2023-02-07
fix(linux): Fix autopkgtests 🍒

ermshiperete:
Autopkgtests couldn't find pkg-config, so this change adds that as test dependency.

(🍒 picked from PR #8180 )

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/control
+++ b/linux/debian/control
@@ -37,8 +37,8 @@ Build-Depends:
  metacity,
  gawk,
 Standards-Version: 4.6.1
-Vcs-Git: https://github.com/keymanapp/keyman.git [linux/debian]
-Vcs-Browser: https://github.com/keymanapp/keyman/tree/master/linux/debian
+Vcs-Git: https://github.com/keymanapp/keyman.git -b stable-16.0 [linux/debian]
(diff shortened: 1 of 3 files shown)


linux/debian/control
```diff
@@ -37,8 +37,8 @@ Build-Depends:
  metacity,
  gawk,
 Standards-Version: 4.6.1
-Vcs-Git: https://github.com/keymanapp/keyman.git [linux/debian]
-Vcs-Browser: https://github.com/keymanapp/keyman/tree/master/linux/debian
+Vcs-Git: https://github.com/keymanapp/keyman.git -b stable-16.0 [linux/debian]
+Vcs-Browser: https://github.com/keymanapp/keyman/tree/stable-16.0/linux/debian
```
darcywong00, line 41: Should these stay master?

ermshiperete: I think it doesn't matter. We don't use this information anywhere, it's just used on Debian with their packaging infrastructure. Keeping the file content the same on stable and master simplifies cherry-picking IMO.

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.45-alpha](https://keyman.com/downloads/releases/alpha/17.0.45)

PR #8188, opened 2023-02-08
chore(common): builder internal dependencies

mcdurdin:
Fixes the scenario where `builder_describe_outputs` is called after `builder_parse`, so that internal dependencies will be built as required.

Note: `--force-deps` will also now rebuild internal dependencies.

@keymanapp-test-bot skip


resources/builder.inc.sh
```diff
@@ -698,24 +711,30 @@ _builder_add_chosen_action_target_dependencies() {
...
       local dep_output=${_builder_internal_dep[$action_target]}
-      if [[ ! -z ${_builder_dep_path[$dep_output]+x} ]] &&
-          [[ ! -e "$KEYMAN_ROOT/${_builder_dep_path[$dep_output]}" ]]; then
-        if ! _builder_item_in_array "$dep_output" "${_builder_chosen_action_targets[@]}"; then
-          _builder_chosen_action_targets+=($dep_output)
-          new_actions+=($dep_output)
+      # If there is a defined output for this dependency
+      if [[ ! -z ${_builder_dep_path[$dep_output]+x} ]]; then
+        # If the output for the dependency is missing, or we have --force-deps
+        if [[ ! -e "$KEYMAN_ROOT/${_builder_dep_path[$dep_output]}" ]] || builder_is_full_dep_build; then
```
mcdurdin, line 720: This change allows for internal deps to be rebuilt with --force-deps

resources/builder.inc.sh
```diff
@@ -698,24 +711,30 @@ _builder_add_chosen_action_target_dependencies() {
...
         fi
       fi
     fi
     i=$((i + 1))
   done
 
   if [[ ${#new_actions[@]} -gt 0 ]]; then
-    echo "Automatically running following required actions with missing outputs:"
+    if builder_is_full_dep_build; then
+      echo "Automatically running all dependent actions due to --force-deps:"
```
mcdurdin, line 734:
```suggestion
      echo "Automatically running all dependency actions due to --force-deps:"
```

resources/builder.inc.sh
```diff
@@ -1126,7 +1151,7 @@ builder_is_dep_build() {
 # `--deps` parameter (which is the default).
 #
 builder_is_quick_dep_build() {
-  if builder_is_dep_build && [[ $_builder_build_deps == --deps ]]; then
+  if [[ $_builder_build_deps == --deps ]]; then
```
mcdurdin, line 1154: This function was never used so I thought it safe to tweak like this.

resources/builder.inc.sh
```diff
@@ -1137,7 +1162,7 @@ builder_is_quick_dep_build() {
 # corresponds to the `--force-deps`` parameter.
 #
 builder_is_full_dep_build() {
-  if builder_is_dep_build && [[ $_builder_build_deps == --force-deps ]]; then
+  if [[ $_builder_build_deps == --force-deps ]]; then
```
mcdurdin, line 1165: Only used internally and builder_is_dep_build was already checked anyway

resources/builder.inc.sh
ermshiperete: I think "dependency actions" is easier to understand than "dependent actions"

ermshiperete (approved):
LGTM, although I'm not sure how the tests are run...

mcdurdin:
> LGTM, although I'm not sure how the tests are run...

Currently, these tests are only run manually. I am planning to add a "Common Tests" build configuration soon.

PR #8191, opened 2023-02-08
chore(linux): Update changelog from Debian

ermshiperete:
This overlaps with PR #8180, but it updates the changelog to a patched version on Debian, so we still need this change.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 16.0.139](https://keyman.com/downloads/releases/stable/16.0.139)

PR #8197, opened 2023-02-09
chore(common): automatically apply _builder_check_color

mcdurdin:
`builder_check_color()` is now an internal function, called automatically, and should never need to be called from a `build.sh` script. All scripts updated accordingly.

@keymanapp-test-bot skip

### DIFF
--- a/resources/builder.inc.sh
+++ b/resources/builder.inc.sh
@@ -19,7 +19,7 @@ function _builder_init() {
   _builder_findRepoRoot
   _builder_setBuildScriptIdentifiers
 
-  if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; then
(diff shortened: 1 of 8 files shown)


resources/builder.inc.sh
```diff
@@ -19,7 +19,7 @@ function _builder_init() {
   _builder_findRepoRoot
   _builder_setBuildScriptIdentifiers
 
-  if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]]; then
+  if [[ -n "$TERM" ]] && [[ "$TERM" != "dumb" ]] && [[ "$TERM" != "unknown" ]] && [ -t 1 ]; then
```
mcdurdin, line 22: Improves automatic check of redirection for color in build scripts.

jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.48-alpha](https://keyman.com/downloads/releases/alpha/17.0.48)

Issue #7985, opened 2023-01-06
feat(core): ldml C++ side for all new sections 🙀


srl295:
see #7630  (#7532)

- list subtables
- key2 subtables

https://github.com/keymanapp/keyman/issues/7042

srl295:
will support #7535

srl295:
done for now

PR #8206, opened 2023-02-10
feat(core): ldml C++ validate list / key2  🙀

srl295:
- need helper objs still

For: #7985

@keymanapp-test-bot skip

### DIFF
--- a/core/src/kmx/kmx_plus.cpp
+++ b/core/src/kmx/kmx_plus.cpp
@@ -575,15 +575,27 @@ COMP_KMXPLUS_LAYR_Helper::getKey(KMX_DWORD key) const {
 
 bool
 COMP_KMXPLUS_KEY2::valid(KMX_DWORD _kmn_unused(length)) const {
-  // TODO-LDML more to do here
-  DebugLog("TODO-LDML: key2");
+  if (header.size < sizeof(*this)
+      + (keyCount  * sizeof(COMP_KMXPLUS_KEY2_KEY))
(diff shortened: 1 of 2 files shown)


rc-swag:
Late to the party here but I am not following what `key2` is from the spec.

srl295:
“keys” is the keybag minimized / optimized for keyman. “key2” is a
different structure that contains all of the attributes. For example ids
and such.

core/src/kmx/kmx_plus.cpp
```diff
@@ -575,15 +575,27 @@ COMP_KMXPLUS_LAYR_Helper::getKey(KMX_DWORD key) const {
...
+  // TODO-LDML: further validation in the COMP_KMXPLUS_KEY2_Helper helper obj
   return true;
 }
 
 bool
 COMP_KMXPLUS_LIST::valid(KMX_DWORD _kmn_unused(length)) const {
-  // TODO-LDML more to do here
-  DebugLog("TODO-LDML: list");
+  if (header.size <
+      sizeof(*this) + (sizeof(COMP_KMXPLUS_LIST_ITEM) * listCount) + (sizeof(COMP_KMXPLUS_LIST_INDEX) * indexCount)) {
```
mcdurdin, line 593: Can you make this test match the style of lines 578-582?

mcdurdin (approved):

core/src/kmx/kmx_plus.cpp

PR #8209, opened 2023-02-10
fix(linux): Fix path to symbols file in packaging GHA

ermshiperete:
@keymanapp-test-bot skip

PR #8202, opened 2023-02-09
docs(linux): Update packaging documentation

ermshiperete:
- remove obsolete `packageName` variable
- add section on how to build with Docker, similar to how the package build is done on GHA
- updated "uploading to Debian" section
- some reformatting

@keymanapp-test-bot skip

PR #8210, opened 2023-02-10
chore(common): Report status if no platform needs build

ermshiperete:
This works around a problem where status.keyman.com keeps a PR in the "Waiting for build" status because no status messages appear in the PR because no files changed that affect any platform (e.g. changes in GHA or [documentation](https://github.com/keymanapp/keyman/pull/8202)) which causes no platform builds to be triggered.

**Edit:** #8209 is no longer an example where this happens because I had to change additional files that triggered a build.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

mcdurdin:
:+1:

This also helps with more explicit status check messages on the PR itself, so I like.

PR #8211, opened 2023-02-10
chore: ldml merge from master 🙀

srl295:
Epic: #7054 

@keymanapp-test-bot skip


core/src/meson.build
```diff
@@ -58,7 +58,7 @@ if compiler.get_id() == 'emscripten'
   links = []
 endif
 
-lib = library('kmnkbp0',
+kmx_files = files(
```
srl295, line 61: @ermshiperete did I get #8211 right here?

ermshiperete: Looks correct.

ermshiperete (approved):

Issue #7535, opened 2022-10-25
feat(developer): ldml: emit test data 🙀


srl295:
- emit test data from kmc

https://github.com/keymanapp/keyman/issues/7042 🙀

srl295:
room to improve, but think this issue is now done

PR #8212, opened 2023-02-10
feat(common): ldml read keyboard testdata 🙀

srl295:
- early scaffolding
- with some basic tests

#7535

@keymanapp-test-bot skip

PR #8219, opened 2023-02-13
feat(developer): ldml emit test data from kmc, integrate into core 🙀

srl295:
- developer: add a `-T` option that reads/emits test json instead of kmx
- core: add a mechanism for generating .json test data for xml source files that have them - not used yet.

for #7535 

@keymanapp-test-bot skip


srl295:
I left `@@keys` and `@@expected` in the .xml source for now so i think this is redy for review even though the test data isn't 'executed' yet.

i'm planning on using the JSON parser that's already in core (!) to read this data on C++

srl295:
though this doesn't actually execute the tests, i am marking this ready to review because it does have important scaffolding.

srl295:
Merging into #8212

PR #8247, opened 2023-02-17
fix(linux): Checkout before  running apt-install

ermshiperete:
We need to check out the source code before we can run the apt-install action in the packaging GHA.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #8261, opened 2023-02-20
chore: consolidate script helper functions

mcdurdin:
@keymanapp-test-bot skip

Refactors the following helper functions in shellHelperFunctions.sh:

* `log_error` --> remove, unused
* `log_warning` --> `builder_warn`
* `echo_heading` --> `builder_heading`
* `fail` --> `builder_die`
* `die` --> `builder_die`
* `warn` --> `builder_warn`

It may be easier to review this PR commit-by-commit.

### DIFF
--- a/android/KMAPro/build-play-store-notes.sh
+++ b/android/KMAPro/build-play-store-notes.sh
@@ -35,7 +35,7 @@ DEFAULT_RELEASE_NOTE="* Additional bug fixes and improvements"
 FILTERED_LINES=$( grep '^\s*\*.*$' "$KEYMAN_ROOT/android/help/about/whatsnew.md" || [[ $? == 1 ]] ) # Continue if grep has no matches
 if [ -z "$FILTERED_LINES" ]; then
   FILTERED_LINES="$DEFAULT_RELEASE_NOTE"
-  warn "Warning: whatsnew.md empty so using default release note: '$FILTERED_LINES'"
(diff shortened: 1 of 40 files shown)


darcywong00 (approved):
lgtm

jahorton (approved):
The main concern I have here at the moment is that this is propagating `builder_` functions to non-builder-based scripts.

The color setting, in particular, is configured with the expectation of builder-style parameters... which are likely to be unsupported by such non-builder-based scripts.  Granted, this _is_ just a color thing, but might this have any ramifications for our CI builds or other parts of our tooling?  Just wondering, since we do have explicit logic and handling to ensure we can disable color-coding as part of `builder`.

Granted, a lot of the functions that are being merged into the newer `builder_` functions never did have an off-switch for their color... so maybe that's not actually an issue, then.  On further inspection... yeah, all of the previously-used functions never did have an "off".

----

So... tl;dr:  LGTM, I guess.  Still, I wanted to voice the concerns in case this might highlight a possible issue for not-yet-builder-based scripts.

mcdurdin:
> So... tl;dr: LGTM, I guess. Still, I wanted to voice the concerns in case this might highlight a possible issue for not-yet-builder-based scripts.

All our scripts reference build-utils.sh, even if they are not yet builder-based, so it should be fine. Also, the color variables have all been defined in builder.inc.sh for some time, so this just completes that consolidation :grin:

keyman-server:

PR #8263, opened 2023-02-20
chore(developer): ignore Esc key when compiling

mcdurdin:
Long ago, ESC was a useful way to stop the compiler. These days, it's so fast that it's hard to break it that way anyway. It's also watching in the background, so that's bad... because if you have keyboard compiles happening in the background and hit Esc at the wrong time, you cancel the build. So... delete it!

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #8265, opened 2023-02-20
feat(developer): touch layout compiler for LDML 🙀

mcdurdin:
@keymanapp-test-bot skip


developer/src/kmc-keyboard/src/compiler/touch-layout-compiler.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    for(let row of layer.row) {
+      y++;
+
+      let fileRow: TouchLayout.TouchLayoutRow = {id: y, key: []};
+      fileLayer.row.push(fileRow);
+
+      const keys = row.keys.split(' ');
+      for(let key of keys) {
+        const keydef = source.keyboard.keys?.key?.find(x => x.id == key);
```
srl295, line 44: need to support 'not found' unless that's handled in validation

developer/src/kmc-keyboard/src/compiler/touch-layout-compiler.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        const fileKey: TouchLayout.TouchLayoutKey = {
+          id: this.translateKeyIdentifierToTouch(keydef.id) as TouchLayout.TouchLayoutKeyId,
+          text: keydef.to,
```
srl295, line 48: gap / switch keys won't have a `to`

developer/src/kmc-keyboard/src/compiler/touch-layout-compiler.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const map = {
+      none:         'default',
+      shift:        'shift',
+      caps:         'caps',
+      altR:         'rightalt',
+      "altR shift": 'rightalt-shift',
+      "shift altR": 'rightalt-shift', // TODO-LDML is this required?
+    };
+
```
srl295, line 82:
```suggestion
  modifier = (modifier||'').split(/\b/).sort().join(' ').trim();
```

developer/src/kmc-keyboard/src/compiler/touch-layout-compiler.ts
```diff
@@ -0,0 +1,104 @@
...
+    if(Object.hasOwn(map, modifier)) {
+      return (map as any)[modifier];
+    }
+
+    // TODO-LDML: Other layer names will be used unmodified, is this sufficient?
+    return id;
+  }
+
+  private translateKeyIdentifierToTouch(id: string): string {
```
srl295, line 91: shouldn't this be matching on the *text* not assuming the id?

so if we have `<row keys="a z e r t y"/>` the first key would get `K_A` which is fine assuming `<key id='a' to='a'/>` but not fine otherwise.

developer/src/kmc-keyboard/src/compiler/visual-keyboard-compiler.ts
```diff
@@ -30,7 +30,7 @@ export default class VisualKeyboardCompiler {
     vk: VisualKeyboard.VisualKeyboard,
     layer: LDMLKeyboard.LKLayer
   ) {
-    // TODO: consider consolidation with keys.ts?
+    // TODO-LDML: consider consolidation with keys.ts?
```
srl295, line 33: yes

srl295 (approved):
looking good so far, some suggestions though

developer/src/kmc-keyboard/src/compiler/touch-layout-compiler.ts
mcdurdin: The principle here was to re-use the 0-9, A-Z, a-z identifiers, just for coherence. We could give everything a `T_` prefix, but there is some value in matching these standard identifiers, for example making digit keys visible to the Keyman app.

if `<key id='a' to='foo'/>` is defined, then `K_A` would match to that key, so I think it still works.

Issue #4273, opened 2021-01-15
feat(linux): Add support for Wayland


ermshiperete:
Currently Keyman doesn't work properly when Wayland instead of X11 is used (see #4122).

ermshiperete:
This happens on Debian Buster 10.7. On Ubuntu 20.04 things work in gedit, but not in gnome-terminal.

To see whether Wayland or X11 is used, run: 

```
loginctl show-session $(awk '/tty/ {print $1}' <(loginctl)) -p Type | awk -F= '{print $2}'
```

To allow Wayland login on Ubuntu 20.04, edit `/etc/gdm3/custom.conf` and set `WaylandEnable` to true (or comment the line). Then restart gdm3: `sudo systemctl restart gdm3`.

ermshiperete:
A little more research shows that Keyman 14.0 kind of works with Wayland: it doesn't survive an ibus restart, which means after installing it's necessary to reboot (or maybe logout/login would be sufficient). 

But even then things don't work properly: With Eurolatin, typing `n` + `>` produces `nŋ`, and with Khmer Angkor typing xEjmr doesn't produce the desired results.

I was testing this in a VM, and I just noticed that my host machine behaves strangely, so that might have an impact...

Note that Debian Buster doesn't have neither `python3-raven` nor `python3-sentry-sdk`, so this will have to be manually downloaded and installed from https://packages.debian.org/bullseye/all/python3-sentry-sdk.

darcywong00:
Is this a pre-requisite for releasing packages on 21.04?  See https://community.software.sil.org/t/how-does-one-go-about-installing-keyman-14-on-ubuntu-21-04/4728

bizulk:
Hello /

Is there a the roadmap for wayland integration ?
By default Ubuntu Linux 20 is using wayland and many embedded Linux devices.
Thanks for answer.

ermshiperete:
> Is there a the roadmap for wayland integration ?

Yes, if everything goes as planned Keyman will work with Wayland in the upcoming version 15. We're currently working on it. In fact, most things already seem to work when using Wayland, but there seems to be a bug in ibus that prevents surrounding text from working properly which results in characters not being replaced.

ermshiperete:
On Ubuntu 21.10 Impish `ibus-keyman` crashes when running under Wayland when trying to switch to a Keyman keyboard.

ermshiperete:
See also #6172.

[...]

PR #8274, opened 2023-02-20
fix(linux): Collect consecutive backspaces

ermshiperete:
When using Wayland `im-wayland.so` (which is part of libgtk) gets loaded into the client app instead of `im-ibus.so`. This means that our ordered output doesn't work with Wayland.

This change works around a problem with keyboards that use multiple consecutive backspace actions. Because method calls get processed asynchronously some backspaces get lost. In clients that support surrounding text we can count the consecutive backspaces and then do one call to delete surrounding text with the appropriate number of characters.

This can be tested with "Khmer Angkor" keyboard by typing `xEjmr`, or with "Vedic Sanskrit Devanagari Phonetic (ITRANS)" keyboard by typing `shrI`.

Part of #4273. See https://github.com/keymanapp/keyman/issues/4273#issuecomment-1241703095 in particular.

@keymanapp-test-bot skip

This should eventually be tested, but since other aspects are not yet working with Wayland, testing doesn't make much sense yet.

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -99,6 +99,7 @@ struct _IBusKeymanEngine {
 
   commit_queue_item commit_queue[MAX_QUEUE_SIZE];
   commit_queue_item *commit_item;
+  guint             consecutive_backspaces;
 };
 
 struct _IBusKeymanEngineClass {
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.53-alpha](https://keyman.com/downloads/releases/alpha/17.0.53)

PR #8281, opened 2023-02-21
chore(linux): Improve deb-packaging GHA

ermshiperete:
This change eliminates the need to checkout the Keyman source repo on the self-hosted runner. The checkout previously timed out, and the only reason we need it was the `apt-install` action. So now we directly call `apt install` instead of using the action.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.53-alpha](https://keyman.com/downloads/releases/alpha/17.0.53)

PR #8289, opened 2023-02-22
chore: use builtin instead of dirname for perf

mcdurdin:
@keymanapp-test-bot skip

Note, I have not replaced all uses of `dirname`, just those where we can be confident that it won't have side-effects (such as `..` in a path or a missing `/`).


ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.55-alpha](https://keyman.com/downloads/releases/alpha/17.0.55)

PR #8303, opened 2023-02-23
chore(core): make kmx_test_source available as a library

mcdurdin:
Linux builds require kmx_test_source.cpp in order to run tests. So make this available even when other tests are disabled.

@keymanapp-test-bot skip


core/tests/unit/meson.build
```diff
@@ -1,5 +1,4 @@
-
-node = find_program('node', required: false)
+node = find_program('node', required: true)
```
mcdurdin, line 1: Unrelated but should be there...

darcywong00 (approved):
lgtm

PR #8318, opened 2023-02-25
spec(core): ldml update spec for either shift/ctrl/alt  🙀

srl295:
- match Keyman constanats

#7533

> _I know now what is bothering me about defining "either ctrl" == "both ctrl". When we define a bitmask such as LDML_KEYS_MOD_CTRL | LDML_KEYS_MOD_LSHIFT, we are not expecting the user to be pressing LCtrl + RCtrl + LShift, but rather either Ctrl + LShift. This makes bitmask comparisons less than trivial because we are always going to have to account for these. I do wonder if it would be better to go the 'keyman way' and have separate bitflags for non-chiral ctrl, non-chiral alt (even though that was an accident of history)?_

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM

PR #8320, opened 2023-02-26
feat(developer): kmc .kpj support 🙀

mcdurdin:
Adds support for building .kpj projects. Required minor tweak to kmcomp to allow it to build both .kmx and .js from .kmn command line. Currently, calls kmcomp.exe to build .kmn files, until we support a cross-platform kmc-kmn C++ project.

@keymanapp-test-bot skip

### DIFF
--- a/common/web/types/src/util/compiler-interfaces.ts
+++ b/common/web/types/src/util/compiler-interfaces.ts
@@ -42,6 +42,8 @@ export interface CompilerCallbacks {
   loadLdmlKeyboardTestSchema(): Buffer;
   reportMessage(event: CompilerEvent): void;
   loadKvksJsonSchema(): Buffer;
+  loadKpjJsonSchema(): Buffer;
+  forceDirectories(dir: string): void;
 };
 
 /**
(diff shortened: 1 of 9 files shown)


common/web/types/src/util/compiler-interfaces.ts
```diff
@@ -42,6 +42,8 @@ export interface CompilerCallbacks {
   loadLdmlKeyboardTestSchema(): Buffer;
   reportMessage(event: CompilerEvent): void;
   loadKvksJsonSchema(): Buffer;
+  loadKpjJsonSchema(): Buffer;
+  forceDirectories(dir: string): void;
```
darcywong00, line 46: From only the name, I was unclear that  `forceDirectories` creates directories. What about something like `forceCreateDirectories`?

mcdurdin: I'm going to pull this back out of here, because it's used by kmc, which has fs access anyway.

mcdurdin: This will be done in a follow-up PR

darcywong00 (approved):
lgtm

Issue #2761, opened 2020-02-29
feat(developer): Use Folders instead of Projects


mcdurdin:
**Is your feature request related to a problem? Please describe.**

There is very little unique content stored in a .kpj file. We encourage a directory structure already. Why not formalize this and move away from projects and instead open a 'folder'. Then populate the 'project' view with the contents of the folder.

Proposed by @MattGyverLee:
> The package (.pkg) solves a real issue of cherry-picking files/keyboards for the release. The Keyman (.kmn) file connects all the keyboard bits (OSK, Touch, Icon) to the keyboard.
>
> What does the .prj file really do for us other than duplicate data from other sources and be out-of-date when you make changes with Developer closed? It seems like we could just open the folder, let it auto-populate folder contents on open, and the contents of the .prj file could be reduced to a “.env” file. 
>
> Is there a use-case for people keeping data (keyboards) from different root directories in a project? I can, of course, see a use case for different roots in a package.

Commentary by Marc:

This is a good suggestion. The .kpj.user currently maintains state -- but it can always be thrown away and is easily transitioned to a folder concept. The .kpj served more of a purpose in the past. We've really come full-circle on how the projects interact with keyboards and packages. This seems like it would be a good future feature.

I'm not sure how we'd transition from .kpj to .env (or whatever it is called). We can't assume that existing files are structured by folder. Perhaps it could be a test when loading the project and if the project meets certain structural requirements offer to upgrade it (and offer reasons why not if it can't be upgraded).

Considerations:
* Project Options remain important: these would need to be stored in a .env?
* build Folder can be a bit confusing -- needs to be cleanly ignored in the project view (shaded down similar to .gitignored files in VS Code, perhaps).
* multi-keyboard packages (not super recommended anyway) in the existing repository reference files from outside the folder. These are the biggest use case for projects.
* This would mean two sets of 'project' style code to maintain in Developer. Some re-abstraction would be necessary to make this maintainable.

mcdurdin:
More commentary on the issue:

Marc, I spent half an hour yesterday helping someone find the keyboard they were working on.
The *.prj was on google drive, but it was linked to a .kmn in My documents and now showing a default touch layout. Once I moved the KMN, osk and touch to the folder, I had to rename the files to match and "remove" the OSK and Touch from the kmn config and re-add them before it found and showed them. This would be very scary if i didn't know that "remove" doesn't delete.

[...]

PR #8326, opened 2023-02-27
feat(developer): Add support for building a folder to kmc 🙀

mcdurdin:
Relates to #2761.

kmc now supports passing in a folder and building all Keyman source files found in the folder, even if a .kpj is not present in the folder.

If a .kpj file exists, `KeymanDeveloperProject/Options/Version` will be checked to determine behaviour. If this field is missing or "1.0", then the classical behaviour persists.

If the Version field is "2.0", the `Files` key is ignored, and .kmn, .xml, .model.ts, and .kps files under `sourcePath` are enumerated and added to the in-memory project for build.

If a .kpj is not present in the folder, a virtual project is constructed in memory with the following defaults:

* `version`: `2.0`
* `buildPath`: `$PROJECTPATH/build`
* `sourcePath`: `$PROJECTPATH/source`
* `compilerWarningsAsErrors`: `false`
* `warnDeprecatedCode`: `true`
* `checkFilenameConventions`: `true`

It is still considered an error to include both a keyboard and a model in the same file, but a version 2.0 project will not include the field determining its type in `Options/ProjectType`; this will instead be checked during build.

The call to `kmc` can reference a folder or a .kpj file. The .kpj file must have the same basename as the folder name (e.g. `khmer_angkor/khmer_angkor.kpj`).

WARNING: this does not add support for 2.0 projects to Keyman Developer; it is likely that opening a version 2.0 project in Developer will currently corrupt it.

WARNING: it is not anticipated that kmcomp.exe will be updated to support version 2.0 projects -- rather kmcomp.exe will be removed before release of v17, and replaced with kmc.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #8329, opened 2023-02-27
fix(linux): Fix ibus-keyman/build.sh

ermshiperete:
Paths that describe outputs are supposed to be relative to root, not absolute. Also, it turns out we don't use the `--output` parameter, so this change removes it.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.58-alpha](https://keyman.com/downloads/releases/alpha/17.0.58)

PR #8338, opened 2023-02-28
chore(linux): Use dependency on core in ibus-keyman/build.sh

ermshiperete:
@keymanapp-test-bot skip

### DIFF
--- a/core/build.sh
+++ b/core/build.sh
@@ -102,7 +102,7 @@ builder_describe_outputs \
   configure:wasm     build/wasm/$CONFIGURATION/build.ninja \
   build:x86          build/x86/$CONFIGURATION/src/libkmnkbp0.a \
   build:x64          build/x64/$CONFIGURATION/src/libkmnkbp0.a \
-  build:arch         build/arch/$CONFIGURATION/src/libkmnkbp0.a \
+  build:arch         build/arch/$CONFIGURATION/src/libkmnkbp0-static.a \
   build:wasm         build/wasm/$CONFIGURATION/src/libkmnkbp0.a
 
 # Target path is used by Linux build, e.g. --target-path keyboardprocessor
--- a/linux/ibus-keyman/build.sh
+++ b/linux/ibus-keyman/build.sh
@@ -19,10 +19,8 @@ builder_describe \
   "test" \
   "install                   install artifacts" \
   "uninstall                 uninstall artifacts" \
-  "--debug,-d                Debug build"
-# We can't yet depend on core until it moved to the new build.sh syntax
-# (currently it doesn't know some parameters that we're passing)
-#  "@/core configure build"
+  "--debug,-d                Debug build" \
+  "@/core :arch"
 
 builder_parse "$@"
 
--- a/docs/settings/linux/tasks.json
+++ b/docs/settings/linux/tasks.json
@@ -3,16 +3,39 @@
   "tasks": [
     {
       "type": "shell",
+      "label": "core: clean",
+      "command": "${workspaceFolder}/core/build.sh",
+      "args": [
+        "--debug",
+        "clean"
+      ],
+      "group": "build",
+      "detail": "clean core"
+    },
+    {
+      "type": "shell",
+      "label": "core: configure",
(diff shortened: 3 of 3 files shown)


ermshiperete:
There's still a problem with dependency handling: after runing `core/build.sh clean` and `linux/ibus-keyman/build.sh clean`, I run `linux/ibus-keyman/build.sh build` and get:

```
[linux/ibus-keyman] build.sh launched with: <build>
Automatically running following required actions with missing outputs:
* configure:engine
## [linux/ibus-keyman] configure starting...
[linux/ibus-keyman] Skipping dependency core for configure
The Meson build system
Version: 0.61.2
Source dir: $KEYMANROOT/linux/ibus-keyman
Build dir: $KEYMANROOT/linux/build/x86_64/release
Build type: native build
Project name: ibus-keyman
Project version: 17.0.58
C compiler for the host machine: ccache cc (gcc 11.3.0 "cc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0")
C linker for the host machine: cc ld.bfd 2.38
C++ compiler for the host machine: ccache c++ (gcc 11.3.0 "c++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0")
C++ linker for the host machine: c++ ld.bfd 2.38
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Run-time dependency ibus-1.0 found: YES 1.5.26
Run-time dependency gtk+-3.0 found: YES 3.24.33
Run-time dependency x11 found: YES 1.7.5
Run-time dependency json-glib-1.0 found: YES 1.6.6
Library libkmnkbp0 found: YES
Program env found: YES (/usr/bin/env)
Header <ibus.h> has symbol "IBUS_CAP_PREFILTER" with dependency ibus-1.0: YES 
Configuring config.h using configuration

meson.build:36:0: ERROR: Include dir ../../core/build/arch/release/include does not exist.

A full log can be found at $KEYMANROOT/linux/build/x86_64/release/meson-logs/meson-log.txt
## [linux/ibus-keyman] configure failed
```

core/build.sh
```diff
@@ -102,7 +102,7 @@ builder_describe_outputs \
   configure:wasm     build/wasm/$CONFIGURATION/build.ninja \
   build:x86          build/x86/$CONFIGURATION/src/libkmnkbp0.a \
   build:x64          build/x64/$CONFIGURATION/src/libkmnkbp0.a \
-  build:arch         build/arch/$CONFIGURATION/src/libkmnkbp0.a \
+  build:arch         build/arch/$CONFIGURATION/src/libkmnkbp0-static.a \
```
ermshiperete, line 105: The dynamic library is `libkmnkbp0.so`. I don't know if this change is valid for Mac?

mcdurdin: We are about to split the linux and mac builds here so we can resolve this separately.

linux/ibus-keyman/build.sh
```diff
@@ -19,10 +19,8 @@ builder_describe \
   "test" \
   "install                   install artifacts" \
   "uninstall                 uninstall artifacts" \
-  "--debug,-d                Debug build"
-# We can't yet depend on core until it moved to the new build.sh syntax
-# (currently it doesn't know some parameters that we're passing)
-#  "@/core configure build"
+  "--debug,-d                Debug build" \
+  "@/core :arch"
```
mcdurdin, line 23: This won't work the way you are expecting -- we can't specify dependency targets yet, so this would attempt to build core if you specified `:arch` on ibus-keyman, which wouldn't work. I will think this through...

mcdurdin (changes requested):

core/build.sh
mcdurdin: I think I will aim to produce both static and dynamic libraries for Core for all architectures. We already do this for Windows builds, and that will resolve this inconsistency everywhere.

ermshiperete:
Abandoning this PR since our current build system doesn't yet support what we need, i.e. limiting the dependency build to a target.

PR #8341, opened 2023-03-01
fix(common): builder: skip dependency build for clean action

mcdurdin:
We should not attempt to build dependencies when running `clean`. Note: if your `clean` action requires dependencies, then something is very, very wrong and you should fix that!

This should resolve the failed Keyman Developer build on master.

@keymanapp-test-bot skip


sgschantz (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.59-alpha](https://keyman.com/downloads/releases/alpha/17.0.59)

PR #8342, opened 2023-03-01
feat(core): Build both architectures for mac and generate a 'fat' library for them

mcdurdin:
@keymanapp-test-bot skip


mcdurdin:
* Test build on M1: https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_macOS/369883?hideProblemsFromDependencies=false&hideTestsFromDependencies=false
* Test build on Intel: https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_macOS/369870?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

mcdurdin:
Jenkins packaging fails with:

```
dh_missing: warning: usr/lib/x86_64-linux-gnu/libkmnkbp0.a exists in debian/tmp but is not installed to anywhere 
```

@ermshiperete?

core/build.sh
```diff
@@ -38,16 +38,22 @@ case $BUILDER_OS in
       ":x64    64-bit Windows (x64) build"
     )
     ;;
-  mac|linux)
+  mac)
     archtargets+=(
-      ":arch   Linux or mac build -- current architecture"
+      ":mac           Mac all architectures fat library build"
+      ":mac-x86_64    Mac Intel build"
+      ":mac-arm64     Mac arm64 (M1) build"
```
darcywong00, line 45: Could you elaborate why mac couldn't have 
```
  archtargets+=(
    ":x86    32-bit Mac (x86) build"
    ":x64    64-bit Mac (x64) build"
   ) 
```

Or will it mean later on there will also be
```
":linux-x86_64   "
":linux-arm64    "
```

mcdurdin: Mac architectures are x86_64 (Intel) and arm64 (M1) -- not x86 (Intel) and x64 (Intel -- technically AMD/Intel). Mac needs a single static library that contains both architectures, and the linker chooses the appropriate architecture to link in.

I don't think we need the same breakdown on Linux at present -- each architecture is built separately and there is no need to provide two architectures in the single distributable.

darcywong00 (approved):
lgtm

sgschantz (approved):
looks good!

mcdurdin:
> Jenkins packaging fails with:
> 
> ```
> dh_missing: warning: usr/lib/x86_64-linux-gnu/libkmnkbp0.a exists in debian/tmp but is not installed to anywhere 
> ```

I've opted to make the 'build both library types' setting apply only to the mac build for now. We can add the change to linux as a separate PR later if we want it, which is better process anyway.

ermshiperete:
> Jenkins packaging fails with:
> 
> ```
> dh_missing: warning: usr/lib/x86_64-linux-gnu/libkmnkbp0.a exists in debian/tmp but is not installed to anywhere 
> ```
> 
> @ermshiperete?

The .a file would have to be added to `linux/debian/libkmnkbp-dev.install`

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.60-alpha](https://keyman.com/downloads/releases/alpha/17.0.60)

mcdurdin:
> The .a file would have to be added to `linux/debian/libkmnkbp-dev.install`

I think if both static and dynamic libraries are needed for Linux, I will leave it to you to make the relevant changes to core build :grin:

ermshiperete:
https://github.com/keymanapp/keyman/pull/8375

PR #8344, opened 2023-03-01
chore(common): cleanup build.inc.sh echo

mcdurdin:
Makes the output of log messages from builder scripts more consistent, and uses color more effectively to highlight specific parts of lines.

Note: updates documentation for builder_describe_internal_dependency from parent PR #8342.

Sample:
image

* Highlights top-most script with bright-white+bold identifier
* Dependencies have grey identifiers
* Blue for headings
* Purple for dependency/parameter info
* Yellow for warnings
* Red for errors

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/build-utils.md
+++ b/resources/build/build-utils.md
@@ -237,6 +237,24 @@ include.
 Use the longer form of `if ...; then` rather than the shorter `[ ... ] && `
 pattern, for consistency and readability.
 
+# Internal dependencies
+
+All build scripts have a set of automatic internal dependencies:
+
+* `build` depends on `configure`
+* `test`, `install`, and `publish` depend on `build`
+
+These dependencies will be added to the list of targets for the build if you
+have described outputs for them, and the outputs do not exist. The build order
+is determined by the order in which `builder_start_action` is called in the
+script for each action.
+
+You can also define your own internal dependencies with
+`builder_describe_internal_dependency`. This allows you to define dependencies
+across targets. Use this judiciously; for example, Keyman Core uses this to
+build both x86_64 and arm64 targets, and test only the appropriate architecture
+on macOS.
+
 # Standard builder parameters
 
 The following parameters are pre-defined and should not be overridden:
@@ -245,6 +263,8 @@ The following parameters are pre-defined and should not be overridden:
 * `--color`: forces on ANSI color output for the script
 * `--no-color`: forces off ANSI color output for the script
 * `--verbose`, `-v`: verbose mode, sets the [`$builder_verbose`] variable
+
+--------------------------------------------------------------------------------
 
 # Builder API functions and variables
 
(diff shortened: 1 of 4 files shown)


mcdurdin:
iOS build fail due to non-standard way of including builder.inc.sh:

image

Will fix the source -- we want to ensure that builder.inc.sh is always included correctly because otherwise `$THIS_SCRIPT` and other variables will not be defined correctly.

resources/builder.inc.sh
```diff
@@ -143,17 +147,47 @@ function builder_heading() {
...
+  if [[ ! -z ${COLOR_RED+x} ]]; then
+    case $color in
+      white) color="$COLOR_WHITE" ;;
+      grey) color="$COLOR_GREY" ;;
+      green|success) color="$COLOR_GREEN" ;;
+      blue|heading) color="$COLOR_BLUE" ;;
+      yellow|warning) color="$COLOR_YELLOW" ;;
+      red|error) color="$COLOR_RED" ;;
+      teal) color="$COLOR_TEAL" ;;
+      setmark) mark="$HEADING_SETMARK" color="$COLOR_PURPLE" ;;
```
darcywong00, line 169: What about bright white and purple (separate from setmark)?

resources/builder.inc.sh
```diff
@@ -808,11 +838,10 @@ _builder_parameter_error() {
   local program="$1"
   local type="$2"
   local param="$3"
-  echo "$COLOR_RED$program: invalid $type: $param$COLOR_RESET"
+  builder_echo red "$program: invalid $type: $param"
```
darcywong00, line 841: Generally, is `builder_die` used for `builder_echo red`?

resources/build/build-utils.md
```diff
@@ -388,6 +408,35 @@ script. If not specified, dependencies will be built for all actions on all
...
+```bash
+builder_describe_internal_dependency \
+  mac:build mac-x86_64:build \
+  mac:build mac-arm64:build
+```
+
+**Note:** actions and targets must be fully specified, and this _must_ be called
+before either builder_describe_outputs or builder_parse in order for
+dependencies to be resolved.
+
```
darcywong00, line 438: Do the new colors also get added to https://github.com/keymanapp/keyman/blob/1bc31c8792b460233ab015c59567ef2d5e6bf966/resources/build/build-utils.md#formatting-variables ?

resources/builder.inc.sh
mcdurdin: It's possible to have a script emit an error before failing. But generally, yes, `builder_die` is the most common way we would want to emit red 'error' text.

resources/builder.inc.sh
mcdurdin: Can do

resources/build/build-utils.md
mcdurdin: Yes, may be a good idea, perhaps under a new `builder_echo` topic

mcdurdin (commented):
chore: address review comments

darcywong00 (approved):
lgtm

keyman-server:

PR #8287, opened 2023-02-22
chore: move from vcvarsall.bat to vsdevcmd.bat

mcdurdin:
@keymanapp-test-bot skip

Fixes #8076.

This page was helpful: https://renenyffenegger.ch/notes/development/tools/scripts/personal/vsenv_bat

Issue #8076, opened 2023-01-24
chore(windows,developer):  Use VsDevCmd.bat as replacement for vcvarsall.bat


SabineSIL:
In `developer/src/kmcmplib/build.bat`  (l.110) we have a comment about `vcvarsall.bat` stating we would need to find a different solution for this.:
```
rem Standalone build, so we'll make the environment available to the caller
rem Also setup
rem Note: Visual Studio 2022 doesn't provide vcvarsall.bat, so we'll have to find a different solution

```
There there is **VsDevCmd.bat** available in VS2022. This might be used instead. 

See comment **https://github.com/keymanapp/keyman/pull/6980#discussion_r1060401824   in feat(developer): cross-platform kmx compiler 🚑 https://github.com/keymanapp/keyman/pull/6980**

mcdurdin:
We need to fix these paths on the build agents as well, as they have hard-coded refs to 2019 in about 4 build steps...

mcdurdin:
Build agents fixed in 7 configs:

* Keyman for Windows Release (two steps)
* Keyman for Windows Test OSK Render
* Keyman for Windows Test
* Keyman Developer Release (two steps)
* Keyman Developer Test
* Keyboards Test (was on 2017)
* KeymanWeb Test PR Regressions (was on 2017) -- currently disabled

Replaced the following (or similar):

```bat
CALL "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsamd64_x86.bat"
```

with:

```bat
CALL %VisualStudioEnvSetupBat%
```

Where `VisualStudioEnvSetupBat` has been configured as a Keyman project-wide parameter:

```
"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\Tools\vsdevcmd.bat" -no_logo -startdir=none
```

PR #8357, opened 2023-03-02
chore(common): remove vswhere

mcdurdin:
With #8287, we no longer use a local copy of vswhere.exe but instead rely on a standard location for it from the Visual Studio installer.

Relates to #8076.

@keymanapp-test-bot skip


ermshiperete (approved):

keyman-server:

PR #8365, opened 2023-03-03
chore(developer): cleanup API interfaces and types for kmcmplib

mcdurdin:
Introduces kmcmplibapi.h which declares all external functions and types for kmcmplib, and updates kmcmplib and kmcmpdll to use these declarations consistently.

Note that kmcmp_CompileKeyboardToBuffer declares a void* instead of PFILE_KEYBOARD, because this is a large and complex type defined in both kmcmpdll and kmcmplib. In the future, when kmcmpdll existing legacy compiler is removed, we can make the kmcmplib type more public.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmcmpdll/Compiler.cpp
+++ b/developer/src/kmcmpdll/Compiler.cpp
@@ -96,10 +96,13 @@
   sizes and alignments have been manually verified.
 */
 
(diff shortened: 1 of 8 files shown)


SabineSIL:
lgtm

SabineSIL (approved):

Issue #7881, opened 2022-12-08
refactor(android/engine): Merge SystemKeyboard and InappKeyboard in KMManager 🛋


darcywong00:
From 17.0 team planning meeting

Refactor the WET code in KMManager and friends.
Reference on iOS implementation: #1347

Check if this impacts:
* KMSample1
* KMSample2
* FV Android
* Keyman Engine API and documentation

darcywong00:
If possible, @mcdurdin would also like to move the KMManager package from `com.tavultesoft.kmea` to `com.keyman.engine.manager`

darcywong00:
Keeping this open as there's more refactoring to do

darcywong00:
More suggestions by @jahorton on what to refactor:

>* toggleSuggestionBanner
>* setSpacebarText
>* resetContext has some WET code, even if it can't go full DRY
>* updateText
>* probably setNumericLayer
>* The fields in this block: [https://github.com/keymanapp/keyman/blob/ba7fc5435e46deaa805ed276c5ae0f06486ec412/[…]android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java](https://github.com/keymanapp/keyman/blob/ba7fc5435e46deaa805ed276c5ae0f06486ec412/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java#L178-L185) (save the InAppKeyboard and SystemKeyboard ones)

----
> More meaty, but almost certainly worth the time investment required:
updateSelectionRange - the if (ic!=null) branch - is there a reason that can't run for the in-app keyboard too?  That we can't use the same thing when using the keyboard within the app?  Digging into that would be pretty insightful.

darcywong00:
Evalaluate at A17S27 if this is complete now

darcywong00:
This issue is resolved by the attached PR's

PR #8370, opened 2023-03-06
refactor(android/engine): Start consolidating KMKeyboards 🛋

darcywong00:
Potentially more progress towards #7881 

I'm unsure if this is a good refactoring, but this replaces members InAppKeyboard and SystemKeyboard with a list.

It simplifies KMManager.getKMKeyboard(), but makes the other accesses a little unweirdly.

@keymanapp-test-bot skip - Will write tests if we proceed


darcywong00:
Will abandon this PR. From Slack, @mcdurdin would prefer eliminating the distinction of "InAppKeyboard" and "SystemKeyboard" if possible. I think that means not running both keyboards in the same process at the same time.

PR #8371, opened 2023-03-06
feat(common): builder.inc.sh inheritable options

mcdurdin:
Appending a `+` to a builder option will automatically pass that option to child scripts.

@keymanapp-test-bot skip

### DIFF
--- a/android/KMEA/build.sh
+++ b/android/KMEA/build.sh
@@ -35,7 +35,9 @@ builder_describe "Builds Keyman Engine for Android." \
   "build" \
   "test             Runs lint and unit tests." \
   ":engine          Builds Engine" \
-  "--ci             Don't start the Gradle daemon. For CI"
+  "--ci             Don't start the Gradle daemon. For CI" \
+  --debug \
+  --upload-sentry
 
 # parse before describe_outputs to check debug flags
 builder_parse "$@"
--- a/android/build.sh
+++ b/android/build.sh
@@ -42,7 +42,7 @@ builder_parse "$@"
 builder_run_child_actions clean
 
 if builder_start_action clean; then
-  builder_echo "Cleaning /android/upload"
+  echo "Cleaning /android/upload"
   rm -rf "$KEYMAN_ROOT/android/upload"
   builder_finish_action success clean
 fi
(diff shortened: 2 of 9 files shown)


android/build.sh
```diff
@@ -35,35 +38,14 @@ builder_describe \
...
-#  "--ci                                     Don't start the Gradle daemon. Use for CI" \
-#  "--upload-sentry                          Uploads debug symbols, etc, to Sentry"
-
 # This script also responsible for cleaning up /android/upload
-if builder_start_action clean; then
+builder_run_child_actions clean
 
-  builder_heading "Cleanup /android/upload"
+if builder_start_action clean; then
+  builder_echo "Cleaning /android/upload"
```
darcywong00, line 45: CI said builder_echo undefined. Should this be echo or builder_heading?

mcdurdin: `builder_echo` will be defined soon ... forgot that it is in another PR

android/KMEA/build.sh
```diff
@@ -35,7 +35,9 @@ builder_describe "Builds Keyman Engine for Android." \
   "build" \
   "test             Runs lint and unit tests." \
   ":engine          Builds Engine" \
-  "--ci             Don't start the Gradle daemon. For CI"
+  "--ci             Don't start the Gradle daemon. For CI" \
+  --debug \
```
darcywong00, line 39: Looking from the CI build, it seems like the dependency on `@/web` is only making a release build.  `--debug` doesn't get passed to `/web/build.sh`?

mcdurdin: True. `--debug` is an option that we want to be passing through to dependency builds. I need to do some work on that particular feature -- I'll make a sub-PR of this one.

I actually think we should never spec `--debug` in `builder_describe` but it should always be available, and _always_ be passed to child scripts and dependency scripts. (If the script doesn't use it, it doesn't hurt!)

Note: `-d` is shorthand for `--debug`.

`--debug` may be the only option we do that with for now, but there may be others in the future?

android/build.sh
```diff
@@ -26,6 +26,9 @@ builder_describe \
   build \
   test \
   "publish                                  Publishes the APKs to the Play Store." \
+  --ci+ \
+  --debug+ \
```
darcywong00, line 30: Some merge conflicts to resolve. But does this line go away with #8382?

mcdurdin:
Replaced by #8384.

PR #8375, opened 2023-03-06
chore(core): Create both dynamic and static libs on Linux

ermshiperete:
This change creates both dynamic and static libs on Linux. 

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM

Well, that was nice and easy!

keyman-server:

PR #8376, opened 2023-03-06
chore(core): Remove obsolete `--target-path` option

ermshiperete:
Linux package builds no longer require a different target path, so we can remove the `--target-path` parameter.

@keymanapp-test-bot skip


mcdurdin (approved):
I _love_ it. Now we are just a bit closer to being able to make core a dependency of linux!

Still want to avoid running core tests when doing linux packaging builds though...

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.63-alpha](https://keyman.com/downloads/releases/alpha/17.0.63)

Issue #8355, opened 2023-03-02
 feat(developer): kmcompx - Use consistent line endings (either \n or \r\n)


SabineSIL:
In kmcompdll we use \r\n as line ending which works for windows. We need to ensure that this works on Linux as well ( as there is no \r)

see  https://github.com/keymanapp/keyman/pull/6980#discussion_r975516282  cross-platform kmx compiler 🚑 https://github.com/keymanapp/keyman/pull/6980

PR #7330, opened 2022-09-20
epic: kmcompx - kmcmpdll cross-platform compiler 🚑

mcdurdin:
@keymanapp-test-bot skip

PR #8380, opened 2023-03-07
chore(developer): use consistent line endings \r\n

SabineSIL:
fixes: #8355
In kmcompdll we use \r\n as line ending which works for windows. We need to ensure that this works on Linux as well ( as there is no \r) 

see https://github.com/keymanapp/keyman/pull/6980#discussion_r975516282 cross-platform kmx compiler 🚑 https://github.com/keymanapp/keyman/pull/6980, [#7330](https://github.com/keymanapp/keyman/pull/7330)
@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmcmplib/src/Compiler.cpp
+++ b/developer/src/kmcmplib/src/Compiler.cpp
@@ -3389,12 +3389,20 @@ KMX_DWORD ReadLine(FILE* fp_in , PKMX_WCHAR wstr, KMX_BOOL PreProcess)
 
   if (cur == fsize)
 
-  // S: Is replacing "\r\n" with "\n" here sufficient or do we need changes at other places as well when we skip "\r"?
+  // \r\n is still added here even though Linux doesn`t use \r.
+  // This is to ensure to still have a working windows-only-version
   u16ncat(str, u"\r\n", _countof(str));  // I3481     // Always a "\r\n" to the EOF, avoids funny bugs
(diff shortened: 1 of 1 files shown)


developer/src/kmcmplib/src/Compiler.cpp
```diff
@@ -3389,12 +3389,20 @@ KMX_DWORD ReadLine(FILE* fp_in , PKMX_WCHAR wstr, KMX_BOOL PreProcess)
...
   if (len == 0) return CERR_EndOfFile;
 
+  // neccessary to add this block for using on non-windows platforms (removes all \r for platforms that use \n instead of \r\n)
+  for (p = str, n = 0; n < len; n++, p++) {
+    if (*p == L'\r')
+      *p = L' ';
+  }
+
+  // \r is still left in this block even though Linux doesn`t use \r.
+  // This is to ensure to still have a working windows-only-version
```
mcdurdin, line 3405: This shouldn't be necessary because we've just stripped out all `\r` above.

mcdurdin (approved):

developer/src/kmcmplib/src/Compiler.cpp
SabineSIL: true it is not necessary, but if  we should ever want to use \r\n again it is easier to still leave it in the function than including it again at the right position.

PR #8382, opened 2023-03-07
chore(common): clean up `--debug` usage in build scripts

### LINKED ISSUE #8373: chore(common): Make `--debug` a globally inheritable option for builder
True. `--debug` is an option that we want to be passing through to dependency builds. I need to do some work on that particular feature -- I'll make a sub-PR of this one.

I actually think we should never spec `--debug` in `builder_describe` but it should always be available, and _always_ be passed to child scripts and dependency scripts. (If the script doesn't use it, it doesn't hurt!)

Note: `-d` is shorthand for `--debug`.

`--debug` may be the only option we do that with for now, but there may be others in the future?

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/8371#discussion_r1125981899_

mcdurdin:
Fixes #8373.

Replaces #8378 (basing on master; android script updates will be added to #7407).

Note that the core functionality requested in #8373 is already in place; this commit just does some cleanup to make this easier to maintain in the future:

* Uses `builder_is_debug_build` function to test for debug build (instead of `builder_has_option --debug`, which would also work but using the other function avoids typo issues with the option name).

* Renames `$_builder_debug` to `$_builder_debug_internal` to reduce naming confusion

* Documents `$builder_debug`, `builder_is_debug_build`, and clarifies usage of `--debug` option.

* Logs command line for dep builds (we could hide this in the future, but it certainly doesn't hurt for now!)

@keymanapp-test-bot skip

### DIFF
--- a/common/predictive-text/unit_tests/test.sh
+++ b/common/predictive-text/unit_tests/test.sh
@@ -23,8 +23,7 @@ builder_describe "Runs all tests for the language-modeling / predictive-text lay
   ":libraries  Runs unit tests for in-repo libraries used by this module"\
   ":headless   Runs this module's headless user tests" \
   ":browser    Runs this module's browser-based user tests" \
-  "--ci        Uses CI-based test configurations & emits CI-friendly test reports" \
-  "--debug,-d  Activates developer-friendly debug mode for unit tests where applicable"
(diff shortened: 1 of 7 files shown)


darcywong00 (approved):
lgtm

keyman-server:

PR #8384, opened 2023-03-08
feat(common): builder.inc.sh inheritable options

darcywong00:
Appending a `+` to a builder option will automatically pass that option to child scripts.

I wanted to apply #8371 to master branch so #7407 can remain strictly Android changes. (This can replace #8371)

TODO on #7407 - add inheritable options syntax.

@keymanapp-test-bot skip


keyman-server:

PR #8385, opened 2023-03-08
chore(core): rebuild kmx binary test fixtures without version data 🚑

mcdurdin:
In order to be able to provide a more robust round-trip comparison of existing .kmx binary fixtures with compiler output, this PR rebuilds them with the `-no-compiler-version` kmcomp flag, to omit the `TSS_COMPILEDVERSION` and `TSS_KEYMANCOPYRIGHT` stores.

---

A rider commit adds `_reserved` member to `COMP_KEY`. Added here in order to get all common changes in this PR series into a single PR.

In order to allow us to reference the padding in the FILE_KEY structure, for reproducible and cross-platform consistent builds, turn the padding into an `_reserved` (must always be zero) field.

Changes to compiler to utilise this will be in a follow-up PR.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

PR #8387, opened 2023-03-08
chore(developer): cross-platform fixups pre-wasm build of kmcmplib 🚑

mcdurdin:
This is not enough for building to WASM, but wanted to split out the bigger changes into their own PRs. This is mostly busy-work.

One serious bug fix:
* u16string_from_string was not converting from UTF-16 correctly

Apart from that, just tidies up a few bits and pieces in kmcmplib:
* type coercion for json-validator.cpp (probably already fixed upstream)
* cleanup of build scripts for new builder variable names
* various compatibility macros for MSVC vs clang
* safer math expressions
* file name casing in headers
* replacement of various ms-only std funcs
* removed some unused variables and consts
* tightened up search for keyboards to get only keyboards in /source/ folders (may be possible to improve further in the future?)

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmcmplib/src/Compiler.cpp
+++ b/developer/src/kmcmplib/src/Compiler.cpp
@@ -84,11 +84,11 @@
 #include "../../../../common/windows/cpp/include/ConvertUTF.h"
 
 #include "debugstore.h"
-#include "namedcodeconstants.h"
+#include "NamedCodeConstants.h"
 
 #include "xstring.h"
 
-#include "edition.h"
+#include "Edition.h"
 
 #include "CharToKeyConversion.h"
 #include "CasedKeys.h"
@@ -99,6 +99,7 @@
 #include "UnreachableRules.h"
 #include "CheckForDuplicates.h"
 #include "kmx_u16.h"
+#include "filesystem.h"
 #include <CompMsg.h>
 
 /* These macros are adapted from winnt.h and legacy use only */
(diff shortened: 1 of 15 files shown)


developer/src/kmcmplib/src/Compiler.cpp
```diff
@@ -99,6 +99,7 @@
 #include "UnreachableRules.h"
 #include "CheckForDuplicates.h"
 #include "kmx_u16.h"
+#include "filesystem.h"
```
mcdurdin, line 102:
```suggestion
```

mcdurdin (commented):
This belongs in #8389

developer/src/kmcmplib/src/NamedCodeConstants.cpp
```diff
@@ -264,14 +264,14 @@ int NamedCodeConstants::GetCode_IncludedCodes(const KMX_WCHAR *codename)
...
+ //HangulVBase = 0x1161,
+ //HangulTBase = 0x11A7,
  HangulLCount = 19,
  HangulVCount = 21,
- HangulTCount = 28,
- HangulNCount = HangulVCount * HangulTCount,   // 588
- HangulSCount = HangulLCount * HangulNCount;   // 11172
+ HangulTCount = 28;
+ //HangulNCount = HangulVCount * HangulTCount;   // 588
+ //HangulSCount = HangulLCount * HangulNCount;   // 11172
```
rc-swag, line 274: I'm  not sure what is going on here with the and above

mcdurdin: These are unused constants, but I didn't want to delete them because they are helpful references.

rc-swag (approved):
lgtm

Issue #8381, opened 2023-03-07
bug(developer): rule sorting algorithm may sort touch or other non-character keys wrongly


mcdurdin:
If touch keys or non-character keys are used in a store and then placed into a single rule, they are currently treated as identical for purposes of ordering, because `VKToChar` necessarily returns `0` for these keys. This means that rules may be misgrouped, as they are supposed to be grouped by key first, then by context length, then by line number.

https://github.com/keymanapp/keyman/blob/0c2c4f42067b465b205bbd2a9d50c4ac6e60492e/developer/src/kmcmpdll/Compiler.cpp#L929-L951

(A secondary issue is that the ordering is unstable across C++ compilers when working with an expanded `any()` and using these keys (or when the same char is used more than once -- see for example [gff_geez.kmn:47](https://github.com/keymanapp/keyboards/blob/master/release/gff/gff_geez/source/gff_geez.kmn#L47). But this instability can be fixed in the kmcompx work).

The ordering is important for the KeymanWeb compiler, which uses it to optimize search by key and then by context.

We need to verify that this can go wrong currently with a unit test, before attempting a fix.

mcdurdin:
> We need to verify that this can go wrong currently with a unit test, before attempting a fix.

I have created a sample keyboard which demonstrates that this failure cannot occur.

```
c rule_order_8381 generated from template at 2023-11-01 12:49:33
c with name "rule_order_8381"
store(&VERSION) '10.0'
store(&NAME) 'rule_order_8381'
store(&COPYRIGHT) '©'
store(&KEYBOARDVERSION) '1.0'
store(&TARGETS) 'any'
store(&BITMAP) 'rule_order_8381.ico'
store(&VISUALKEYBOARD) 'rule_order_8381.kvks'
store(&LAYOUTFILE) 'rule_order_8381.keyman-touch-layout'

begin Unicode > use(main)

group(main) using keys

+ [T_A] > 'aa'
+ [T_B] > 'bb'
+ [T_C] > 'cc'

'cc' + [T_C] > 'xx'
'bb' + [T_B] > '??'
'aa' + [T_A] > '!!'
```

This has the second set of keys out of order, so the .kmx gives the following rule order:

```
  "cc"  + [2] > "xx"  c line(20)
  "bb"  + [1] > "??"  c line(21)
  "aa"  + [0] > "!!"  c line(22)
   + [0] > "aa"  c line(16)
   + [1] > "bb"  c line(17)
[...]

PR #8388, opened 2023-03-08
fix(developer): improve .kmx output consistency for cross-platform 🚑

mcdurdin:
This fixes two issues:

1. `COMP_KEY` structure had padding that was not always zeroed out. Made this explicit.

2. Rule sorting could be inconsistent when re-sorting rules that use `+ any() > ...`, which are expanded at compile time into multiple rules, and which ended up with identical sort keys due to lack of precision. This relates to #8381 but only adds differentiation for the otherwise ambiguous rule sort keys, and does not fix that issue, which requires further investigation.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #8398, opened 2023-03-09
chore(core): shared meson config 🚑

mcdurdin:
Shares meson configuration between Core and Kmcmplib, so that we can hopefully reign in configuration flag insanity before it takes off.

Does not touch meson.build files under /linux because those are very different. Worth reviewing to see if it makes sense to use the same shared config there too.

@keymanapp-test-bot skip


ermshiperete (approved):

ermshiperete:
> Does not touch meson.build files under /linux because those are very different. Worth reviewing to see if it makes sense to use the same shared config there too.

My guess is that we can use the shared meson file for linux as well. I went with a minimal set of settings for Linux because my impression was that some of the flags and settings are the default and thus set anyway.

PR #8400, opened 2023-03-09
chore(core): breadth-first build

### LINKED ISSUE #8395: chore(core,developer): build.sh needs to configure all targets first before any builds
Currently, we clean,configure,build,test,install for each target independently (depth-first). We need to go to breadth-first, clean for all targets, configure for all targets, etc.

This applies to core and developer/kmcmplib

mcdurdin:
Fixes #8395 for core.

@keymanapp-test-bot skip


ermshiperete:
I don't think breadth-first will work in all cases, e.g. if A depends on B and A's configure target (in meson) expects to find a library that gets only build in B's build target.

ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.64-alpha](https://keyman.com/downloads/releases/alpha/17.0.64)

PR #8409, opened 2023-03-10
refactor(core): consolidate wasm defs

### LINKED ISSUE #8010: refactor(core): Review wasm build scripts
The wasm.build scripts for Linux, Windows, and Mac (`core/wasm.build.*.in`) greatly differ which seems really odd to me. I'd have thought  that they would differ in the paths and how the executable gets called. They specify different parameters and the entries in the `[binaries]` section are different on Linux and Mac where I'd think they could be the same. This is highly confusing for someone not very familiar with wasm/meson.

It would be good to refactor the wasm.build scripts to be as similar as possible and remove unnecessary lines. (I assume that the parameters that the Linux build specifies are already defined in `wasm.defs.build` and so are not needed)

The same build scripts got copied to `developer/src/kmcmplib/` in the `feature-kmcompx` branch.

mcdurdin:
Fixes #8010.

Note that #8398 moves other settings into a shared resources folder, and once that hits master, it would be good to do another round of consolidation to share the wasm defs in resources as well.

@keymanapp-test-bot skip

### DIFF
--- a/core/wasm.build.linux.in
+++ b/core/wasm.build.linux.in
@@ -1,7 +1,4 @@
 [binaries]
-c = ['$EMSCRIPTEN_BASE/emcc.py', '-s', 'WASM=1', '-O2']
-cpp = ['$EMSCRIPTEN_BASE/em++.py', '-s', 'WASM=1','-O2']
+c = ['$EMSCRIPTEN_BASE/emcc.py']
+cpp = ['$EMSCRIPTEN_BASE/em++.py']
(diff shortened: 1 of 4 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.65-alpha](https://keyman.com/downloads/releases/alpha/17.0.65)

Issue #8161, opened 2023-02-02
feat(common): support jis and abnt2 🙀 


srl295:
followon to #7965 

support `jis` and `abnt2` hardware layouts

srl295:
https://github.com/keymanapp/keyman/issues/8552 to continue reviewing VK for abnt2 and jis

PR #8412, opened 2023-03-11
chore(developer): ldml March 2023 update to tech-preview 🙀 

### LINKED ISSUE #8160: fix(developer): ldml: support >1 layers lists 🙀
followon to #7965

right now we support a single `<layers>` element in compile. Should support >1 - ~multiple~ only one hardware layer, but multiple touch layers.

srl295:
- significantly, doesn't include any of the transform updates, but has other updates
- latest CLDR: https://github.com/unicode-org/cldr/commit/dc112dab589bb5c91e2d0c048dd48d85c3652912
- `<layers form="hardware" hardware="iso"/>` is now just `<layers form="iso"/>`
- Fixes #8160 at least for hardware
- normalizes a number of test cases
- note that #8161 is still unfixed, changed the TODO to an internal error instead of a warning.

@keymanapp-test-bot skip


srl295:
should build now

core/tests/unit/ldml/keyboards/fr-t-k0-azerty.xml
```diff
@@ -123,28 +124,28 @@
...
+			<row keys="base W X C V B N" />  <!--TODO:  + bksp -->
 			<row keys="numeric extra space enter" />
 		</layer>
 
 		<layer id="numeric">
 			<row keys="1 2 3 4 5 6 7 8 9 0" />
-			<row keys="hyphen slash colon semi-colon open-paren close-paren dollar amp at quote" />
-			<row keys="symbol period comma question bang quote bksp" />
+			<row keys="hyphen slash colon semi-colon open-paren close-paren dollar amp at double-quote" />
+			<row keys="symbol period comma question bang double-quote" /> <!--TODO:  + bksp -->
```
mcdurdin, line 141: Why is double-quote here twice?

developer/src/kmc-keyboard/src/compiler/keys.ts
```diff
@@ -71,15 +72,17 @@ export class KeysCompiler extends SectionCompiler {
 
     // Finally, kmap
     // Use LayerMap + keys to generate compiled keys for hardware
-    const theLayers = this.keyboard.layers?.[0]; // TODO-LDML: handle >1 layers. #8160
-
-    if(theLayers?.form == 'hardware') {
+    const hardwareLayers = this.keyboard.layers?.filter(({form}) => form !== 'touch');
```
mcdurdin, line 75: Given we are doing this twice -- here and on line 43 -- we should have a local helper function `this.hardwareLayers()`

developer/src/kmc-keyboard/src/compiler/keys.ts
```diff
@@ -71,15 +72,17 @@ export class KeysCompiler extends SectionCompiler {
...
+    if (hardwareLayers.length > 1) {
+      // validation should have already caught this
+      throw Error(`Internal error: Expected 0 or 1 hardware layer, not ${hardwareLayers.length}`);
```
mcdurdin, line 78: Are we supporting multiple hardware layers? I thought we were, just not more than one of each form?

developer/src/kmc-keyboard/test/fixtures/sections/keys/maximal.xml
```diff
@@ -24,7 +24,7 @@
       <row keys="Q q W w" />
     </layer>
   </layers>
-  <layers form="hardware" hardware="abnt2">
+  <layers form="iso">
```
mcdurdin, line 27:
```suggestion
  <layers form="abnt2">
```
?

developer/src/kmc-keyboard/test/test-keys.ts
```diff
@@ -10,92 +10,97 @@ const K = Constants.USVirtualKeyCodes;
...
+    {
+      // should compile minimal keys data
+      subpath: 'sections/keys/minimal.xml',
+      callback(sect) {
+        const keys = <Keys> sect;
+        assert.ok(keys);
+        assert.equal(compilerTestCallbacks.messages.length, 0);
+        assert.equal(keys.keys.length, 1);
+        assert.equal(keys.flicks.length, 1); // there's always a 'null' flick
+        assert.equal(keys.keys[0].to.value, '🪦');
```
mcdurdin, line 23: Can we use a Unicode value here rather than a char, as I get tofu on Windows by default.

mcdurdin (approved):
Minor questions only

developer/src/kmc-keyboard/src/compiler/keys.ts
srl295: Briefly How would that work ? Or different widths ?

mcdurdin:
```xml
<layers form="iso">...</layers>
<layers form="us">...</layers>
```

Would give the dev the ability to spec different layouts for 101 vs 102 key keyboards, which could be helpful (e.g. on a 102 key layout can use the extra key more effectively?)

But it is also a bit of a footgun for docs and support.

srl295:
rebased to pickup fixed (and actually running) tests

mcdurdin:
> rebased to pickup fixed (and actually running) tests

Rebase on the PR details doesn't re-run tests :grin: 

PR rebase isn't quite the same as a local git rebase and force-push. It doesn't incorporate the base branch's changes, unless there is a merge conflict. This detail is hidden by default in the UI, which is confusing... I have just switched on the option "Always suggest updating pull request branches" so let's see if that helps!

core/tests/unit/ldml/keyboards/fr-t-k0-azerty.xml
srl295: that's in the upstream in ldml. I plan to rename it soon-ish as the 'test keyboard', to make way for   https://unicode-org.atlassian.net/browse/CLDR-12026  as the actual azerty.

developer/src/kmc-keyboard/src/compiler/keys.ts
srl295: mm this was rejected in _Keyboard-SC February 13/14, 2023_ 

> only a single hardware layout…Avoids confusion since we don't know what's actually plugged in

developer/src/kmc-keyboard/test/fixtures/sections/keys/maximal.xml
srl295: abnt2 not supported, so changed to iso since this is now being validated

rc-swag:
> Not sure which is the best option to use here!

Well, the `Update with rebase` now sounds like the local rebase and force push.  Or is it still behaving differently?
Which of those two options?
Since being with the keyman repo I actually haven't looked at the history log at all. Only individual files in the IDE.  However, I do find all the commits grouped together (what the `rebase` essentially gives you) easier to follow.

developer/src/kmc-keyboard/test/test-keys.ts
srl295: done

srl295:
> 

oh hi… missed this…  should re-run since i added another commit.

srl295:
> And ... we now have:
> 
> image
> 
> Not sure which is the best option to use here!

This happened because I edited the PR to change the base before finishing the force push which contained the updated base.

developer/src/kmc-keyboard/src/compiler/keys.ts
mcdurdin: OK

mcdurdin:
> This happened because I edited the PR to change the base before finishing the force push which contained the updated base.

Yeah, but given the GH PR UX allows you to do this, it will happen ... I am not fond of force push anyway because of the messiness it generates for reviewers. I encourage team members to avoid force push for PRs that are not in draft.

srl295:
> > This happened because I edited the PR to change the base before finishing the force push which contained the updated base.
> 
> 
> 
> Yeah, but given the GH PR UX allows you to do this, it will happen ... I am not fond of force push anyway because of the messiness it generates for reviewers. I encourage team members to avoid force push for PRs that are not in draft.

Won't that make merge commits then making it harder to review?

mcdurdin:
> Won't that make merge commits then making it harder to review?

It's mostly an issue for PRs that are out of draft and have already had review started on them. Force push really breaks the review process there. Changes from merge commits don't show up in the review (except where there may be manual corrections to merge conflicts etc).

srl295:
Ok. Will probably keep in draft a bit more till I’m sure what the merge
order is. And try using merge commits. Thanks.. sorry to break process

mcdurdin:
It's all good -- just trying to avoid grief for reviewers :grin:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.68-alpha](https://keyman.com/downloads/releases/alpha/17.0.68)

PR #8415, opened 2023-03-12
chore(common): support shorthand for build.sh

### LINKED ISSUE #8396: chore(common): make it easier to specify complex builder commands
Currently, if we want to do a bunch of commands on a single target, it's a little painful:

```
./build.sh clean:wasm configure:wasm build:wasm test:wasm
```

Given no action, only `build` is run, which makes sense to maintain, so `./build.sh :wasm` only does `./build.sh build:wasm`. (We definitely don't want _all_ actions to run by default, because there are scary ones such as `publish`.

Given no target, all targets are run, which makes sense to maintain.

It'd be nice to be able to say ... "all these actions for this target", e.g. `./build.sh clean,configure,build,test:wasm`

It'd be nice to be able to do shorthand, e.g. `./build.sh cl co b t`

And to combine, e.g. `./build.sh cl,co,b,t:w`

Ambiguity in shorthand would be an error, not first-past-the-post.

mcdurdin:
Fixes #8396.

@keymanapp-test-bot skip

### DIFF
--- a/resources/builder.inc.sh
+++ b/resources/builder.inc.sh
@@ -152,10 +152,9 @@ builder_echo() {
   local color=white message= mark=
   if [[ $# -gt 1 ]]; then
     color="$1"
-    message="$2"
-  else
-    message="$1"
+    shift
   fi
+  message="$*"
 
   if [[ ! -z ${COLOR_RED+x} ]]; then
     case $color in
@@ -1159,7 +1158,7 @@ _builder_parse_expanded_parameters() {
 
     # Expand shorthand parameters
 
-    new_action=`_builder_expand_shorthand $action "${_builder_actions[@]}"` ||
+    new_action=$(_builder_expand_shorthand $action "${_builder_actions[@]}") ||
       case $? in
         1)
           action=$new_action
@@ -1170,7 +1169,7 @@ _builder_parse_expanded_parameters() {
           ;;
       esac
 
-    new_target=`_builder_expand_shorthand $target "${_builder_targets[@]}"` ||
+    new_target=$(_builder_expand_shorthand $target "${_builder_targets[@]}") ||
       case $? in
         1)
           target=$new_target
--- a/resources/builder.inc.sh
+++ b/resources/builder.inc.sh
@@ -1312,7 +1312,7 @@ _builder_parse_expanded_parameters() {
     echo "${_params[@]}"
     builder_echo setmark "build.sh parameters: <${_params[@]}>"
     if [[ ${#builder_extra_params[@]} -gt 0 ]]; then
-      builder_echo grey "build.sh extar parameters: <${builder_extra_params[@]}>"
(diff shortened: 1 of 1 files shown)


resources/builder.inc.sh
```diff
@@ -1049,9 +1156,39 @@ builder_parse() {
...
 
+    # Expand shorthand parameters
+
+    new_action=$(_builder_expand_shorthand $action "${_builder_actions[@]}") ||
+      case $? in
+        1)
+          action=$new_action
+          ;;
+        *)
+          builder_warn "Parameter $action has $? matches, could mean any of {$new_action}"
```
darcywong00, line 1167: Any reason you used builder_warn instead of builder_die? (The example in l.261 uses die)

resources/builder.inc.sh
```diff
@@ -1163,15 +1300,20 @@ builder_parse() {
...
   else
     # This is a top-level invocation, not a dependency build, so we want to
     # track which dependencies have been built, so they don't get built multiple
     # times.
-    builder_echo setmark "build.sh parameters: <${_builder_params[@]}>"
+    # TODO: consider printing expanded builder parameters instead of shorthand
+    echo "${_params[@]}"
+    builder_echo setmark "build.sh parameters: <${_params[@]}>"
+    if [[ ${#builder_extra_params[@]} -gt 0 ]]; then
+      builder_echo grey "build.sh extar parameters: <${builder_extra_params[@]}>"
```
darcywong00, line 1315:
```suggestion
      builder_echo grey "build.sh extra parameters: <${builder_extra_params[@]}>"
```

resources/builder.inc.sh
```diff
@@ -1049,9 +1156,39 @@ builder_parse() {
...
+      esac
+
     _builder_item_in_array "$action" "${_builder_actions[@]}" && has_action=1 || has_action=0
     _builder_item_in_array "$target" "${_builder_targets[@]}" && has_target=1 || has_target=0
 
+    if (( has_action )) || (( has_target )); then
+      # Document parameter expansion for end use
+      _params[$n]=$action$target
+    fi
+    n=$((n + 1))
```
darcywong00, line 1190: Is this the same as
`n=$(n + 1)`?

resources/builder.inc.sh
mcdurdin: I just liked it in yellow rather than red :grin: -- it looks friendlier for this scenario.

resources/builder.inc.sh
mcdurdin: No, `$()` is a subshell, `$(())` is arithmetic expression evaluation.

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.67-alpha](https://keyman.com/downloads/releases/alpha/17.0.67)

Issue #8435, opened 2023-03-14
feat(core): ldml support repertoire tests 🙀


srl295:
followon to #7535

support repertoire tests in the test runner

depends on #8495

srl295:
```
failure summary: pt-BR-k0-abnt2-test.json\pt-test\repertoire\latn-repertoire
[16:34:31 ](https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_Windows/376474?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=376474_1942_1942&logView=flowAware)  FAILED
```

So it prints out at the end of the big log, what exactly failed. Latin repertoire.  I'll rerun this on windows locally and debug it.

srl295:
`- FAIL: No string for repertoire test: �` (U+FFFD)

Hmmmm sounds like a bug.  (Sounds like default codepage somewhere)

mcdurdin:
Does #8441 fix this?

srl295:
@mcdurdin at an initial level yes

PR #8441, opened 2023-03-15
feat(core): ldml repertoire test, initial ICU integration 🙀

srl295:
- icu4c as a required, statically linked, dependency, only on the ldml test (for now

For: #8435 and #8495

Note, f2b8b9b10d056fc7ba88acbbaa3be26cdead1324 does add build files from `mesonbuild/wrapdb`, includes license file (MIT)

@keymanapp-test-bot skip


srl295:
Trying to see how this builds.

Had to run:

```shell
$ sudo ln -s /usr/local/opt/icu4c/lib/pkgconfig/* /opt/X11/lib/pkgconfig/
```

due to meson issues, see https://bugs.freedesktop.org/show_bug.cgi?id=111126#c8

may be able to do: `meson… --build.pkg-config-path "$PKG_CONFIG_PATH"`

which seems very redundant. c'mon meson…

core/tests/unit/ldml/meson.build
```diff
@@ -4,6 +4,10 @@
 # Authors:      Marc Durdin
 #
 
+pkgconfig = 'pkg-config'
+icu4c = dependency('icu-uc', required: false)
```
mcdurdin, line 8: How will this work on Windows builds?

srl295: Choco? I'll see. Making it optional as a PoC.

mcdurdin: Is this intended just for tests?

I would be happy to include icu as a dependency but we need to think about how we want to add it strategically because adding dependencies is always complicated for a cross-platform project.

srl295: Yea for tests. I think it's ultimately a part of developer tools. Which will have wasm linked icu4c or icu4x available. So this is just a way to bring up and prove out.

srl295:
- 🤦  the mac cross compile stuff means…… the homebrew linkage doesn’t work because it’s the “wrong” architecture (arm vs x64)
- WIP. It's able to get 1-1 keys (a, b, c etc.) but fails for Khmer with "No string for repertoire test: ា" since tiny….xml does not have a single key with ា
- I originally excluded strings in UnicodeSet but wonder if I should rethink that. Instead of `[ħ ថា]` it would be `[ħ {ថា}]` i.e. expecting that there's a key (flick etc) that can generate "ថា" wholesale as a string
- no transforms yet.

srl295:
```
$ ls -lh core/build/mac-x86_64/release/tests/unit/ldml/ldml
-rwxr-xr-x 1 srl295 staff 1.2M Mar 31 13:14 core/build/mac-x86_64/release/tests/unit/ldml/ldml
$ nm core/build/mac-x86_64/release/tests/unit/ldml/ldml | fgrep _72 | head
0000000100041620 T _T_CString_int64ToString_72
0000000100041570 T _T_CString_integerToString_72
00000001000416e0 T _T_CString_stringToInteger_72
00000001000414f0 T _T_CString_toLowerCase_72
0000000100041530 T _T_CString_toUpperCase_72
000000010005fbf0 T _UDataMemory_createNewInstance_72
000000010005fb90 T _UDataMemory_init_72
000000010005fda0 T _UDataMemory_isLoaded_72
000000010005fc50 T _UDataMemory_normalizeDataPointer_72
000000010005fc80 T _UDataMemory_setData_72
$ nm core/build/mac-x86_64/release/tests/unit/ldml/ldml | fgrep _72 | wc -l
1350
$ otool -L core/build/mac-x86_64/release/tests/unit/ldml/ldml
core/build/mac-x86_64/release/tests/unit/ldml/ldml:
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.36.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
```

^^ statically linked

core/subprojects/packagefiles/icu-minimal/meson.build
```diff
@@ -0,0 +1,94 @@
...
+uconfig.set('UCONFIG_NO_FILE_IO', 1)
+uconfig.set('UCONFIG_NO_LEGACY_CONVERSION', 1) # turn off file based codepage conversion
+uconfig.set('UCONFIG_NO_NORMALIZATION', 1) # TODO-LDML: may want this
+uconfig.set('UCONFIG_NO_BREAK_ITERATION', 1) # TODO-LDML: may want this
+uconfig.set('UCONFIG_NO_IDNA', 1)
+uconfig.set('UCONFIG_NO_COLLATION', 1)
+uconfig.set('UCONFIG_NO_FORMATTING', 1)
+uconfig.set('UCONFIG_NO_TRANSLITERATION', 1)
+uconfig.set('UCONFIG_NO_REGULAR_EXPRESSIONS', 1) # TODO-LDML: Will probably want this for transforms #7375
+uconfig.set('UCONFIG_NO_SERVICE', 1)
```
srl295, line 44: knobs and switches :gear:

mcdurdin:
From [Windows build](https://build.palaso.org/viewLog.html?buildId=376150&buildTypeId=Keyman_Common_KPAPI_TestPullRequests_Windows):

```
[06:33:00 ](https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_Windows/376150?buildTab=log&focusLine=1644&logView=flowAware&linesState=1644)  FAILED: tests/unit/ldml/ldml.exe.p/ldml_test_source.cpp.obj
[06:33:00 ](https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_Windows/376150?buildTab=log&focusLine=1645&logView=flowAware&linesState=1645)  "cl" "-Itests\unit\ldml\ldml.exe.p" "-Itests\unit\ldml" "-I..\..\..\tests\unit\ldml" "-Iinclude" "-I..\..\..\include" "-I..\..\..\..\common\include" "-Isrc" "-I..\..\..\src" "-I..\..\..\..\developer\src\ext\json" "-Isubprojects\icu\source\common" "-I..\..\..\subprojects\icu\source\common" "/MT" "/nologo" "/showIncludes" "/Zc:__cplusplus" "/W3" "/WX" "/EHsc" "/std:c++14" "/permissive-" "/O2" "/Gw" "/source-charset:utf-8" "-DU_STATIC_IMPLEMENTATION" "-DKMN_KBP" "-DKMN_KBP_EXPORTING" "-D_SCL_SECURE_NO_WARNINGS" "-D_CRT_SECURE_NO_WARNINGS" "-DUNICODE" "-DKMN_KBP_STATIC" "-DHAVE_ICU4C" "/Fdtests\unit\ldml\ldml.exe.p\ldml_test_source.cpp.pdb" /Fotests/unit/ldml/ldml.exe.p/ldml_test_source.cpp.obj "/c" ../../../tests/unit/ldml/ldml_test_source.cpp
[06:33:00 ](https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_Windows/376150?buildTab=log&focusLine=1646&logView=flowAware&linesState=1646)  ../../../tests/unit/ldml/ldml_test_source.cpp(620): error C2220: the following warning is treated as an error
[06:33:00 ](https://build.palaso.org/buildConfiguration/Keyman_Common_KPAPI_TestPullRequests_Windows/376150?buildTab=log&focusLine=1647&logView=flowAware&linesState=1647)  ../../../tests/unit/ldml/ldml_test_source.cpp(620): warning C4267: 'argument': conversion from 'size_t' to 'int32_t', possible loss of data
```

mcdurdin:
I'm keen to see the before-and-after sizes, in particular for the WASM build.

mcdurdin (changes requested):
I'm not quite ready to approve it yet :grin: would like to see the build passing on all platforms and have the additional docs on the subprojects folder (as well as file size delta data)

srl295:
> I'm not quite ready to approve it yet 😁 would like to see the build passing on all platforms and have the additional docs on the subprojects folder (as well as file size delta data)

100%.  I'm trying to think of how to make it work on Debian.

What would you think of falling back to `pkg-config` (i.e. including existing icu) if meson was too old? Or some kind of build switch.  Then Windows,  and anyone else with later meson, can download icu via the wrap, and debian can include the icu4c package which is already available there.

core/subprojects/icu-minimal.wrap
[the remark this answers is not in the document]
srl295: And once we have a later meson they will be 99% deleted. But yes, i'll add a note.

mcdurdin:
> What would you think of falling back to `pkg-config` (i.e. including existing icu) if meson was too old? Or some kind of build switch. Then Windows, and anyone else with later meson, can download icu via the wrap, and debian can include the icu4c package which is already available there.

Possible? @ermshiperete can you weigh in?

core/subprojects/packagefiles/icu/meson.build
```diff
@@ -0,0 +1,98 @@
...
+# cpp_native = meson.get_compiler('cpp', native: true)
+
+have_elf_h = cpp.has_header('elf.h')
+
+if host_machine.system() == 'windows'
+  add_project_arguments('-DU_PLATFORM_USES_ONLY_WIN32_API', language: 'cpp')
+  add_project_arguments('-DWIN32', '-DWIN64', '-D_MBCS', language: 'cpp')
+  add_project_arguments('-DU_PLATFORM_USES_ONLY_WIN32_API', language: 'c')
+  add_project_arguments('-DWIN32', '-DWIN64', '-D_MBCS', language: 'c')
+endif
```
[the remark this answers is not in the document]
srl295: this is preexisting from wrapdb.  I think `WIN32` is meant in a more generic sense here.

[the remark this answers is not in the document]
mcdurdin: Yes, `WIN32` is always set. but `WIN64` should only be set for x64 (and ARM64) targets.

core/tests/unit/ldml/ldml_test_source.cpp
```diff
@@ -482,6 +481,164 @@ int LdmlJsonTestSource::load(const nlohmann::json &data) {
...
+  UErrorCode status = U_ZERO_ERROR;
+  icu::UnicodeString pattern(chars.data(), chars.length());
+  uset = std::unique_ptr<icu::UnicodeSet>(new icu::UnicodeSet(pattern, status));
+  if (U_FAILURE(status)) {
+    // could be a malformed syntax isue
+    std::cerr << "UnicodeSet c'tor problem " << u_errorName(status) << std::endl;
+    return 1;
+  }
+  std::cout << "Got UnicodeSet of " << uset->size() << " char(s)." << std::endl;
+  #if (U_ICU_VERSION_MAJOR_NUM >= 70)
```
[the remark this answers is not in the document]
srl295: > > What would you think of falling back to `pkg-config` (i.e. including existing icu) if meson was too old? Or some kind of build switch. Then Windows, and anyone else with later meson, can download icu via the wrap, and debian can include the icu4c package which is already available there.
> 
> Possible? @ermshiperete can you weigh in?

see https://github.com/keymanapp/keyman/pull/8441/commits/0693e6436df27ceb8544df5bb7f4283ac353f420#diff-cb5653763d2d663a6a5174df34f9fe7fa49d0f2d6f522d6f468309b404220726R7-R8

it would work like that, behind a conditional.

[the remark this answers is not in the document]
srl295: the v70 has to do with the hasStrings function. But the general issue is dependent on where the spec finally lands. Like to leave this as a TODO-LDML for this clause.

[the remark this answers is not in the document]
srl295: There's another way to rewrite it without hasStrings that i'll come back to.

srl295:
> From [Windows build](https://build.palaso.org/viewLog.html?buildId=376150&buildTypeId=Keyman_Common_KPAPI_TestPullRequests_Windows):
> 
> ```
> 06:33:00   FAILED: tests/unit/ldml/ldml.exe.p/ldml_test_source.cpp.obj
> 06:33:00   "cl" "-Itests\unit\ldml\ldml.exe.p" "-Itests\unit\ldml" "-I..\..\..\tests\unit\ldml" "-Iinclude" "-I..\..\..\include" "-I..\..\..\..\common\include" "-Isrc" "-I..\..\..\src" "-I..\..\..\..\developer\src\ext\json" "-Isubprojects\icu\source\common" "-I..\..\..\subprojects\icu\source\common" "/MT" "/nologo" "/showIncludes" "/Zc:__cplusplus" "/W3" "/WX" "/EHsc" "/std:c++14" "/permissive-" "/O2" "/Gw" "/source-charset:utf-8" "-DU_STATIC_IMPLEMENTATION" "-DKMN_KBP" "-DKMN_KBP_EXPORTING" "-D_SCL_SECURE_NO_WARNINGS" "-D_CRT_SECURE_NO_WARNINGS" "-DUNICODE" "-DKMN_KBP_STATIC" "-DHAVE_ICU4C" "/Fdtests\unit\ldml\ldml.exe.p\ldml_test_source.cpp.pdb" /Fotests/unit/ldml/ldml.exe.p/ldml_test_source.cpp.obj "/c" ../../../tests/unit/ldml/ldml_test_source.cpp
> 06:33:00   ../../../tests/unit/ldml/ldml_test_source.cpp(620): error C2220: the following warning is treated as an error
> 06:33:00   ../../../tests/unit/ldml/ldml_test_source.cpp(620): warning C4267: 'argument': conversion from 'size_t' to 'int32_t', possible loss of data
> ```
> 
> From WASM build:
> 
> ```
> 06:17:53   Downloading icu-minimal source from https://github.com/unicode-org/icu/releases/download/release-72-1/icu4c-72_1-src.tgz
> 06:17:53   Download size: 26303933
> 06:17:55   Downloading: ..........
> 06:17:55   
> 06:17:55   tests/unit/ldml/meson.build:9:0: ERROR: Subproject exists but has no meson.build file
> ```
> 
> From Linux build:
> 
> ```
> 06:15:45   Downloading icu-minimal source from https://github.com/unicode-org/icu/releases/download/release-72-1/icu4c-72_1-src.tgz
> 06:15:45   Download size: 26303933
> 06:15:47   Downloading: ..........
> 06:15:47   
> 06:15:47   tests/unit/ldml/meson.build:9:0: ERROR: Subproject exists but has no meson.build file
> ```

There's no line 620 in that source file.

mcdurdin:
> There's no line 620 in that source file.

https://github.com/keymanapp/keyman/pull/8441/files#diff-b3dcd5df28d0cfcea5425d941f1e3618be88ea8c5a0501a2fd66687cd129405aR620

core/tests/unit/ldml/ldml_test_source.cpp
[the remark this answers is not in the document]
ermshiperete: Ubuntu 20.04 Focal comes with ICU 66, so  things would be easier if we can make that the minimum instead of v70.

ermshiperete:
> > What would you think of falling back to `pkg-config` (i.e. including existing icu) if meson was too old? Or some kind of build switch. Then Windows, and anyone else with later meson, can download icu via the wrap, and debian can include the icu4c package which is already available there.
> 
> Possible? @ermshiperete can you weigh in?

Yes, I think it would be beneficial to be able to use `pkg-config` for Debian/Ubuntu packaging. Otherwise the download step would have to happen when we create the source package because during a package build it's not possible to access the Internet. Additionally, Debian prefers dynamic libraries [[1]](https://wiki.debian.org/StaticLinking), [[2]](https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#s-sharedlibs-static).

core/subprojects/packagefiles/icu/source/io/meson.build
ermshiperete, line 1: Do we need this file if we don't include it in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/layoutex/meson.build
ermshiperete, line 1: Do we need this file if we don't include it in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/test/meson.build
ermshiperete, line 1: Do we need this file if we don't include it in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/ctestfw/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/meson.build
ermshiperete, line 1: Do we need this file if we don't include it in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/genbrk/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/genccode/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/gencmn/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/icuinfo/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/icupkg/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/pkgdata/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/subprojects/packagefiles/icu/source/tools/toolutil/meson.build
ermshiperete, line 1: Do we need this file if we don't include tools in `core/subprojects/packagefiles/icu/source/meson.build`?

core/tests/unit/ldml/ldml_test_source.cpp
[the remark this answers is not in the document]
srl295: ICU 66 is fine. This doesn't error, just has a different code path for <70

core/subprojects/packagefiles/icu-minimal.zip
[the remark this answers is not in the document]
srl295: We need this file because older meson doesn't support reading directly from the meson.build files.

core/subprojects/packagefiles/icu/source/io/meson.build
srl295: it's for consistency with upstream, and we might want to include it for tests later.

srl295:
we could have it prefer `pkg-config` on linux.  For now it's a meson version check.   Let's see if this actually builds.

ermshiperete:
> we could have it prefer `pkg-config` on linux. For now it's a meson version check. Let's see if this actually builds.

Yes, it would be good to have it prefer `pkg-config` on linux since the meson version check won't help with packaging on Jammy and newer. Ubuntu 22,.04 Jammy comes with [Meson 0.61](https://packages.ubuntu.com/search?suite=jammy&arch=any&mode=exactfilename&searchon=names&keywords=meson), upcoming 23.04 Lunar with [Meson 1.0.1](https://packages.ubuntu.com/search?suite=lunar&arch=any&mode=exactfilename&searchon=names&keywords=meson).

srl295:
Will fix

core/tests/unit/ldml/ldml_test_source.cpp
```diff
@@ -31,6 +31,15 @@
 #include "utfcodec.hpp"
 
 #include "ldml_test_source.hpp"
+#include "ldml_test_utils.hpp"
+
+// Optional ICU4C
+#if defined(HAVE_ICU4C)
+// ?!
```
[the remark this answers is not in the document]
srl295: fixed

core/subprojects/packagefiles/icu/meson.build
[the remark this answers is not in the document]
srl295: Filed in https://github.com/mesonbuild/wrapdb/issues/967 

Actually, ICU's own C++ build (in msvc) does this. So… it's compatible  🤷

srl295:
@mcdurdin 

> I'm not quite ready to approve it yet 😁 would like to see the build passing on all platforms

Should be passing now.

>  and have the additional docs on the subprojects folder

Done

>  (as well as file size delta data)

Should be ±0 for core, as this is linked into the ldml tests, not into the core lib yet.

srl295:
With this change:

`1.2M ./core/build/mac-x86_64/release/tests/unit/ldml/ldml`

Without this change:

`406K ./core/build/mac-x86_64/release/tests/unit/ldml/ldml`

srl295:
the penultimate commit has the builds. the ultimate one is only docs

mcdurdin:
New meson warning in build that we should address:

```
WARNING: Project targets '>=0.53.0' but uses feature introduced in '0.55.0': Local wrap patch files without patch_url.
```

mcdurdin:
ldml.wasm doubles in size from 495948 to 954686 bytes. We can't go to release with a 500kb jump in size for web platform for this dependency. I am assuming that the size increase is largely due to ICU?

For other platforms, size is not so important, but for web it is very important -- and right now we are pushing a lot of size boundaries in various areas and we need to keep working on it.

I think I'd like to see this resolved before it disappears into a flood of TODO items if possible.

core/subprojects/README.md
```diff
@@ -0,0 +1,21 @@
...
+- These are subprojects accessible to the `core` meson project.
+
+## icu
+
+### Why are there several files for ICU?
+
+- Ideally we would use meson's [wrapdb](https://mesonbuild.com/Wrapdb-projects.html), and in fact the files here are sourced from that project.
+- However, it requires a much newer meson version to work properly: 0.55 at least.
+- So the file `packagefiles/icu-minimal.zip` is generated from the files in [`packagefiles/icu`](./packagefiles/icu/)
+- Regenerate it with these commands:
```
mcdurdin, line 14: OK, so now that we are looking to use pkgconfig on Linux, we should be able to assume meson 0.55 for other platforms, so we don't need these files?

(There's also the question of which min ver of Debian to support for Keyman 17.0 (@ermshiperete?)

srl295:
> ldml.wasm doubles in size from 495948 to 954686 bytes. We can't go to release with a 500kb jump in size for web platform for this dependency. I am assuming that the size increase is largely due to ICU?


Yes. 


> For other platforms, size is not so important, but for web it is very important -- and right now we are pushing a lot of size boundaries in various areas and we need to keep working on it.

Any rough guidelines on boundaries ?

> I think I'd like to see this resolved before it disappears into a flood of TODO items if possible.

So as mentioned one challenge is the transform syntax, thus requirements, thus implementation is not final. Over slicing can be tricky to back out. 

Could we say that the size must be acceptable before transforms go into core depending on icu? Or is the ldml tests size itself a concern?

Just trying to understand the time and product dependencies.

core/subprojects/README.md
srl295: That's a question. What's the minimum meson for the non linux builds? I'll then simplify the files accordingly.

srl295: More specifically:
- at one level we can get rid of the zip file
- at another level we can get rid of most of the files, and only need to override somewhat for the trimming. 

Upstream in wrap db isn't very configurable. I'm planning to contribute some improvements (separately).

mcdurdin:
> Any rough guidelines on boundaries ?

So the current size of KeymanWeb is (before line compression) 392,929 bytes. This is already much bigger than I want it to be. A big chunk of this is ES6 polyfills for downlevel support (double-whammy here -- the people on the oldest devices can least afford the large download size but most need it). But there are also architectural choices and designs in KMW that are contributing to this.

See #3796 for more on this topic.

My preference would be for the LDML keyboard support to add at most 100KB to the size of KMW, but I recognise that this is probably unachievable. Okay, actually, my preference would be for the LDML keyboard code to _reduce_ the size of KMW but now I am really being silly.

> Could we say that the size must be acceptable before transforms go into core depending on icu?

My primary reason for blocking on web download size now is that I think that otherwise it will turn into something we don't care enough about and we'll end up going to release with a monster.

> Or is the ldml tests size itself a concern?

No, I don't care about unit test file size.

core/subprojects/README.md
mcdurdin: > What's the minimum meson for the non linux builds?

Min meson version for Keyman builds on Windows is already 1.0 and can be 1.0 on macOS as well. It is only Debian which blocks us here.

(Used to be that IE on Windows was our blocker. Today it is always one of Debian, Safari, or Android. How times have changed.)

Notes:

* mac build docs do not currently state a min meson version; brew.sh has [1.0.1](https://formulae.brew.sh/formula/meson#default) as current version.
* Windows build docs spec 1.0 as min version, and suggest using pip to install as at time of writing, chocolatey only had 0.55. choco now lists [1.0.1](https://community.chocolatey.org/packages/meson) also.

srl295:
- [ ] `tests/unit/ldml/meson.build:13:2: ERROR: Pkg-config binary for machine MachineChoice.HOST not found. Giving up.` — Skip pkg-config if the target is `wasm`.

core/tests/unit/ldml/meson.build
```diff
@@ -4,6 +4,25 @@
 # Authors:      Marc Durdin
 #
 
+# ICU4C is used for repertoire tests
+
+
+if build_machine.system() == 'linux'
```
srl295, line 10: change to if target == linux?

core/subprojects/packagefiles/icu/source/meson.build
```diff
@@ -0,0 +1,20 @@
...
+subdir('i18n')
+# subdir('io')
+#subdir('layout')
+#subdir('layoutex')
+
+# subdir('tools')
+
+# subdir('data')
+
+# subdir('test')
```
srl295, line 20: delete these subdirs

srl295:
@mcdurdin ptal

core/subprojects/README.md
srl295: I didn't change the minimum because that would break debian. However I changed the wrapfile to a minimum of 0.63.0 (it's what is on my local mac via homebrew) and I know it will function.

srl295: actually the mac build agent is 0.58.1

srl295: actually the mac build agent is 0.58.1

srl295:
🤦  but the linux server (even if it's building for wasm) still has the backlevel meson, which requires the .zipfile for patching. I'll add the .zipfile back in I suppose.

srl295:
reverted the zipfile…

mcdurdin:
> 🤦 but the linux server (even if it's building for wasm) still has the backlevel meson, which requires the .zipfile for patching. I'll add the .zipfile back in I suppose.

@ermshiperete we should be able to update meson on ba-bionic64ta right? It's only on packaging agents we can't do that.

core/subprojects/icu-minimal.wrap
```diff
@@ -0,0 +1,17 @@
+[wrap-file]
+directory = icu
+source_url = https://github.com/unicode-org/icu/releases/download/release-72-1/icu4c-72_1-src.tgz
```
mcdurdin, line 3: I thought @ermshiperete requested v66?

mcdurdin (commented):
So, just waiting on the clarification on meson version on linux agent (not packaging agent) as to whether we can avoid zip altogether.

srl295:
As a minimum, yes. Sent from my iPadOn Apr 12, 2023, at 9:43 PM, Marc Durdin ***@***.***> wrote:﻿
@mcdurdin commented on this pull request.

So, just waiting on the clarification on meson version on linux agent (not packaging agent) as to whether we can avoid zip altogether.

In core/subprojects/icu-minimal.wrap:
> @@ -0,0 +1,17 @@
+[wrap-file]
+directory = icu
+source_url = https://github.com/unicode-org/icu/releases/download/release-72-1/icu4c-72_1-src.tgz

I thought @ermshiperete requested v66?

srl295:
all passed

srl295:
@mcdurdin  down to +58 bytes, ok?  😆 

> Warning: keymanweb.js is 58 bytes (0%) larger than 17.0.86

mcdurdin:
So @srl295 are you okay with waiting on @ermshiperete's response next week before merging this, or is it blocking something else?

srl295:
[18:46:21 ](https://build.palaso.org/buildConfiguration/Keyman_Developer_Test/377658?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=377658_6812_6812&logView=flowAware)    1) KmpCompiler
[18:46:21 ](https://build.palaso.org/buildConfiguration/Keyman_Developer_Test/377658?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=377658_6813_6813&logView=flowAware)         should generates a valid .kmp (zip) file:
[18:46:21 ](https://build.palaso.org/buildConfiguration/Keyman_Developer_Test/377658?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=377658_6814_6814&logView=flowAware)       Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\BuildAgent\work\7ac43416c45637e9\keyman\developer\src\kmc-package\build\test\test-package-compiler.js)
[18:46:21 ](https://build.palaso.org/buildConfiguration/Keyman_Developer_Test/377658?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=377658_6815_6815&logView=flowAware)        at listOnTimeout (node:internal/timers:557:17)
[18:46:21 ](https://build.palaso.org/buildConfiguration/Keyman_Developer_Test/377658?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=377658_6816_6816&logView=flowAware)        at processTimers (node:internal/timers:500:7)


seems like something unrelated, rerunning

mcdurdin:
> seems like something unrelated, rerunning

I saw this on one of our agents with another build, checking the agents now. We may have to adjust the test timeout for that specific test.

ermshiperete:
> @ermshiperete we should be able to update meson on ba-bionic64ta right? It's only on packaging agents we can't do that.

Probably. I can't think of a reason why it should not work although I have the feeling I'm overlooking something. I guess we can try and find out if it works :smile:

srl295:
> > @ermshiperete we should be able to update meson on ba-bionic64ta right? It's only on packaging agents we can't do that.
> 
> Probably. I can't think of a reason why it should not work although I have the feeling I'm overlooking something. I guess we can try and find out if it works 😄

so, where are we on this? PR as it is requires icu to be installed beforehand if you're on linux, otherwise builds its own.  Ready to merge?

ermshiperete:
> so, where are we on this? PR as it is requires icu to be installed beforehand if you're on linux, otherwise builds its own. Ready to merge?

Fine for me (once the agent got updated), as long as it is possible to build with ICU 66 on linux.

core/subprojects/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+- These are subprojects accessible to the `core` meson project.
+
+## icu
+
+- Ideally we would use meson's [wrapdb](https://mesonbuild.com/Wrapdb-projects.html), and in fact the files here are sourced from that project.
+- However, it requires a much newer meson version to work properly: 0.55 at least.
+- Patches for icu (i.e. meson build files) are in [`packagefiles/icu`](./packagefiles/icu/)
+- The file `packagefiles/icu-minimal.zip` is generated from the files in [`packagefiles/icu`](./packagefiles/icu/)
+- Regenerate it with these commands:
```
mcdurdin, line 13: So, given we can build with meson 1.0+ on our ba-bionic64ta agent now (which is where we are limited on meson version), we should be able to rely on system-supplied ICU (v66) for Ubuntu and thus simplify the integration here, am I right?

If so, does that simplify the ICU integration any further than what we have now? Then this should be ready to merge!

srl295: We can drop the zip file (again!) that's the simplification we can do now. We can't use upstream wrap db AND minimize size yet. But I hope to upstream changes to make the icu configurable so we can someday depend on the wrap

mcdurdin: That's a good step forward, really happy to not include the .zip!

srl295: > That's a good step forward, really happy to not include the .zip!

You ok if I force push just to drop the last couple commits? That will remove the zip I re added.

mcdurdin: ok

srl295: Nevermind. That's not going to work with the merges.  Anyway, I did a revert and also another merge

core/subprojects/README.md
```diff
@@ -0,0 +1,11 @@
+# core/subprojects
+
+## What's this? What _is_ this?
+
+- These are subprojects accessible to the `core` meson project.
+
+## icu
+
+- Ideally we would use meson's [wrapdb](https://mesonbuild.com/Wrapdb-projects.html), and in fact the files here are sourced from that project.
+- However, it requires a much newer meson version to work properly: 0.55 at least.
```
mcdurdin, line 10: This statement is not quite correct now (given we are 1.0 on all platforms except for linux packaging agents?)

mcdurdin (approved):

core/subprojects/packagefiles/icu/meson.build
```diff
@@ -3,7 +3,7 @@ project(
   'c',
   'cpp',
   version: '72.1',
```
ermshiperete, line 5: I thought you'd use ICU 73 (as mentioned in `icu-minimal.wrap` !?

srl295: fixed

core/subprojects/README.md
srl295: I replaced it with a different statement.

srl295: When I wrote this, I still thought there was a way to customize the build without modifying the patchfiles. There isn't, but maybe someday there will be. (I filed the upstream ticket.)

core/subprojects/README.md
```diff
@@ -0,0 +1,12 @@
...
+## What's this? What _is_ this?
+
+- These are subprojects accessible to the `core` meson project.
+
+## icu
+
+- When building for Linux, [ICU](https://icu.unicode.org) must already be installed. The system ICU will be used instead of the subproject.
+- Patches for icu (i.e. meson build files) are in [`packagefiles/icu`](./packagefiles/icu/)
+- It would be preferable to use meson's [wrapdb](https://mesonbuild.com/Wrapdb-projects.html), and in fact the patchfiles here are sourced from that project.
+- However, Having our own patchfiles allows for slimming down ICU, at least until the wrapdb version is more configurable [mesonbuild/wrapdb#950](https://github.com/mesonbuild/wrapdb/issues/950)
```
ermshiperete, line 12:
```suggestion
- However, having our own patchfiles allows for slimming down ICU, at least until the wrapdb version is more configurable [mesonbuild/wrapdb#950](https://github.com/mesonbuild/wrapdb/issues/950)
```

ermshiperete (approved):

keyman-server:

PR #8442, opened 2023-03-16
fix(android/engine): Add builder output for configure

darcywong00:
Follow-on to #7407 

@bharanidharanj was getting exceptions loading test builds of "Keyman for Android". Turns out the KMEA builder configure step wasn't consistently triggering, so the engine got built missing the embedded KeymanWeb dependencies.

## User Testing
**Setup** - Install the PR build of Keyman for Android

* **TEST_KEYMAN** - Verifies Keyman app runs fine
1. Launch Keyman for Android
2. Verify OSK appears behind the "Get Started" menu
3. Verify exceptions **do NOT** appear about being unable to copy an asset file
[Screenshot of an exception](https://user-images.githubusercontent.com/19683143/225273142-d7f6221b-c5b2-41e1-9f81-40fa1c31d5b1.png) as reported

### DIFF
--- a/android/Samples/KMSample1/build.sh
+++ b/android/Samples/KMSample1/build.sh
@@ -29,6 +29,7 @@ builder_describe "Build KMSample1 app for Android." \
   "clean" \
   "configure" \
   "build" \
+  "test" \
   ":app                   KMSample1" \
   "--ci                   Don't start the Gradle daemon. Use for CI"
 
@@ -75,6 +76,10 @@ fi
 if builder_start_action build:app; then
   ./gradlew clean $SAMPLE_FLAGS
 
-
   builder_finish_action success build:app
+fi
+
+if builder_start_action test:app; then
+
+  builder_finish_action success test:app
 fi
--- a/android/Samples/KMSample1/build.sh
+++ b/android/Samples/KMSample1/build.sh
@@ -80,6 +80,6 @@ if builder_start_action build:app; then
 fi
 
 if builder_start_action test:app; then
-
+  # TODO: define tests
   builder_finish_action success test:app
 fi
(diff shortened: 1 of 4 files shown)


darcywong00:
I also notice top top level Android `./build.sh test` fails because a test step isn't defined in the Sample and Test apps. Should I define empty steps in those child scripts?

mcdurdin:
> I also notice top top level Android `./build.sh test` fails because a test step isn't defined in the Sample and Test apps. Should I define empty steps in those child scripts?

Yes. It's by design for parent/child scripts that all actions should be defined in the child scripts.

mcdurdin (approved):
LGTM. Good fix -- describing the outputs of the build is important!

mcdurdin:
> Turns out the KMEA builder configure step wasn't consistently triggering, so the engine got built missing the embedded KeymanWeb dependencies.

The call in CI should probably be `build.sh configure build` -- we like CI to be explicit about what it is doing.

android/Samples/KMSample1/build.sh
```diff
@@ -75,6 +76,10 @@ fi
...
   ./gradlew clean $SAMPLE_FLAGS
 
-
   builder_finish_action success build:app
 fi
+
+if builder_start_action test:app; then
+
+  builder_finish_action success test:app
+fi
```
mcdurdin, line 85: We can technically skip this -- it won't error -- but it's not a bad idea to include it

```suggestion
if builder_start_action test:app; then
  # TODO: define tests
  builder_finish_action success test:app
fi
```

bharanidharanj:
# Test Results

* **TEST_KEYMAN (PASSED):** Tested with the attached PR build (keyman 17.0.69-alpha-test-8442) in Android 12 / API 31 emulator and here is my observation: 1. Verified the OSK appears behind the "Get Started" menu. 2. Also, verified the error message did not appear on the screen.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.69-alpha](https://keyman.com/downloads/releases/alpha/17.0.69)

PR #6048, opened 2021-12-15
chore(common): Check in crowdin strings for Kibaku 📜

darcywong00:
This checks in the crowdin strings for the locale `ckl` for Kibaku. Does not include macOS.

- [x] TODO: @sgschantz to check on adding ckl (ckl-NG?) locale in XCode

## User Testing
For the listed platforms:
1. Load the PR build
2. Set the UI/locale to Kibaku
3. Verify the UI is Kibaku
Note: The crowdin translator left some strings in English, so that's out of our control

* **TEST_ANDROID**
1. Verify Kibaku strings are like the screenshot
kibaku

* **TEST_IOS** - Wait until Shawn addresses the XCode TODO before testing

* **TEST_LINUX**
1. start Keyman configuration with: `LANGUAGE=ckl_NG km-config`

* **TEST_WINDOWS**
1. start Keyman configuration and select UI language 'Nya Nasara'

PR #8447, opened 2023-03-16
chore(common): Update crowdin strings for Kibaku

darcywong00:
Follow-on to #6048

This updates crowdin strings for locale `ckl`.
@sgschantz - In the previous PR, you said the locale was unavailable in XCode for iOS. Do you know if that's still the case for iOS and macOS?

@keymanapp-test-bot skip

(May add iOS and macOS tests if those locales can be added)


mcdurdin (approved):

sgschantz:
I still don't see Kibaku or Cibak or anything similar available in Xcode.

PR #8456, opened 2023-03-16
chore(developer): fixup tests and build for kmcmplib

### LINKED ISSUE #8407: chore(developer): make kmcmplib tests independent of keyboards repo and legacy compiler
Currently, kmcmplib tests depend on both the [legacy compiler](https://github.com/keymanapp/keyman/blob/2ad068909007b8f59e531f0e5d555e84effe348c/developer/src/kmcmplib/tests/kmcompxtest.cpp#L69) and a local clone of the [keyboards repository](https://github.com/keymanapp/keyman/blob/2ad068909007b8f59e531f0e5d555e84effe348c/developer/src/kmcmplib/tests/get_test_source_exclude_some_kmn.bat#L24).

This is problematic long-term for a number of reasons:

1. The tests cannot run fully on linux or macOS, because of the legacy compiler dependency.
2. The legacy compiler is going away.
3. CI is made more complicated and slower by needing to clone the large keyboards repo.
4. The set of keyboards in the keyboards repository changes over time.

My proposed solution is:

1. Do a shallow clone of the keyboards repo [at a specific commit](https://stackoverflow.com/a/43136160/1836776) to ensure we match the keyboard source with the compiled artifacts, during test step, into a temp folder.
2. Keep a local copy of the .kmx files, as compiled by the legacy compiler, with `-no-compiler-version` and `-use-legacy-compiler` flags. Zipped, the .kmx artifacts are 4.2mb (we could remove three vietnamese keyboards to reduce to 2.4mb).

Query whether zipping is necessary -- 8.8mb unzipped without the three vietnamese keyboards? Primary concern: I feel like it makes a lot of binary file noise in the repo for day-to-day development.

It would be helpful to make this into a script so we can re-run the prep at any time (perhaps even with a specific version of the legacy compiler), without hardcoded paths.

We can then use kmcompx to compile the same set of keyboards and binary compare as we do currently, unzipping as part of the build process.

mcdurdin:
Fixes #8407.

Updates compilation of kmcmplib to support new MSVC version, and includes a set of precompiled fixtures from the legacy compiler that we can use to test against, as well as scripting to generate these fixtures.

Other minor tidy-ups also included here such as global .gitignore.

For performance, it's best to have a local clone of the https://github.com/keymanapp/keyboards repo, and set your `KEYBOARDS_ROOT` environment variable to point to the clone. The kmcmplib test configuration will clone _that_ repo and then check it out to a specific commit, to avoid messing up your own clone. (In future, we may consolidate these steps for all of Developer)

@keymanapp-test-bot skip


mcdurdin:
This is building -- but Developer test and release build configs in CI do not currently run kmcmplib tests at this point. I will correct this in a separate PR.

SabineSIL (approved):
lgtm

Issue #5111, opened 2021-05-19
bug(common/core/desktop): .kmx files don't work on big endian architecture


ermshiperete:
Running unit tests for keyman-keyboardprocessor fail on s390x architecture because it has a different byte order.

mcdurdin:
Resolving this should just be a matter of updating the `CopyKeyboard` and `FixupKeyboard` functions in common/core/desktop/src/kmx/kmx_file.cpp to reverse the relevant WORD and DWORD values in the stream at load (including, probably UTF-16 strings). The existing tests in common/core/desktop may well be sufficient to test the result on the target platforms.

PR #8476, opened 2023-03-20
chore(linux): Prevent building on s390x

ermshiperete:
This reverts commit 07ae306fd41e1c7d0e4274020e3656ac2c7afed8 which got merged in PR #6500 after a [discussion on the Debian mailing list](https://lists.debian.org/debian-input-method/2022/03/msg00095.html).

However, Keyman currently doesn't work on s390x architecture because we don't support big-endian (#5111). This shows in test failures during package build (cf [build log](https://buildd.debian.org/status/fetch.php?pkg=keyman&arch=s390x&ver=16.0.138-4&stamp=1676139609&raw=0)).

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/changelog
+++ b/linux/debian/changelog
@@ -1,3 +1,10 @@
+keyman (16.0.139-2) unstable; urgency=medium
+
+  * Don't build on s390x because Keyman doesn't work on big-endian architectures
+   (upstream bug https://github.com/keymanapp/keyman/issues/5111)
+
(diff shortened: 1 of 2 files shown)


ermshiperete:
> LGTM. Do you have a ref for the mailing list discussion?

https://lists.debian.org/debian-input-method/2022/03/msg00095.html. That doesn't explicitly talk about libkmnkbp, and I didn't search my  emails to reconstruct how this lead to making the change, but that was the trigger.

keyman-server:

PR #837, opened 2018-05-08
[android] cleanup console log

darcywong00:
Fixes #748 

* Remove nearly all uses of `Log.d()`
* Disable `window.console.log()` calls in `keyboard.html`

PR #8483, opened 2023-03-21
refactor(android/engine): Consolidate dispatchKey

darcywong00:
Continues refactor for #7881 

This consolidates `dispatchKey` into `KMKeyboardJSHandler.java`

## User Testing
**Setup** - Install the PR build of Keyman for Android on a physical device. Also pair an external keyboard (USB or bluetooth) to verify dispatchKey functionality. For all the test steps, use the external keyboard for typing and **NOT** the touch keyboard.

* **TEST_INAPP_KEYBOARD** - Verifies dispatchKey with the InApp keyboard
1. Launch Keyman for Android
2. With the external keyboard, type "qwerty"
3. Verify the text is output to the Keyman app
4. With the external keyboard, press the <kbd>ENTER</kbd> key
5. Verify a newline is output in the Keyman app
6. With the external keyboard, press the <kbd>TAB</kbd> key
7. Verify nothing happens (The Keyman app doesn't output tabspace)

* TEST_SYSTEM_KEYBOARD** - Verifies dispatchKey with the System keyboard
1. Launch Keyman for Android
2. From the "Get Started" menu, set Keyman as a system keyboard and enable as the default system keyboard
3. Exit the Keyman app
4. Launch a text application like Google Keep
5. Select the title field
6. With Keyman as the system keyboard, type from the external keyboard "qwerty"
7. Verify the text is output in the title field
8. With the external keyboard, press the <kbd>TAB</kbd> key
9. Verify Google Keep advances from the title to the the body of the app
10. Continue typing on the external keyboard
11. Verify the text is output to the body of the app
12. From the external keyboard, press the <kbd>ENTER</kbd> key
13. Verify a newline is output

### DIFF
--- a/android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
+++ b/android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
@@ -209,28 +209,30 @@ public abstract class KMKeyboardJSHandler {
     Handler mainLoop = new Handler(Looper.getMainLooper());
     mainLoop.post(new Runnable() {
       public void run() {
-        if (SystemKeyboard == null) {
+        if (k == null) {
           KMLog.LogError(TAG, "dispatchKey failed: SystemKeyboard is null");
           return;
         }
 
-        if (SystemKeyboard.subKeysWindow != null) {
+        if (k.subKeysWindow != null) {
           return;
         }
 
-        InputConnection ic = IMService.getCurrentInputConnection();
+        InputConnection ic = (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP) ?
+          KMTextView.activeView.onCreateInputConnection(new EditorInfo()) :
+          KMManager.getInputMethodService().getCurrentInputConnection();
         if (ic == null) {
-          if (isDebugMode()) {
-            Log.w(HANDLER_TAG, "insertText failed: InputConnection is null");
+          if (KMManager.isDebugMode()) {
+            Log.w(TAG, "insertText failed: InputConnection is null");
           }
           return;
         }
 
-        SystemKeyboard.dismissHelpBubble();
-        SystemKeyboard.setShouldShowHelpBubble(false);
+        k.dismissHelpBubble();
+        k.setShouldShowHelpBubble(false);
 
         // Handle tab or enter since KMW didn't process it
-        Log.d(HANDLER_TAG, "dispatchKey called with code: " + code + ", eventModifiers: " + eventModifiers);
+        Log.d(TAG, "dispatchKey called with code: " + code + ", eventModifiers: " + eventModifiers);
         if (code == KMScanCodeMap.scanCodeMap[KMScanCodeMap.KEY_TAB]) {
           KeyEvent event = new KeyEvent(0, 0, 0, KeyEvent.KEYCODE_TAB, 0, eventModifiers, 0, 0, 0);
           ic.sendKeyEvent(event);
--- a/android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
+++ b/android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
@@ -210,11 +210,19 @@ public abstract class KMKeyboardJSHandler {
     mainLoop.post(new Runnable() {
       public void run() {
         if (k == null) {
-          KMLog.LogError(TAG, "dispatchKey failed: SystemKeyboard is null");
+          KMLog.LogError(TAG, "dispatchKey failed: Keyboard is null");
           return;
         }
 
         if (k.subKeysWindow != null) {
+          return;
+        }
+
+        if (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP &&
+            (KMTextView.activeView == null || KMTextView.activeView.getClass() != KMTextView.class)) {
+          if (KMTextView.activeView == null && KMManager.isDebugMode()) {
+            Log.w(TAG, "dispatchKey failed: activeView is null");
+          }
           return;
         }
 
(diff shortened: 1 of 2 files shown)


android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
```diff
@@ -205,7 +205,63 @@ public void setIsChiral(boolean isChiral) {
...
+            (KMTextView.activeView == null || KMTextView.activeView.getClass() != KMTextView.class)) {
+          if (KMTextView.activeView == null && KMManager.isDebugMode()) {
+            Log.w(TAG, "dispatchKey failed: activeView is null");
+          }
+          return;
+        }
+
+        InputConnection ic = (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP) ?
+          KMTextView.activeView.onCreateInputConnection(new EditorInfo()) :
+          KMManager.getInputMethodService().getCurrentInputConnection();
```
mcdurdin, line 231: This feels like it should be a helper function that will probably be used elsewhere too. FWIW, I see 8 calls to `getCurrentInputConnection()` in our current master source.

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
```diff
@@ -205,7 +205,63 @@ public void setIsChiral(boolean isChiral) {
...
+          return;
+        }
+
+        if (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP &&
+            (KMTextView.activeView == null || KMTextView.activeView.getClass() != KMTextView.class)) {
+          if (KMTextView.activeView == null && KMManager.isDebugMode()) {
+            Log.w(TAG, "dispatchKey failed: activeView is null");
+          }
+          return;
+        }
```
mcdurdin, line 227: Can we split this into separate tests to make it clearer what is happening and that one of these conditions is an error case? Also, are similar checks going to be repeated elsewhere?

```suggestion
        if (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP) {
          if (KMTextView.activeView == null) {
            if (KMManager.isDebugMode()) {
              Log.w(TAG, "dispatchKey failed: activeView is null");
            }
            return;
          }
          if (KMTextView.activeView.getClass() != KMTextView.class)) {
            return;
          }
        }
```

Separate question on `isDebugMode`: is there any need to predicate on that for emitting warnings to the log? The log is not generally visible to end users anyway, and it simplifies our code significantly to remove all those `if (KMManager.isDebugMode())` checks!

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
```diff
@@ -205,7 +205,63 @@ public void setIsChiral(boolean isChiral) {
...
+        // Handle tab or enter since KMW didn't process it
+        Log.d(TAG, "dispatchKey called with code: " + code + ", eventModifiers: " + eventModifiers);
+        if (code == KMScanCodeMap.scanCodeMap[KMScanCodeMap.KEY_TAB]) {
+          KeyEvent event = new KeyEvent(0, 0, 0, KeyEvent.KEYCODE_TAB, 0, eventModifiers, 0, 0, 0);
+          if (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP) {
+            KMTextView textView = (KMTextView)KMTextView.activeView;
+            textView.dispatchKeyEvent(event);
+          } else {
+            ic.sendKeyEvent(event);
+          }
```
mcdurdin, line 251: This also feels like it should be a helper function, especially as it is repeated immediately below! c.f. `keyDownUp`

mcdurdin (changes requested):
Looking good, a few minor tweaks requested

darcywong00:
Back to draft to address review comments

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
darcywong00: I had to dig through the PR chronicles to #837 where most of the debug logging got removed. I think we maybe we convert these to Sentry logs (shouldn't be happening?)

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
darcywong00: I can refactor some of the calls. The rest are used for system keyboard apps:
* Keyman
* KMSample2
* FV Android

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
darcywong00: I originally accepted the suggestion, but later turned into a helper `inappKMTextViewInvalid()` to use within `insertText` and `dispatchKey`.

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
darcywong00: Refactored to use keyDownUp. I included special handling for InApp with KEYCODE_TAB.

Otherwise, dispatchKeyEvent(KEYCODE_TAB) was inserting newlines in the Keyman app.

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -459,6 +460,29 @@ public static void setInputMethodService(InputMethodService service) {
...
+   * @return InputConnection
+   */
+  protected static InputConnection getInputConnection() {
+    if (inputConnection != null) {
+      return inputConnection;
+    }
+
+    // Determine input connection
+    inputConnection = (IMService != null) ? IMService.getCurrentInputConnection() :
+      KMTextView.activeView.onCreateInputConnection(new EditorInfo());
```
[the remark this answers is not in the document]
darcywong00: I had to adjust `getInputConnection` to save off the input connection for `KEYBOARD_TYPE_INAPP`...

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardJSHandler.java
```diff
@@ -205,15 +197,59 @@ public void setIsChiral(boolean isChiral) {
...
+    return true;
+  }
+
+  private void keyDownUp(int keyEventCode, int eventModifiers) {
     if (k.keyboardType == KeyboardType.KEYBOARD_TYPE_INAPP) {
       KMTextView textView = (KMTextView)KMTextView.activeView;
-      textView.keyDownUp(KeyEvent.KEYCODE_ENTER);
+      if (keyEventCode == KeyEvent.KEYCODE_TAB) {
+        KeyEvent event = new KeyEvent(0, 0, 0, KeyEvent.KEYCODE_TAB, 0, eventModifiers, 0, 0, 0);
+        textView.dispatchKeyEvent(event);
```
darcywong00, line 245: This is to account for the subtle different in IN_APP handling of KEYCODE_TAB 

Currently from master:
https://github.com/keymanapp/keyman/blob/610b9b22579fcf1cd17cf307d8243ee73f243f45/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java#L2255-L2261

bharanidharanj:
# Test Results

* **TEST_INAPP_KEYBOARD (PASSED):** Tested with the attached PR build (Keyman 17.0.73-alpha-test-8483) in my Android Mobile device (Ver 11.0) and here is my observation: 1. For these test, I using the external (USB) keyboard attached to my Mobile. 2. Typing 'qwerty' in the external keyboard produces the same letters in the keyman text area. 3. Pressing Enter key outputs a new line in the text area. 4. I noticed that pressing Tab key makes the Predictive text blinking. Other than this, nothing happens. 

   


* **TEST_SYSTEM_KEYBOARD (PASSED):** 1. Verified that the qwerty name appears on the title field, using extneral (USB) Keyboard. 2. Pressing Tab key moves the cursor to the body of the app. 3. Able to type in the body of the app. 3. Pressing Enter key adds new line. 

   


* **TEST_FV_KEYBOARD (PASSED):** 1. Able to see 'qwerty' word in the title field. 2. Pressing Tab key moves the cursor to the body of the app. 3. Able to type in the body of the app. 4. Pressing Enter key, adds new line.

mcdurdin (approved):
LGTM!

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.76-alpha](https://keyman.com/downloads/releases/alpha/17.0.76)

PR #8491, opened 2023-03-22
chore(linux): Run and ignore autopkgtests on s390x

ermshiperete:
This change allows the autopkgtests to run on s390x architecture but immediately exits the test with a successful return code.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #8498, opened 2023-03-23
chore(common): use mac /usr/bin/stat rather than homebrew version

mcdurdin:
@keymanapp-test-bot skip

I am not 100% sure that this changed due to the macOS upgrade, but stat was now finding the coreutils version rather than the macOS version.

Later, we should probably standardize on the coreutils version so we can use it on all platforms.

### DIFF
--- a/resources/shellHelperFunctions.sh
+++ b/resources/shellHelperFunctions.sh
@@ -126,7 +126,7 @@ write_download_info() {
 
   FILE_EXTENSION="${BASE_FILE##*.}"
 
-  FILE_SIZE=$(stat -f"%z" "${BASE_PATH}/${BASE_FILE}")
+  FILE_SIZE=$(/usr/bin/stat -f"%z" "${BASE_PATH}/${BASE_FILE}")
   MD5_HASH=$(md5 -q "${BASE_PATH}/${BASE_FILE}")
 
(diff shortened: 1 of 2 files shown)


resources/shellHelperFunctions.sh
```diff
@@ -126,7 +126,7 @@ write_download_info() {
 
   FILE_EXTENSION="${BASE_FILE##*.}"
 
-  FILE_SIZE=$(stat -f"%z" "${BASE_PATH}/${BASE_FILE}")
+  FILE_SIZE=$(/usr/bin/stat -f"%z" "${BASE_PATH}/${BASE_FILE}")
```
darcywong00, line 129: Does this also apply to?
https://github.com/keymanapp/keyman/blob/471ebcea1e7ca4559338122983371612ac3f167a/mac/Keyman4MacIM/write-download_info.sh#L84

mcdurdin: Fixed

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.77-alpha](https://keyman.com/downloads/releases/alpha/17.0.77)

PR #8503, opened 2023-03-23
feat(core): use kmc to build kmx tests

mcdurdin:
Removes the Windows-only kmcomp.exe dependency, and instead now uses kmc to always build .kmx files for unit tests.

debug_api.cpp was using an outdated .kmx file, which is why the line number offsets have changed. (The perils of binary fixtures!)

No longer embeds files for wasm unit tests; instead uses nodefs to gain access to filesystem, which simplifies building the tests.

Note: see #8529 follow-up for changes which fix Linux builds.

@keymanapp-test-bot skip


common/include/build.sh
```diff
@@ -0,0 +1,41 @@
...
+#-------------------------------------------------------------------------------------------------------------------
+
+if builder_start_action clean; then
+  rm -f keymanversion_build.h .keymanversion_build.configured
+  builder_finish_action success clean
+fi
+
+if builder_start_action configure; then
+  # no-op, keeping 'configure' for deps
+  touch .keymanversion_build.configured
```
darcywong00, line 34: Is this just a placeholder file?

core/tests/unit/kmx/meson.build
```diff
@@ -15,22 +15,22 @@ else
...
+source_path = meson.current_source_dir()
+test_path = meson.current_build_dir()
+tests_flags = []
 
 if cpp_compiler.get_id() == 'emscripten'
-  tests_flags = ['--embed-file', join_paths(meson.current_source_dir(),'@/')]
+  # '-sMODULARIZE', '-sEXPORT_ES6',
+  tests_flags += ['-lnodefs.js', '-sEXPORTED_RUNTIME_METHODS=[\'UTF8ToString\']']
+
+  # tests_flags = ['--embed-file', join_paths(meson.current_source_dir(),'@/')]
```
rc-swag, line 26: Can these commented #test_flags assignments be removed?

common/include/build.sh
mcdurdin: Yes, it's just here to allow us to know that configuration is done. We have a configure action because all dependency scripts are required to have configure and build actions.

core/tests/unit/kmnkbd/debug_api.cpp
```diff
@@ -315,8 +316,8 @@ void test_store_offsets() {
       // store_offsets: 5 = store #, 1 = index into store, 6 = store #, 0 = index into store
```
darcywong00, line 316: Is this now?
```suggestion
      // store_offsets: 6 = store #, 1 = index into store, 6 = store #, 0 = index into store
```

darcywong00 (approved):
lgtm

rc-swag (approved):
lgtm

PR #8505, opened 2023-03-24
chore(linux): Revert "Run and ignore autopkgtests on s390x"

ermshiperete:
This basically reverts commit 2f716b3e517ecc4ad4c3a86479533d37fc55a4bc (#8491 .

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/changelog
+++ b/linux/debian/changelog
@@ -1,3 +1,9 @@
+keyman (16.0.139-4) unstable; urgency=medium
+
+  * debian/tests: Revert previous change and ignore s390x from autopkgtests
+
+ -- Eberhard Beilharz <eb1@sil.org>  Fri, 24 Mar 2023 16:05:07 +0100
+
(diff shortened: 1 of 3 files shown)


ermshiperete:
Debian is currently in [hard freeze](https://release.debian.org/testing/freeze_policy.html). Only packages with (successful) autopkgtests will migrate from _unstable_ to _testing_. We have autopkgtests, but those for s390x were not successful, so the previous PRs (and uploads to Debian) tried to get successful builds, first by running the test on s390x (which fails because libkmnkbp0 can be build, but doesn't work due to #5111). Then I tried to ignore s390x inside of the tests, but that fails because of dependencies. 
So this PR reverts to a non-hacky state and I have requested an unblock for the Keyman package ([Debian bug #1033571](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1033571)).

keyman-server:

PR #8509, opened 2023-03-24
feat(linux): Add dbus system service

### LINKED ISSUE #4273: feat(linux): Add support for Wayland

ermshiperete:
This implements a dbus system service that allows to provide the missing functionality when running under Wayland.

Fixes #4273.

# User Testing

## Preparations

- The tests should be run on these Linux platforms:
  - **GROUP_JAMMY_X11**: Ubuntu 22.04 Jammy with Gnome Shell and X11
  - **GROUP_JAMMY_WAYLAND**: Ubuntu 22.04 Jammy with Gnome Shell and Wayland
  - **GROUP_LUNAR_WAYLAND**: Ubuntu 23.04 Lunar with Gnome Shell and Wayland

- Reboot
- Install Nanum fonts for typing Korean:

  ```bash
  sudo apt install fonts-nanum
  ```

- Change "Complex text layout" setting in LibreOffice Writer:

  <details>
  <summary>Expand to see the steps</summary>

  - In LO Tools/Options, go to the Language Settings/Languages tab. Under "Default Languages for Documents" check the box for "Asian" and select "Korean (RoK)" as language, and check "Complex text layout" and select "Khmer" as language.

    Screenshot from 2022-06-23 12-13-56

  - You'll also have to change the default style: go to Styles/Manage Styles, right-click on "Default Paragraph Style" and select Modify. Go to the Font tab. For Asian Text Font, select Family "NanumMyeongjo" and Language "Korean (RoK)". For CTL Font, select Family "Khmer Mondulkiri" (or another "Khmer" font) and Language "Khmer".

    Screenshot from 2022-06-23 12-11-35
  </details>
</details>

 - Install "IPA (SIL)", "Korean KORDA Jamo (SIL)" , and "Khmer Angkor" keyman keyboards

## Tests

- **TEST_IPA_WRITER:** Open LibreOffice Writer. Switch to "IPA (SIL)" keyboard. Type `n>`. Verify that the result is "ŋ".
- **TEST_IPA_GEDIT:** Open gedit. Switch to "IPA (SIL)" keyboard. Type `n>`. Verify that the result is "ŋ".
- **TEST_KO_WRITER:** Open LibreOffice Writer. Switch to "Korean KORDA Jamo (SIL)" keyboard. Type `han<space>geul<space>`. Verify that the result is "한글".
- **TEST_KO_GEDIT:** Open gedit. Switch to "Korean KORDA Jamo (SIL)" keyboard. Type `han<space>geul<space>`. Verify that the result is "한글".
- **TEST_KM_WRITER:** Open LibreOffice Writer. Switch to "Khmer Angkor" keyboard. Type `xEjmr`. Verify that the output is "ខ្មែរ".
- **TEST_KM_GEDIT:** Open gedit. Switch to "Khmer Angkor" keyboard. Type `xEjmr`. Verify that the output is "ខ្មែរ".
- **TEST_CONTEXT:**
  - Open gedit
  - Switch to "IPA (SIL)" keyboard.
  - Type
    ```
    an
    m
    ```
  - click after `n`
  - type `>`
  - verify that the output is
    ```
    aŋ
    m
    ```


linux/ibus-keyman/meson.build
```diff
@@ -9,10 +9,10 @@ project('ibus-keyman', 'c', 'cpp',
 cc = meson.get_compiler('c')
 conf = configuration_data()
 
-ibus =        dependency('ibus-1.0',             version: '>= 1.2.0')
-gtk =         dependency('gtk+-3.0',             version: '>= 2.4')
-x11 =         dependency('x11',                  version: '>= 1.6')
-json_glib =   dependency('json-glib-1.0',        version: '>= 1.0')
+ibus =           dependency('ibus-1.0',             version: '>= 1.2.0')
+gtk =            dependency('gtk+-3.0',             version: '>= 2.4')
+json_glib =      dependency('json-glib-1.0',        version: '>= 1.0')
```
darcywong00, line 14: Is x11 still used on older distros?

linux/ibus-keyman/src/test/meson.build
```diff
@@ -3,7 +3,7 @@ keymanutil_sources = [
   util_files,
 ]
 
-keymanutil_deps = [ibus, gtk, x11, json_glib, kmnkbp_lib]
+keymanutil_deps = [ibus, gtk, json_glib, kmnkbp_lib]
```
darcywong00, line 6: Does this need systemd?

linux/ibus-keyman/meson.build
ermshiperete: Yes, X11 can still be used, even on newer distros, but this code no longer has a dependency on it (turns out that the way I implemented capslock handling for Wayland also works on X11 and so we no longer need the distinction between X11 and Wayland. In fact, the new code works better - previously when we programmatically turned caps lock off the caps lock indicator was still on until after the next key press. With the new code it immediately turns off).

linux/ibus-keyman/src/test/meson.build
ermshiperete: No. We don't have any tests in this directory that access dbus (which is the reason for the systemd dependency).

linux/ibus-keyman/tests/scripts/teardown-tests.sh
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -eu
-PID_FILE=$1
 
-echo "Shutting down processes..."
-bash "$PID_FILE" > /dev/null 2>&1
-rm "$PID_FILE"
-echo "Finished shutdown of processes."
+. "$(dirname "$0")/test-helper.sh"
```
darcywong00, line 4: @mcdurdin has been having me rename included scripts as `test-helper.inc.sh`.

linux/ibus-keyman/tests/scripts/test-helper.sh
```diff
@@ -92,3 +92,136 @@ function link_test_keyboards() {
...
+    # mutter-Message: 18:56:15.422: Using Wayland display name 'wayland-1'
+    mutter --wayland --headless --no-x11 --virtual-monitor 1024x768 &> "$TMPFILE" &
+    echo "kill -9 $! || true" >> "$PID_FILE"
+    sleep 1s
+    export WAYLAND_DISPLAY
+    WAYLAND_DISPLAY=$(grep "Using Wayland display" "$TMPFILE" | cut -d"'" -f2)
+    rm "$TMPFILE"
+    echo "export WAYLAND_DISPLAY=\"$WAYLAND_DISPLAY\"" >> "$ENV_FILE"
+  else
+    echo "Starting Xvfb..."
```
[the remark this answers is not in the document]
ermshiperete: Yes. Updated the output.

mcdurdin:
I hope to review this in next few days

bharanidharanj:
# Test Results

### GROUP_LUNAR_WAYLAND: Ubuntu 23.04 Lunar with Gnome Shell and Wayland

* **TEST_IPA_WRITER (PASSED):** Tested with the attached PR build (17.0.108-1~PR-8509+lunar48) in Ubuntu 23.04 Lunar with Gnome Shell and Wayland OS and here is my observation: 1. Verified that the typing n> produces the output ŋ in LibreOffice Writer Document.
* **TEST_IPA_GEDIT (PASSED):** Verified that the typing n> produces the output ŋ in Text Editor.
* **TEST_KO_WRITER (PASSED):** Verified that typing han,space>geul produces the output 한글 in LibreOffice Writer Document.
* **TEST_KO_GEDIT (PASSED):** Verified that typing han,space>geul produces the output 한글 in Text Editor.
* **TEST_KM_WRITER (PASSED):** Verified that typing xEjmr produces the ouput ខ្មែរ in LibreOffice Writer Document.
* **TEST_KM_GEDIT (PASSED):** Verified that typing xEjmr produces the ouput ខ្មែរ in Text Editor.
* **TEST_CONTEXT (FAILED):** Noticed that after typing > after the letter n, it is showing wrong output. Seems to be an issue. 

..in LibreOffice Writer

   


..in Text Editor

linux/ibus-keyman/tests/km-dbus-test-server.c
```diff
@@ -0,0 +1,140 @@
...
+loop() {
+  sd_bus_slot *slot = NULL;
+  sd_bus *bus       = NULL;
+  int ret;
+
+  ret = sd_bus_open_user(&bus);
+
+  if (ret < 0) {
+    g_error("Failed to connect to system bus: %s", strerror(-ret));
+    return;
```
darcywong00, line 77: Does this need `exit_loop(slot, bus);`?

linux/ibus-keyman/tests/scripts/test-helper.inc.sh
```diff
@@ -0,0 +1,229 @@
...
+    fi
+    echo "Running on Wayland:"
+    echo "Starting mutter..."
+    TMPFILE=$(mktemp)
+    # mutter-Message: 18:56:15.422: Using Wayland display name 'wayland-1'
+    mutter --wayland --headless --no-x11 --virtual-monitor 1024x768 &> "$TMPFILE" &
+    echo "kill -9 $! || true" >> "$PID_FILE"
+    sleep 1s
+    export WAYLAND_DISPLAY
+    WAYLAND_DISPLAY=$(grep "Using Wayland display" "$TMPFILE" | cut -d"'" -f2)
```
darcywong00, line 156: I notice this is how the previous code was. But why wouldn't the two lines be combined?
```suggestion
    export WAYLAND_DISPLAY=$(grep "Using Wayland display" "$TMPFILE" | cut -d"'" -f2)
```

linux/ibus-keyman/src/keyman-system-service-client.c
```diff
@@ -0,0 +1,90 @@
...
+
+void set_capslock_indicator(
+  guint32 capsLock
+) {
+  sd_bus_error error   = SD_BUS_ERROR_NULL;
+  sd_bus_message* msg  = NULL;
+  sd_bus* bus          = NULL;
+  int result;
+
+  if (capsLock > 1) {
```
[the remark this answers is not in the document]
ermshiperete: Any value besides 0 and 1 is invalid.

linux/ibus-keyman/tests/km-dbus-test-server.c
ermshiperete: Not really, since opening the user bus failed `bus` is not set, nor `slot`, but for consistency it would be better to add it.

linux/ibus-keyman/tests/scripts/test-helper.inc.sh
ermshiperete: Combining it masks failures. See https://www.shellcheck.net/wiki/SC2155.

darcywong00 (commented):
lgtm - I think @mcdurdin wants to get to this too.

ermshiperete:
@keymanapp-test-bot skip GROUP_LUNAR_WAYLAND TEST_CONTEXT

I created #8828 for the failing test.

mcdurdin:
@ermshiperete note the format change for your test skip (the format you used skipped _all_ tests)

linux/ibus-keyman/tests/km-dbus-test-server.c
mcdurdin: This is because [C++ RAII](https://stackoverflow.com/questions/161177/does-c-support-finally-blocks-and-whats-this-raii-i-keep-hearing-about) means we don't get nice things (`try`/`finally`) -- otherwise we need to wrap these variables in a class for automatic destruction, which is tedious boilerplate. Another pattern I see used is `goto loop_cleanup`, which at least can keep the cleanup inside the same function, and is thus a little clearer, a little shorter, and avoids accidentally missing the `return` statement, or calling the cleanup twice (which in the current model would be _bad_ because it would result in double-frees):

```
  if (ret < 0) {
    g_error("Failed to connect to system bus: %s", strerror(-ret));
    goto cleanup;
  }

  // ... function continues ...

cleanup:
  if (bus)  sd_bus_release_name(bus, KEYMAN_TESTSVC_BUS_NAME);
  if (slot) sd_bus_slot_unref(slot);
  if (bus)  sd_bus_close_unref(bus);
}
```

linux/ibus-keyman/tests/km-dbus-test-server.c
```diff
@@ -0,0 +1,141 @@
...
+      return;
+    }
+    if (exitFlag) {
+      g_debug("Exiting loop");
+      exit_loop(slot, bus);
+      return;
+    }
+
+    if (ret > 0)  // we processed a request, try to process another one, right-away
+      continue;
```
mcdurdin, line 116:
```suggestion
    if (ret > 0) {
      // we processed a request, try to process another one, right-away
      continue;
    }
```

linux/ibus-keyman/tests/km-dbus-test-server.c
```diff
@@ -0,0 +1,141 @@
...
+    if (ret < 0) {
+      g_error("Failed to process bus: %s", strerror(-ret));
+      exit_loop(slot, bus);
+      return;
+    }
+    if (exitFlag) {
+      g_debug("Exiting loop");
+      exit_loop(slot, bus);
+      return;
+    }
```
mcdurdin, line 113:
```suggestion
    if (exitFlag) {
      // `exitFlag` can be modified by the callback function
      // `on_exit_method` which can be called by `sd_bus_process`
      g_debug("Exiting loop");
      exit_loop(slot, bus);
      return;
    }
```

linux/ibus-keyman/src/keyman-system-service-client.c
```diff
@@ -0,0 +1,90 @@
...
+#define KEYMAN_BUS_NAME       "com.keyman.SystemService1"
+#define KEYMAN_INTERFACE_NAME "com.keyman.SystemService1.System"
+#define KEYMAN_OBJECT_PATH    "/com/keyman/SystemService1/System"
+
+static void
+finish(sd_bus_error* error, sd_bus_message* msg, sd_bus * bus) {
+  if (error) { sd_bus_error_free(error); }
+  if (msg)   { sd_bus_message_unref(msg); }
+  if (bus)   { sd_bus_unref(bus); }
+}
```
mcdurdin, line 15: I think I'd prefer to see the `goto cleanup` pattern which I've noted in another comment -- it's less vulnerable to copypasta errors.

linux/ibus-keyman/tests/scripts/run-tests.sh
```diff
@@ -57,68 +52,22 @@ function run_tests() {
...
-  elif [ -d "${COMMON_ARCH_DIR}"/debug ]; then
-    COMMON_ARCH_DIR=${COMMON_ARCH_DIR}/debug
+  if [ -d "$(dirname "$0")/../../../build/$(arch)/debug" ]; then
+    CONFIG=debug
+  elif [ -d "$(dirname "$0")/../../../build/$(arch)/release" ]; then
+    CONFIG=release
   else
-    echo "Can't find neither ${COMMON_ARCH_DIR}/release nor ${COMMON_ARCH_DIR}/debug"
-    exit 2
+    echo "Can't find neither ../../../build/$(arch)/debug nor ../../../build/$(arch)/release"
```
mcdurdin, line 60:
```suggestion
    echo "Cannot find ../../../build/$(arch)/debug or ../../../build/$(arch)/release"
```

Yay English grammar (or ... "Can find neither a nor b" is technically okay but weird)

linux/ibus-keyman/tests/stop-test-server.c
```diff
@@ -0,0 +1,34 @@
+#include <systemd/sd-bus.h>
+
+void
+finish(sd_bus_error* error, sd_bus_message* msg, sd_bus * bus) {
+  if (error) { sd_bus_error_free(error); }
+  if (msg)   { sd_bus_message_unref(msg); }
+  if (bus)   { sd_bus_unref(bus); }
+}
```
mcdurdin, line 8: same `goto cleanup` pattern preferred I think

linux/keyman-system-service/build.sh
```diff
@@ -0,0 +1,73 @@
...
+  "build" \
+  "test" \
+  "install                   install artifacts" \
+  "uninstall                 uninstall artifacts"
+
+builder_parse "$@"
+
+cd "$THIS_SCRIPT_PATH/"
+
+if builder_has_option --debug; then
```
mcdurdin, line 25:
```suggestion
if builder_is_debug_build; then
```

linux/keyman-system-service/build.sh
```diff
@@ -0,0 +1,73 @@
...
+MESON_PATH="build/$(uname -m)/$MESON_TARGET"
+
+builder_describe_outputs \
+  configure "${MESON_PATH}/build.ninja" \
+  build "${MESON_PATH}/src/keyman-system-service"
+
+if builder_start_action clean; then
+  rm -rf "$THIS_SCRIPT_PATH/build/"
+  builder_finish_action success clean
+fi
```
mcdurdin, line 42:
```suggestion
builder_run_action clean rm -rf "$THIS_SCRIPT_PATH/build/"
```

linux/keyman-system-service/build.sh
```diff
@@ -0,0 +1,73 @@
...
+
+if builder_start_action configure; then
+  cd "$THIS_SCRIPT_PATH"
+  # shellcheck disable=SC2086
+  meson setup "$MESON_PATH" --werror --buildtype $MESON_TARGET "${builder_extra_params[@]}"
+  builder_finish_action success configure
+fi
```
mcdurdin, line 49:
```suggestion
# shellcheck disable=SC2086
builder_run_action configure meson setup "$MESON_PATH" --werror --buildtype $MESON_TARGET "${builder_extra_params[@]}"
```

linux/keyman-system-service/src/KeyboardDevice.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+  this->dev = nullptr;
+  this->fd  = -1;
+  this->hasCapsLockLed = -1;
+  this->debug          = false;
+}
+
+KeyboardDevice::~KeyboardDevice()
+{
+  libevdev_free(this->dev);
```
mcdurdin, line 21: Is this safe to call in failure cases (e.g. `Initialize` fails and `this->dev` is never set)?

linux/keyman-system-service/src/ServerMethods.cpp
```diff
@@ -0,0 +1,80 @@
...
+  void *data
+) {
+  std::list<KeyboardDevice*> *kbd_devices = static_cast<std::list<KeyboardDevice*>*>(data);
+  bool state = false;
+  for (KeyboardDevice *kbdDevice : *kbd_devices) {
+    if (!kbdDevice->HasCapsLockLed()) {
+      syslog(LOG_USER | LOG_NOTICE, "%s: No kbdDevice or no CapsLockLed", __FUNCTION__);
+      continue;
+    }
+    state |= kbdDevice->GetCapsLockLed();
```
mcdurdin, line 77:
```suggestion
    state ||= kbdDevice->GetCapsLockLed();
```

Technically a bool should be a logical-or, not a bitwise-or. Can we document that this function will return true if _any_ keyboard device has its caps lock led lit?

linux/keyman-system-service/src/dbus-service.c
```diff
@@ -0,0 +1,121 @@
...
+static int32_t
+on_set_caps_lock_indicator(
+  sd_bus_message *msg,
+  void *user_data,
+  sd_bus_error *ret_error
+) {
+  int32_t ret;
+  uint32_t state;
+
+  ret_error = NULL;
```
mcdurdin, line 23: This is a no-op. Did you mean to set the value rather than the pointer?

```suggestion
  *ret_error = NULL;
```

linux/keyman-system-service/src/dbus-service.c
```diff
@@ -0,0 +1,121 @@
...
+
+static int32_t
+on_get_caps_lock_indicator(
+  sd_bus_message *msg,
+  void *user_data,
+  sd_bus_error *ret_error
+) {
+  uint32_t state;
+
+  ret_error = NULL;
```
mcdurdin, line 43: Did you mean to do this?

```suggestion
  *ret_error = NULL;
```

linux/keyman-system-service/src/dbus-service.c
```diff
@@ -0,0 +1,121 @@
...
+  SD_BUS_VTABLE_END
+};
+
+static int finish(int ret, void *kbd_devices, sd_bus_slot *slot, sd_bus *bus) {
+  close_kbd_devices(kbd_devices);
+  sd_bus_slot_unref(slot);
+  sd_bus_unref(bus);
+
+  return ret < 0 ? 1 : 0;
+}
```
mcdurdin, line 62: Again, would prefer a `goto cleanup` pattern

linux/keyman-system-service/src/dbus-service.c
```diff
@@ -0,0 +1,121 @@
...
+
+  for (;;) {
+    // Process requests
+    ret = sd_bus_process(bus, NULL);
+    if (ret < 0) {
+      syslog(LOG_USER | LOG_NOTICE, "Failed to process bus: %s", strerror(-ret));
+      return finish(ret, kbd_devices, slot, bus);
+    }
+    if (ret > 0) // we processed a request, try to process another one, right-away
+      continue;
```
mcdurdin, line 110:
```suggestion
    if (ret > 0) {
      // we processed a request, try to process another one, right-away
      continue;
    }
```

linux/scripts/dist.sh
```diff
@@ -33,6 +33,8 @@ dch keyman --newversion "${VERSION}" --force-bad-version --nomultimaint
 dpkg-source --tar-ignore=*~ --tar-ignore=.git --tar-ignore=.gitattributes \
```
mcdurdin, line 33: Still crying about how awful this exclusion model is :grin:

mcdurdin (changes requested):
This looks really solid. I have one request around resource cleanup patterns, because I think the separate-function pattern is a bit fragile. But I didn't really see any other significant issues. Good job!

linux/keyman-system-service/src/dbus-service.c
ermshiperete: Refactored to make use of C++ RAII.

linux/keyman-system-service/src/ServerMethods.cpp
mcdurdin: Ha ha, my bad, mixing languages again!

linux/ibus-keyman/tests/km-dbus-test-server.c
mcdurdin: > I don't like gotos, so I put the whole thing in a class to benefit from C++ RAII 😄

I thought there was a good chance that you would say that! 🤣

linux/ibus-keyman/src/KeymanSystemServiceClient.cpp
```diff
@@ -0,0 +1,97 @@
...
+  if (result < 0) {
+    g_error("%s: Can't get connection: %s", __FUNCTION__, strerror(-result));
+  }
+}
+
+KeymanSystemServiceClient::~KeymanSystemServiceClient() {
+  if (error) { sd_bus_error_free(error); }
+  if (msg)   { sd_bus_message_unref(msg); }
+  if (bus)   { sd_bus_unref(bus); }
+}
```
mcdurdin, line 40: In the pattern you have now, if you call `SetCapsLockIndicator` twice and both times an error is returned, you'll leak `error`. It would be good to make note of this -- that the class functions are intended to be called just once in the class lifecycle, for future maintainers.

linux/ibus-keyman/tests/KmDbusTestServer.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+  dbus  = g_test_dbus_new(G_TEST_DBUS_NONE);
+
+  // Add the private directory with our in-tree service files.
+  g_test_dbus_add_service_dir(dbus, KEYMAN_TEST_SERVICE_PATH);
+
+  std::cout << "Add service dir to " << KEYMAN_TEST_SERVICE_PATH << std::endl;
+
+  // Start the private D-Bus daemon
+  g_test_dbus_up(dbus);
```
mcdurdin, line 58: What happens if any of these functions fail?

linux/keyman-system-service/build.sh
```diff
@@ -0,0 +1,66 @@
...
+  cd "$MESON_PATH"
+  ninja install
+  builder_finish_action success install
+fi
+
+if builder_start_action uninstall; then
+  cd "$MESON_PATH"
+  ninja uninstall
+  builder_finish_action success uninstall
+fi
```
mcdurdin, line 66: You _could_ move `cd "$MESON_PATH"` out of the if blocks and then turn these into `builder_run_action` statements if you wanted to. But it's not important.

mcdurdin (approved):
LGTM!

linux/keyman-system-service/build.sh
ermshiperete: Well, there's one complication: If these gets run as part of `linux/build.sh clean` it will fail because it can't cd into `$MESON_PATH` because that got deleted in clean action. 

The other a bit unfortunate thing is that if I run e.g. `linux/keyman-system-service/build.sh build` the output will also show the actions that come after the build action which is a bit confusing:

```
...
[linux/keyman-system-service] ## build starting...
[8/8] Linking target src/keyman-system-service
[linux/keyman-system-service] ## build completed successfully
builder_run_action test meson test --print-errorlogs
builder_run_action install ninja install
builder_run_action uninstall ninja uninstall
```

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.111-alpha](https://keyman.com/downloads/releases/alpha/17.0.111)

linux/keyman-system-service/build.sh
mcdurdin: > The other a bit unfortunate thing is that if I run e.g. linux/keyman-system-service/build.sh --debug build the output will also show the actions that come after the build action which is a bit confusing:

This has been fixed in f8535c469effd4b57fed0e3731f6f48d64ba8af6 (#8834)

mcdurdin: > Well, there's one complication: If these get run as part of linux/build.sh clean it will fail because it can't cd into $MESON_PATH because that got deleted in the clean action.

Ouch. Yeah, that's a shame... well, it looked good at the time!

PR #8513, opened 2023-03-24
feat(developer): support jis and abnt2 🙀

srl295:
Fixes: #8161 

@keymanapp-test-bot skip


common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -161,16 +161,42 @@ export const ISOVirtualKeyMap: KeyMap = [
   [ k.K_SPACE ],
 ];
 
+export const JISVirtualKeyMap: KeyMap = [
+  // [Hankaku/Zenkaku] 1 2 3 4 5 6 7 8 9 0 - ^ ¥ [bksp]
+  [ k.K_1, k.K_2, k.K_3, k.K_4, k.K_5, k.K_6, k.K_7, k.K_8, k.K_9, k.K_0, k.K_HYPHEN, k.K_EQUAL, k.K_oE2 /* YEN? */ ],
```
srl295, line 166: Not 1000.0‰ sure about that k_oE2 at the end there.

core/tests/unit/ldml/keyboards/meson.build
```diff
@@ -44,15 +59,34 @@ foreach kbd : tests
...
+    input: join_paths(ldml_data, kbd + '.xml'),
+    output: kbd + '.kmx',
+  )
+  configure_file(
+    command: kmc_cmd + ['build-test-data', '@INPUT@', '--out-file', '@OUTPUT@'],
+    input: join_paths(ldml_testdata, kbd + '-test.xml'),
     output: kbd + '-test.json',
-    input: kbd + '-test.xml'
   )
 endforeach
```
srl295, line 92: copypasta and probably more complex than it should be. There are 3 different kinds of sample keyboard files at present. 

The good thing here is it's loading a subset directly out of CLDR!

srl295:
```console
  tests\unit\ldml\keyboards\meson.build:77:2: ERROR: Command "C:\WINDOWS\system32\cmd.exe /c copy C:/BuildAgent/work/99b311828f4ee7c/keyman/core/../resources/standards-data/ldml-keyboards/techpreview/3.0/ja-Latn.xml C:\BuildAgent\work\99b311828f4ee7c\keyman\core\build\x86\release\tests\unit\ldml\keyboards\ja-Latn.xml" failed with status 1.
```

srl295:
fixing the first set of paths to forward slashes works.

srl295:
i can repro locally

mcdurdin:
iOS build failure just due to agent upgrade, will be resolved with merge to master, no need to rebuild because it isn't used here anyway.

mcdurdin:
> `tests\unit\ldml\keyboards\meson.build:77:2: ERROR: Command "C:\WINDOWS\system32\cmd.exe /c copy C:/BuildAgent/work/99b311828f4ee7c/keyman/core/../resources/standards-data/ldml-keyboards/techpreview/3.0/ja-Latn.xml C:\BuildAgent\work\99b311828f4ee7c\keyman\core\build\x86\release\tests\unit\ldml\keyboards\ja-Latn.xml" failed with status 1.`
> 
> Hrmm. I'lltry to debug this locally.

https://stackoverflow.com/a/61128721/1836776 is one possible answer. But we are now on meson 0.53+ so we may also be able to use [`configure_file`](https://mesonbuild.com/Reference-manual_functions.html#configure_file) with the `copy` keyword argument, which would be better.

srl295:
Filed https://github.com/keymanapp/keyman/issues/8524 for the path issue

core/tests/unit/ldml/keyboards/meson.build
```diff
@@ -44,15 +62,34 @@ foreach kbd : tests
...
+  )
+endforeach
+
+
+foreach kbd : tests_from_cldr
+  configure_file(
+    command: copy_cmd + ['@INPUT@', '@OUTPUT@'],
+    input: join_paths(ldml_data, kbd + '.xml'),
+    output: kbd + '.xml'
+  )
```
mcdurdin, line 84:
```suggestion
  configure_file(
    copy: true,
    input: ldml_data / kbd + '.xml',
    output: kbd + '.xml'
  )
```

core/tests/unit/ldml/keyboards/meson.build
```diff
@@ -26,16 +28,32 @@ tests = [
...
+ldml_data = join_paths(ldml_root, '3.0')
+ldml_testdata = join_paths(ldml_root, 'test')
 kmc_cmd = [node, kmc_root]
 
 # Build all keyboards in output folder
 
-foreach kbd : tests
+foreach kbd : tests_needing_copy
   configure_file(
     command: copy_cmd + ['@INPUT@', '@OUTPUT@'],
```
mcdurdin, line 58:
```suggestion
    copy: true,
```

core/tests/unit/ldml/keyboards/pt-BR-k0-abnt2.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE keyboard SYSTEM "../dtd/ldmlKeyboard.dtd">
+<keyboard locale="pt-BR-t-k0-abnt2" conformsTo="techpreview">
+	<locales>
+		<locale id="pt" />
+	</locales>
+	<version number="0.0.0" />
+	<names>
+		<!-- this is the Romaji (English) layer of the JIS keyboard -->
```
mcdurdin, line 9: This comment is wrong?

core/tests/unit/ldml/ldml_test_source.cpp
```diff
@@ -398,6 +398,9 @@ void LdmlJsonTestSource::set_key_from_id(key_event& k, const std::u16string& id)
 
   // First, find the string
   KMX_DWORD strId = kmxplus->strs->find(id);
+  if (strId == 0) {
+    std::cerr << "ERROR: could not find string for " << id << std::endl;
+  }
   assert(strId != 0);
   if (strId == 0) { // will also get here if id is empty.
     return;
```
mcdurdin, line 406:
```suggestion
  if (strId == 0) {
    // will also get here if id is empty.
    std::cerr << "ERROR: could not find string for " << id << std::endl;
    assert(false);
    return;
```

I don't like the multiple ifs :grin:

mcdurdin (changes requested):
I think this is pretty close

core/tests/unit/ldml/keyboards/ja-Latn-test.xml
```diff
@@ -0,0 +1,26 @@
...
+        </test>
+        <test name="test2">
+			<startContext to="" /> <!-- empty startContext -->
+			<keystroke key="open-square"/>
+			<keystroke key="8"/>
+			<keystroke key="9"/>
+			<keystroke key="0"/>
+			<keystroke key="pipe"/>
+			<check result="[890|"/>
+        </test>
```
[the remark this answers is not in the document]
srl295: indeed. reformatted.

srl295:
got some of the comments. I'll wait till i'm near my ISO keyboard to try to understand the rest.

Also yes i need to delete the duplicate cldr files

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -161,16 +161,42 @@ export const ISOVirtualKeyMap: KeyMap = [
...
+  // [tab] Q W E R T Y U I O P @«`» [ [enter]
+  [ k.K_Q, k.K_W, k.K_E, k.K_R, k.K_T, k.K_Y, k.K_U, k.K_I, k.K_O, k.K_P, k.K_BKQUOTE, k.K_LBRKT  ],
+  // [caps] A S D F G H J K L ; : ] [enter]
+  [ k.K_A, k.K_S, k.K_D, k.K_F, k.K_G, k.K_H, k.K_J, k.K_K, k.K_L, k.K_COLON, k.K_QUOTE, k.K_BKSLASH ],
```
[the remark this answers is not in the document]
srl295: you're right.

common/web/types/src/consts/virtual-key-constants.ts
srl295: @miloush maybe?  The windows docs said that the ABNT2 key was also used for RO / backslash.

srl295:
didn't get near the ISO kbd today.

IF this passes on windows i'm inclined to merge…

mcdurdin (approved):
LGTM, review of ABNT2 key codes still reqd

common/web/types/src/consts/virtual-key-constants.ts
miloush: OK so the OADG standard says that OADG109 and OADG109A keyboards have the extra key next to backspace and next to right shift (OADG104 has neither).

The standard specifies the scan codes as follows:

| key | Set 1 make | Set 1 break | Set 2 make | Set 2 break | Set 3 make | Set 3 break |
|--|--|--|--|--|--|--|
|next to backspace|7D| FD| 6A| F0 6A |5D|F0 5D|
|next to right shift|73|F3|51|F0 51|51|F0 51|

The standards also shows AX and J-3100 keyboards but I take the liberty to assume you are not interested in those (again, just saying JIS is very vague).

I do have an OADG109A keyboard and can confirm that the scan codes match the standard, set 1 on Windows. The key caps in the standard are as follows:

| key | OADG109 | OADG109A | my hardware |
|--|--|--|--|
| next to backspace | image | image |  image |
| next to right shift | image | image | image |

The standard does not formally define the key outputs other than in these figures.

Now the meaty part. As you probably know, the SC to VK mapping on Windows is part of the driver dll and as such is theoretically arbitrary, although in practice Windows ships _mostly_ one mapping. Unfortunately _mostly_ applies especially to the east Asian keyboards. You can change whether you have 101/106/109 etc. key in the settings. That changes which VKs are generated.

For Japanese, there are two settings which result in two different mappings:
| setting | next to backspace | next to shift |
|--|--|--|
| English keyboard (101/102 key) | invalid (0xFF) | VK_ABNT_C1 (0xC1) |
| Japanese keyboard (106/109 key) | VK_OEM_5 (0xDC) | VK_OEM_102 (0xE2) |

I hope this helps, good luck.

miloush: This is by the way the reason why once you are on EA hardware, such as Japanese laptop in this case, you won't be able to type common characters that are on these VK when using any of the western layouts. I am hoping that LDML will be able to fix that.

miloush: As for keys right to L, you get VK_OEM_PLUS, VK_OEM_1, VK_OEM_6 for 106/109 setting and VK_OEM_1, VK_OEM_7, VK_OEM_5 for 101/102 setting. The latter corresponds to colon, quote and backslash on US layout indeed. I don't have anyone to ask at the moment but I would assume that if you have one of these keyboards, you would use the 106/109 setting.

mcdurdin: Thank you @miloush! About to fly to Thailand, will be back in 10 days, so I probably won't have much useful to say until then :grin:, but your feedback here is much appreciated. I have always struggled to correlate my reality to the various standards out there!

> I hope this helps, good luck.

🤣 Thank you 🍀

mcdurdin: FWIW, on Windows, Keyman translates to scan codes and then back to a 'standard' kbdus-based layout (with tweak for abnt2 extra key). I don't remember exactly what we do with Japanese keyboards, would need to test to be sure.

srl295: @miloush and all, i'm going to merge this but filed https://github.com/keymanapp/keyman/issues/8552 to fine tune the ABNT2 and JIS scancodes.   Per the discussion over on Keyboard-SC, we should be using metadata from CLDR once available.  This PR has other structural and build improvements, so I'd like to move it forward…thanks…

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.82-alpha](https://keyman.com/downloads/releases/alpha/17.0.82)

PR #8529, opened 2023-03-27
refactor(common): move core baseline tests into common

### LINKED ISSUE #8499: chore(core): rearrange kmx tests
There are too many files in the core/tests/unit/kmx folder. There are several different tests. Suggest rearranging to make it clearer which fixtures belong to which tests, etc.

### LINKED ISSUE #8504: refactor(linux): avoid using tests from Core in Linux source
Right now, we have a deep cross-module dependency from Linux to on test fixtures in Core. I am trying to cleanup these fixtures in #8503, but this breaks the Linux build -- because we no longer store any .kmx in the repo.

```
  ――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
  Can't find kmx files in /build/linux/ibus-keyman/tests/../../../core/tests/unit/kmx
  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
```

Deep dependencies are fragile because there is no expectation that other modules depend on the test fixtures, and so changes cause ripple effects, and also make it difficult to maintain the tests.

Two parts to the solution for this:

1. Move the .kmn fixtures to common and building them there as a separate build step, otherwise copy the fixtures to the Linux build
2. Update Linux build to build the fixtures also.

See also:
* #8499

mcdurdin:
Fixes #8504.
Fixes #8499.

Decided to use the name `baseline` for this set of tests. They can be used by any project in the repo now. Tests will be treated as shared, and not modified (except in the case of a buggy test).

* Moves .kmn files, add `&NAME` line (don't use meson templating for these files)
* Generates reference .kmx files using kmcomp for use as a baseline with kmcmplib (kmc) tests, and for linux tests
* Updates references in core, kmcmplib, and linux builds
* Renames all the .kmn / .kmx to use a standardized, clean name format (this was needed for building packages in the core tests)

@keymanapp-test-bot skip


mcdurdin:
@ermshiperete can you check and let me know what else is needed for Linux packaging to be able to access common/?

mcdurdin:
iOS build failure due to agent upgrade, fixed on master. iOS has no deps on these changes anyway.

rc-swag (approved):
lgtm

darcywong00 (approved):
lgtm

PR #8536, opened 2023-03-28
chore(linux): Improve robustness of test cleanup

### LINKED ISSUE #8517: chore(linux): stray ibus-engine-keyman left running post CI builds
ps aux gives:

```
bob       970553  0.0  0.0 258316   624 ?        Sl   Feb21   0:00 /var/lib/TeamCity/agent/work/99b311828f4ee7c/keyman/linux/build/x86_64/release/tests/../src/ibus-engine-keyman
bob      2660136  0.0  0.9 310128 40100 ?        Sl   Feb10   6:46 /var/lib/TeamCity/agent/work/99b311828f4ee7c/keyman/linux/build/x86_64/release/tests/../src/ibus-engine-keyman
bob      2689737  0.0  0.0 258244   624 ?        Sl   Feb10   0:00 /var/lib/TeamCity/agent/work/99b311828f4ee7c/keyman/linux/build/x86_64/release/tests/../src/ibus-engine-keyman
```

We can of course just kill these three, and it may be due to an old issue given their start date, but wanted to track it just in case.

ermshiperete:
Closes #8517.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/tests/scripts/run-tests.sh
+++ b/linux/ibus-keyman/tests/scripts/run-tests.sh
@@ -56,7 +56,7 @@ function run_tests() {
 
   echo > "$PID_FILE"
   TEMP_DATA_DIR=$(mktemp --directory)
-  echo "rm -rf ${TEMP_DATA_DIR}" >> "$PID_FILE"
+  echo "rm -rf ${TEMP_DATA_DIR} || true" >> "$PID_FILE"
 
   COMMON_ARCH_DIR=
   [ -d "${TOP_SRCDIR}"/../../core/build/arch ] && COMMON_ARCH_DIR=${TOP_SRCDIR}/../../core/build/arch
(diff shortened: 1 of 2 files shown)


linux/ibus-keyman/tests/scripts/run-tests.sh
```diff
@@ -97,23 +97,23 @@ function run_tests() {
     TMPFILE=$(mktemp)
     # mutter-Message: 18:56:15.422: Using Wayland display name 'wayland-1'
     mutter --wayland --headless --no-x11 --virtual-monitor 1024x768 &> "$TMPFILE" &
-    echo "kill -9 $!" >> "$PID_FILE"
+    echo "kill -9 $! || true" >> "$PID_FILE"
```
darcywong00, line 100: Are these safe to proceed if the `kill -9` fails?

mcdurdin (approved):
LGTM

linux/ibus-keyman/tests/scripts/run-tests.sh
ermshiperete: Yes. It's just a cleanup task, so if `kill` fails it means that this one process keeps hanging around. But if we don't proceed then there will be more processes hanging around, so it's better to continue.

Issue #8525, opened 2023-03-27
bug(windows): Close button appears outside of Add/remove language dialog after changing the UI into Deutsch(German)/Kannada/French languages


bharanidharanj:
### Describe the bug

I noticed that when the UI is changed to German, by clicking the Add/remove language button, the close button appears outside of the Add/remove language window. 

### Reproduce the bug

1. Install Windows 17.0.77-alpha build. 
2. Open Keyman Configuration dialog. 
3. Install any Keyboard. (eg., Khmer Angkor)
4. Change the UI language into Deutsch (German). 
5. In the Keyboard Layouts tab, Click Add/remove language button. 

Here, I noticed that the 'Close' button appears outside of the Add/remove language dialog. 

Note: I am seeing the same behavior when I change the UI language into Kannada or French or Myanmar. 

..German UI
 <img width="300" src="https://user-images.githubusercontent.com/19683143/227967878-b0c5f433-067d-44ef-a21e-a5294595dc94.PNG">

..Kannada UI

   <img width="300" src="https://user-images.githubusercontent.com/19683143/227971855-33da436d-ffaf-4d69-8003-74d5c11be491.PNG">

..French UI

   <img width="300" src="https://user-images.githubusercontent.com/19683143/227971894-ca9460a8-7098-4321-af6d-a722695dbd51.PNG">

..Myanmar UI

   <img width="300" src="https://user-images.githubusercontent.com/19683143/227971954-2ac0eddd-a5d6-429b-88d4-70c1bc3eedc4.PNG">

### Expected behavior

Close button should appear inside the Add/remove language dialog, even the UI is changed to Deutsch (German).  

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for iPhone and iPad
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [X] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

17.077-alpha

### Operating system

Window 10 OS

rc-swag:
I can replicate this. 
Some further notes for the investigation. If you hit cancel instead of pressing the install button again, although the keyboard layout tab is not updated with the Galaxie keyboard. If you exit Keyman Configuration and come back in, the keyboard is actually installed but with no language.  Other keyboards would have installed but used systems default language instead.

bharanidharanj:
[...]

PR #8561, opened 2023-04-03
fix(windows): use right instead of left alignment for popup buttons

rc-swag:
Fixes: #8525 
The fixed layout of the popup means we need to use a specified starting point for the child div that contains the `add language` and `close` button. However, it needs to be right justified instead of left justified to allow for the variable-sized buttons. The buttons will be longer based on the locale text chosen.

# User Testing

# TEST_BUTTONS_ALIGN

1. Install Windows build from this PR
2. Open Keyman Configuration dialog.
3. Install any Keyboard. (eg., Khmer Angkor)
4. Change the UI language into Deutsch (German).
5. In the Keyboard Layouts tab, Click Add/remove language button.
6. Verify that the `Add Language` and `Close` Button now align inside the pop-up box.

7. Change the UI to other Languages and verify steps 5-6. for those languages.

### DIFF
--- a/windows/src/desktop/kmshell/xml/config.css
+++ b/windows/src/desktop/kmshell/xml/config.css
@@ -1199,8 +1199,8 @@ th
 .list_languages_add {
   padding-top: 10px;
   position: fixed;
-  top: 45%;
-  left: 49%;
+  top: 223px; /* Have to use fixed position can't use flex or grid inside a fixed child */
+  right: 281px;
 }
 
 .modify .keyboard_language {


bharanidharanj:
# Test Results

* **TEST_BUTTONS_ALIGN (PASSED):** Tested with the attached PR build (Keyman 17.0.83-alpha-test-8561)  in Windows 10 OS and here is my observation: 1. Changed UI language into German, Kannada and French and I noticed that the Close button is now algined inside the dialog. Seems to be fixed and working fine with this build. Thanks. 

..German UI

   


..Kannada UI

   


..French UI

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.83-alpha](https://keyman.com/downloads/releases/alpha/17.0.83)

PR #8571, opened 2023-04-10
chore(developer): AddWarning and SetError cleanup

### LINKED ISSUE #8570: bug(developer): `SetError` returns `FALSE` so we need to ensure it is only used in functions that return a `bool`
This is a long-standing issue in kmcmplib Compiler.cpp.

Note: perhaps done best by moving the relevant functions to a separate .cpp and only defining it there.

Also:
- [ ] `kmcmp_CompileKeyboardFile` returns a `bool` value so clean up its defn
- [ ] `kmcmp_CompileKeyboardFileToBuffer` returns `CERR_CannotReadInfile` but needs to use `SetError` which returns FALSE instead.

mcdurdin:
Fixes #8570.

* Splits compiler interface functions that return bool into CompilerInterfaces.cpp, and moves `SetError` define to that unit so it can't be used accidentally elsewhere
* Audit of `AddWarning` calls and only those that are used in bool functions now `return FALSE` (with new `AddWarningBool`) whereas all others `return CERR_Break`
* Adds basic unit test for kmcmplib apis
* Fixes `-d` flag which should not have been overridden in build.sh
* Fixes return types for public apis in kmcmplib
* Moves test_assert.h and test_color.h from core to common

@keymanapp-test-bot skip


developer/src/kmcmplib/src/CompilerInterfaces.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  PKMX_STR p;
+
+  if ((p = strrchr_slash(pszInfile)) != nullptr)
+  {
+    strncpy(kmcmp::CompileDir, pszInfile, (int)(p - pszInfile + 1));  // I3481
+    kmcmp::CompileDir[(int)(p - pszInfile + 1)] = 0;
+  }
+  else
+    kmcmp::CompileDir[0] = 0;
```
darcywong00, line 96: I'm seeing a mix of coding styles for braces..
```
if // l.90
{
}
else
```

and 
```
if {  // e.g. l. 105
}
```

We prefer the latter, right?

developer/src/kmcmplib/src/CompilerInterfaces.cpp
```diff
@@ -0,0 +1,368 @@
...
+    SetError(CERR_CannotReadInfile);
+  }
+
+  fseek( fp_in,0,SEEK_SET);
+  if (str[0] == UTF8Sig[0] && str[1] == UTF8Sig[1] && str[2] == UTF8Sig[2])
+    fp_in = UTF16TempFromUTF8(fp_in, TRUE);
+  else if (str[0] == UTF16Sig[0] && str[1] == UTF16Sig[1])
+    fseek( fp_in,2,SEEK_SET);
+  else
+    fp_in = UTF16TempFromUTF8(fp_in, FALSE);
```
darcywong00, line 214: Is there an opportunity to refactor this with ll. 117-122?

developer/src/kmcmplib/src/CompilerInterfaces.cpp
mcdurdin: Yes, where I edit, I try to fixup. If I haven't edited a line, it remains the old style (l.90). I may have missed some spots...

developer/src/kmcmplib/src/CompilerInterfaces.cpp
mcdurdin: Yes, separate PR though.

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.85-alpha](https://keyman.com/downloads/releases/alpha/17.0.85)

PR #8579, opened 2023-04-11
fix(developer): kmc should not validate as well as compile keyboard

### LINKED ISSUE #8512: bug(developer): kmc should not validate as well as compile ldml keyboard
```console
HINT 100008: Locale 'pt-BR-t-k0-abnt2' is not minimal or correctly formatted and should be 'pt-t-k0-abnt2'
HINT 100008: Locale 'pt-BR-t-k0-abnt2' is not minimal or correctly formatted and should be 'pt-t-k0-abnt2'
```

“There now, I've said it twice.”

I think we could dedup compiler messages that are exactly identical. At least before output to the screen. Note that part of the cause is that the same validation is done twice, such as in keys and in layr.

mcdurdin:
Fixes #8512.

Note: the compile phase is identical to the validation phase at this point, so no need to run validate prior to compile:

https://github.com/keymanapp/keyman/blob/b34cdb9c8405eb0cb5a3803c16f7ceb4f6f92338/developer/src/kmc-keyboard/src/compiler/compiler.ts#L113-L121

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/src/activities/BuildLdmlKeyboard.ts
+++ b/developer/src/kmc/src/activities/BuildLdmlKeyboard.ts
@@ -55,9 +55,6 @@ function buildLdmlKeyboardToMemory(inputFilename: string, options: BuildActivity
   if (!source) {
     return [null, null, null];
   }
-  if (!k.validate(source)) {
-    return [null, null, null];
-  }
   let kmx = k.compile(source);
   if (!kmx) {
     return [null, null, null];


srl295:
didn't see this one!

srl295 (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.87-alpha](https://keyman.com/downloads/releases/alpha/17.0.87)

PR #8588, opened 2023-04-12
chore(common): add missing deps and files for npm publish

mcdurdin:
The build/ folder for most packages was missing due to being listed in .gitignore.

Also, @keymanapp/ldml-keyboard-constants package was not being published despite being a dependency of @keymanapp/kmc-keyboard.

Instead of relying on npm's use of .gitignore / .npmignore (which has seriously wonky behaviours), we list the files that should be included in package.json.

NPM Wonky behaviours: it seems that .npmignore causes .gitignore in the same folder to be ignored. But, higher level .gitignore files still affect the files included in the package, which makes specification-by-exclusion painful (negating exclusions, etc). We do not recommend use of .npmignore anywhere for this reason.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.87-alpha](https://keyman.com/downloads/releases/alpha/17.0.87)

PR #8590, opened 2023-04-13
fix(developer): kmc should not compile twice 🙀

srl295:
- don’t call validate() before compiler()

Fixes: #8512

@keymanapp-test-bot skip


mcdurdin:
Dup of #8579 which was already linked on the issue!

srl295:
dup'd by #8579

PR #8606, opened 2023-04-17
fix(developer): package editor no longer loses RTL flag for LMs

### LINKED ISSUE #8595: bug(developer): rtl gets unchecked for lexical-models
>  Thanks @DavidLRowe , I'm a tad surprised because I was very sure that I had checked that box ... and I believe we have uncovered a bug!   My Keyman Developer version is 16.0.139, these steps bring out a bug where "Is Right-to-left" will become _unchecked_ : 

1. Open a lexical model .kps file.
2. Go to the "Lexical Models" section/tab.
3. Check "Is Right-to-left"
4. Go to the "Source" section/tab (do nothing).
5. Return to "Lexical Models" , "Is Right-to-left" is now _unchecked!_

"Is Right-to-left" remains checked if you go back and forth to any of the other tabs.  I'll open a bug report for KMD.

I'll make a new commit with "Is Right-to-left" checked, I guess it will be fine so long as the "Source" section is not visited.

_Originally posted by @dyacob in https://github.com/keymanapp/lexical-models/issues/196#issuecomment-1506217855_

mcdurdin:
Fixes #8595.

# User Testing

* **TEST_RTL_CHECKBOX:** Following the steps below, ensure that the "Is Right-to-left" checkbox for the lexical model always matches the source view value for `RTL`.

1. Create a new lexical model and open the .kps file.
2. Go to the "Lexical Models" section/tab.
3. Check "Is Right-to-left" by clicking on it.
4. Go to the "Source" section/tab and verify that `<RTL>True</RTL>` is present in the lexical models section.
5. Return to "Lexical Models", "Is Right-to-left" should remain checked.
6. Repeat these steps, with the RTL checkbox _off_ and confirm that the Source and Lexical Models tabs match up.

### DIFF
--- a/common/windows/delphi/packages/PackageInfo.pas
+++ b/common/windows/delphi/packages/PackageInfo.pas
@@ -2213,7 +2213,7 @@ begin
     lexicalModel := TPackageLexicalModel.Create(Package);
     lexicalModel.Name := XmlVarToStr(ALexicalModel.ChildValues[SXML_PackageLexicalModel_Name]);
     lexicalModel.ID := XmlVarToStr(ALexicalModel.ChildValues[SXML_PackageLexicalModel_ID]);
-    lexicalModel.RTL := ANode.ChildNodes.IndexOf(SXML_PackageLexicalModel_RTL) >= 0;
+    lexicalModel.RTL := ALexicalModel.ChildNodes.IndexOf(SXML_PackageLexicalModel_RTL) >= 0;
     lexicalModel.Languages.LoadXML(ALexicalModel);
     Add(lexicalModel);
   end;


rc-swag (approved):
lgtm

bharanidharanj:
# Test Results

* **TEST_RTL_CHECKBOX (PASSED):** Tested with the attached PR build (Keyman 17.0.89-alpha-test-8606) in Windows 10 OS and here is my observation: 1. Verified that enabling RTL check box in Lexical Models tab, shows that <RTL>True</RTL> is present in the "Source" tab. Returned to Lexical Models tab, and noticed that the RTL check box remains checked.  2. Disabling "RTL" checkbox  in Lexical Models tab, shows nothing in the Source tab. Returned to Lexicon Models tab, and noticed that the RTL check box remains unchecked. Seems to be working as expected. 

..RTL box checked

   


..RTL box unchecked

PR #8607, opened 2023-04-17
fix(developer): package editor no longer loses RTL flag for LMs 🍒

### LINKED ISSUE #8595: bug(developer): rtl gets unchecked for lexical-models

mcdurdin:
Fixes #8595.

Cherry-pick of #8606.

@keymanapp-test-bot skip

### DIFF
--- a/common/windows/delphi/packages/PackageInfo.pas
+++ b/common/windows/delphi/packages/PackageInfo.pas
@@ -2211,7 +2211,7 @@ begin
     lexicalModel := TPackageLexicalModel.Create(Package);
     lexicalModel.Name := XmlVarToStr(ALexicalModel.ChildValues[SXML_PackageLexicalModel_Name]);
     lexicalModel.ID := XmlVarToStr(ALexicalModel.ChildValues[SXML_PackageLexicalModel_ID]);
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

keyman-server:

PR #8608, opened 2023-04-18
chore(android/engine): Cleanup deprecated APIs

darcywong00:
Some API cleanup to go with #7881

1. ~Since our minimum SDK is 21 (Lollipop 5.0), this cleans up some of the logic checking for SDK version (always greater than KITKAT and never less than JELLY_BEAN_MR1)~ Handled in #8612
2. The KMManager.addKeyboard() API using HashMap was marked intended to deprecate in 15.0, so this is now cleaned up. help.keyman.com was already marked in keymanapp/help.keyman.com#257
3. Removing the KMManager.addKeyboard impacted one of the unit test helpers.

## User Testing
**Setup** - Install the PR builds of Keyman for Android and FirstVoices for Android on a 5.0 emulator (SDK 21)

* **TEST_KEYMAN_ADD_KEYBOARD** - Verifies addKeyboard still works in Keyman app
1. Launch Keyman for Android
2. From the "Get Started" menu --> Add a keyboard for your language --> Install from keyman.com
3. Search for and install sil_cameroon_qwerty keyboard, choose one of the languages during the language selection.
4. Verify sil_cameroon_qwerty keyboard gets installed

* **TEST_FV_ADD_KEYBOARD** - Verifies addKeyboard still works in FirstVoices app
1. Launch FirstVoices for Android
2. From the FirstVoices "Setup" menu, add a region --> keyboard
3. From the "Setup" menu, enable FirstVoices as a system keyboard and set as the default system keyboard
4. Launch a separate app (e.g. Google Chrome)
5. Select a textfield
6. With FirstVoices as the system keyboard, type
7. Verify the system keyboard functions


android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -1473,24 +1473,9 @@ public static boolean addKeyboard(Context context, Keyboard keyboardInfo) {
     return KeyboardPickerActivity.addKeyboard(context, keyboardInfo);
   }
 
-  // Intend to deprecate in Keyman 15.0
+  // This API is deprecated in Keyman 17.0
```
mcdurdin, line 1476: We should just remove the function altogether rather than leaving a broken stub. That way, devs still using this API function will get a compile-time failure rather than a mysterious runtime failure.

android/KMEA/app/src/test/java/com/keyman/engine/view/FunctionalTestHelper.java
```diff
@@ -66,7 +66,21 @@ static void installCustomKeyboard(File aKPMFile) throws IOException, JSONExcepti
...
+      firstKbd.get(KMManager.KMKey_LanguageID),
+      firstKbd.get(KMManager.KMKey_LanguageName),
+      firstKbd.get(KMManager.KMKey_KeyboardVersion),
+      firstKbd.containsKey(KMManager.KMKey_CustomHelpLink) ? firstKbd.get(KMManager.KMKey_CustomHelpLink) : "",
+      "",
+      true,
+      firstKbd.containsKey(KMManager.KMKey_Font) ? firstKbd.get(KMManager.KMKey_Font) : "",
+      firstKbd.containsKey(KMManager.KMKey_OskFont) ? firstKbd.get(KMManager.KMKey_OskFont) : "");
+
+    KMManager.addKeyboard(ApplicationProvider.getApplicationContext(), installedKbd);
```
mcdurdin, line 83: We need to fix the return value from `kmpProcessor.processKMP()` and also any other locations where a `Map<String,String>` is used, e.g. the events like `onPackageInstalled()` to use `Keyboard[]` rather than a `Map<String, String>`.

So... all that to say: can we split the deprecation part of this PR into a separate PR? Because it looks like we need to do more work on the API (and internals too), otherwise API consumers are going to have to do this kind of fragile refactoring.

mcdurdin (changes requested):
So ... I like the removal of old code but I don't think we are ready to remove the old `AddKeyboard` endpoint yet.

android/KMEA/app/src/test/java/com/keyman/engine/view/FunctionalTestHelper.java
darcywong00: >  can we split the deprecation part of this PR into a separate PR?

I've split the old SDK calls to a separate #8612. When I resume this PR it'll be for addressing the API

darcywong00: Sadness - the Map<String, String> ripples out to the other places like KeyboardEventHandler and CloudCatalog.. Too much refactoring to tackle now.

darcywong00:
Alas, too much refactoring to tackle for 17.0

PR #8622, opened 2023-04-18
refactor(developer): move build activities

mcdurdin:
Moves build activities under commands/build/, and build-test-data action under commands/buildTestData/, and adds `BuildTestDataOptions` interface so that buildTestData.ts does not depend on BuildActivity.ts. Clarifies which activites are used where.

@keymanapp-test-bot skip


developer/src/kmc/src/commands/build.ts
```diff
@@ -1,8 +1,8 @@
 import * as fs from 'fs';
 import { Command } from 'commander';
-import { BuildActivityOptions } from '../activities/BuildActivity.js';
-import { buildActivities } from '../activities/buildActivities.js';
-import { BuildProject } from '../activities/BuildProject.js';
+import { BuildActivityOptions } from './build/BuildActivity.js';
+import { buildActivities } from './build/buildActivities.js';
```
darcywong00, line 4: Maybe this was asked before, is there a reason `buildActivities.ts` isn't CamelCase like all the other files?(buildTestData.ts too)

darcywong00 (approved):
lgtm

developer/src/kmc/src/commands/build.ts
mcdurdin: One is a module containing a class, the other is a module containing a variable.

PR #8623, opened 2023-04-18
feat(developer): add warning options to kmc

mcdurdin:
Adds --compiler-warnings-as-errors and --no-warn-deprecated-code options to kmc command line and interfaces.

Note that most compilers do not yet honour these options, only kmc-kmn.

Both of these should be handled in error filtering and reporting rather than at the compiler level, so management of these may be factored out of kmcmplib and into kmc later (although this goal may be blocked by kmcmpdll's need to also use them).

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #8625, opened 2023-04-19
feat(developer): use new message infrastructure in kmc-kmn

mcdurdin:
Moves to using the `reportMessage` callback for reporting all errors in kmc-kmn.

This required some patching to common/web/types in order to get the new constants in place and add support for line numbers to `CompilerEvent`, so at the same time I refactored `compilerErrorSeverityName` into the module.

After the refactor, coverage tests were failing because they included the test/ folder, so cleaned that up at the same time.

@keymanapp-test-bot skip


developer/src/kmc-kmn/src/compiler/messages.ts
mcdurdin, line 1: This file matches kmc-keyboard messages.ts format.

darcywong00 (approved):
lgtm

keyman-server:

PR #8626, opened 2023-04-19
refactor(developer): move shared test helpers into common

mcdurdin:
Sadly, a lot of plumbing required to share code. But it should work now; these test helpers are used only in dev so should not require publication.

Will have to monitor the `npm publish` step because I am not 100% sure this isn't going to break even though it's only a `devDependency`.

@keymanapp-test-bot skip


developer/src/common/web/test-helpers/index.ts
```diff
@@ -1,7 +1,10 @@
 import * as fs from 'fs';
-import * as path from 'path';
 import { CompilerEvent, CompilerCallbacks } from '@keymanapp/common-types';
 
+// TODO: schemas are only used by kmc-keyboard for now, so this works at this
+// time, but it's a little fragile if we need them elsewhere in the future
+const SCHEMA_BASE = '../../../../kmc-keyboard/build/src/';
```
darcywong00, line 6: That's a curious path to have `/build/src/`

mcdurdin: It relates to how tsc builds -- the build/ folder is the output base folder, then the full source tree is replicated under it (in this case, src/, test/, ...)

package-lock.json
```diff
@@ -567,13 +569,67 @@
...
       "license": "MIT"
     },
+    "developer/src/common/web/test-helpers": {
+      "name": "@keymanapp/developer-test-helpers",
+      "license": "MIT",
+      "dependencies": {
+        "@keymanapp/common-types": "*"
+      },
+      "devDependencies": {
+        "@types/node": "^10.14.6",
```
darcywong00, line 579: Should this be the same version as l. 585? 10.17.60

mcdurdin: Don't review package-lock.json -- it's auto generated. Look only at the various package.json files :grin:

darcywong00 (approved):
lgtm

keyman-server:

Issue #8150, opened 2023-02-02
feat(developer): Add hints, warnings, and errors to ts-based package compiler


mcdurdin:
We need to ensure that all the error checking that is currently done in the pascal-based package compiler is replicated in the typescript version.

---

- [x] Note: ts-based package compiler needs to have significant additional work on it, as we do not currently emit any warnings or errors. Suggest bundling this error into that work.
- [x] #8657
- [x] #8658 (CompilePackage.pas:155)
- [x] #8659
- [ ] #7353
- [x] #8730
- [x] #8752
- [x] #8753
- [x] #8754
- [x] #8756
- [x] #8765
- [x] #8772
- [x] #8773
- [x] #8766
- [x] #8774
- [x] #8775
- [x] #8779
- [x] #8780
- [x] #8776
- [x] #8777
- [x] #8767
- [x] #8643 (CompilePackage.pas:293)
- [x] #8787
- [x] #8788
- [x] #8789
- [x] #8790
- [x] #8768
- [x] #8782
- [x] #8781
- [x] #8731
- [ ] #8672
- [x] #8791

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/issues/8119#issuecomment-1413295606_

mcdurdin:
e.g. this should be a friendly error message:

```
C:\Projects\keymanapp\keyboards\release\k\khmer_angkor\source>kmc build khmer_angkor.kps
Building khmer_angkor.kps
Error: ENOENT: no such file or directory, open 'C:\Projects\keymanapp\keyboards\release\k\khmer_angkor\build\khmer_angkor.js'
    at Object.openSync (node:fs:601:3)
    at Module.readFileSync (node:fs:469:35)
    at file:///C:/Users/mcdurdin/AppData/Roaming/npm/node_modules/@keymanapp/kmc/node_modules/@keymanapp/kmc-package/build/src/kmp-compiler.js:222:27
    at Array.forEach (<anonymous>)
    at KmpCompiler.buildKmpFile (file:///C:/Users/mcdurdin/AppData/Roaming/npm/node_modules/@keymanapp/kmc/node_modules/@keymanapp/kmc-package/build/src/kmp-compiler.js:198:20)
    at BuildPackage.build (file:///C:/Users/mcdurdin/AppData/Roaming/npm/node_modules/@keymanapp/kmc/build/src/activities/BuildPackage.js:19:38)
[...]

PR #8628, opened 2023-04-19
feat(developer): warn if .kps file includes a non-binary .kvk file

mcdurdin:
Relates to #8150.

Legacy .kmp compiler would transform xml-format .kvk files into a binary .kvk file; now we want that to remain the responsibility of the keyboard compiler, so we'll warn the few users who are still doing this.

@keymanapp-test-bot skip


common/web/types/src/main.ts
```diff
@@ -8,6 +8,8 @@ export { default as KMXPlusBuilder}  from './kmx/kmx-plus-builder/kmx-plus-build
 export { default as KvkFileReader } from './kvk/kvk-file-reader.js';
 export { default as KvksFileReader } from './kvk/kvks-file-reader.js';
 export { default as KvkFileWriter } from './kvk/kvk-file-writer.js';
+export * as KvkFile from './kvk/kvk-file.js';
+export * as KvksFile from './kvk/kvk-file.js';
```
mcdurdin, line 12: Not currently using this but exporting it for future consumers.

common/web/types/src/kvk/kvk-file.ts
```diff
@@ -5,6 +5,7 @@ import * as r from 'restructure';
 //
 
 export const BUILDER_KVK_HEADER_IDENTIFIER = 0x464B564B; // 'KVKF', little-endian
+export const KVK_HEADER_IDENTIFIER_BYTES = [0x4B, 0x56, 0x4B, 0x46];
```
darcywong00, line 8: Two questions.
1. Wondering why no `BUILDER_` prefix?
2. Is this like `BUILDER_KVK_HEADER_IDENTIFIER` in big-endian? e.g. `0x4B564B46`?

developer/src/kmc-package/src/kmp-compiler.ts
```diff
@@ -294,4 +297,15 @@ export default class KmpCompiler {
...
     return zip.generateAsync({type: 'binarystring', compression:'DEFLATE'});
   }
+
+  /**
+   * Legacy .kmp compiler would transform xml-format .kvk files into a binary .kvk file; now
+   * we want that to remain the responsibility of the keyboard compiler, so we'll warn the
+   * few users who are still doing this
+   */
+  private warnIfKvkFileIsNotBinary(filename: string, data: Buffer) {
+    if(filename.match(/\.kvk$/) && data.compare(Buffer.from(KvkFile.KVK_HEADER_IDENTIFIER_BYTES), 0, 3, 0, 3) != 0) {
```
darcywong00, line 307: Prob not really needed, but I kinda feel this could go in a helper `isBinaryKvkFile()` that returns a boolean.

Then `warnIfKvkFileIsNotBinary` would print the warning if true.

developer/src/kmc-package/test/test-package-compiler.ts
```diff
@@ -143,7 +145,49 @@ describe('KmpCompiler', function () {
...
+    assert.lengthOf(callbacks.messages, 1);
+    assert.deepEqual(callbacks.messages[0].code, CompilerMessages.WARN_FileIsNotABinaryKvkFile);
   });
 
+  it('should not warn if a binary kvk file is included', async function() {
+    this.timeout(10000); // building a zip file can sometimes be slow
+
+    callbacks.clear();
+
+    // const kmpPath = makePathToFixture('khmer_angkor', 'build', 'khmer_angkor.kmp');
```
darcywong00, line 178: Can kmpPath be removed? l.156 too

darcywong00 (approved):
lgtm -

common/web/types/src/kvk/kvk-file.ts
mcdurdin: 1. I left `BUILDER_` out because it's a define not used by the builder. But perhaps better to be removing a number of the `BUILDER_` prefixes in the future from other objects as well.
2. Yes, it's in array form for use in `Buffer` comparisons.

developer/src/kmc-package/src/kmp-compiler.ts
mcdurdin: Will do this refactor in the future if we reuse the test elsewhere.

keyman-server:

Issue #8361, opened 2023-03-03
chore(common): add new PR control commands to manage builds


mcdurdin:
It would be useful to be able to skip certain builds or do only builds without artifacts etc. Rationale: many of our minor changes are triggering many builds, and while this is a good default, we can reduce the burden on the build agents by being a little more circumspect when we know certain platforms will not be impacted. We can also reduce the artifact burden and build times by skipping release builds when they are not needed.

These commands would override the defaults -- so you could force a build where it would not automatically run otherwise, for example. The following commands would be supported:

* `skip`: Skip build altogether for a given target:

  ```
  Build-bot: skip windows,developer
  ```

* `build`: Just compile and run unit tests on given target:

  ```
  Build-bot: build windows,developer
  ```

* `release`: Make a full installer for a given target and make it available as an artifact on the PR, good for user testing:

  ```
  Build-bot: release windows,developer
  ```

**QUESTION:** Do we need to support additional commands or parameters for commands? For example, Developer has a set of low-cost unit tests, plus a very long regression test suite (25+ minutes), which we could do with `test(quick)`, `test(full)` and the default (`test`) would be `test(quick)`. Something like that?

  - use `test-quick` vs `test-full` vs `test` perhaps, and we may be able to pass that into build.sh in the future as well.
  - PR builds default to quick tests (but full installer builds?) -- but can override
  - nightly builds do full tests
  - master-merge builds do not build installers

Multiple commands in a comment should be supported.

The following build targets should be defined:
* `android`
* `common` (for tests)
* `core` (for tests)
* `developer`
* `ios`
* `linux`
* `mac`
* `web`
* `windows`

**QUESTION:** Am I missing any? Perhaps the ability to specify platform for common + core would be helpful?

  - ability to skip build of sample apps would help

Multiple targets should be comma separated, suggest no spaces between to support future parameters.

The commands could be:
* `Build-bot: skip [target[,...]]` -- don't build anything for these target(s)
* `Build-bot: test [target[,...]]` -- only compile and run automated tests for these target(s)
* `Build-bot: release [target[,...]]` -- `test` + build an installable artifact for these target(s), if applicable (i.e. `release common` would not be an error but would only run tests for `common`)

These commands will be checked on all comments in a PR and are consolidated; later commands override earlier commands.

**IDEA:** reduce set of default builds, e.g. don't build Developer,Android,iOS if Web builds?

mcdurdin:
[...]

PR #8635, opened 2023-04-20
chore(developer): only run keyboards-repo tests for a full-test

mcdurdin:
If kmcmplib build includes the --full-test flag, then configure and run tests that include pulling in the keyboards repo; otherwise skip all of that.

This will require CI changes to run full tests, relating to #8361.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmcmplib/meson_options.txt
+++ b/developer/src/kmcmplib/meson_options.txt
@@ -1,0 +1,1 @@
+option('full_test', type: 'boolean', value: false)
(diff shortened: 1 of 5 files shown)


developer/src/kmcmplib/meson_options.txt
```diff
@@ -0,0 +1 @@
+option('full_test', type: 'boolean', value: false)
```
darcywong00, line 1: How does this get set to `true` on CI?

mcdurdin: It doesn't, yet. See my comment at top:

> This will require CI changes to run full tests, relating to https://github.com/keymanapp/keyman/issues/8361.

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.92-alpha](https://keyman.com/downloads/releases/alpha/17.0.92)

PR #8637, opened 2023-04-20
refactor(developer): kmc-model to use `CompilerEvent`

mcdurdin:
Replaces the existing error reporting in kmc-model with `CompilerEvent` style messages. Some of the existing error reporting was kinda nice, but consistency trumps pretty here.

One downside at present is the global `filename` and `line` variables in the model-compiler-errors.ts unit. This is fragile, and should probably be refactored further in the future.

Where it made sense, I converted various throws into `ModelCompilerError` instances. If it appeared to be an internal error, I left it as-is, but the external API interfaces wrap all unknown exceptions and convert them into a compiler error message, so these APIs (`compileModel` and `loadFromFilename`) now have a guarantee that they will never throw.

At this point, all of the error messages in the model compiler are unrecoverable -- they will halt the compile entirely. However, the pathway is there to support recoverable errors through the normal `callbacks.reportMessage` mechanisms.

The message-counting mechanism is currently missing from this, as it belongs in the API consumer, rather than in the low-level reporting infrastructure. This means that models with large numbers of warnings are going to generate scads of messages for now.

Note that the _users_ of `CompilerCallbacks` are subject to change in the future, once we have all the compilers using the callbacks.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.92-alpha](https://keyman.com/downloads/releases/alpha/17.0.92)

PR #8641, opened 2023-04-21
feat(common): kmx file reader

mcdurdin:
Read a .kmx file into memory in Typescript. Currently only reads stores from the file, as these are the most common things needed for other processes. Will add support for reading other metadata and rules in future when required.

@keymanapp-test-bot skip


common/web/types/src/kmx/kmx-file-reader.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as r from 'restructure';
+
+export class KmxFileReader {
+  private readonly rString = new r.String(null, 'utf16le');
+
+  private readString(source: Uint8Array, offset: number): string {
+    if(offset == 0) {
+      // the file has a null string here, which is not the same as an empty
+      // string ('')
+      return null;
```
darcywong00, line 11: Is this ever an error to be flagged/logged?

darcywong00: I guess it's up to the caller to handle?

darcywong00 (approved):
lgtm

common/web/types/src/kmx/kmx-file-reader.ts
mcdurdin: > I guess it's up to the caller to handle?

Yes, up to the caller. It's not an error in the file format.

keyman-server:

PR #8651, opened 2023-04-24
feat(developer): use CompilerEvent for messages in all of kmc

mcdurdin:
Cleans up majority of remaining kmc-* modules to use CompilerEvent messages instead of console.log and friends, adding InfrastructureMessages class for kmc CLI messages.

Consolidates schema loading functions into a single `loadSchema` function in the CompilerCallbacks interface and implementations.

Moves responsibility for instantiating NodeCompilerCallbacks out of each individual BuildActivity and into higher level modules.

Fixes return value for failing builds.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc-keyboard/src/compiler/compiler.ts
+++ b/developer/src/kmc-keyboard/src/compiler/compiler.ts
@@ -66,7 +66,7 @@ export default class Compiler {
       return null;
     }
     try {
-      if (!reader.validate(source, this.callbacks.loadLdmlKeyboardSchema())) {
+      if (!reader.validate(source, this.callbacks.loadSchema('ldml-keyboard'))) {
         return null;
       }
     } catch(e) {
@@ -98,7 +98,7 @@ export default class Compiler {
       // TODO-LDML: The unboxed data doesn't match the schema anymore. Skipping validation, for now.
 
       // try {
-      //   if (!reader.validate(source, this.callbacks.loadLdmlKeyboardTestSchema())) {
+      //   if (!reader.validate(source, this.callbacks.loadSchema('ldml-keyboard-test'))) {
       //     return null;
       //   }
       // } catch(e) {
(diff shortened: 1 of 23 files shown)


developer/src/kmc-keyboard/src/compiler/compiler.ts
```diff
@@ -98,7 +98,7 @@ export default class Compiler {
       // TODO-LDML: The unboxed data doesn't match the schema anymore. Skipping validation, for now.
 
       // try {
-      //   if (!reader.validate(source, this.callbacks.loadLdmlKeyboardTestSchema())) {
+      //   if (!reader.validate(source, this.callbacks.loadSchema('ldml-keyboard-test'))) {
```
darcywong00, line 101: Should this be `ldml-keyboardtest`?

reader-callback-test.ts l. 24

developer/src/kmc/src/messages/messages.ts
```diff
@@ -0,0 +1,39 @@
...
+    `Unrecognised input file ${o.filename}, expecting ${o.extensions}, or project folder`);
+  static ERROR_FileTypeNotRecognized = SevError | 0x0004;
+
+  static Error_OutFileNotValidForProjects = () => m(this.ERROR_OutFileNotValidForProjects,
+    `--out-file should not be specified for project builds`);
+  static ERROR_OutFileNotValidForProjects = SevError | 0x0005;
+
+  static Info_FileBuiltSuccessfully = (o:{filename:string}) => m(this.INFO_FileBuiltSuccessfully,
+    `${o.filename} built successfully.`);
+  static INFO_FileBuiltSuccessfully = SevInfo | 0x0006;
```
darcywong00, line 33: With the status masking, wouldn't `SevInfo | 0x0003` be available?

Is this the expected status for successful compile/build?

mcdurdin: * All base codes are unique (when | with their namespace) -- it's just safer to always do this than to accidentally end up using the same code twice (also, in theory we might want to increase/decrease severity of a code in future?)
* This is an informational message. The return status for kmc will be 0 on success, 1 on failure (there may be other statuses in future).

developer/src/kmc-keyboard/src/compiler/compiler.ts
mcdurdin: Good catch; it's commented out code for now so I didn't get compiler error on it. I will leave it for now because it's an easy fix when we uncomment the code (and saves another build run)

darcywong00 (approved):
lgtm

keyman-server:

PR #8652, opened 2023-04-24
chore(developer): fixup deps for kmc modules

mcdurdin:
@keymanapp-test-bot skip

Note: also adds rudimentary top-level developer build script, just for building kmc + friends. Its primary use right now is to run `developer/build.sh test` as that will run tests on all modules (building happens through child dep relationships for kmc anyway)


darcywong00 (approved):
lgtm

keyman-server:

PR #8654, opened 2023-04-24
fix(common): builder - pass deps flags to child scripts

### LINKED ISSUE #8650: bug(common): `--no-deps` does not get passed through to child builds (build.sh)
A parent script called with `--no-deps` does not pass the `--no-deps` parameter through to child scripts. (Or `--force-deps` either for that matter)

mcdurdin:
@keymanapp-test-bot skip

Fixes #8650.


resources/build/tests/trees/test.sh
```diff
@@ -2,16 +2,26 @@
...
 if [ $(ls child?.* | wc -l) -ne 2 ]; then
   echo unexpected output file count
   exit 1
+fi
+
+echo "--- Test that --no-deps is passed correctly to child scripts ---"
+./build.sh --no-deps
+if [ -f ./dep.configure ] || [ -f ./dep.build ]; then
+  echo unexpected dep.configure or dep.build file, should not have built this because of --no-deps flag
+  exit 1
```
darcywong00, line 26: > A parent script called with --no-deps does not pass the --no-deps parameter through to child scripts. (Or --force-deps either for that matter)

Should we also add a test for `--force-deps`?

mcdurdin: I didn't want to add the extra code required to verify --force-deps into the test -- it's the same flag variable, so if --no-deps is passed through, then --force-deps will be too.

darcywong00 (approved):
lgtm

keyman-server:

PR #8660, opened 2023-04-24
chore(common): Move the_99 test fixture keyboard

darcywong00:
From the Sprint A17S11 design meeting, the team plans to reorganize keyboard fixtures in the repo.

/common/test/keyboards/ would following the convention:
/common/test/keyboards/issue/{issue number}/

So this moves the_99.kmp to that hierarchy. The builld.sh script seems to just "copy" the kmp file to a build/ folder.

@keymanapp-test-bot skip


darcywong00:
Ah I also see CI failed because it's expecting an issue.kpj file...

darcywong00:
> 1. The script does not cd "$THIS_DIR" (or cd "$THIS_SCRIPT_PATH", for that matter). It will fail if not run from its specific directory, so far as I can tell.

Script updated to cd "$THIS_DIR".  I expect many of these test keyboards old not worth converting to builder scripts.

> Ah I also see CI failed because it's expecting an issue.kpj file...

I tweaked the top build.sh to skip invalid/ and issue/ folders. 

We'll update issue/ targets when other keyboards get moved there

common/test/keyboards/issue/6254/build.sh
```diff
@@ -6,7 +6,7 @@ set -u
 ## START STANDARD BUILD SCRIPT INCLUDE
```
mcdurdin, line 6: The issue folder name could be `6254-invalid-version-99` -- that way it's easy to remember which is which.

mcdurdin (approved):

PR #8664, opened 2023-04-25
chore(developer): test message correctness in compiler modules

### LINKED ISSUE #8656: chore(developer): apply uniqueness tests for CompilerEvent messages to new modules
kmn-keyboard, I believe, has a unit test to verify that all CompilerEvent messages in the module are unique and have the approriate severity matching their name, and that there are no typos in repeated names. We need to apply this to all the messages in all the modules.

mcdurdin:
Fixes #8656.

Fixes some dependencies as well, also removing obsolete `/// <reference>` which broke tsc in some cases.

Per the design of the unit test, `CompilerMessages` and its sibling types need to be clean objects that contain only message builder functions and message constants. This is probably a good design long-term, so there has been some minor refactoring to deal with this.

Added first unit test to kmc itself, but currently have disabled c8 until we have sufficient tests to make coverage useful.

@keymanapp-test-bot skip


developer/src/kmc/build.sh
```diff
@@ -51,40 +51,48 @@ fi
 if builder_start_action clean; then
   rm -rf ./build/ ./tsconfig.tsbuildinfo
   builder_finish_action success clean
-else
+fi
+
+function copy_schemas() {
```
darcywong00, line 56: 1. This makes the diff easy to see.  Just checking if you wanted to move the function higher to keep all the builder actions together.
2. Do you want to use internal name  `_copy_schemas()`?

mcdurdin: No, I think it's fine there for now; no need to use `_` here because this script isn't imported into other scripts like builder.inc.sh is.

darcywong00 (approved):
lgtm

developer/src/common/web/test-helpers/verifyCompilerMessagesObject.ts
```diff
@@ -0,0 +1,66 @@
...
+  const m = source as Record<string,any>;
+
+  let codes: number[] = [];
+
+  for(let key of keys) {
+
+    // Verify each object member matches the pattern we expect
+
+    if(typeof m[key] == 'function') {
+      const o = /^(Info|Hint|Warn|Error|Fatal)_([A-Za-z0-9_]+)$/.exec(key);
```
srl295, line 33: It makes me wonder if this should really be derived from the output of `compilerErrorSeverityName()`.

developer/src/common/web/test-helpers/verifyCompilerMessagesObject.ts
```diff
@@ -0,0 +1,66 @@
...
+      expect(o).to.be.instanceOf(Array, `Expected member ${key} to be a valid message function name`);
+
+      const c = o[1].toUpperCase() + '_' + o[2];
+      expect(m[c]).to.be.a('number', `Expected constant name ${c} to exist`);
+
+      let v = m[key]('','','','','','','','','','','','' /* ignore arguments*/);
+      expect(v.code).to.equal(m[c], `Function ${key} returns the wrong code`);
+    }
+    else if(typeof m[key] == 'number') {
+      const o = /^(INFO|HINT|WARN|ERROR|FATAL)_([A-Za-z0-9_]+)$/.exec(key);
```
srl295, line 43: same, should be derived.

developer/src/common/web/test-helpers/verifyCompilerMessagesObject.ts
```diff
@@ -0,0 +1,66 @@
...
+      expect(m[f]).to.be.a('function', `Expected function name ${f} to exist`);
+    } else {
+      assert.fail(`Unexepected compiler message member ${key}`);
+    }
+
+    // Verify severify masks
+
+    if(typeof m[key] == 'number') {
+      const o = /^(INFO|HINT|WARN|ERROR|FATAL)_([A-Za-z0-9_]+)$/.exec(key);
+      expect(o).to.be.instanceOf(Array);
```
srl295, line 56: WET here as far as the error types

srl295 (approved):
LGTM with some dehydration comments

mcdurdin:
> LGTM with some dehydration comments

I am not seeing a clear way to DRY out the messages at present -- we need the constants, and the helper functions. We could refactor to a single object I guess? ``ERROR_foo: [0x1000, ()=>`...`}``? In any case, we can do a follow-up PR on that (but I would do it soon to avoid embedding this current design too deep)

mcdurdin:
So, I can dry the messages out from this structure:

```ts
  static Info_BuildingFile = (o:{filename:string}) => m(this.INFO_BuildingFile,
    `Building ${o.filename}`);
  static INFO_BuildingFile = SevInfo | 0x0002;
```

To something like this:

```ts
  static Info_BuildingFile = {
    code: SevInfo | 0x0002, 
    v: function(o:{filename:string}) { return m(this, `Building ${o.filename}`); } 
  }
```

But not sure I can get much better -- due to arrow functions not being methods, we have to go the wordy `function(){ return x }` approach instead. So it's kinda sad either way; usage is cleaner with the first approach, and unit tests deal with WETness, so I'm inclined to leave it as is.

srl295:
> > LGTM with some dehydration comments
> 
> I am not seeing a clear way to DRY out the messages at present -- we need the constants, and the helper functions. We could refactor to a single object I guess? `` ERROR_foo: [0x1000, ()=>`...`] ``? In any case, we can do a follow-up PR on that (but I would do it soon to avoid embedding this current design too deep)

100% there.  by WET I meant the repeated `/^(INFO|HINT|WARN|ERROR|FATAL)` regex, which would be derivable by the enumerated error classes.

keyman-server:

mcdurdin:
> 100% there. by WET I meant the repeated `/^(INFO|HINT|WARN|ERROR|FATAL)` regex, which would be derivable by the enumerated error classes.

Low benefit on this I think -- I'll focus on other things as it's literally a 30 second patch to the test in the event that we ever change the error classes. :grin:

developer/src/common/web/test-helpers/verifyCompilerMessagesObject.ts
```diff
@@ -0,0 +1,66 @@
...
+//   * that constants have the right error mask for their type
+//
+// This means we don't have to worry about DRYing out the constants, which would
+// probably require us to have a preprocessing step, which has its own
+// negatives.
+//
+// These are used in each of the test-messages.ts units to verify the
+// CompilerMessages classes
+
+const toTitleCase = (s: string) => s.substring(0, 1).toUpperCase() + s.substring(1).toLowerCase();
```
srl295, line 19: Doesn't handle Greek titlecase, of course https://unicode-org.github.io/icu/userguide/transforms/general/rules.html#combinations

mcdurdin: Good news then that I don't plan to use Greek (or German) characters in my compiler message identifiers!

PR #8666, opened 2023-04-25
chore(developer): kmc-keyboard should not use any node-specific modules

mcdurdin:
One of a set of updates for kmc-*, as all of them currently have either wasm-fs or node core module references. This is the easiest patch.

@keymanapp-test-bot skip

### DIFF
--- a/common/web/types/src/main.ts
+++ b/common/web/types/src/main.ts
@@ -28,3 +28,5 @@ export { TouchLayoutFileWriter, TouchLayoutFileWriterOptions } from './keyman-to
 export * as KPJ from './kpj/kpj-file.js';
 export { KPJFileReader } from './kpj/kpj-file-reader.js';
 export { KeymanDeveloperProject } from './kpj/keyman-developer-project.js';
+
+export * as util from './util/util.js';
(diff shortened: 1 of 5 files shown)


darcywong00 (approved):
lgtm

common/web/types/src/main.ts
```diff
@@ -28,3 +28,5 @@ export { TouchLayoutFileWriter, TouchLayoutFileWriterOptions } from './keyman-to
 export * as KPJ from './kpj/kpj-file.js';
 export { KPJFileReader } from './kpj/kpj-file-reader.js';
 export { KeymanDeveloperProject } from './kpj/keyman-developer-project.js';
+
+export * as util from './util/util.js';
```
srl295, line 32: would be clearer as `fs-util.ts` or `path-util.ts` in light of the encapsulation mentioned.

mcdurdin: Note that follow-up PR #8674 refactors the fs-related functions in util.ts away again, moving responsibility for path and fs ops into the callbacks (we may be able to do better later but avoiding adding extra deps for now). :grin:

keyman-server:

Issue #5272, opened 2021-06-11
chore(android): work through hints in Android Studio


mcdurdin:
There are a great many hints reported in Android Studio, some of which are highlighting bugs. Let's clean these up, remove unused dependencies, etc.

PR #8667, opened 2023-04-25
chore(android): Move namespace to Gradle files

darcywong00:
Addresses a small part of #5272.

Gradle warns that declaring package names in AndroidManifest.xml files are deprecated.
They should now be declared as namespaces in the build.gradle files.

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.94-alpha](https://keyman.com/downloads/releases/alpha/17.0.94)

PR #8668, opened 2023-04-25
fix(common): builder parse multi-param order

mcdurdin:
If a build script specified, e.g. :kmc-keyboard, :kmc-package, :kmc as three targets, in that order, then passing :kmc as a parameter would fail, because :kmc-keyboard, :kmc-package would be emitted to the match result before finding :kmc.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #8675, opened 2023-04-26
chore(android/engine): Remove jvm args

darcywong00:
More gradle cleanup

The Robolectric issue seems resolved and jvm workaround can be removed.

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.95-alpha](https://keyman.com/downloads/releases/alpha/17.0.95)

PR #8698, opened 2023-05-02
chore(common): `set -eu` in build-utils.sh

### LINKED ISSUE #8693: chore: move `set -eu` into build-utils.sh
Currently, all our build scripts have a tedious prolog. I want to minimize that and maximize script safety by moving the `set -eu` line into build-utils.sh. There are a number of scripts that fail when I do this, though, including:

* /resources/build/history/history-utils.sh:
  ```
  /Users/mcdurdin/buildAgent/work/99b311828f4ee7c/keyman/resources/build/history/history-utils.sh: line 11: DEBUG: unbound variable
  ```

* /mac/build.sh:
  ```
  ./build.sh: line 113: KME4M_BUILD_PATH: unbound variable
  ```

* /linux/scripts/dist.sh:
  ```
  ./scripts/dist.sh: line 20: $1: unbound variable
  ```

I originally had this change bundled into #8692 but will make it a separate task to avoid a cascade of issues.

mcdurdin:
Fixes #8693.

Enforces `-e` (abort script on error) and `-u` (error on unset variable use) on all scripts that source build-utils.sh. This may result in some downstream build failures which we'll need to address before merge to master.

@keymanapp-test-bot skip

### DIFF
--- a/common/predictive-text/unit_tests/test.sh
+++ b/common/predictive-text/unit_tests/test.sh
@@ -58,7 +58,7 @@ if builder_start_action test:libraries; then
 
   pushd "$KEYMAN_ROOT/common/models/templates"
   echo
-  echo "### Running $builder_term common/models/templates) tests"
+  echo "### Running $(builder_term common/models/templates) tests"
   if builder_has_option --ci; then
     npm run test -- -reporter mocha-teamcity-reporter
   else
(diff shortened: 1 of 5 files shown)


darcywong00 (approved):
lgtm

Does this mean we can then start removing `set -eu` in child scripts?

mcdurdin:
> Does this mean we can then start removing `set -eu` in child scripts?

Yes, but just do this as part of general maintenance, not as a special PR :grin:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.101-alpha](https://keyman.com/downloads/releases/alpha/17.0.101)

PR #8705, opened 2023-05-03
chore(linux): Fix build script

ermshiperete:
Previously we didn't correctly change directories which could lead to the error "cd: ../build/x86_64/release: No such file or directory".

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/build.sh
+++ b/linux/ibus-keyman/build.sh
@@ -33,6 +33,8 @@ else
 fi
 MESON_PATH="../build/$(uname -m)/$MESON_TARGET"
 
+cd "$THIS_SCRIPT_PATH"
+
 builder_describe_outputs \
   configure "${MESON_PATH}/build.ninja" \
   build "${MESON_PATH}/src/ibus-engine-keyman" \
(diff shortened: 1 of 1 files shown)


linux/ibus-keyman/build.sh
```diff
@@ -33,6 +33,8 @@ else
 fi
 MESON_PATH="../build/$(uname -m)/$MESON_TARGET"
```
darcywong00, line 34: Do you want to define MESON_PATH relative to THIS_SCRIPT_PATH? Then you could leave the other lines intact

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.99-alpha](https://keyman.com/downloads/releases/alpha/17.0.99)

linux/ibus-keyman/build.sh
```diff
@@ -33,6 +33,8 @@ else
...
 
+cd "$THIS_SCRIPT_PATH"
+
 builder_describe_outputs \
   configure "${MESON_PATH}/build.ninja" \
   build "${MESON_PATH}/src/ibus-engine-keyman" \
```
mcdurdin, line 40: I notice you have two build outputs? I don't think this is supported

linux/ibus-keyman/build.sh
mcdurdin: For performance reasons, it is good to make `builder_describe_outputs` absolute (i.e. to root of repo). I notice that it's missing from build-utils.md at present....

linux/ibus-keyman/build.sh
ermshiperete: At least it seems to work :smile:

mcdurdin: I think it'll just ignore one of the outputs

ermshiperete: If I delete either one of them it'll rebuild.

mcdurdin: I'm not sure how that could work? Reviewing the source of `builder_describe_outputs` shows essentially it results in: `_builder_dep_path[$action$target]="$path"` So I can't see how it would add both items to the array.

ermshiperete: Ahh, you're right. I have to run `./build.sh test` to test if this works :smile: - and it doesn't. I tried with `./build.sh build`, but that doesn't consider the build outputs specified in `build.sh` and always calls `ninja` which then rebuilds the outputs if necessary.

PR #8707, opened 2023-05-04
chore(window): rename core functions ⛏️

### LINKED ISSUE #8703: chore(windows): Rename functions if "Core" in name is redundant ⛏️
The following functions can be renamed or even removed now that legacy core has been removed. #8593 

-  // Core was mainly as a point of difference vs the windows engine core while the feature flag was present. So could remove the core name 
ReleaseKeyboardMemoryCore
ReleaseStateMemoryCore

- // These functions are about saving and restoring the keyboard option list in the core.  used in the case of testing a keystroke 
// Due to the optimisation of just using the cached core actions in the case of the !Updateable  call there is no longer a need for these 3  
// functions and there test cases so can be removed as dead code. 
UpdateKeyboardOptionsCore
SaveKeyboardOptionsCore
RestoreKeyboardOptionsCore
DisposeKeyboardOptionsCore

- // Used for loading and saving options between the core loaded keyboard and the windows registry. Core was mainly as a point of difference vs the windows engine core while the feature flag was present.
// it does cary more meaning and suggest a name the suggest the flow of data LoadKeyboardOptionsREGtoCore. SaveKeyboardOptionCoretoReg 
LoadKeyboardOptionsREGCore
SaveKeyboardOptionREGCore

- // Was named when refactoring the ProcessEvent call to be better optimised around the TSF not updatable call
Process_Event_Core

- // The Core was mainly as a point of difference vs the windows engine core while the feature flag was present. So could remove the core name 
MapKeyboardCore
MapKeyRuleCore

rc-swag:
Fixes #8703 
This follows on the removal of addins. 

I made the decision to remove 

1. These functions are about saving and restoring the keyboard option list in the core used in the case of testing a keystroke in the NonUpdateable case. Due to the optimisation of just using the cached core actions in the case of the !Updateable call there is no longer a need for these 3 functions and their test cases so can be removed as dead code.
UpdateKeyboardOptionsCore
SaveKeyboardOptionsCore
RestoreKeyboardOptionsCore
DisposeKeyboardOptionsCore

2.  Removed the core name
MapKeyboardCore
MapKeyRuleCore

3. 
LoadKeyboardOptionsREGCore  -> LoadKeyboardOptionsRegistrytoCore
SaveKeyboardOptionREGCore -> SaveKeyboardOptionCoretoRegistry

# User Testing

**TEST_KEYBOARDOPTIONS_CHANGE**:  

This test requires IPA (SIL) keyboard to be installed (or a keyboard that has modifiable keyboard options).
Step 1. Go to `Keyman Configuration`, select the IPA (SIL) keyboard, click on `Keyboard options` and choose `Before`. 
Step 2. Open Notepad, start typing anything and notepad should NOT crash.
Step 3. Type `>`  `b` 
Step 4. **Expected output**  `ɓ`
Step 5. Exit Notepad and Keyman
Step 6. Go to `Keyman Configuration`, select the IPA (SIL) keyboard, click on `Keyboard options` and choose `After`.
Step 7. Type `>`  `b` 
Step 8. **Expected output**  `ɓ`

Note: `Before` `After` setting. This will change the sequence to produce the IPA character for example `>b` => `ɓ`  for before and if the setting is after it is `b>` => `ɓ`

**TEST_PRESERVEDKEYS** 
This is to test the key rules which contain <kbd>AltGr</kbd> or <kbd>Shift</kbd> + <kbd>AltGr</kbd> work correctly.

* Open LibreOffice or Notepad 
* Observere output of  ឱ​

* Observere output of ᧨


windows/src/engine/keyman32/keyboardoptions.cpp
```diff
@@ -19,8 +19,8 @@
 */
 #include "pch.h"
 
-BOOL IntLoadKeyboardOptionsCore(LPCSTR key, LPINTKEYBOARDINFO kp, km_kbp_state* const state);
-void IntSaveKeyboardOptionREGCore(LPCSTR REGKey, LPINTKEYBOARDINFO kp, LPCWSTR key, LPWSTR value);
+BOOL IntLoadKeyboardOptionsRegistrytoCore(LPCSTR key, LPINTKEYBOARDINFO kp, km_kbp_state* const state);
+void IntSaveKeyboardOptionCoretoRegistry(LPCSTR REGKey, LPINTKEYBOARDINFO kp, LPCWSTR key, LPWSTR value);
```
ermshiperete, line 23: Do you want to add an `s` to make the names between loading and saving consistent? (Similar in the other method names)

```suggestion
void IntSaveKeyboardOptionsCoretoRegistry(LPCSTR REGKey, LPINTKEYBOARDINFO kp, LPCWSTR key, LPWSTR value);
```

ermshiperete (commented):
I'm wondering about the `Load/SaveKeyboardOptions*` methods - is the "Core" relevant here, or would they be better called `LoadKeyboardOptionsFromRegistry` and `SaveKeyboardOptionsToRegistry`?

bharanidharanj:
* **TEST_KEYBOARDOPTIONS_CHANGE (PASSED):** Tested with the attached PR build (17.0.87-alpha-test-8707) in Windows 10 OS and here is my observation: 1. Installed IPA (SIL) keyboard. Selected 'Before' option from Keyboard options. Verified that after typing >b, it shows ɓ on the Notepad. Then, I selected 'After' option from Keyboard options. Verified that after typing b>, it shows ɓ on Notepad. Working as expected.

bharanidharanj:
# Test Results

* **TEST_PRESERVEDKEYS (PASSED):** Tested this as per the instructions in LibreOffice writer document and it showing the expected result.

mcdurdin (approved):

ermshiperete (changes requested):
LGTM but flagging as "Request changes" so that my questions don't slip off the radar with multiple comments after them...

windows/src/engine/keyman32/keyboardoptions.cpp
rc-swag: Loading and Saving are not exactly symmetrical. The Loading works on a list of options whereas saving saves only a single option, it is a deliberate option to leave the `s` off.

Issue #5283, opened 2021-06-14
feat(developer): modernise command line tooling


mcdurdin:
streamlining the use of the Keyman Developer tooling:

1. Native command-line Keyman development tools for Linux, macOS and Windows.
2. `apt install keyman-compiler` (or something like that)
3. A good set of documentation on 'getting started' (or refreshing knowledge) with the command line tools.
4. Update kmcomp and kmconvert to use the "generate" paradigm common to development tools these days (it's all there, just needs rewording and organising):
       kmdev generate project foo  # generates a new keyboard project from a template, supporting all platforms by default
       kmdev duplicate project foo bar  # copies the foo project, renaming files to bar as needed

Something like that is my long-term goal for the development tools. Then, a simple `kmdev build [foo]` will build the target package which can be distributed either directly to users or uploaded to Keyman Cloud for wider distribution.

**Update:** Design doc: https://docs.google.com/document/d/1ra7jQgg0ZDC8QcVIFRyA_Xw0kya3HS5ITSjJBsSNTtQ/edit#heading=h.me4lkdiw0zz5

---

**Keyman Developer:**
 - OS: any 
 - Keyman Developer version: 15.0

mcdurdin:
Intention:

Design: 16.0.
Implementation: Future.

darcywong00:
@mhosken has some thoughts on what he'd find convenient:

>ideally having kmcomp do .kmn->.kmp with a minimal package would be nice (as well as all the rest)
>
> so I would just type `kmcomp -o xyz.kmp xyz.kmn` would be wonderful
> if I want more, then I would have to create a .kps
> 
> also kmcomp -o xyz.kps xyz.kmn generating a template .kps (as would be used in the previous command) so that I can add whatever, would be nice too

mcdurdin:
> ideally having kmcomp do .kmn->.kmp with a minimal package would be nice (as well as all the rest)

[kmconvert](https://help.keyman.com/developer/current-version/context/kmconvert) generates a minimal project .kpj and .kps with the `template` parameter. Does this do what you need @mhosken?

mcdurdin:
> so I would just type `kmcomp -o xyz.kmp xyz.kmn` would be wonderful

[...]

PR #8716, opened 2023-05-05
feat(developer): uset api from wasm! 🙀

srl295:
For: #7234

@keymanapp-test-bot skip


developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -65,8 +66,11 @@ export class Compiler {
       this.compileKeyboardFile = this.wasmModule.cwrap('kmcmp_Wasm_CompileKeyboardFile', 'boolean', ['string', 'string',
         'number', 'number', 'number', 'string']);
       this.setCompilerOptions = this.wasmModule.cwrap('kmcmp_Wasm_SetCompilerOptions', 'boolean', ['number']);
+      this._parseUnicodeSet = this.wasmModule.cwrap('kmcmp_Wasm_ParseUnicodeSet', 'number', [ 'string', 'number', 'number']);
```
mcdurdin, line 69: I think we should keep the member names consistent -- perhaps rename `compileKeyboardFile` and `setCompilerOptions` to `_compileKeyboardFile` and `_setCompilerOptions` as well?

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -107,4 +111,72 @@ export class Compiler {
...
+   */
+  public async parseUnicodeSet(pattern: string, bufferSize: number) : Promise<UnicodeSet> {
+    if (!bufferSize) {
+      bufferSize = 100;
+    }
+
+    const initOk = await this.init();
+    if (!initOk) {
+      throw Error(`WASM machinery didn't start up properly`);
+    }
```
mcdurdin, line 129: init should be done separately to keep the async out of this func; also we have `reportMessage` rather than `throw` for handling errors? Perhaps we need a design discussion on error handling throughout kmc as it is currently expanding rapidly.

```suggestion
    if(!this.wasmModule) {
      this.callbacks.reportMessage(CompilerMessages.Fatal_MissingWasmModule());
      return null;
    }
```

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -107,4 +111,72 @@ export class Compiler {
...
+    // Module seems to be the usual name for it
+    const Module = this.wasmModule;
+    const buf = Module.asm.malloc(bufferSize * 2 * Module.HEAPU32.BYTES_PER_ELEMENT);
+    const rc = this._parseUnicodeSet(pattern, buf, bufferSize);
+    if (rc < 0) {
+      throw new UnicodeSetError(rc);
```
mcdurdin, line 135: I think we want to be reporting through the normal `reportMessage` callback and defining a Error_ message

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -107,4 +111,72 @@ export class Compiler {
...
+  }
+}
+
+// from kmcmplib.h
+export const KMCMP_USET_OK = 0;
+export const KMCMP_ERROR_SYNTAX_ERR = -1;
+export const KMCMP_ERROR_HAS_STRINGS = -2;
+export const KMCMP_ERROR_UNSUPPORTED_PROPERTY = -3;
+export const KMCMP_ERROR_UNSUPPORTED = -4;
+export const KMCMP_FATAL_OUT_OF_RANGE = -5;
```
mcdurdin, line 164: We should translate these to kmc error values in messages.ts.

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -107,4 +111,72 @@ export class Compiler {
...
+}
+
+export class UnicodeSetError extends Error {
+  code: number;
+  constructor(code:number) {
+    super();
+    this.code = code;
+    this.message = `UnicodeSet error: ${code}`;
+  }
+}
```
mcdurdin, line 156: Use messages.ts instead.

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -65,8 +66,11 @@ export class Compiler {
...
       this.setCompilerOptions = this.wasmModule.cwrap('kmcmp_Wasm_SetCompilerOptions', 'boolean', ['number']);
+      this._parseUnicodeSet = this.wasmModule.cwrap('kmcmp_Wasm_ParseUnicodeSet', 'number', [ 'string', 'number', 'number']);
     }
-    return this.compileKeyboardFile !== undefined && this.setCompilerOptions !== undefined;
+    return this.compileKeyboardFile !== undefined
+        && this.setCompilerOptions  !== undefined
+        && this._parseUnicodeSet    !== undefined;
   }
 
   public run(infile: string, outfile: string, callbacks: CompilerCallbacks, options?: CompilerOptions): boolean {
```
mcdurdin, line 76: I actually think this needs more refactoring to put the `callbacks` parameter into the constructor, which will avoid having to pass it to each endpoint.

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -107,4 +111,72 @@ export class Compiler {
...
+/**
+ * Represents a parsed UnicodeSet
+ */
+export class UnicodeSet {
+  pattern: string;
+  ranges: number[][];
+  constructor(pattern: string, ranges: number[][]) {
+    this.pattern = pattern;
+    this.ranges = ranges;
+  }
```
mcdurdin, line 175:
```suggestion
  constructor(public pattern: string, public ranges: number[][]) {
  }
```

developer/src/kmc-kmn/test/test-wasm-uset.ts
```diff
@@ -0,0 +1,70 @@
...
+describe('Compiler class', function() {
+  let consoleLog: any;
+
+  beforeEach(function() {
+    consoleLog = sinon.spy(console, 'log');
+  });
+
+  afterEach(function() {
+    consoleLog.restore();
+  });
```
mcdurdin, line 18: I don't think we need sinon.spy in this unit. As we should never be `console.log`ging directly from any of our funcs anyway.

mcdurdin (changes requested):
This is looking good. May need some discussion around error handling.

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: Except I think setCompilerOptions is called directly, whereas parseUnicodeSet has a shim JS function?

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: that might be good,  failed ot startup would definitely be fatal.

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: yes

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: +1

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: +1

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: that makes sense.   Callbacks and an async init.

developer/src/kmc-kmn/test/test-wasm-uset.ts
srl295: OK.

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: OK. I guess `run` and `runCompiler` are the real API.

I'll prefix it with `wasm`.  so `wasmCompileKeyboardFile`

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: nice!

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: done

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: done

srl295:
@mcdurdin  First chunk of review comments addressed.  I will refactor the messages, wanted to make sure I didn't gum up the machinery first.

srl295:
@mcdurdin merge from master worked fine 😅

srl295:
@mcdurdin  maybe I didn't formally re-request.  Anwyay , up to date w/ master, i think this should be ready to go?

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,35 +47,72 @@ const baseOptions: CompilerOptions = {
  */
 let callbackProcIdentifier = 0;
 
+/**
+ * The wrapped functions
+ */
+class WrappedWasmFuncs {
```
mcdurdin, line 53: :+1:

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,35 +47,72 @@ const baseOptions: CompilerOptions = {
...
+  compileKeyboardFile?: (a0: string, a1: string, a2: number, a3: number, a4: number, a5: string) => boolean;
+  parseUnicodeSet?: (a0: string, a1: number, a2: number) => number;
```
mcdurdin, line 55: I think we should give the actual parameter names here for posterity -- grab them from the C++ code

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,35 +47,72 @@ const baseOptions: CompilerOptions = {
...
 
-  public run(infile: string, outfile: string, callbacks: CompilerCallbacks, options?: CompilerOptions): boolean {
-    this.callbacks = callbacks;
-
-    if(!this.wasmModule) {
+  /**
+   * Verify that wasm is spun up OK.
+   * @returns true if OK
+   */
+  public verifyInitted() : boolean {
```
mcdurdin, line 100:
```suggestion
  public verifyInitialized() : boolean {
```

mcdurdin, line 100:
```suggestion
  public verifyInitialized() : boolean {
```

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,35 +47,72 @@ const baseOptions: CompilerOptions = {
...
+class WrappedWasmFuncs {
+  compileKeyboardFile?: (a0: string, a1: string, a2: number, a3: number, a4: number, a5: string) => boolean;
+  parseUnicodeSet?: (a0: string, a1: number, a2: number) => number;
+  setCompilerOptions?: (shouldAddCompilerVersion: number) => boolean;
+
+  constructor(wasmModule: any) {
+    this.compileKeyboardFile = wasmModule.cwrap('kmcmp_Wasm_CompileKeyboardFile', 'boolean', ['string', 'string', 'number', 'number', 'number', 'string']);
+    this.parseUnicodeSet = wasmModule.cwrap('kmcmp_Wasm_ParseUnicodeSet', 'number', ['string', 'number', 'number']);
+    this.setCompilerOptions = wasmModule.cwrap('kmcmp_Wasm_SetCompilerOptions', 'boolean', ['number']);
+  }
```
mcdurdin, line 62: I think we should consider a `throw` here if this fails as it's a fatal error, and then we can eliminate `ok()`?

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,35 +47,72 @@ const baseOptions: CompilerOptions = {
...
+   */
+  get ok(): boolean {
+    return this.parseUnicodeSet !== undefined
+      && this.setCompilerOptions !== undefined
+      && this.compileKeyboardFile !== undefined;
+  }
+};
+
 export class Compiler {
   wasmModule: any;
```
mcdurdin, line 75: We can put `wasmModule` into `WrappedWasmFuncs` as well.

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,35 +47,72 @@ const baseOptions: CompilerOptions = {
...
     callbackProcIdentifier++;
   }
 
-  public async init(): Promise<boolean> {
+  public async init(callbacks: CompilerCallbacks): Promise<boolean> {
+    if(!this.callbacks) {
+      this.callbacks = callbacks;
+    }
     if(!this.wasmModule) {
       this.wasmModule = await loadWasmHost();
```
mcdurdin, line 90: My suggestion is:

First, move `wasmModule` into `WrappedWasmFuncs` (probably rename that class then to say `WasmWrapper`).

Then, something like this:

```
try {
  this.wasm = new WasmWrapper();
} catch(e: Error) {
  this.callbacks.reportMessage(CompilerMessages.Fatal_MissingWasmModule({e}));
  return false;
}
```

Means we can just do `if(!this.wasm) { throw }` as the input condition for other functions in the class. I'm happy with a throw for those cases, because those are bugs in our code rather than an input or environment issue (e.g. missing wasm module is an environment issue and that's handled by `init()` which returns a polite error).

developer/src/kmc-kmn/src/compiler/messages.ts
```diff
@@ -47,11 +47,26 @@ export class CompilerMessages {
   static Fatal_UnexpectedException = (o:{e: any}) => m(this.FATAL_UnexpectedException, `Unexpected exception: ${(o.e ?? 'unknown error').toString()}\n\nCall stack:\n${(o.e instanceof Error ? o.e.stack : (new Error()).stack)}`);
   static FATAL_UnexpectedException = SevFatal | 0x1000;
 
-  static Fatal_MissingWasmModule = () => m(this.FATAL_MissingWasmModule, `Could not instanatiate WASM compiler module`);
+  static Fatal_MissingWasmModule = () => m(this.FATAL_MissingWasmModule, `Could not instantiate WASM compiler module or not initted`);
```
mcdurdin, line 50:
```suggestion
  static Fatal_MissingWasmModule = () => m(this.FATAL_MissingWasmModule, `Could not instantiate WASM compiler module or initialization failed`);
```

We'll be able to add the exception here also as a parameter.

mcdurdin (changes requested):
I like the refactoring of the wasm stuff into a separate class; can we take it a step further per my comments below? Cheers!

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: I'll put a reference in too, since I keep forgetting they are in CompilerInterfaces.cpp (for the _Wasm versions)

common/web/types/src/util/compiler-interfaces.ts
```diff
@@ -32,6 +32,18 @@ export function compilerErrorSeverityName(code: number): string {
...
+ * Format the error code number
+ * example: "FATAL:0x03004"
+ */
+export function compilerErrorFormatCode(code: number): string {
+  const severity = code & CompilerErrorSeverity.Severity_Mask;
+  const severityName = compilerErrorSeverityName(severity);
+  const errorCode = code & CompilerErrorSeverity.Error_Mask;
+  const errorCodeString = Number(errorCode).toString(16).padStart(5,'0');
+  return `${severityName}:0x${errorCodeString}`;
+}
```
mcdurdin, line 45: This is fine for now but may need updating as we consolidate the error reporting further as part of #5283. It doesn't match what is generated from NodeCompilerCallbacks at this point (which itself needs work):

https://github.com/keymanapp/keyman/blob/7746cadd980acba251788ab3800ed37874563c9f/developer/src/kmc/src/messages/NodeCompilerCallbacks.ts#L31-L38

common/web/types/src/util/compiler-interfaces.ts
```diff
@@ -137,3 +149,10 @@ export interface CompilerCallbacks {
  * @returns
  */
 export const CompilerMessageSpec = (code: number, message: string) : CompilerEvent => { return { code, message } };
+
+/**
+ * @param e Error-like
+ */
+export function compilerExceptionToString(e?: any) : string {
+  return `${(e ?? 'unknown error').toString()}\n\nCall stack:\n${(e instanceof Error ? e.stack : (new Error()).stack)}`;
+}
```
mcdurdin, line 158: :+1:

We need to use this in various message classes across kmc-*

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,33 +47,89 @@ const baseOptions: CompilerOptions = {
  */
 let callbackProcIdentifier = 0;
 
+/**
+ * Pointer in wasm-space
+ */
+type WasmPtr = number;
```
mcdurdin, line 53: :+1:

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,33 +47,89 @@ const baseOptions: CompilerOptions = {
...
+    }
+    this.compileKeyboardFile = this.Module.cwrap('kmcmp_Wasm_CompileKeyboardFile', 'boolean', ['string', 'string', 'number', 'number', 'number', 'string']);
+    this.parseUnicodeSet = this.Module.cwrap('kmcmp_Wasm_ParseUnicodeSet', 'number', ['string', 'number', 'number']);
+    this.setCompilerOptions = this.Module.cwrap('kmcmp_Wasm_SetCompilerOptions', 'boolean', ['number']);
+
+    if (this.parseUnicodeSet == undefined
+      || this.setCompilerOptions == undefined
+      || this.compileKeyboardFile == undefined) {
+        throw Error(`some wasm functions did not load properly.`);
+      }
```
mcdurdin, line 78:
```suggestion
    if (this.parseUnicodeSet === undefined
      || this.setCompilerOptions === undefined
      || this.compileKeyboardFile === undefined) {
        throw Error(`some wasm functions did not load properly.`);
    }
```

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,33 +47,89 @@ const baseOptions: CompilerOptions = {
...
-  public async init(): Promise<boolean> {
-    if(!this.wasmModule) {
-      this.wasmModule = await loadWasmHost();
-      this.compileKeyboardFile = this.wasmModule.cwrap('kmcmp_Wasm_CompileKeyboardFile', 'boolean', ['string', 'string',
-        'number', 'number', 'number', 'string']);
-      this.setCompilerOptions = this.wasmModule.cwrap('kmcmp_Wasm_SetCompilerOptions', 'boolean', ['number']);
+  public async init(callbacks: CompilerCallbacks): Promise<boolean> {
+    if(!this.callbacks) {
+      this.callbacks = callbacks;
     }
```
mcdurdin, line 103: Why not overwrite with parameter in all cases? Seems like it may be confusing otherwise.

```suggestion
    this.callbacks = callbacks;
```

developer/src/kmc-kmn/test/test-wasm-uset.ts
```diff
@@ -0,0 +1,67 @@
...
+  it('should start', async function() {
+    const compiler = new Compiler();
+    const callbacks = new TestCompilerCallbacks();
+    assert(await compiler.init(callbacks));
+    assert(compiler.verifyInitted());
+  });
+
+  it('should compile a basic uset', async function() {
+    const compiler = new Compiler();
+    // const callbacks = new TestCompilerCallbacks();
```
mcdurdin, line 18:
```suggestion
```

mcdurdin (approved):
LGTM, only minor tweaks requested now

common/web/types/src/util/compiler-interfaces.ts
srl295: Didn't see that. I looked in common and didn't see such a function. Yes it would make sense to consolidate. 

The use case was trying to track down why the wrong err was thrown.

developer/src/kmc-kmn/src/compiler/compiler.ts
srl295: makes sense.

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,33 +47,89 @@ const baseOptions: CompilerOptions = {
...
-    return this.compileKeyboardFile !== undefined && this.setCompilerOptions !== undefined;
+    if(!this.wasm) {
+      try {
+        this.wasm = await WasmWrapper.load();
+      } catch(e: any) {
+        this.callbacks.reportMessage(CompilerMessages.Fatal_MissingWasmModule({e}));
+        return false;
+      }
+    }
+    return this.verifyInitted();
```
srl295, line 112:
```suggestion
    return this.verifyInitialized();
```

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -47,33 +47,89 @@ const baseOptions: CompilerOptions = {
...
+      this.callbacks.reportMessage(CompilerMessages.Fatal_MissingWasmModule({}));
+      return false;
+    }
+    return true;
+  }
 
-    if(!this.wasmModule) {
-      this.callbacks.reportMessage(CompilerMessages.Fatal_MissingWasmModule());
+  public run(infile: string, outfile: string, options?: CompilerOptions): boolean {
+    if(!this.verifyInitted()) {
```
srl295, line 132:
```suggestion
    if(!this.verifyInitialized()) {
```

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -92,10 +148,10 @@ export class Compiler {
 
   private runCompiler(infile: string, outfile: string, options: CompilerOptions): boolean {
     try {
-      if(!this.setCompilerOptions(options.shouldAddCompilerVersion)) {
+      if (!this.wasm.setCompilerOptions(options.shouldAddCompilerVersion ? 1 : 0)) {
```
srl295, line 151: This was actually a hidden type mismatch.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.106-alpha](https://keyman.com/downloads/releases/alpha/17.0.106)

PR #8742, opened 2023-05-10
feat(common): add `builder_run_action` shorthand function

mcdurdin:
I got a bit tired of the longhand form of `builder_start_action` and `builder_finish_action` so thought I would wrap the normal path mode of this into a handy little one-liner. Use as appropriate (and read the docs!)

While I was there I added the missing docs for `builder_describe_outputs`.

@keymanapp_test_bot skip

### DIFF
--- a/developer/src/kmc-model/build.sh
+++ b/developer/src/kmc-model/build.sh
@@ -1,33 +1,28 @@
 #!/usr/bin/env bash
-#
-# Compiles the kmc lexical model compiler.
-#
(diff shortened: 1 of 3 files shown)


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.104-alpha](https://keyman.com/downloads/releases/alpha/17.0.104)

PR #8784, opened 2023-05-12
chore(linux): Fix installation build step on TC

ermshiperete:
When building on TC we install into a temporary directory. This broke with the recent introduction of build.sh because that now calls `make install` whereas previously we explicitly called `make install-temp` on TC. Instead of adding a `install-temp` action to `build.sh` we now check in the Makefile if sudo is set.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/Makefile
+++ b/linux/keyman-config/Makefile
@@ -6,7 +6,14 @@ default: clean version man langtags
 langtags:
 	cd buildtools && python3 ./build-langtags.py
 
-install: # run as sudo
+install:
+	if [ -n "${SUDO_USER}" ]; then \
+		make install-sudo; \
+	else \
+		make install-temp; \
+	fi
+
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

mcdurdin (approved):
Makes sense. Does this fix the issue that @srl295 experienced and reported in Slack?

ermshiperete:
> Makes sense. Does this fix the issue that @srl295 experienced and reported in Slack?

Yes.

srl295:
> > Makes sense. Does this fix the issue that @srl295 experienced and reported in Slack?
> 
> Yes.

Glad i said something!

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.107-alpha](https://keyman.com/downloads/releases/alpha/17.0.107)

Issue #8493, opened 2023-03-23
chore(developer): kmcmplib cleanup notes


mcdurdin:
Taking notes on things that need cleanup in kmcmplib as I work on kmc-kmn:

- [ ] `kmcmp_CompileKeyboardFile` returns a `bool` value so clean up its defn
- [ ] Ensure all inputs/outputs are UTF-8 (e.g. msgproc)

mcdurdin:
Two remaining tasks, will leave those to A17S16 as they are not critical path.

PR #8796, opened 2023-05-14
refactor(developer): move file write out of WriteCompiledKeyboard and CompileKeyboard merge

mcdurdin:
Relates to #8493.

Stage one of the kmcmplib filesystem refactoring, as we move to writing to buffers and allowing caller to write to disk. Moves responsibility for writing data out of WriteCompiledKeyboard and into kmcmp_CompileKeyboardFile. This means that there are now no file writes in Compiler.cpp, except for the temp UTF16fromUTF8 (which I will also change to a in-memory buffer "real soon now").

Note that responsibility for writing in kmcmp_CompileKeyboardFile is very much temporary as we'll move it out of the library altogether in a subsequent commit.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmcmplib/src/CompilerInterfaces.cpp
+++ b/developer/src/kmcmplib/src/CompilerInterfaces.cpp
@@ -144,9 +144,16 @@ EXTERN bool kmcmp_CompileKeyboardFile(char* pszInfile,
   bool result = CompileKeyboardHandle(fp_in, &fk);
   if(result) {
     KMX_DWORD msg;
-    if ((msg = WriteCompiledKeyboard(&fk, fp_out)) != CERR_None) {
+    KMX_BYTE* data = nullptr;
+    size_t dataSize = 0;
+    if ((msg = WriteCompiledKeyboard(&fk, &data, dataSize)) != CERR_None) {
       result = FALSE;
       AddCompileError(msg);
+    } else {
+      if(fwrite(data, 1, dataSize, fp_out) != dataSize) {
+        AddCompileError(CERR_UnableToWriteFully);
+      }
+      delete[] data;
     }
   } else {
     AddCompileError(CERR_InvalidValue);
--- a/developer/src/kmcmplib/src/CompilerInterfaces.cpp
+++ b/developer/src/kmcmplib/src/CompilerInterfaces.cpp
@@ -165,33 +165,27 @@ EXTERN bool kmcmp_CompileKeyboardFile(char* pszInfile,
   }
 
   FILE_KEYBOARD fk;
-  bool result = CompileKeyboard(pszInfile, &fk, ASaveDebug, ACompilerWarningsAsErrors, AWarnDeprecatedCode,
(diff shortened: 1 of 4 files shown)


developer/src/kmcmplib/src/CompilerInterfaces.cpp
```diff
@@ -133,105 +132,70 @@ EXTERN bool kmcmp_CompileKeyboardFile(char* pszInfile,
...
-  if(result) {
-    KMX_DWORD msg;
-    if ((msg = WriteCompiledKeyboard(&fk, fp_out)) != CERR_None) {
-      result = FALSE;
-      AddCompileError(msg);
-    }
-  } else {
-    AddCompileError(CERR_InvalidValue);
-  }
+  bool result = CompileKeyboardHandle(fp_in, static_cast<PFILE_KEYBOARD>(pfkBuffer));
```
ermshiperete, line 136: We don't need `AddCompileError` if `CompileKeyboardHandle` failed?

developer/src/kmcmplib/src/CompilerInterfaces.cpp
```diff
@@ -133,105 +132,70 @@ EXTERN bool kmcmp_CompileKeyboardFile(char* pszInfile,
...
-    fp_in = UTF16TempFromUTF8(fp_in, FALSE);
-
-  kmcmp::CodeConstants = new kmcmp::NamedCodeConstants;
+  KMX_DWORD msg;
+  KMX_BYTE* data = nullptr;
+  size_t dataSize = 0;
+  if ((msg = WriteCompiledKeyboard(&fk, &data, dataSize)) != CERR_None) {
+    AddCompileError(msg);
+  } else {
+    if(fwrite(data, 1, dataSize, fp_out) != dataSize) {
```
ermshiperete, line 185: What happens if we call `fwrite` if `fp_out` is NULL?

ermshiperete (changes requested):

developer/src/kmcmplib/src/CompilerInterfaces.cpp
mcdurdin: `CompileKeyboardHandle` already calls `AddCompileError` if it fails, with more detail.

developer/src/kmcmplib/src/CompilerInterfaces.cpp
mcdurdin: `SetError` is a macro that emits `return FALSE;`, so if `fp_out == NULL` we'll never get here (line 176). I want to remove `SetError` because it is so misleading -- did the initial round of work in #8570.

Now is probably as good a time as any -- it's only used in a handful of places, all within CompilerInterfaces.cpp :grin: so expect a PR to arrive soon.

ermshiperete (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.112-alpha](https://keyman.com/downloads/releases/alpha/17.0.112)

PR #8849, opened 2023-05-23
chore(linux): Add missing dependency to GHA

ermshiperete:
Since we reference `sourcepackage` in the upload-to-llso step it should be listed in the `needs` list.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.111-alpha](https://keyman.com/downloads/releases/alpha/17.0.111)

PR #8856, opened 2023-05-25
chore(developer): remove SetError macro

### LINKED ISSUE #8853: chore(developer): `SetError` is a macro that emits `return FALSE;`  Please remove
`SetError` is a macro that emits `return FALSE;`, so if `fp_out == NULL` we'll never get here (line 176). I want to remove `SetError` because it is so misleading -- did the initial round of work in #8570.

Now is probably as good a time as any -- it's only used in a handful of places, all within CompilerInterfaces.cpp :grin: so expect a PR to arrive soon.

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/pull/8796#discussion_r1203714784_

Plan to replace with `AddCompileError(); return FALSE;` to make the control flow explicit.

mcdurdin:
Fixes #8853.

The `SetError` macro emitted `return FALSE;` which was confusing for reviewing code flow. Replaced with explicit `AddCompileMessage()` and `return FALSE;`.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmcmplib/src/CompilerInterfaces.cpp
+++ b/developer/src/kmcmplib/src/CompilerInterfaces.cpp
@@ -11,8 +11,6 @@
 #include "../../../../common/windows/cpp/include/ConvertUTF.h"
 #include "../../../../common/windows/cpp/include/keymanversion.h"
 
-#define SetError(err)       { if(AddCompileError(err) || (err & CERR_FATAL)) return FALSE; }
-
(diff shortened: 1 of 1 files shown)


ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.113-alpha](https://keyman.com/downloads/releases/alpha/17.0.113)

PR #8899, opened 2023-05-31
refactor(developer): rearrange kmcmplib interface source

### LINKED ISSUE #8889: chore(developer): Move emscripten wrappers into their own .cpp. Also move CompileKeyboardHandle into its own .cpp
Move emscripten wrappers into their own .cpp. Also move CompileKeyboardHandle into its own .cpp, so CompilerInterfaces.cpp has only C public API functions listed.

mcdurdin:
Fixes #8889.

No code changes, just moves WASM interfaces into CompilerInterfacesWasm.cpp, and CompileKeyboardHandle is renamed to CompileKeyboardBuffer and moved into its own source file.

@keymanapp-test-bot skip


keyman-server:

PR #8907, opened 2023-06-01
refactor(developer): move filename consistency check to kmc

### LINKED ISSUE #8883: chore(developer): Move filename consistency check from kmcmplib into kmc

mcdurdin:
Fixes #8883.

kmcmplib no longer has any filesystem access, so it cannot verify if a referenced filename in a source file has the same case as the actual filename on disk (a risk when moving projects between platforms). So I opted to move this to the `loadFile` callback in kmc, which is the only place where filesystem is actually accessed, and added corresponding unit test.

Small additional fixes here:

1. Move from `Buffer` to `Uint8Array` in all kmc-* modules, so that we remove that barrier to running on web.
2. Use `callbacks.loadFile` instead of `callbacks.fs.readFileSync`, so that we can be sure to run the filename consistency check.
3. Fixed kps parser silently swallowing xml errors on load.
4. Added silent mode to NodeCompilerCallbacks so we could cleanly test the new filename consistency hint.
5. Noted a location where we still have NodeJS deps in kmc-ldml.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #8914, opened 2023-06-02
chore(common): remove url module ref from common/web/types

mcdurdin:
The url module is a node module. We need to move responsibility for resolving the path of the LDML XML <import> statements out of common/web/types, and into the ultimate consumer, so it's now surfaced as an option, along with a helper constant that reports the import.meta.url-relative base path of the standard imports that are compiled into common/web/types.

This hopefully means we can use this module in both browser and node contexts without trouble.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc-ldml/test/test-testdata-e2e.ts
+++ b/developer/src/kmc-ldml/test/test-testdata-e2e.ts
@@ -14,7 +14,7 @@ describe('testdata-tests', function() {
     const jsonFilename = makePathToFixture('test-fr.json');
 
     // Compile the keyboard
-    const testData = loadTestdata(inputFilename, compilerTestOptions);
+    const testData = loadTestdata(inputFilename, {...compilerTestOptions, debug: true, addCompilerVersion: false});
     assert.isNotNull(testData);
 
     const jsonData = JSON.parse(readFileSync(jsonFilename, 'utf-8'));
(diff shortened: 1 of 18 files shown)


mcdurdin:
[Web build failed](https://build.palaso.org/buildConfiguration/Keymanweb_TestPullRequests/386205?buildTab=log&focusLine=1119&logView=flowAware&linesState=673) because language-processor.js unit test is still cjs and was importing common/web/types, which had a ref to import.meta.url in declaring a const, which is not a valid url in CJS:

```ts
export const LDMLKeyboardXMLDefaultImportsURL = new URL(`../import/`, import.meta.url);
```

Side-stepping by making the URL metadata read a class member instead of a exported const in the module.

developer/src/kmc-ldml/test/test-testdata-e2e.ts
```diff
@@ -14,7 +14,7 @@ describe('testdata-tests', function() {
     const jsonFilename = makePathToFixture('test-fr.json');
 
     // Compile the keyboard
-    const testData = loadTestdata(inputFilename, {debug: true, addCompilerVersion: false});
+    const testData = loadTestdata(inputFilename, compilerTestOptions);
```
darcywong00, line 17: Is `{debug: true, addCompilerVersion: false}` set somewhere else?

mcdurdin: Good catch! Probably doesn't matter for this test but I should have ... included it.
```suggestion
    const testData = loadTestdata(inputFilename, {...compilerTestOptions, debug: true, addCompilerVersion: false});
```

keyman-server:

PR #8932, opened 2023-06-05
fix(core): Fix compiling with GCC 13

### LINKED ISSUE #8920: bug(core): support GCC 13 compilation
### Describe the bug

The v16.0.139 code compiles correctly with GCC 12, but using GCC 13 the `core` fails with a message ([full log](https://dpaste.com/77XLBL3V5)):

> ../../common/include/km_types.h:22:9: error: ‘uint32_t’ does not name a type

It can be solved adding library to [core/include/keyman/keyboardprocessor_bits.h](https://github.com/keymanapp/keyman/blob/master/core/include/keyman/keyboardprocessor_bits.h):

```c
#include <stdint.h>
```

Not `<stdint>` because this header is used in both C and C++ code.

Related to Arch Linux distribution, but might be useful to others.

P.S. Probably, a [bug in GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107134).

### Keyman version

16.0.139

### Operating system

Arch Linux

ermshiperete:
Fixes #8920.

@keymanapp-test-bot skip

### DIFF
--- a/core/include/keyman/keyboardprocessor_bits.h
+++ b/core/include/keyman/keyboardprocessor_bits.h
@@ -66,4 +66,5 @@
 #define USE_CHAR16_T
 #endif
 
+#include <stdint.h>
 #include <km_types.h>


darcywong00 (approved):
lgtm

mcdurdin:
Please wait before merge -- mac agents are both offline due to network outage, would like to confirm that mac core builds successfully with this change.

Developer build failure due to a network glitch, got past the core build so should be fine.

keyman-server:

Issue #8913, opened 2023-06-01
docs(core): Document how to build Core on Linux


ermshiperete:
With the current kmc/ldml work, the instructions for dependencies and how to  build no longer work. One error I'm getting when trying to run `core/build.sh configure` is that it complains that it can't find `emcc.py`. This might have to do with how I installed emscripten (`sudo apt install emscripten`).

mcdurdin:
Yeah, we don't really want emscripten on the path because it pollutes the path with its own versions of python, node, etc. It's pretty awful. So instead we use an env var:

https://github.com/keymanapp/keyman/blob/6d9377fed6aeaee3dbe564c09d97b2e5b4b0253b/core/commands.inc.sh#L117-L131

ermshiperete:
`commands.inc.sh` tries to set `EMSCRIPTEN_BASE` if it's not already set. The problem is that it looks for `emcc`, but then in `wasm.build.linux.in` we specify `emcc.py` - which fails if the two files are not in the same directory.

mcdurdin:
Hmm, I thought they would be in the same directory. But could we update `commands.inc.sh` to look for `emcc.py` and would that fix it? All three `wasm.build.*.in` files look for `emcc.py`, `em++.py`, and `emar.py`.

ermshiperete:
`emcc.py` wouldn't be found (because it's not on the path), but at least this would then indicate that you'll have to set `EMSCRIPTEN_BASE`.

mcdurdin:
Does #8934 resolve any remaining problems for you @ermshiperete? I think that if emcc.py is not on the path, then the instructions do work?

ermshiperete:
> Does #8934 resolve any remaining problems for you @ermshiperete? I think that if emcc.py is not on the path, then the instructions do work?

Unfortunately not. The documentation still doesn't tell which dependencies need to be installed and how to build core. 

With the dependencies installed that are currently in the documentation and running `core/build.sh` (or `core/build.sh configure`) I get errors that it can't find `tsc` which it wants for `common/tools/hextobin`. And there are probably more things I'm missing.

PR #8934, opened 2023-06-06
chore(core): look for emcc.py, not emcc

mcdurdin:
Relates to #8913.

@ermshiperete, can you verify that the shellcheck directive does what we want here for SC2148?

@keymanapp-test-bot skip

### DIFF
--- a/core/commands.inc.sh
+++ b/core/commands.inc.sh
@@ -1,4 +1,7 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
+# no hashbang for .inc.sh
+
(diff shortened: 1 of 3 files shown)


ermshiperete (approved):

ermshiperete:
> @ermshiperete, can you verify that the shellcheck directive does what we want here for SC2148?

Yes, that seems to work.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.120-alpha](https://keyman.com/downloads/releases/alpha/17.0.120)

PR #8695, opened 2023-04-28
spec(core): spec/impl for transform and vars 🙀

srl295:
Includes:
- #8966
- #8942 
- #8943 

core:
- spec updates for transforms `tran`/`bksp` and variables `vars`, removed `finl` and folded `ordr` into subtable of `tran`/`bksp`

common
- pattern-parser objects for parsing variables and regex
- KMX+ updates to support above
- utility functions on `Vars` to support variable operations
- utility functions on messages for formatting subexceptions

developer:
- refactor compiler system make compilation async
- refactor compiler sections to add async init (for unicodeset bringup)
- refactor compiler sections to have a dependency system
- make `strs`/`elem`/`list` actual compilers ('empty compilers' since they take no input)
- fixes to test harnesses to support the above
- `vars` and `tran`/`bksp` writing/rewriting
- builder for `vars`, and initial builder for `tran`/`bksp` (not fully proven)

For: #7377

@keymanapp-test-bot skip

Issue #7377, opened 2022-09-28
feat(developer): compiler updates for ldml transforms 🙀


srl295:
Updates to compiler for anything pertaining to transforms / UnicodeSets

Blocks #7375

#7042

srl295:
https://github.com/keymanapp/keyman/pull/8695 review parkinglot:
- [ ] https://github.com/keymanapp/keyman/pull/8695/files#r1223611629 merge common IDs

srl295:
I think the only remaining _known_ issue here is:

- [ ] need to support reorder -> externalized `uset` section

srl295:
final todos:
- [ ] support `<reorder from="\u{0127 0128}">` as equivalent to `<reorder from="\u{0127}\u{0128}">
- [ ] refactor KmnCompiler instances for uset

srl295:
Markers are in separate tickets, so declaring this one done for now. re basic implementation.

PR #8943, opened 2023-06-06
feat(core): validate tran 🙀

srl295:
- new helper for tran
- also for bksp
- fix a bug (not surprising) in build-tran (ref #8695)

For: #7377

@keymanapp-test-bot skip


core/src/kmx/kmx_plus.cpp
```diff
@@ -375,11 +376,100 @@ COMP_KMXPLUS_TRAN::valid(KMX_DWORD _kmn_unused(length)) const {
...
+    DebugLog("tran helper: validating '%c%c%c%c' newTran=%p", DEBUG_IDENT(newTran->header.ident), newTran);
+  }
+  is_valid = true;
+  if (newTran == nullptr) {
+    DebugLog("tran helper: invalid, newTran=%p", newTran);
+    // null = invalid
+    is_valid = false;
+    // No assert here: just a missing layer
+    return false;
+  }
```
mcdurdin, line 399:
```suggestion
  is_valid = true;
  if (newTran == nullptr) {
    DebugLog("tran helper: invalid, newTran=%p", newTran);
    // null = invalid
    is_valid = false;
    // No assert here: just a missing layer
    return false;
  }
  DebugLog("tran helper: validating '%c%c%c%c' newTran=%p", DEBUG_IDENT(newTran->header.ident), newTran);
```

Just reduces the verbosity of the code a little

core/src/kmx/kmx_plus.cpp
```diff
@@ -375,11 +376,100 @@ COMP_KMXPLUS_TRAN::valid(KMX_DWORD _kmn_unused(length)) const {
...
+  tran = newTran;
+  const uint8_t *rawdata = reinterpret_cast<const uint8_t *>(tran);
+  rawdata += LDML_LENGTH_TRAN;  // skip past non-dynamic portion
```
mcdurdin, line 402: Are we validating that we aren't pointing or reading outside buffers anywhere?

mcdurdin (approved):

core/src/kmx/kmx_plus.cpp
srl295: Yes the validity check on the buffer size happens in …TRAN.valid

srl295:
@mcdurdin  would like to merge this and the child PRs once #8943  passes tests which I expect shortly!

core/src/kmx/kmx_plus.cpp
mcdurdin: So this can't be called before TRAN.valid? Just wanting to make sure we don't accidentally pass bogus data to this function.

srl295: "A programmer looks both ways on a one way street". 

It wouldn't hurt to call newTran.valid() although I'm remembering that requires the end of memory region info which we don't have here.

srl295: TRAN.valid() is called by the macro that initialized the "tran" field in the kmx_plus object. And there it's called with the end of memory pointer to validate the end. So there shouldn't be a path where an invalid tran is passed to this function.

mcdurdin: > "A programmer looks both ways on a one way street".

In this part of the world, everyone looks both ways on a one way street. Or doesn't look at all.

PR #8948, opened 2023-06-07
fix(linux): Fix display of installed keyboard version

ermshiperete:
Previously if there was a newer version available on keyman.com we showed that version number instead of the version number of the installed keyboard.

# User Testing

## Preparations

- The tests can be run on any Linux version
- Reboot
- Download EuroLatin (SIL) version 2.0.1: <https://keyman.com/go/package/download/sil_euro_latin?version=2.0.1&tier=stable>
- Open Keyman Configuration
- If you have any EuroLatin (SIL) keyboards installed, uninstall them
- Install the just downloaded keyboard by clicking the "Install keyboard" button and selecting the downloaded file

## Tests

**TEST_VERSION_NO**: Correct version number shows

- Verify that the "Version" column shows version 2.0.1 for the EuroLatin (SIL) keyboard

### DIFF
--- a/linux/keyman-config/keyman_config/kmpmetadata.py
+++ b/linux/keyman-config/keyman_config/kmpmetadata.py
@@ -51,7 +51,6 @@ def print_info(info):
         print(type(e))    # the exception instance
         print(e.args)     # arguments stored in .args
         print(e)          # __str__ allows args to be printed directly,		pass
-        pass
 
 
 def print_system(system):
@@ -67,7 +66,6 @@ def print_system(system):
         print(type(e))    # the exception instance
         print(e.args)     # arguments stored in .args
         print(e)          # __str__ allows args to be printed directly,		pass
-        pass
 
 
 def print_options(options):
@@ -83,7 +81,6 @@ def print_options(options):
         print(type(e))    # the exception instance
         print(e.args)     # arguments stored in .args
         print(e)          # __str__ allows args to be printed directly,		pass
-        pass
 
 
 def print_keyboards(keyboards):
@@ -107,7 +104,6 @@ def print_keyboards(keyboards):
         print(type(e))    # the exception instance
         print(e.args)     # arguments stored in .args
         print(e)          # __str__ allows args to be printed directly,		pass
-        pass
 
 
 def determine_filetype(kblist, filename):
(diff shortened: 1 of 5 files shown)


linux/keyman-config/keyman_config/kmpmetadata.py
mcdurdin, line 1: Changes in this file don't appear to be related to the fix?

linux/keyman-config/keyman_config/list_installed_kmp.py
```diff
@@ -122,32 +122,21 @@ def get_kmp_version(packageID):
...
+    version = _get_kmp_version_internal(packageID, InstallLocation.Shared, version)
+    version = _get_kmp_version_internal(packageID, InstallLocation.User, version)
     return version
 
 
+def _get_kmp_version_internal(packageId, location, previousVersion):
+    kmp = get_installed_kmp(location)
+    if packageId in kmp:
+        version = kmp[packageId]['kmpversion']
+        if previousVersion and previousVersion < version or not previousVersion:
```
mcdurdin, line 135: String comparison of version numbers is not safe: `'2.0.1' > '10.0'`.

Also, this if statement is kinda strange, mixing `and` and `or not` like that, perhaps could do instead (still need to fix the version comparison though):

```suggestion
        if not previousVersion or previousVersion < version:
```

mcdurdin (changes requested):
Need to fixup the version comparison first I think. I am not 100% sure where the fix for the PR title is, either :grin:

bharanidharanj:
# Test Results

* **TEST_VERSION_NO (PASSED):** Tested with the attached PR build (17.0.120-1~PR-8948+lunar4) in Ubuntu 23.04 Lunar Linux OS (VM) and here is my observation: 1. Verified that the Version number appears correctly under the version column.

linux/keyman-config/keyman_config/list_installed_kmp.py
ermshiperete: We changed the version number comparison a while ago, but looks this one slipped through...

linux/keyman-config/keyman_config/kmpmetadata.py
ermshiperete: Right. I put the refactorings in a separate commit, but probably should have created a separate PR as well...

ermshiperete:
> Need to fixup the version comparison first I think. I am not 100% sure where the fix for the PR title is, either grin

Maybe misleading. What I meant was that this change fixes the method that returns the version number. With that the correct version number will be displayed :smile:

mcdurdin (approved):

keyman-server:

PR #8976, opened 2023-06-09
feat: dependency checker and git-search

mcdurdin:
Tooling to check dependency declaration consistency between build.sh, package.json, tsconfig.json. Left as a draft PR for now but it highlights many inconsistencies in common/web and developer/src modules.

For now setup to work only with ESM.

Also, helper script git-search.sh lets us search changes in a branch more easily.

@keymanapp-test-bot skip


mcdurdin:
Closing this, will revisit when we need it

PR #8999, opened 2023-06-12
fix(developer): keep kmc-kmn messages within namespace

### LINKED ISSUE #8998: bug(developer): kmc-kmn messages overlap other namespaces
The kmc-kmn messages start at 0x1000 for new messages, but we only have 0x0000-0x0FFF available to play with. These codes have been allocated in 17.0 so we need to fix this before we hit release.

https://github.com/keymanapp/keyman/blob/459f42b273f440a0a99a9b13652807e01249ad7c/developer/src/kmc-kmn/src/compiler/messages.ts#L39

vs 

https://github.com/keymanapp/keyman/blob/8bfd4c0bae2fe599cebb845422ea3ce004388727/common/web/types/src/util/compiler-interfaces.ts#L61-L91

mcdurdin:
Fixes #8998.

kmc-kmn messages were broken down into sub-namespaces, but the range overlapped the reserved namespace mask range. Fixed the message values, added a new test to the verifyCompilerMessagesObject function to verify that messages don't creep outside their namespace, and defined new masks (in CompilerErrorSeverity enum... hmm) to help.

I opted to remove the leading `0` in the messages for kmc-kmn as I had to correct a number of them anyway, but not the messages in remaining units at this time (it's not an error, but just slightly misleading as we only have 12 bits, not 16 to play with).

It is safe to reassign these messages as they were only assigned in 17.0 alpha.

@keymanapp-test-bot skip

### DIFF
--- a/common/web/types/src/util/compiler-interfaces.ts
+++ b/common/web/types/src/util/compiler-interfaces.ts
@@ -15,8 +15,13 @@ export enum CompilerErrorSeverity {
   Error =         0x300000, // Severe error where we can't continue
   Fatal =         0x400000, // OOM or should-not-happen internal problem
 
-  Severity_Mask = 0xF00000, // includes reserved bits
-  Error_Mask =    0x0FFFFF,
+  // Mask values for mapping compiler errors
+  // TODO: make this a separate enum?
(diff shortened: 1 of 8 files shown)


jahorton (approved):

keyman-server:

Issue #725, opened 2018-04-03
Build the master: error


chandrab699:
I am getting the below error while building the master. Please help.

```
Build started 4/2/2018 10:30:57 PM.
     1>Project "c:\keyman\windows\src\buildtools\getfilelocks\getfilelocks.sln" on node 1 (Clean target(s)).
     1>ValidateSolutionConfiguration:
         Building solution configuration "Release|Win32".
     1>Project "c:\keyman\windows\src\buildtools\getfilelocks\getfilelocks.sln" (1) is building "c:\keyman\windows\src\buildtools\getfilelocks\getfil
       elocks.vcxproj" (2) on node 1 (Clean target(s)).
     2>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.Cpp.Current.targets(108,3): error MSB4019: Th
       e imported project "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Platforms\Win32\Platform.targets" w
       as not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [c:\keyman\windows\src\buildtool
       s\getfilelocks\getfilelocks.vcxproj]
     2>Done Building Project "c:\keyman\windows\src\buildtools\getfilelocks\getfilelocks.vcxproj" (Clean target(s)) -- FAILED.
     1>Done Building Project "c:\keyman\windows\src\buildtools\getfilelocks\getfilelocks.sln" (Clean target(s)) -- FAILED.

Build FAILED.

       "c:\keyman\windows\src\buildtools\getfilelocks\getfilelocks.sln" (Clean target) (1) ->
       "c:\keyman\windows\src\buildtools\getfilelocks\getfilelocks.vcxproj" (Clean target) (2) ->
         C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.Cpp.Current.targets(108,3): error MSB4019:
       The imported project "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Platforms\Win32\Platform.targets"
        was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [c:\keyman\windows\src\buildto
       ols\getfilelocks\getfilelocks.vcxproj]

    0 Warning(s)
    1 Error(s)

[...]

PR #9001, opened 2023-06-13
fix(linux): Fix creation of markdown help files for Linux

ermshiperete:
Fixes [help.keyman.com #725](https://github.com/keymanapp/help.keyman.com/issues/725).

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/help-keyman-com.sh
+++ b/resources/build/help-keyman-com.sh
@@ -83,7 +83,7 @@ function help_product_path {
 function generate_markdown_help {
   if [ $platform == 'linux' ]; then
     pushd $KEYMAN_ROOT/linux/keyman-config > /dev/null
-    ./build-help.sh --md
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
LGTM -- building a bunch more than needs to be built for just .md files though?

ermshiperete:
> LGTM -- building a bunch more than needs to be built for just .md files though?

Yes, although it's the same as before (`build-help.sh` called `scripts/reconf.sh` which did a build as well; now we do the build first and when we call `build-help.sh` we pass `--no-reconf` to skip the `reconf.sh` script). The difference to before is that we temporarily install the schema which is required now with keyman-system-service.

The TC build could be improved if we'd swap the build steps to upload the help before we build for Launchpad (which blows away the build).

keyman-server:

Issue #8950, opened 2023-06-07
bug(developer): node v16 fails tests


srl295:
Problems trying to run `kmc` with node v16:

> Could not instantiate WASM compiler module: `unexpected section <Tag>`

passes with v18

```
Program cp found: YES (/usr/local/opt/coreutils/libexec/gnubin/cp)
Configuring k_002_tinyu32.kmx with command
Running command: /Users/srl295/.nvm/versions/node/v16.20.0/bin/node /Users/srl295/src/keymanapp/keyman/core/../developer/src/kmc build /Users/srl295/src/keymanapp/keyman/core/tests/unit/ldml/keyboards/k_002_tinyu32.xml --out-file /Users/srl295/src/keymanapp/keyman/core/build/wasm/release/tests/unit/ldml/keyboards/k_002_tinyu32.kmx
--- stdout ---
INFO 5002: Building /Users/srl295/src/keymanapp/keyman/core/tests/unit/ldml/keyboards/k_002_tinyu32.xml
FATAL 403001: Could not instantiate WASM compiler module or initialization failed: RuntimeError: Aborted(CompileError: WebAssembly.instantiate(): unexpected section <Tag> (enable with --experimental-wasm-eh) @+2187). Build with -sASSERTIONS for more info.

Call stack:
RuntimeError: Aborted(CompileError: WebAssembly.instantiate(): unexpected section <Tag> (enable with --experimental-wasm-eh) @+2187). Build with -sASSERTIONS for more info.
    at abort (file:///Users/srl295/src/keymanapp/keyman/developer/src/kmc-kmn/build/src/import/kmcmplib/wasm-host.js:146:19)
    at file:///Users/srl295/src/keymanapp/keyman/developer/src/kmc-kmn/build/src/import/kmcmplib/wasm-host.js:184:326
FATAL 400017: The compiler for 'vars' failed to initialize.
FATAL 405001: Unexpected exception: Error: Internal error: section bksp depends on uninitialized dependency vars

Call stack:
Error: Internal error: section bksp depends on uninitialized dependency vars
    at file:///Users/srl295/src/keymanapp/keyman/developer/src/kmc-ldml/build/src/compiler/compiler.js:144:31
    at Array.forEach (<anonymous>)
    at LdmlKeyboardCompiler.compile (file:///Users/srl295/src/keymanapp/keyman/developer/src/kmc-ldml/build/src/compiler/compiler.js:140:44)
[...]

PR #9008, opened 2023-06-13
chore: add engine clause to package.json

srl295:
- will warn if used in node !18
- per https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines this could be an error if `engine-strict` was enabled. 

```console
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: undefined,
npm WARN EBADENGINE   required: { node: '^18.x' },
npm WARN EBADENGINE   current: { node: 'v16.20.0', npm: '8.19.4' }
npm WARN EBADENGINE }
```

@keymanapp-test-bot skip

#8950


srl295:
it will warn if you're on 19 also, so we'll be prompted to fix when ever we bump again.

srl295:
I made it 18.x to be generic and not try to give a specific version.

mcdurdin (approved):
LGTM -- we should probably check and if necessary refresh the docs/build files at the same time?

srl295:
> LGTM -- we should probably check and if necessary refresh the docs/build files at the same time?

you mean, whenever this changes again?

I don't think this change necessitates anything, just causes a fail-fast

mcdurdin:
I mean where we are non-specific about Node version in the current build config docs -- e.g. docs/build/macos.md (and seems linux.md doesn't even mention it but we need it for hextobin for core). Just seems a good trailer to tag onto this patch.

srl295:
@mcdurdin will take a look.

srl295:
@ermshiperete @mcdurdin will we need node.js in the debian/control file?

ermshiperete:
> will we need node.js in the debian/control file?

This would mean that we'd have to repackage node.js 18 for Ubuntu 20.04 Focal and 22.04 Jammy. 

22.10 Kinetic and later come with node.js 18 in the default repos.

mcdurdin:
> will we need node.js in the debian/control file?

No, because Keyman for Linux doesn't use Node.

docs/build/linux-ubuntu.md
```diff
@@ -49,6 +49,10 @@ sudo apt install devscripts equivs
 sudo mk-build-deps --install linux/debian/control
 ```
 
+### Node.js
+
+Node.js v18 is required for the core build.
```
mcdurdin, line 54:
```suggestion
Node.js v18 is required for Core build, Web tests, and Developer command line tools.
```

I think I'd like to spend a day on the dev docs at some point to get them consistent once more -- the Win/mac docs also need this info...

mcdurdin (approved):

Issue #8993, opened 2023-06-12
feat(windows): Online Update - Scheduled Services - Stage 2


rc-swag:
Keyman for Windows online updates - **Scheduled Services** - Stage1

## Table of Contents
1. [Introduction](#introduction)
2. [Detail](#detail)

# Introduction 

The ideal end goal of the online updated work is to automate the updated process with zero user interaction. 
There are many obstacles to achieving this some of which are out of Keyman control. 
However, we can make steps towards this goal. 
This issue is for stage 2 adding a Scheduled Service will:
1. Reduce the need for requesting elevated permissions from the user.
2. Revisit the change made in #13153 
3. Allow the check for updates to be done on a guaranteed interval independent of Keyman running.
- Elevation - installing keyboards and Keyman both require elevation - will be solved by the scheduled task.
- Files in use - updating Keyman files often requires a restart (we can’t avoid).
      - Keyman32.dll will be locked with SetWindowsHookEx hooks – with no clean way of releasing it from memory in all processes
      - kmtip.dll can be locked as it is selected as an input method

# Detail

See the [original design notes](https://docs.google.com/document/d/1b2rybSf2188Vbgx8tystQn6351bZfROYkXzPYEAi-ro/edit#)

See  Stage 1: #10038  with changes to the Keyman Configuration

# Functional Spec
In this Stage, two scheduled tasks shall be added one to check for updates, the second to start the install process as admin avoiding the need to ask the user for `Elevated` permissions.

1. Auto Update scheduled task should always be installed, but do nothing if auto updates are off
    1. Check for updates should be triggered by user but run as an elevated scheduled task
2. Note we must be aware of user vs machine preferences!
    1. e.g. keyboard is installed for machine – but language preferences are per-user
    2. Keyman options are nearly all per-user (except base keyboard I think)
3. There are two scheduled tasks:
    1. **Check for updates** – checks online for an update, and downloads it to a cache - replaces equiv functionality in kmshell
        1. Keyman Configuration can trigger the same schedule task download via check for `updates now`.
        2. Scheduled task is the admin user so that it can store the cache data in folder that the installer running as admin can also access with out having to prompt the user for elevated access.
        3. Scheduled task downloads metadata + binaries and puts them in a cache folder (e.g. ProgramData\Keyman\Updates)
        4. Once it has finished, it signals Keyman Configuration (e.g. SendMessage) OR can just check for the cache download on start up. Keyman.exe could also poll the to see if the download is available hourly.
[...]

PR #9013, opened 2023-06-14
feat:(windows) create versioned and proxy dlls for keyman32 and keyman64 📲

rc-swag:
Fixes: #8993

- The rename existing `keyman32` Visual Studio(VS) project project (which built the dll)  `keyman32-ver`. 
- The rename existing `keyman64` VS project to `keyman64-ver`
- When built these will include a version identification in its name. The name is created by: TODO keyman32-ver{BuildNumber}-{alpha/beta}-[local]. It will also include the `local` tag if it is a local build.  Using the `VERSION_WITH_TAG` build variable
- Created a proxy dll for keyman32 and keyman64.  (Needed because existing IMX keyboards will need the keyman32 and keyman64 name)
- proxy dlls are only loaded (lazy loaded) if there is a IMX keyboard callback needed, they are also unloaded when the IMX keyboard is unloaded.


rc-swag:
This pull request has been replaced by #9202

PR #9019, opened 2023-06-14
feat(core): implementation of non-regex components of tran 🙀

srl295:
- some scaffolding for loading transforms
- add transform_test
- simple test, simple impl - non-regex, simple strings
- scaffolding for bksp
- bksp transform test, but disabled for now

For: #7377

@keymanapp-test-bot skip


core/src/ldml/ldml_processor.cpp
```diff
@@ -84,66 +84,23 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
...
-          str = kplus.strs->get(element.element);
-          // TODO-LDML: actually a UnicodeSet here
-        } else {
-          str = element.get_string();  // Get single UTF-16
-        }
-        list.push_back(str);  // add the string to the end
-      }
-      // Now, add the list to the map
-      transforms.add(list, tostr);
+  // load transforms
```
mcdurdin, line 97:
```suggestion
  // load bksp transforms
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -84,66 +84,23 @@ ldml_processor::ldml_processor(path const & kb_path, const std::vector<uint8_t>
...
+  if (kplus.bksp != nullptr && kplus.bksp->groupCount > 0) {
+    bksp_transforms.reset(km::kbp::ldml::load_transform_groups(kplus, kplus.bksp, kplus.bkspHelper));
+    if (!transforms) {
```
mcdurdin, line 100:
```suggestion
    if (!bksp_transforms) {
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -246,6 +214,7 @@ ldml_processor::process_event(
           }
         }
         if (last_char == 0UL) {
+          // couldn't find the last char
```
mcdurdin, line 217:
```suggestion
          /*
            We couldn't find a character at end of context (context is empty),
            so we'll pass the backspace keystroke on to the app to process; the
            app might want to use backspace to move between contexts or delete 
            a text box, etc. Or it might be a legacy app and we've had our caret
            dumped in somewhere unknown, so we will have to depend on the app to
            be sensible about backspacing because we know nothing.
          */
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -294,25 +263,33 @@ ldml_processor::process_event(
...
+          ctxtstr.append(ctxt[i]);
         }
-        // Now, add in the updated text
-        const std::u32string outstr32 = kmx::u16string_to_u32string(outputString);
-        for (size_t i = 0; i < outstr32.length(); i++) {
-          state->context().push_character(outstr32[i]);
-          state->actions().push_character(outstr32[i]);
+        const size_t matchedContext = transforms->apply(ctxtstr, outputString);
+
+        if (matchedContext > 0) {
```
mcdurdin, line 275: So we are not currently working with markers in the keyboard on this I guess?

core/src/ldml/ldml_processor.cpp
```diff
@@ -294,25 +263,33 @@ ldml_processor::process_event(
...
+          }
+          // Now, add in the updated text
+          const std::u32string outstr32 = kmx::u16string_to_u32string(outputString);
+          for (size_t i = 0; i < outstr32.length(); i++) {
+            state->context().push_character(outstr32[i]);
+            state->actions().push_character(outstr32[i]);
+          }
+          // Add a marker so we don't retransform this text.
+          state->context().push_marker(0x0);
+          state->actions().push_marker(0x0);
```
mcdurdin, line 291: I'm a little puzzled by this -- why do we need to do this?

core/src/ldml/ldml_transforms.cpp
```diff
@@ -6,64 +6,191 @@
...
+      if (subOutput.length() >= output.length()) {
+        output = subOutput; // simple: all output
+      } else {
+        if (subMatched <= output.length()) {
+          output.resize(output.length() - subMatched);
+          output.append(subOutput);
+        } else {
+          output = subOutput;
+        }
       }
```
mcdurdin, line 101: Hmm, we have two identical output - subOutput statements in this if/else conglomeration. Another mild refactor may be called for. This matches the existing logic you have:

```suggestion
      // now update 'output'
      if (subOutput.length() >= output.length() || subMatched > output.length()) {
        output = subOutput; // simple: all output
      } else {
        output.resize(output.length() - subMatched);
        output.append(subOutput);
      }
```

But, I am not sure I understand the logic here -- should we be comparing `subOutput.length()` to `output.length()` at all?  Is this change more correct?

```suggestion
      // now update 'output'
      if (subMatched >= output.length()) {
        output = subOutput; // simple: all output
      } else {
        output.resize(output.length() - subMatched);
        output.append(subOutput);
      }
```

core/src/ldml/ldml_transforms.cpp
```diff
@@ -6,64 +6,191 @@
...
+   * - s/z$/zz/ -> match=2, output='zz'
+   *      (but could contract to match=0, output='' as already present)
+   * - s/zz$/zw/ -> match=2, output='zw'
+   *      (but could contract to match=1, output='w')
+   */
+  size_t matched              = 0;
+  std::u16string updatedInput = input;  // modified input
+  for (auto group = transform_groups.begin(); group < transform_groups.end(); group++) {
+    // for each transform group
+    // break out once there's a match
```
mcdurdin, line 68: We don't seem to be breaking out of the inner loop with first match. A group should only be applying the first matching transform. A cleaner structure would be to search the group for a matching transform, and then do the application of that found transform outside the inner loop.

pseudocode:

```
transform = group.findFirstMatch(updatedInput, &subMatched);
if(transform) {
  ... do the transform
}
```

or 

```
transform = findFirstMatchInGroup(group, updatedInput, &subMatched);
if(transform) {
  ... do the transform
}
```

core/tests/unit/ldml/keyboards/k_003_transform.xml
```diff
@@ -3,55 +3,58 @@
...
-      <transform from="beq" to="n" />
+      <transform from="^${e}" to="ê"/>
     </transformGroup>
   </transforms>
   <transforms type="backspace">
     <transformGroup>
+      <!-- 'ab' deleted as a group -->
       <transform from="ab" />
     </transformGroup>
     <transformGroup>
```
mcdurdin, line 56: Is this reorder group really intended to be under bksp?

core/tests/unit/ldml/keyboards/k_006_backspace.xml
```diff
@@ -0,0 +1,61 @@
...
+    <transformGroup>
+      <transform from="^${e}" to="ê"/>
+    </transformGroup>
+  </transforms>
+  <transforms type="backspace">
+    <transformGroup>
+      <!-- 'ab' deleted as a group -->
+      <transform from="ab" />
+    </transformGroup>
+    <transformGroup>
```
mcdurdin, line 56: Ditto do we really want this reorder group under bksp?

core/tests/unit/ldml/test_transforms.cpp
```diff
@@ -0,0 +1,129 @@
...
+      tr.addTransformGroup(st);
+    }
+
+    // now test
+
+    // see if we can match the same
+    {
+      std::u16string src(u"ta");
+      bool res = tr.apply(src);
+      zassert_string_equal(src, std::u16string(u"e"));
```
mcdurdin, line 93: Can we add a comment on the expected flow?

```suggestion
      // ta --> tbb --> tccc --> tcd --> e
      zassert_string_equal(src, std::u16string(u"e"));
```

mcdurdin (changes requested):
This looks like we are heading in the right direction. I wonder if we want to be baking in use of `std::u16string` given we are going to need to be playing with markers soon anyway? It may be working with vectors of our utf32|marker struct will be cleaner and safer? (Except regex is coming to break us as well)

srl295:
Yea it was a vector at first in the old implementation.

srl295:
> Yea it was a vector at first in the old implementation. 

I'm wondering even about something like icu "Replaceable" ( not literally, but the idea of a mutator and a mutatee ) so the reorder engine or the transform engine or the marker engine can be a virtual interface.

srl295:
> This looks like we are heading in the right direction. I wonder if we want to be baking in use of `std::u16string` given we are going to need to be playing with markers soon anyway? It may be working with vectors of our utf32|marker struct will be cleaner and safer? (Except regex is coming to break us as well)

yeah, it may break, I'm trying to just see that I can implement before putting in the regex side.

core/src/ldml/ldml_processor.cpp
srl295: aha! left over from 'old transforms', will remove

srl295, line 291:
```suggestion
```

core/src/ldml/ldml_transforms.cpp
srl295: This is to resize the output buffer though.

core/tests/unit/ldml/keyboards/k_003_transform.xml
srl295: oops, no.  thanks

core/src/ldml/ldml_transforms.cpp
srl295: Have not restructured. but I did make it break out of the inner loop.

mcdurdin (approved):
LGTM I think (although I am still a fan of segmenting the loop to find transform from the actual transform work, which to my mind just reduces the complexity of reading through the algorithm.

And I'm still not 100% sure I understand the actual transform performed in ldml_transforms.cpp ll.75-99, particularly what the purpose of the variables are (what is updatedInput, subOutput vs output, subMatched vs matched). Perhaps some additional comments on what is happening might help?

srl295:
@mcdurdin linux er seems unrelated

mcdurdin:
> linux error seems unrelated. 

See #9053.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.127-alpha](https://keyman.com/downloads/releases/alpha/17.0.127)

PR #9029, opened 2023-06-16
chore(common): Bump crowdin github action to 1.3.2

darcywong00:
The crowdin [workflow](https://github.com/keymanapp/keyman/actions/workflows/crowdin.yml) started failing 2 months ago with the error:
> null: Unrecognized field "customSegmentation" (class com.crowdin.client.sourcefiles.model.OtherFileImportOptions), not marked as ignorable (one known property: "contentSegmentation"])

This is fixed by bumping the crowdin GHA which updates crowdin-cli (fixed in version 3.6.4)

See https://github.com/crowdin/crowdin-cli/issues/379

@keymanapp-test-bot skip

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -15,7 +15,7 @@ jobs:
       uses: actions/checkout@v3
 
     - name: crowdin action
-      uses: crowdin/github-action@1.0.4
+      uses: crowdin/github-action@1.3.2
       with:
         upload_sources: true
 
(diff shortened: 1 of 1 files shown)


mcdurdin:
Did the workflow raise errors? Wondering why we weren't notified of the error?

mcdurdin (approved):

darcywong00:
> Did the workflow raise errors? Wondering why we weren't notified of the error?

I don't think the GHA does any notification of pass/fail.

I only realized cause my forked repo (which doesn't have keys) got notified the inactive GHA would be disabled.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.125-alpha](https://keyman.com/downloads/releases/alpha/17.0.125)

mcdurdin:
> I don't think the GHA does any notification of pass/fail.

Is there any way we can raise a notification on failure? It's not good for errors to be silently swallowed.

PR #9038, opened 2023-06-19
refactor(developer): kmc-kmn now builds .js when needed 🗜

mcdurdin:
kmc-kmn calls into kmc-kmw to build .js when it is needed, rather than kmc-kmw depending on kmc-kmn. This means examining the `&targets` system store to determine which files need to be generated.

This also implements the `&displayMap` rewrite for both .kvks and .keyman-touch-layout as part of the KMW compiler.

I suspect that the next step is to place kmc-kmw into a subfolder of kmc-kmn, given they are part of the same process, and there is no practical way to separate out the .js generation from the .kmx generation.

@keymanapp-test-bot skip


developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -10,15 +10,25 @@ import { UnicodeSetParser, UnicodeSet, Osk } from '@keymanapp/common-types';
...
+import { WriteCompiledKeyboard } from '@keymanapp/kmc-kmw';
 
 export interface CompilerResultFile {
   filename: string;
   data: Uint8Array;
 };
 
+export const COMPILETARGETS_KMX =   0x01;
+export const COMPILETARGETS_JS =    0x02;
+export const COMPILETARGETS__MASK = 0x03;
```
darcywong00, line 22: Do we note that these match kmcmplibapi.h?

developer/src/kmcmplib/src/Compiler.cpp
```diff
@@ -1146,6 +1149,45 @@ KMX_DWORD ProcessSystemStore(PFILE_KEYBOARD fk, KMX_DWORD SystemID, PFILE_STORE
...
+    for(auto p: KMXKeymanTargets) {
+      if(p == token) result |= COMPILETARGETS_KMX;
+    }
+    for(auto p: KMWKeymanTargets) {
+      if(p == token) result |= COMPILETARGETS_JS;
+    }
+
+    token = u16tok(nullptr, u" ", &ctx);
+
+    // Future: consider warnings on invalid compile targets?
```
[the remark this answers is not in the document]
mcdurdin: Yes, see CompileKeyboardBuffer.cpp:45

developer/src/kmc-kmn/package.json
```diff
@@ -61,6 +62,7 @@
     "exclude-after-remap": true,
     "exclude": [
       "src/import/",
+      "src/kmw-compiler",
```
mcdurdin, line 65: This will come back out at some point -- once we have coverage for kmw-compiler (if we ever get there).

jahorton (approved):

developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -254,6 +302,24 @@ export class KmnCompiler implements UnicodeSetParser {
...
   }
 
+  private loadDisplayMapping(kmnFilename:string, displayMapFilename: string): Osk.PuaMap {
+    // Remap using the osk-char-use-rewriter
+    let mapping: any;
+
+    displayMapFilename = this.callbacks.resolveFilename(kmnFilename, displayMapFilename);
+    try {
+      // Expected file format: displaymap.schema.json
+      // TODO: verify with schema
```
mcdurdin, line 312: This comment will be deleted.

developer/src/kmc-kmn/test/kmw/test-compiler-manual.ts
```diff
@@ -26,21 +25,19 @@ if(!await kmnCompiler.init(callbacks)) {
   process.exit(1);
 }
 
+// TODO: this needs rewrite due to circular deps
```
mcdurdin, line 28:
```suggestion
```

Issue #8955, opened 2023-06-08
feat(developer): KeymanWeb Typescript Compiler 🗜


mcdurdin:
## TODO

Future build infrastructure:
* `# TODO: remove c8 exclusion of coverage of kmw-compiler folder`

**various files:**
* [x] #8956
* [x] Sort out discrepancy between `color` and `style` in `VisualKeyboardFont` --> #9054
* [x] #8957
* [x] #8958
* [x] Lifecycle for global variables, put them all in one place at least, to start with (#9046)
* [x] Invert kmc-kmn and kmc-kmw dependency -- make kmc-kmn dependent on kmc-kmw rather than vice-versa --> #9038

Future common/web/types work:
* Review consistency of various common/web/types readers and writers (group task would be good?)

**javascript-strings.ts:**
* [x] Move constant declarations such as VKeyNames into separate units (in a _consts_ folder?)  (#9046)
* [x] kmx compiler needs to support build-to-memory-buffer
* [x] `// TODO: lifecycle` (#9046)
* [x] Get readonly groups and store metadata from kmx compiler when building for .js --> #9076
* [x] `// TODO: export group + store metadata in debug store, use that` --> #9076
* [x] Remove various commented code sections (mostly Delphi variable declarations) --> #9086
* [x] `// TODO: problems with supp chars` --> #9086
* [x] `// TODO: move elsewhere`

Following TODO items are from the Delphi code:
* `// TODO:  Refactor the section below into JavaScript_String, as it's`
* `// TODO: this doesn't resolve the issue of, e.g. SHIFT+K_SPACE`
* `// TODO: context character level switches instead of full context comparisons`

**validate-layout-file.ts:**
* [x] variable lifecycle (#9046)
* [x] `// TODO lifecycle` (#9046)
* [x] error reporting --> #9087
* [x] `// TODO: callbacks.reportMessage() ReportError(0, CWARN_TouchLayoutMissingLayer, 'Key "'+FId+'" on platform "'+FPlatform.Name+'", layer "'+FLayer.Id+'", platform "'+FPlatform.Name+'", references a missing layer "'+FNextLayer+'".');      ` --> #9087
* [x] `// TODO: ReportError(0, CWARN_TouchLayoutUnidentifiedKey, 'A key on layer "'+FLayer.Id+'" has no identifier.');` --> #9087
[...]

PR #9054, opened 2023-06-21
chore(developer): remove font style/color from kvk 🗜

mcdurdin:
Relates to TODO items in #8955.

While .kvk includes a font color field, it is essentially unused. There was also some sort of support provided for font style in the KeymanWeb keyboard compiler, but this was never streamed into .kvk or .kvks files, so was always a no-op.

Thus, this PR removes any semblance of support for font style and color from the .kvk and .kvks readers and writers. When we write the font color field which is present in .kvk, we always use the default TColor.clWindowText which is what the legacy Delphi-based writer would always have written.

Also sorted out the default font size and name in the .kvk transform from .kvks, and in so doing cleaned up the basic.kvk and basic.js in LDML keyboard compiler to match what we are doing in the .kmn compiler.

@keymanapp-test-bot skip


common/web/types/src/kvk/kvk-file-writer.ts
```diff
@@ -17,19 +17,27 @@ export default class KvkFileWriter {
   }
 
   private build(source: VisualKeyboard) {
+    /**
+     * The visual keyboard text color is included in .kvk files but is never
+     * defined in .kvks files. Thus, it is always written by the legacy .kvk
+     * binary writer VisualKeyboardSaverBinary.pas as TColor.clWindowText (-8),
+     * or in DWORD form 0xFF000008, and we maintain this value for consistency
```
mcdurdin, line 24:
```suggestion
     * binary writer VisualKeyboardSaverBinary.pas as TColor.clWindowText (SystemColor | 8),
     * or in DWORD form 0xFF000008, and we maintain this value for consistency
```

developer/src/kmc-ldml/src/compiler/keymanweb-compiler.ts
```diff
@@ -22,7 +22,7 @@ export class LdmlKeyboardKeymanWebCompiler {
     const vk: VisualKeyboard.VisualKeyboard = vkc.compile(source);
 
     let result =
-      `{F: '${vk.header.unicodeFont.size}pt ${JSON.stringify(vk.header.unicodeFont.name)}', `+
+      `{F: ' 1em ${JSON.stringify(vk.header.unicodeFont.name)}', `+
```
mcdurdin, line 25: We always ignore font size in the vk export to kmw... so `1em` it is. (Fortunately, the whole font string is normally ignored and will be removed in a future version)

windows/src/engine/kmcomapi/processes/package/kpinstallpackage.pas
```diff
@@ -274,6 +274,8 @@ procedure TKPInstallPackage.Execute(const FileName: string; Options: TKPInstallP
             try
               with TKPInstallVisualKeyboard.Create(Context) do
               try
+                // TODO: we should probably try and do this by the keyboard install instead so
+                // that we can deprecate the associatedKeyboard field in .kvk
```
mcdurdin, line 278: TODO: we should probably try and do this by the keyboard install instead so that we can deprecate the associatedKeyboard field in .kvk

jahorton (approved):

PR #9059, opened 2023-06-21
chore(developer): cleanup data passing from wasm 🗜

mcdurdin:
Relates to #8955.

Reorganizes the data structures passed out of kmcmplib via WASM to make it easier to work with the extra metadata and reduce the number of places we have to touch when we add extra metadata fields.

Preparation for supporting the metadata that the kmw compiler requires.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc-kmn/src/compiler/compiler.ts
+++ b/developer/src/kmc-kmn/src/compiler/compiler.ts
@@ -22,7 +22,7 @@ export const COMPILETARGETS_JS =    0x02;
 export const COMPILETARGETS__MASK = 0x03;
 
 /**
- * Data in CompilerResultMetadata comes from kmcmplib
+ * Data in CompilerResultExtra comes from kmcmplib
  */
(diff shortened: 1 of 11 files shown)


developer/src/kmc-kmn/src/compiler/compiler.ts
```diff
@@ -21,21 +21,24 @@ export const COMPILETARGETS_KMX =   0x01;
 export const COMPILETARGETS_JS =    0x02;
 export const COMPILETARGETS__MASK = 0x03;
 
-export interface CompilerResultMetadata {
+/**
+ * Data in CompilerResultMetadata comes from kmcmplib
```
ermshiperete, line 25: Is this comment still correct or should that be renamed as well?

```suggestion
 * Data in CompilerResultExtra comes from kmcmplib
```

ermshiperete (approved):

PR #9061, opened 2023-06-21
fix(windows): testhost improve error message

srl295:
if dwError is 0, don't try to format it
avoid "Error starting Keyman: The operation completed successfully.  [0]"

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/engine/testhost/testhost.cpp
+++ b/windows/src/engine/testhost/testhost.cpp
@@ -26,6 +26,7 @@ INT_PTR CALLBACK    About(HWND, UINT, WPARAM, LPARAM);
 void SizeEditWindow();
 void StartKeyman();
 void StopKeyman();
+void Fail(PCWSTR message);
 
   int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
                      _In_opt_ HINSTANCE hPrevInstance,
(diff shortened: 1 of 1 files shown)


windows/src/engine/testhost/testhost.cpp
```diff
@@ -298,7 +299,9 @@ Fail(PCWSTR message) {
   std::vector<wchar_t> err(bufferLength);
   std::wstring buf;
   DWORD dwError = GetLastError();
-  if (FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, nullptr, dwError, 0, err.data(), bufferLength, nullptr) != 0) {
+  if (dwError == ERROR_SUCCESS) {
+    buf = std::wstring(message); // Don't append "The operation completed successfully."
```
mcdurdin, line 303:
```suggestion
    buf = std::wstring(message) + L": no error code was returned"; // Don't append "The operation completed successfully."
```

Not that it matters very much for this particular in-house program, but I think for clarity we should be at least saying that we got a zero error code.

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.128-alpha](https://keyman.com/downloads/releases/alpha/17.0.128)

PR #9072, opened 2023-06-22
fix(core): fix ldml transforms 🙀

srl295:
- document transform calling code in ldml_transform
- remove obsolete portions
- optimize keystroke code when no transforms are present
- also, fix 'enter' and 'delete' breakage, and other keys passed to the OS like 'tab'

@keymanapp-test-bot skip

#7377

### DIFF
--- a/core/src/ldml/ldml_processor.cpp
+++ b/core/src/ldml/ldml_processor.cpp
@@ -230,72 +230,79 @@ ldml_processor::process_event(
       }
       break;
     default:
-      // from kmx_processor.cpp
-      // Construct a context buffer from the items up until the last KM_KBP_CT_MARKER marker
-      ldml::string_list ctxt;
-      auto cp = state->context();
-      // We're only interested in as much of the context as is a KM_KBP_CT_CHAR.
-      // This will stop at the KM_KBP_CT_MARKER type.
-      uint8_t last_type = KM_KBP_BT_UNKNOWN;
-      for (auto c = cp.rbegin(); c != cp.rend(); c++) {
-        last_type = c->type;
-        if (last_type != KM_KBP_BT_CHAR) {
-          break;
-        }
-        km::kbp::kmx::char16_single buf;
-        const int len = km::kbp::kmx::Utf32CharToUtf16(c->character, buf);
-        const std::u16string str(buf.ch, len);
-        ctxt.push_front(str); // prepend to string
-      }
-      if (last_type != KM_KBP_BT_MARKER) {
-        // There was no beginning-of-translation marker.
-        //Add one.
-        state->context().push_marker(0x0);
-        state->actions().push_marker(0x0);
-      }
-      // Look up the key
-      const std::u16string str = keys.lookup(vk, modifier_state);
-      if (str.empty()) {
-        // not found
-        state->actions().commit(); // finish up and
-        return KM_KBP_STATUS_OK; // Nothing to do- no key
-      }
-      const std::u32string str32 = kmx::u16string_to_u32string(str);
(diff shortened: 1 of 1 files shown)


srl295:
https://github.com/keymanapp/keyman/assets/855219/c2c335c8-c78b-4d7a-aab2-ca735c31b7ee

srl295:
Also fix backspace.

core/src/ldml/ldml_processor.cpp
```diff
@@ -225,77 +225,84 @@ ldml_processor::process_event(
           state->actions().push_backspace(KM_KBP_BT_UNKNOWN);
         } else {
           state->actions().push_backspace(KM_KBP_BT_CHAR, last_char);
+          state->context().pop_back();
         }
-        state->context().pop_back();
```
srl295, line 229: if the context is empty (see prior loop/comment), pop_back() understandably throws

core/src/ldml/ldml_processor.cpp
```diff
@@ -225,77 +225,84 @@ ldml_processor::process_event(
...
-        // TODO-LDML: unroll ctxt into a str
-        std::u16string ctxtstr;
-        for (size_t i = 0; i < ctxt.size(); i++) {
-          ctxtstr.append(ctxt[i]);
+
+        // Look up the key
+        const std::u16string str = keys.lookup(vk, modifier_state);
+        if (str.empty()) {
+          // not found
+          break; // ----- commit and exit
```
srl295, line 264: there's probably a better structure for this at this point.  but at least, this is updated to have fewer return; points.

core/src/ldml/ldml_processor.cpp
```diff
@@ -225,77 +225,86 @@ ldml_processor::process_event(
...
+      {
+        // adapted from kmx_processor.cpp
+
+        /** a copy of the current/changed context, for transform use */
+        ldml::string_list ctxt;
+
+        // Construct a context buffer of all the KM_KBP_BT_CHAR items
+        // Extract the context into 'ctxt' for transforms to process
+        if (!!transforms) {
+          // if no transfors, no reason to do this extraction
```
mcdurdin, line 242:
```suggestion
          // if no transforms, no reason to do this extraction
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -225,77 +225,86 @@ ldml_processor::process_event(
...
+          auto &cp = state->context();
+          // We're only interested in as much of the context as is a KM_KBP_BT_CHAR.
+          uint8_t last_type = KM_KBP_BT_UNKNOWN;
+          for (auto c = cp.rbegin(); c != cp.rend(); c++) {
+            last_type = c->type;
+            if (last_type != KM_KBP_BT_CHAR) {
```
mcdurdin, line 248: Does this cope with markers interleaved in the context?

mcdurdin (approved):
I think LGTM, just one question around backspace actions and markers

core/src/ldml/ldml_processor.cpp
mcdurdin: I don't actually mind `return` -- especially if it reduces depth of algorithm.

core/src/ldml/ldml_processor.cpp
srl295: It just stops when it gets to a non char (such as a marker). will need to revisit w/ markers

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.129-alpha](https://keyman.com/downloads/releases/alpha/17.0.129)

PR #9075, opened 2023-06-23
fix(developer): support option stores in kmc-kmn/kmw 🗜

mcdurdin:
Relates to #8955.

Adds support for option stores to kmc-kmn/kmw-compiler. This required adding the relevant store metadata into the output from kmcmplib, and so I also tidied up the corresponding WASM interfaces slightly more in the process.

Added unit test for the options store.

While testing this, ran into a second bug with the way I ported a constant from Delphi to Typescript in constants.ts, so fixed that and added a corresponding unit test. Small steps, but good ones.

@keymanapp-test-bot skip


developer/src/kmc-kmn/test/kmw/util.ts
```diff
@@ -1,13 +1,16 @@
...
 }
 
 export function extractTouchLayout(js: string): ETLResult|null  {
   let m = /KVKL=(?<kvkl>.+?);[\r\n]/ds.exec(js);
   if(!m) {
-    return null;
+    return {
+      js: js,
+      touchLayout: 'null'
+    };
```
mcdurdin, line 13: we always want to compare the js, even if no touch layout is present in the compiled keyboard.

darcywong00 (approved):
lgtm

PR #9084, opened 2023-06-23
feat(developer,common,core): ldml UTF-32 literals in binary 🙀

### LINKED ISSUE #9050: feat(developer): support UTF-32 single char strings 🙀
- developer: for the compiler
- core: for the C++ implementation
 
In the KMX+ format for LDML, certain fields can be either strings _or_ single UTF-32LE chars.  This saves space, being 4 bytes (plus the flag area) instead of 16 (4 for the char, 4 for the null, 8 for the string table entry)

- 'to' in the [`key2.keys` subtable](https://github.com/keymanapp/keyman/blob/master/core/src/ldml/C7043_ldml.md#key2keys-subtable)
- `element` in the [`elem` table](https://github.com/keymanapp/keyman/blob/master/core/src/ldml/C7043_ldml.md#c704323-elemtransform-variable-and-reorder-element-strings)
- `to` in the [`key2.flick` table](https://github.com/keymanapp/keyman/blob/master/core/src/ldml/C7043_ldml.md#key2flick-flick-element-subtable)

Until now this hasn't been implemented on the compiler side.  Change to always use single char entries for these.

srl295:
- use utf-32 literals in key `to` and flicks `to` and elem `element`

Fixes #9050 
 
@keymanapp-test-bot skip  (but not for long!)


srl295:
when we have lots of reorders this will really keep the string table from exploding

core/tests/unit/ldml/ldml_test_source.cpp
```diff
@@ -565,18 +565,11 @@ LdmlJsonRepertoireTestSource::next_action(ldml_action &fillin) {
...
-    fillin.type = LDML_ACTION_FAIL;
-    return;
-  }
-  assert(strId != 0);
+  KMX_DWORD strId = kmxplus->strs->find(chstr); // not an error if chstr is 0, may be single ch
 
   // OK. Now we can search the keybag
   KMX_DWORD keyIndex = 0;
-  auto *key2 = kmxplus->key2Helper.findKeyByStringTo(strId, keyIndex);
+  auto *key2 = kmxplus->key2Helper.findKeyByStringTo(chstr, strId, keyIndex);
```
srl295, line 572: we need the string id AND the string, so that it can search by character

common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -87,13 +87,20 @@ export class Name extends Section {
...
+  /** string value */
   readonly value: string;
-  constructor(value: string) {
+  /** char value if this is a single-char placeholder item (CharStrsItem) */
+  readonly char?: number;
+
+  constructor(value: string, char?: number) {
     this.value = value;
+    this.char = char;
   }
```
mcdurdin, line 98: It'd be good to have an assertion here, that if `char` is not `undefined`, that it actually is the first char from `value`.

mcdurdin (approved):
LGTM. Nice :grin:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.131-alpha](https://keyman.com/downloads/releases/alpha/17.0.131)

PR #9123, opened 2023-06-28
fix(linux): Fix GHA package builds for non-Jammy dists

### LINKED ISSUE #9024: bug(linux): Wrong dependencies for GHA PR builds
Trying to install Keyman from a GHA PR build on Wasta 20.04 will fail due to wrong dependencies. It looks like it uses the dependencies from Jammy instead of Focal.

ermshiperete:
Fixes #9024.

@keymanapp-test-bot skip

### DIFF
--- a/.github/workflows/deb-packaging.yml
+++ b/.github/workflows/deb-packaging.yml
@@ -113,8 +113,9 @@ jobs:
         path: artifacts
 
     - name: Build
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
LGTM. I checked every digit of the hash and they all looked like they were valid hexadecimal digits.





🤣

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.132-alpha](https://keyman.com/downloads/releases/alpha/17.0.132)

PR #9132, opened 2023-06-30
fix(developer): ensure `delete` uses array semantics where appropriate 🗜

mcdurdin:
The kmcmplib compiler source used `delete` instead of `delete[]` in a number of locations where the corresponding allocation was an array.

This doesn't appear to matter on arrays of primitives on most platforms that we are targeting, but it _is_ undefined behaviour so we should correct it.

https://stackoverflow.com/a/2425749/1836776

@keymanapp-test-bot skip


jahorton (approved):

srl295 (approved):
ah yes, this one!

PR #9133, opened 2023-06-30
fix(developer): `GetDelimitedString` was using wrong `strchr` 🗜

mcdurdin:
Picked this up while running `-fsanitize=address` for trying to track down another bug.

This fixes a buffer overrun (normally invisible) in kmcmplib, where `xstrchr` was being called, which takes a string as its second parameter, but being passed a single character. This was incorrect behaviour for two reasons:

1. Passing a single character doesn't have null termination (big bug!)
2. We were not wanting xstring semantics on the token being parsed here

Replaced with `u16chr`, which does do what we want. Note the casting because `u16chr` only has a `const` version of its input/output at present.

Removed `xstrchr` because it is not used anywhere else in kmcmplib.

@keymanapp-test-bot skip


jahorton (approved):

srl295 (approved):
good catch

PR #9134, opened 2023-06-30
fix(common): specify alignment for `COMP_` structs 🗜

mcdurdin:
Picked up by using `-fsanitize=undefined` on kmcmplib build.

When we read .kmx files, they have no alignment guarantees, so we need to tell the compiler to generate unaligned-safe code for accesses to `COMP_` structure members, because typically we point into the buffer and read at any offset.

A .kmx file will generally be 2-byte aligned, as there are no structures or data types with smaller than 2 byte widths. There is no requirement that this be the case per the .kmx spec, though. So we use a 1 byte alignment attribute for the compiler, so it will generate safe code when reading these structs.

Note that we are assuming that `COMP_KEYBOARD` is aligned because it is always the start of the file, so will be at the start of any buffer which will automatically be aligned correctly.

We should probably review some of our design decisions and structs for KMXPlus given this.

@keymanapp-test-bot skip

### DIFF
--- a/common/include/kmx_file.h
+++ b/common/include/kmx_file.h
@@ -7,6 +7,24 @@
 
 #include <km_types.h>
 
+/*
+  When we read .kmx files, they have no alignment guarantees, so we need to tell
+  the compiler to generate unaligned-safe code for accesses to COMP_ structure
+  members. Note we are assuming that COMP_KEYBOARD is aligned because it is
+  always the start of the file, so will be at the start of any buffer which will
+  automatically be aligned correctly.
+*/
+#ifdef __EMSCRIPTEN__
+typedef KMX_DWORD __attribute__((aligned(1))) KMX_DWORD_unaligned;
+typedef KMX_BOOL  __attribute__((aligned(1))) KMX_BOOL_unaligned;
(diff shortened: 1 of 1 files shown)


jahorton (approved):

srl295 (approved):
Or "How to make a DEC Alpha processor explode"

srl295:
> We should probably review some of our design decisions and structs for KMXPlus given this.

file a followup ticket for this?

mcdurdin:
> file a followup ticket for this?

Yes, two things:

* We should tag any KMXPlus file-based structs with alignment as well
* This attribute is currently only applied on WASM builds (clang-based), but we need to consider for other compilers as well (clang-other platforms, GCC, VC++)?

srl295:
> > file a followup ticket for this?
> 
> Yes, two things:
> 
> * We should tag any KMXPlus file-based structs with alignment as well
> * This attribute is currently only applied on WASM builds (clang-based), but we need to consider for other compilers as well (clang-other platforms, GCC, VC++)?

maybe 2 different tix 

I think there's GCC alignment pragmas that can work in clang.  So GCC + WIN might cover most everyone.

Issue #8069, opened 2023-01-24
chore(common): ldml TODO-LDML cleanup 🙀 


srl295:

mcdurdin:
Here's a snapshot of today's TODO-LDML state. We should take the important ones and turn them into issues, then close this issue. The remaining minor TODO-LDML items can remain in the source.

```
171 results - 56 files

common\web\types\build.sh:
  78    do
  79:     # TODO-LDML: developer/src/inst/download.in.mak needs these also...
  80      CLDR_PATH=$(dirname "$CLDR_INFO_PATH")

  82      mkdir -p "$THIS_SCRIPT_PATH/build/src/import/$CLDR_VER"
  83:     # TODO-LDML: When these are copied, the DOCTYPE will break due to the wrong path. We don't use the DTD so it should be OK.
  84      cp "$CLDR_INFO_PATH" "$CLDR_PATH/import/"*.xml "$THIS_SCRIPT_PATH/build/src/import/$CLDR_VER/"

common\web\types\src\consts\virtual-key-constants.ts:
  2  // Define standard keycode numbers (exposed for use by other modules)
  3: // TODO-LDML: merge with common\web\keyboard-processor\src\text\codes.ts
  4  

common\web\types\src\kmx\element-string.ts:
   74          typeFlag |= constants.elem_flags_type_uset;
   75:         // TODO-LDML: err on max buffer size
   76          const needRanges = sections.usetparser.sizeUnicodeSet(item.segment);

   78          if (!uset) {
   79:           // TODO-LDML - error callback
   80            throw Error(`Could not parse uset ${item.segment}`);

   88            str = unescapeOneQuadString(str);
   89:           // TODO-LDML: any other escape forms here?
   90            elem.value = sections.strs.allocString(str, { singleOk: true });

  104          (ElemElementFlags.type & typeFlag) |
  105:         (tertiary_bases?.[i] == '1' /* TODO-LDML: or 'true'? */ ? ElemElementFlags.tertiary_base : 0) |
  106:         (prebases?.[i] == '1' /* TODO-LDML: or 'true'? */ ? ElemElementFlags.prebase : 0);
  107        array.push(elem);

common\web\types\src\kmx\kmx-plus.ts:
  278          const inner = items.map(i => i.value.value).join('|');
  279:         return `(?:${inner})`; // TODO-LDML: need to escape here
  280        }

common\web\types\src\kmx\string-list.ts:
  82    toStringArray(): string[] {
[...]

PR #9170, opened 2023-07-04
fix(developer): improve coverage, add todos 🙀

srl295:
- skip coverage on some actual internal errors
- add todos where coverage is needed

for: #8069

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc-ldml/src/compiler/compiler.ts
+++ b/developer/src/kmc-ldml/src/compiler/compiler.ts
@@ -67,19 +67,23 @@ export class LdmlKeyboardCompiler {
     const reader = new LDMLKeyboardXMLSourceFileReader(this.options.readerOptions, this.callbacks);
     const data = this.callbacks.loadFile(filename);
     if(!data) {
+      // TODO-LDML: coverage for this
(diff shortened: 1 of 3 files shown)


developer/src/kmc-ldml/src/compiler/compiler.ts
```diff
@@ -67,19 +67,23 @@ export class LdmlKeyboardCompiler {
     const reader = new LDMLKeyboardXMLSourceFileReader(this.options.readerOptions, this.callbacks);
     const data = this.callbacks.loadFile(filename);
     if(!data) {
+      // TODO-LDML: coverage for this
```
mcdurdin, line 70: Do we need coverage TODOs given c8 will have them in its report anyway?

mcdurdin (approved):

developer/src/kmc-ldml/src/compiler/compiler.ts
srl295: We don't but I wanted to distinguish them from false positives. I scanned the coverage report to find these.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.134-alpha](https://keyman.com/downloads/releases/alpha/17.0.134)

PR #9192, opened 2023-07-05
chore(linux): Allow to run tests without integration tests

ermshiperete:
This will allow to run the integration tests on TC separately with lower priority.

@keymanapp-test-bot skip


mcdurdin (approved):
This LGTM, and we should consider it alongside #8361.

keyman-server:

Issue #9120, opened 2023-06-27
feat(core): basic ldml reorder algo implementation 🙀 


srl295:
supports #7375

just add a basic algo for unit tests to make sure I can implement reorders.

srl295:
fairly close on this one

PR #9197, opened 2023-07-05
feat(core): kmxplus uset implementation in core 🙀

srl295:
- this is the uset section, not UnicodeSet from ICU

For: #9120

@keymanapp-test-bot skip


core/src/kmx/kmx_plus.cpp
```diff
@@ -953,6 +953,89 @@ COMP_KMXPLUS_LIST_Helper::getIndex(KMX_DWORD i) const {
...
+  rawdata += sizeof(COMP_KMXPLUS_USET_USET) * uset->usetCount;
+  // entries
+  if (uset->rangeCount > 0) {
+    ranges = reinterpret_cast<const COMP_KMXPLUS_USET_RANGE *>(rawdata);
+  } else {
+    ranges = nullptr;
+  }
+
+  // Now, validate offsets by walking
+  if (is_valid) {
```
mcdurdin, line 1003: This can never be false

core/src/kmx/kmx_plus.cpp
```diff
@@ -953,6 +953,89 @@ COMP_KMXPLUS_LIST_Helper::getIndex(KMX_DWORD i) const {
...
+    usets = nullptr;
+    // not invalid, just empty.
+  }
+  rawdata += sizeof(COMP_KMXPLUS_USET_USET) * uset->usetCount;
+  // entries
+  if (uset->rangeCount > 0) {
+    ranges = reinterpret_cast<const COMP_KMXPLUS_USET_RANGE *>(rawdata);
+  } else {
+    ranges = nullptr;
+  }
```
mcdurdin, line 1000: Are we guaranteed to have checked validity before calling `setUset`? Otherwise, if these values `uset->usetCount` and `uset->rangeCount` are out of range, this could get access violations.

mcdurdin (changes requested):

core/src/kmx/kmx_plus.cpp
srl295: Guaranteed, `section_from_sect()` at 

https://github.com/keymanapp/keyman/pull/9197/files#diff-95985cb44c34779804524113d19f84c84c0df3809790667a95bd3592e3a9e7cbR1087 

checks validity or returns nullptr.  So `uset` would be null.  The key is that that function has the *original ending size* of the kmx+ region, which isn't propagated further.

Same for every section and helper.

At this point:

- the _Helper things should be children of a templatized base class. 
- `COMP_KMXPLUS_*::valid(size)` functions should have `size` parameter removed, because it's used there, the size is checked in the header against the end of the region, and it's not needed further.

core/src/kmx/kmx_plus.cpp
srl295: this is true.  I guess if there were ranges-but-no-usets it could be invalid. 

I kept this in so that it's parallel code structure to the other helper functions.  it's a dead test though ,so i can remove.

srl295 (commented):
thx. revamping a bit

srl295:
PTAL.

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.136-alpha](https://keyman.com/downloads/releases/alpha/17.0.136)

PR #9258, opened 2023-07-12
fix(core): don't use double newlines in debuglog

srl295:
- syslog() will automatically add newlines (per mac and linux manpages)
- there's no reason to add NL *AND* add std::endl
- this fix will cause syslog to have no newline going in, and print only one instead of two newlines on other platforms/environments

Just an annoyance now that I'm using DebugLog a lot.

@keymanapp-test-bot skip

### DIFF
--- a/core/src/debuglog.cpp
+++ b/core/src/debuglog.cpp
@@ -358,7 +358,7 @@ int DebugLog_1(const char *file, int line, const char *function, const char *fmt
           "%ld" TAB   //"TickCount" TAB
           "%s:%d" TAB //"SourceFile" TAB
           "%s" TAB    //"Function"
-          "%s" NL,    //"Message"
+          "%s",    //"Message"
 
           GetTickCount(), //"TickCount" TAB
           file, line,     //"SourceFile" TAB


mcdurdin:
Have you tested on Windows, where this originated? I believe on Windows you'll see the NL is needed in windbg.

srl295:
> Have you tested on Windows, where this originated? I believe on Windows you'll see the NL is needed in windbg.

I'll try that. 

The windows branch of the if has a std::endl but maybe we can add a specific NL if needed instead of double spacing everywhere else

srl295:
> Have you tested on Windows, where this originated? I believe on Windows you'll see the NL is needed in windbg.

what's the best way to test this?  just from the command line test?

mcdurdin:
> what's the best way to test this? just from the command line test?

It's a little complicated, needs a reg setting or two added first:

```
HKCU\Software\Keyman\Keyman Engine
"debug": REG_DWORD = 1
"debug to console": REG_DWORD = 1
```

Then, load testhost.exe into windbg, and run; you should see messages emitted to the debug console.

The appropriate solution would probably be to append `\n` in the `OutputDebugString()` call in `Keyman_WriteDebugEvent` in windows/src/engine/keyman32/DebugEventTrace.cpp. But we should test first to be sure.

And to make matters worse, some debuggers treat each call to `OutputDebugString()` as a log entry, others as just plain text emitted to the debug console, so YMMV anyway.

srl295:
couldn't get the build to work.  I think i might just test OutputDebugString() to see what its behavior is.

mcdurdin:
> couldn't get the build to work. I think i might just test OutputDebugString() to see what its behavior is.

@rc-swag are you able to help on this?

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.193-alpha](https://keyman.com/downloads/releases/alpha/17.0.193)

PR #9273, opened 2023-07-17
feat(developer): support file lists in kmc

### LINKED ISSUE #9162: feat(developer): allow kmc to take a file list parameter (`@file`) 🗜
As kmc works very well with building multiple projects, give it the possibility of a file list parameter, so we can avoid xargs and the like in the keyboard repo build.

mcdurdin:
Fixes #9162.

Adds support for `@files.txt` parameters in kmc. Files listed in this file, separated by line breaks, will be added to the input file list, relative to the path of files.txt.

The file can include comments in lines starting with '#' and ignores whitespace before and after each filename.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/src/commands/build.ts
+++ b/developer/src/kmc/src/commands/build.ts
@@ -7,6 +7,7 @@ import { NodeCompilerCallbacks } from '../util/NodeCompilerCallbacks.js';
 import { InfrastructureMessages } from '../messages/messages.js';
 import { CompilerFileCallbacks, CompilerOptions, KeymanFileTypes } from '@keymanapp/common-types';
 import { BaseOptions } from '../util/baseOptions.js';
+import { expandFileLists } from '../util/fileLists.js';
 
 
 function commandOptionsToCompilerOptions(options: any): CompilerOptions {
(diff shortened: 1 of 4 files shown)


developer/src/kmc/src/commands/build.ts
```diff
@@ -29,7 +30,19 @@ function commandOptionsToCompilerOptions(options: any): CompilerOptions {
...
-    .description('Build a source file into a final file')
+    .description(`Compile one or more source files or projects.`)
+    .addHelpText('after', `
+Supported file types:
+  * folder: Keyman project in folder
+  * .kpj: Keyman project
+  * .kmn: Keyman keyboard
+  * .xml: LDML keyboard
+  * .model.ts: Keyman lexical model
+  * .kps: Keyman keyboard package
```
darcywong00, line 41: Is this also Keyman lexical model package?

mcdurdin: yes

mcdurdin, line 41:
```suggestion
  * .kps: Keyman keyboard or model package
```

jahorton (approved):

PR #9276, opened 2023-07-17
fix(developer): warning suppression should be per file in kmc 🗜

### LINKED ISSUE #9274: bug(developer): warning suppression should only be for current file in kmc 🗜
The message `info KM0500D: More than 100 warnings or errors received; suppressing further messages.` should only suppress messages for the current file. Right now it suppresses any further messages for the session.

Arising from #9242 (which was fixing #9147)

mcdurdin:
Fixes #9274.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #9309, opened 2023-07-19
chore(linux): Fix running tests with Wayland

ermshiperete:
Because we passed in a wrong parameter (`--wayland`) but checked for just `wayland` we never run the tests with Wayland. This change rectifies this.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/tests/meson.build
+++ b/linux/ibus-keyman/tests/meson.build
@@ -57,22 +57,16 @@ teardown_tests = find_program('teardown-tests.sh', dirs: [meson.current_source_d
 run_test = find_program('run-single-test.sh', dirs: [meson.current_source_dir() / 'scripts'])
 find_tests = find_program('find-tests.sh', dirs: [meson.current_source_dir() / 'scripts'])
 
+# Mutter 40.x added the --headless option wich we need in order to be able to run the Wayland tests
+mutter = find_program('mutter', required: false, version: '>=40')
+can_build_wayland = mutter.found()
+
 test(
   'setup-x11',
   setup_tests,
   args: ['--x11', env_file, pid_file],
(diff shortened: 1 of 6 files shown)


linux/ibus-keyman/tests/scripts/teardown-tests.sh
```diff
@@ -3,4 +3,10 @@ set -eu
 
 . "$(dirname "$0")/test-helper.inc.sh"
 
+if [ -v KEYMAN_PKG_BUILD ]; then
+  # Skip setup during package builds - can't run headless and we won't
+  # run the other tests anyway
+  exit 0
+fi
```
darcywong00, line 10: Since this block is also in setup-tests.sh, would it be worth adding this to test-helper.inc.sh?

linux/ibus-keyman/tests/meson.build
```diff
@@ -67,16 +71,6 @@ test(
...
-  args: ['--wayland', env_file, pid_file],
-  env: test_env,
-  priority: -20,
-  is_parallel: false,
-  protocol: 'exitcode'
-)
-
 test(
   'teardown-x11',
   teardown_tests,
```
darcywong00, line 76: Is teardown_tests supposed to occur after all the tests are run? (e.g. towards the end of the file)

linux/ibus-keyman/tests/scripts/teardown-tests.sh
ermshiperete: done

linux/ibus-keyman/tests/meson.build
ermshiperete: The order is determined by the priority.

PR #9310, opened 2023-07-19
fix(linux): Fix logging

ermshiperete:
We can't initialize Sentry during module initialization because our `initialize_sentry` method logs some status info. Setting the logging level later on is then a no-op, breaking verbose and very-verbose logging.

This change fixes this by putting Sentry initialization into a separate method that gets called after we set the log level.

Also refactor some common code into methods that appear in all our executables.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/__init__.py
+++ b/linux/keyman-config/keyman_config/__init__.py
@@ -1,7 +1,7 @@
 import gettext
+import logging
 
 from keyman_config.sentry_handling import SentryErrorHandling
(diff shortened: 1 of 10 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.145-alpha](https://keyman.com/downloads/releases/alpha/17.0.145)

PR #9313, opened 2023-07-20
refactor(linux): Use better way to get username

ermshiperete:
`getpass.getuser()` tries several different ways to get the username and is the recommended way of getting the username.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.144-alpha](https://keyman.com/downloads/releases/alpha/17.0.144)

PR #9328, opened 2023-07-24
docs(core): Document how to build Core on Linux

### LINKED ISSUE #8913: docs(core): Document how to build Core on Linux

ermshiperete:
Closes #8913.

@keymanapp-test-bot skip

### DIFF
--- a/docs/build/linux-ubuntu.md
+++ b/docs/build/linux-ubuntu.md
@@ -80,12 +80,10 @@ git clone https://github.com/emscripten-core/emsdk.git
 cd emsdk
 ./emsdk install latest
 ./emsdk activate latest
-source ./emsdk_env.sh
-export EMSDK_NODE=/usr/bin/node
+export EMSCRIPTEN_BASE=$(pwd)/upstream/emscripten
 ```
 
-*emscripten* comes with an older version of node, so it's important to set
-the `EMSDK_NODE` environment variable to the node version we need.
+**NOTE:** Don't put EMSDK on the path, i.e. don't source `emsdk_env.sh`.
 
 ### Building Keyman Core
 
@@ -114,7 +112,7 @@ mkdir -p build/linux
 docker run -it --rm -v $(pwd)/..:/home/build/build \
   -v $(pwd)/build/linux:/home/build/build/core/build \
   keymanapp/keyman-linux-builder:latest \
-  bash -c 'core/build.sh --debug'
+  wrapper core/build.sh --debug
 ```
 
 - linux
@@ -124,7 +122,7 @@ docker run -it --rm -v $(pwd)/..:/home/build/build \
 cd $(git rev-parse --show-toplevel)
 docker run -it --rm -v $(pwd):/home/build/build \
   keymanapp/keyman-linux-builder:latest \
-  bash -c 'DESTDIR=/home/build linux/build.sh --debug build install'
+  wrapper 'DESTDIR=/home/build linux/build.sh --debug build install'
 ```
 
 ## Keyman for Android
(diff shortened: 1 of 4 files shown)


docs/build/linux-ubuntu.md
```diff
@@ -53,21 +54,42 @@ sudo mk-build-deps --install linux/debian/control
...
+Most dependencies are already installed if you followed the instructions under
+[Prerequisites](#prerequisites). You'll still have to install `emscripten`:
 
-Building:
+```shell
+git clone https://github.com/emscripten-core/emsdk.git
+cd emsdk
+./emsdk install latest
+./emsdk activate latest
+source ./emsdk_env.sh
```
mcdurdin, line 83: We don't want to put EMSDK onto the path, because it pollutes with its own version of node and python and other tools -- instead we want to set the EMSCRIPTEN_BASE environment variable to the root of EMSDK

docs/build/linux-ubuntu.md
```diff
@@ -53,21 +54,42 @@ sudo mk-build-deps --install linux/debian/control
...
+export EMSDK_NODE=/usr/bin/node
+```
+
+*emscripten* comes with an older version of node, so it's important to set
+the `EMSDK_NODE` environment variable to the node version we need.
```
mcdurdin, line 88: This is not necessary if emscripten is not on the path

docs/build/linux-ubuntu.md
```diff
@@ -86,50 +108,55 @@ Once the image is built, it may be used to build parts of Keyman.
 
 ```shell
```
mcdurdin, line 109: I wonder if these should be targets in /linux/build.sh and /core/build.sh?

linux/Dockerfile
```diff
@@ -21,13 +22,28 @@ ENV DEBCONF_NOWARNINGS yes
...
 RUN apt-get -q -y install nodejs
+
+# Install emscripten
+RUN cd /usr/share && \
+  git clone https://github.com/emscripten-core/emsdk.git && \
+  cd emsdk && \
+  ./emsdk install latest && \
+  ./emsdk activate latest && \
+  echo 'source "/usr/share/emsdk/emsdk_env.sh"' >> $HOME/.bashrc && \
+  echo "export EMSDK_NODE=/usr/bin/node" >> $HOME/.bashrc && \
```
mcdurdin, line 45: Again, we specifically don't want emscripten on the path -- use EMSCRIPTEN_BASE to specify the path to emcc etc

mcdurdin (changes requested):
I'd like to make sure we don't end up with emscripten on the path by default :grin:

docs/build/linux-ubuntu.md
```diff
@@ -53,21 +54,40 @@ sudo mk-build-deps --install linux/debian/control
...
-Building:
+```shell
+git clone https://github.com/emscripten-core/emsdk.git
+cd emsdk
+./emsdk install latest
+./emsdk activate latest
+export EMSCRIPTEN_BASE=$(pwd)/upstream/emscripten
+```
+
+**NOTE:** Don't put EMSDK on the path, i.e. don't source `emsdk_env.sh`.
```
mcdurdin, line 86: :+1:

mcdurdin (approved):
LGTM!

keyman-server:

PR #9341, opened 2023-07-25
chore(linux): Update supported Ubuntu versions

ermshiperete:
- remove Ubuntu 22.10 Kinetic which is no longer supported
- add future Ubuntu 23.10 Mantic

@keymanapp-test-bot skip


srl295 (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.148-alpha](https://keyman.com/downloads/releases/alpha/17.0.148)

Issue #7375, opened 2022-09-28
feat(core): support ldml transforms 🙀


srl295:
- #5015 
- #7042 

This ticket is for 

Depends on:
- kmx spec updates for UnicodeSet contents
- compiler updates for transforms

-----
### TODO area
- [x] Markers #9119 
- [x] variables in lists
- [x] variables in keys
- [x] map variables

srl295:
@mcdurdin I think this is ready to close, with the final wrapup in #9121 (for those edge cases) and https://github.com/keymanapp/keyman/issues/9707 (tertiary reorder) and of course normalization

mcdurdin:
Note: we don't normally close issues until the corresponding PRs are merged.

srl295:
@mcdurdin Was trying to segment them. But I think I could consider #9688 as closing it.

PR #9352, opened 2023-07-26
chore(core): update documentation in transform logic and processor 🙀

srl295:
- drop some unneeded variables also, should be no semantic change

For: #7375

@keymanapp-test-bot skip


core/src/ldml/ldml_processor.cpp
```diff
@@ -258,16 +259,16 @@ ldml_processor::process_event(
         // Look up the key
         const std::u16string str = keys.lookup(vk, modifier_state);
         if (str.empty()) {
-          // not found
+          // not found, so pass the keystroke on to the
```
mcdurdin, line 262:
```suggestion
          // not found, so pass the keystroke on to the Engine
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -258,16 +259,16 @@ ldml_processor::process_event(
...
           break; // ----- commit and exit
         }
         // found the correct string - push it into the context and actions
         const std::u32string str32 = kmx::u16string_to_u32string(str);
-        for(size_t i=0; i<str32.length(); i++) {
-          state->context().push_character(str32[i]);
-          state->actions().push_character(str32[i]);
+        for (const auto &ch : str32) {
+          state->context().push_character(ch);
+          state->actions().push_character(ch);
```
mcdurdin, line 271: The potential for desync between actions and context is a real design problem ... not to fix here but one day we should discuss

mcdurdin (approved):
LGTM

core/src/ldml/ldml_processor.cpp
srl295: I'll need to learn how that works.. but sgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.149-alpha](https://keyman.com/downloads/releases/alpha/17.0.149)

PR #9340, opened 2023-07-25
chore(linux): Remove obsolete dist from uploading to launchpad

ermshiperete:
Ubuntu 22.10 Kinetic is no longer supported, so this change omits it from uploading to Launchpad.

@keymanapp-test-bot skip

PR #9354, opened 2023-07-27
chore(linux): Stop building Kinetic on Jenkins 🏠

ermshiperete:
In #9340 I overlooked one place where we mention Kinetic...

@keymanapp-test-bot skip

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -5,7 +5,7 @@
 @Library('lsdev-pipeline-library') _
 
 keymanPackaging {
-  distributionsToPackage = 'bionic focal jammy kinetic'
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
LGTM

Is there a way we can consolidate this? I'm also surprised we still list bionic and focal here?

ermshiperete:
> I'm also surprised we still list bionic and focal here?

Focal is still officially supported, and Bionic is still in extended-LTS. Therefore it's still possible to build for Bionic (in contrast to Kinetic). Since we said we drop Bionic with Keyman 17 I left it in on the stable-16.0 branch.

keyman-server:

Issue #9119, opened 2023-06-27
feat(core): ldml marker implementation 🙀


srl295:
depends on #9118 

spec at https://github.com/keymanapp/keyman/blob/master/core/src/ldml/C9134_ldml_markers.md

-----
### TODO area
- [x] markers in lists, such as:
- [x] markers in keys.longPress
- [x] markers in keys.multiTap
- [x] markers in flicks.to

PR #9365, opened 2023-07-29
feat(common): marker processing🙀

srl295:
- emit the sentinel value, with range checking

#9119

@keymanapp-test-bot skip


srl295:
the pt abnt2 keyboard is temporarily disabled.. it needs its transforms fleshed out, creates a marker validation err

common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -40,6 +47,25 @@ export class MarkerParser {
...
 
+  /**
+   * Marker sentinel as a string - U+FFFF
+   */
+  public static readonly SENTINEL = String.fromCodePoint(constants.marker_sentinel);
+
+  /**
+   * Matches all markers.
+   */
+  public static readonly SENTINEL_ALL_MARKERS = this.SENTINEL + this.SENTINEL;
```
mcdurdin, line 58: This makes me a little squirmy. I'm probably worrying about nothing but it does mean `SENTINEL` has two meanings, whereas in kmx processor `SENTINEL` only has a single meaning.

FWIW, in kmx processor, `SENTINEL` is always followed by a `CODE` to determine the actual meaning of the sentinel, e.g. `SENTINEL CODE_DEADKEY`. I wonder if we should be following the same pattern here to allow for future expansion

common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -56,6 +82,35 @@ export class MarkerParser {
...
+  }
+
+  /** @returns all marker strings as sentinel values */
+  public static toSentinelString(s: string, markers?: OrderedStringList) : string {
+    if (!s) return s;
+    return s.replaceAll(this.REFERENCE, (sub, arg) => {
+      if (arg === MarkerParser.ANY_MARKER_ID) {
+        return MarkerParser.SENTINEL_ALL_MARKERS;
+      }
+      if (!markers) {
```
mcdurdin, line 101: Given this, why is `markers` an optional parameter?

mcdurdin (approved):
LGTM

common/web/types/src/ldml-keyboard/pattern-parser.ts
srl295: May be no markers in that file.

common/web/types/src/ldml-keyboard/pattern-parser.ts
srl295: @mcdurdin i actually couldn't find `U_SENTINEL` or `SENTINEL` in the keyman source…  I did see SENTINEL_CODE_MARKER which is FDD0 but it didn't seem to go with any other sentinel value. 

I did delete SENTINEL_ALL_MARKERS in a future PR

mcdurdin: Yep, we discussed this at standup but for reference, `UC_SENTINEL CODE_DEADKEY <marker>` is the sequence used by kmx.

keyman-server:

PR #9368, opened 2023-07-31
chore(linux): Don't fail on parallel builds

ermshiperete:
If another package build runs on the same build agent at the same time and installs dependencies at the same time, previously the build failed because apt/dpkg was already running. This change checks for the existence of the apt/dpkg lock file and waits until the other process is finished before starting the installation of dependencies.

@keymanapp-test-bot skip


mcdurdin:
Oh Jenkins...

mcdurdin (approved):

keyman-server:

PR #9374, opened 2023-07-31
chore(developer): update markers per review 🙀

srl295:
- remove SENTINEL_ALL_MARKERS

For: #9119

from comment in #9365

@keymanapp-test-bot skip


srl295:
took into draft to update per standup discussion, will be `UC_SENTINEL CODE_DEADKEY <#>`  and also use `0xFFFE` for 'any'

srl295:
from Mon's discussion:

- whither UC_SENTINEL, CODE_xxxx, params 
- https://github.com/keymanapp/keyman/blob/1810434b3454b9fd520d22623f8d0e6f54ca32b9/common/include/kmx_file.h#L198 
- Look at incxstr, decxstr, xstrlen etc
https://github.com/keymanapp/keyman/blob/52869e0d15d10c00ea08911b7838c4c373c6919b/common/web/types/src/kmx/kmx.ts#L262
- Use `CODE_DEADKEY` for markers

So we will end up with:

- `UC_SENTINEL CODE_DEADKEY MARKER_ANY=0xFFFE`   

(not FFFF for any)

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -615,16 +615,18 @@ class Constants {
   }
 
   // ---- marker stuff ----
-  /** sentinel value indicating a marker follows */
-  readonly marker_sentinel = 0xFFFF;
+  /** == UC_SENTINEL, first item of a  */
+  readonly marker_sentinel     = 0xFFFF;
+  /** == CODE_DEADKEY */
+  readonly marker_code_deadkey = 0x0008;
```
mcdurdin, line 621:
```suggestion
  readonly marker_code = 0x0008;
```

Or perhaps code_marker ... up for discussion!

common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -51,11 +51,10 @@ export class MarkerParser {
    * Marker sentinel as a string - U+FFFF
    */
   public static readonly SENTINEL = String.fromCodePoint(constants.marker_sentinel);
-
   /**
-   * Matches all markers.
+   * Marker code as a string - U+0008
    */
-  public static readonly SENTINEL_ALL_MARKERS = this.SENTINEL + this.SENTINEL;
+  public static readonly CODE_DEADKEY = String.fromCodePoint(constants.marker_code_deadkey);
```
mcdurdin, line 57:
```suggestion
  public static readonly MARKER_CODE = String.fromCodePoint(constants.marker_code);
```

I reckon let's avoid polluting the ldml keyboard code with 'deadkey' refs.

common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -88,15 +87,15 @@ export class MarkerParser {
     if (n < MarkerParser.MIN_MARKER_INDEX || n > MarkerParser.ANY_MARKER_INDEX) {
       throw RangeError(`Internal Error: marker index out of range ${n}`);
     }
-    return this.SENTINEL + String.fromCharCode(n);
+    return this.SENTINEL + this.CODE_DEADKEY + String.fromCharCode(n);
```
mcdurdin, line 90:
```suggestion
    return this.SENTINEL + this.MARKER_CODE + String.fromCharCode(n);
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -615,16 +615,18 @@ class Constants {
   }
 
   // ---- marker stuff ----
-  /** sentinel value indicating a marker follows */
-  readonly marker_sentinel = 0xFFFF;
+  /** == UC_SENTINEL, first item of a  */
+  readonly marker_sentinel     = 0xFFFF;
+  /** == CODE_DEADKEY */
```
mcdurdin, line 620:
```suggestion
  /** == kmx_file.CODE_DEADKEY */
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -615,16 +615,18 @@ class Constants {
   }
 
   // ---- marker stuff ----
-  /** sentinel value indicating a marker follows */
-  readonly marker_sentinel = 0xFFFF;
+  /** == UC_SENTINEL, first item of a  */
```
mcdurdin, line 618:
```suggestion
  /** == kmx_file.UC_SENTINEL, always followed by `marker_code`, marker index 0x0001-0xfffe */
```

core/src/kmx/kmx_plus.h
```diff
@@ -310,6 +310,10 @@ static_assert(sizeof(struct COMP_KMXPLUS_TRAN_GROUP) == LDML_LENGTH_TRAN_GROUP,
 static_assert(sizeof(struct COMP_KMXPLUS_TRAN_TRANSFORM) == LDML_LENGTH_TRAN_TRANSFORM, "mismatched size of tran transform");
 static_assert(sizeof(struct COMP_KMXPLUS_TRAN_REORDER) == LDML_LENGTH_TRAN_REORDER, "mismatched size of tran reorder");
 
+// assert some parallel constants
+static_assert(LDML_MARKER_SENTINEL == UC_SENTINEL, "mismatch: LDML_MARKER_SENTINEL");
+static_assert(LDML_MARKER_CODE_DEADKEY == CODE_DEADKEY, "mismatch: LDML_MARKER_CODE_DEADKEY");
```
mcdurdin, line 315:
```suggestion
static_assert(LDML_MARKER_CODE == CODE_DEADKEY, "mismatch: LDML_MARKER_CODE");
```

core/include/ldml/keyboardprocessor_ldml.ts
```diff
@@ -615,16 +615,18 @@ class Constants {
...
+  readonly marker_sentinel     = 0xFFFF;
```
mcdurdin, line 619: should this be prefixed with `marker_` given it's a generic prefix which is currently only used with markers but may be used with other things in the future?

mcdurdin (approved):
Conceptually all lgtm but I think we should discuss names before merge. I've made some suggestions which I am not 100% sure are right, but will need references to those suggestions updated too.

core/include/ldml/keyboardprocessor_ldml.ts
srl295: maybe just uc_sentinel?

common/web/types/src/ldml-keyboard/pattern-parser.ts
srl295: Hm… though, then it matches the rest of the codebase.  

But, I suppose a (1) jsdoc comment and (2) unit tests asserting that it's the same, could be enough documentation.

core/include/ldml/keyboardprocessor_ldml.ts
rc-swag: for future extension I guess code_marker, could then match code_beep, code_return  etc

srl295:
@rc-swag @mcdurdin a thought here…

what if I update kmx_file.h etc:

```c
#define CODE_DEADKEY    0x08
#define CODE_MARKER     CODE_DEADKEY
```

… and then use CODE_MARKER as discussed. No 'DEADKEY' in any new source files.

srl295:
merged a basic.txt (binary format) fix from the next PR, should pass tests. So we can discuss naming and comments :)

core/include/ldml/keyboardprocessor_ldml.ts
mcdurdin: > maybe just uc_sentinel?

yep sounds good, consistent with existing codebase, and not confusing in this context.

mcdurdin:
> … and then use CODE_MARKER as discussed. No 'DEADKEY' in any new source files.

I think it's helpful to have the definition in the new header file, and a reference to the old kmx_file.h there. So happy to have `CODE_MARKER` in kmx_file.h so long as the new file also has an identical definition. Also remembering that the KMXPlus component is independent of the rest of the KMX file and could in the future be delivered without the kmx wrapper.

common/web/types/src/ldml-keyboard/pattern-parser.ts
mcdurdin: Core already uses the term marker synonymously with deadkey. So ldml core can and should use marker. Let's keep the deadkey term internal to the kmx core.

srl295: Fine with me. My only question is whether kmx_file.h should have an alias as well.

mcdurdin: I don't think it needs to at this point. kmx_file.h is used only by the KMX processor which only believes in deadkeys, and it's quite old so we don't want it getting confused ;-)

rc-swag:
> > … and then use CODE_MARKER as discussed. No 'DEADKEY' in any new source files.
> 
> I think it's helpful to have the definition in the new header file, and a reference to the old kmx_file.h there. So happy to have `CODE_MARKER` in kmx_file.h so long as the new file also has an identical definition. 

Yeah sounds good. Also would not want the  kmx_processor or platform code to start using CODE_MARKER mixed with CODE_DEADKEY.  Unless of course a find and replace refactor. The clean swap in terminology at the boundary is ok at them moment.

srl295:
PTAL…

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.155-alpha](https://keyman.com/downloads/releases/alpha/17.0.155)

Issue #7068, opened 2022-08-12
feat(windows): mcompile - use static tables for Windows keyboard data


mcdurdin:
This is a proposal to decouple the Windows system keyboard data from the kbdxx.dll file so that we can use mcompile on other platforms as well. The idea would be to have a program to query the set of Windows kbdxx DLLs that we are interested in, and write each Windows keyboard's data to a datafile. Then mcompile would load this datafile rather than the Windows DLL when running the mcompile process.

Sample call currently:

```
mcompile -d c:\projects\keyman\keyboards\release\sil\sil_ipa\build\sil_ipa.kmx kbdgr.dll 0407 c:\projects\keyman\keyboards\release\sil\sil_ipa\build\sil_ipa_german.kmx
```

mcompile should add two new command line modes:

First, exporting keyboard dll data to file:
```
mcompile -e kbdgr.dll 0407 kbdgr.txt
```

Second, using keyboard dll data from file:
```
mcompile -d c:\projects\keyman\keyboards\release\sil\sil_ipa\build\sil_ipa.kmx kbdgr.txt c:\projects\keyman\keyboards\release\sil\sil_ipa\build\sil_ipa_german.kmx
```

The second part of this project would be to remove Windows-isms from mcompile so that we can build it on any desktop platform.

PR #8558, opened 2023-03-31
Feature(windows): mcompile - use static tables for Windows keyboard data 🐘 

SabineSIL:
fixes #7068

This is a proposal to decouple the Windows system keyboard data from the kbdxx.dll file so that we can use mcompile on other platforms as well. The idea would be to have a program to query the set of Windows kbdxx DLLs that we are interested in, and write each Windows keyboard's data to a datafile. Then mcompile would load this datafile rather than the Windows DLL when running the mcompile process.

@keymanapp-test-bot skip

Issue #3345, opened 2020-07-13
feat(linux): Mnemonic keyboard support


ermshiperete:
Currently Linux doesn't properly support mnemonic keyboards - a US-English keyboard base layout is always assumed.
see #8972, #8973, #8974, #8975, #8977, #8978, #8979

mcdurdin:
The way I'd like to solve this is to use the mnemonic layout recompiler (mcompile) to build positional versions of a mnemonic keyboard. mcompile takes a mnemonic layout and combines it with a Windows base layout to produce a positional layout that should work unmodified on other platforms. We then ensure that the other platforms always treat the base layout as US (with a couple of minor tweaks around modifiers and additional keys not on a US keyboard) so that positional layouts always work consistently.

Note that the Windows core itself is positional now, because mcompile does the hard work.

Right now, mcompile has a dependency on Windows for parsing the base layout, but I think we can build a table of base layouts into mcompile, and once we have that then mcompile itself should be very cross-platform.

mcdurdin:
The same approach applies for mac (won't open an issue on this just now)

ermshiperete:
So `mcompile` gets executed when the keyboard gets installed? And as base layout the default (or first) keyboard/language is used. What happens when the user sets a different default?

mcdurdin:
On Windows, yes, `mcompile` runs when the keyboard is installed. The base layout by default is the first Latin-script Windows keyboard on the system (the user's default keyboard is first keyboard). The user can change the base layout in Keyman Configuration, which then triggers a re-run of mcompile. Keyman on Windows stores each mrecompiled keyboard along with the original in the keyboard folder:

![image](https://user-images.githubusercontent.com/4498365/87478216-babdf400-c66c-11ea-9f42-2fd984e5617a.png)

Note that the `-d` alternative treats ['deadkeys as plain keys'](https://blog.keyman.com/2015/02/an-improved-experience-with-deadkeys-on-european-hardware-keyboards/). Keyman uses this keyboard internally if that option is selected in Configuration.

ermshiperete:
Hmm, implementing an option to select the base keyboard in km-config might be a quicker way to support mnemonic keyboards on Linux.

mcdurdin:
> Hmm, implementing an option to select the base keyboard in km-config might be a quicker way to support mnemonic keyboards on Linux.

[...]

PR #9384, opened 2023-08-01
feat(linux): mnemonic layout support 🐘

SabineSIL:
fixes https://github.com/keymanapp/keyman/issues/7068
see #3345

This is a proposal to rewrite mcompile for Linux. For this we need to query the base keyboard data from the Linux platform, then rewriting the keyboard .kmx using the same approach as is done in mcompile for Windows, but working from the data from the x11 keyboard on Linux.

Ideally, we'd rewrite mcompile to be cross-platform (Windows, Linux, macOS), so that the keyboard interrogation would be separated from the .kmx rewriting, at least to some degree. Nevertheless it would probably be easiest to start from a standalone implementation.

connected to https://github.com/keymanapp/keyman/pull/8558 mcompile - use static tables for windows keyboard data
@keymanapp-test-bot skip


SabineSIL:
Here is the first draft of mcompile for Linux. Before reviewing make sure to read this [document ](https://docs.google.com/document/d/10eNrApYMC6sOj353OmjYcDqmtsI2zD2A9mxui-Y-1D0/edit)  in order to 

- not accidentally review unfinished (deadkey related) code

- learn about differences in files obtained by mcompile-windows and mcompile-linux

ermshiperete (commented):
I started to review these changes, but I didn't get very far today. Here's my first batch of comments.

Also, some of the code is fairly hard to read because the indentation is off. I think it would be good if you'd reformat the files. This can be done e.g. in vscode if the C/C++ Extension Pack is installed: Open a file, then right click in the document and select "Format Document".

linux/mcompile/keymap/deadkey.cpp
```diff
@@ -0,0 +1,468 @@
...
+		do {
+			if( dk == dkVec[i]->KMX_GetDeadCharacter())
+				return true;
+			i++;
+		} while (i < (int) dkVec.size());
+	}
+	return false;
+}
+
+bool query_dk_combinations_for_specific_dk(v_dw_2D * p_dk_ComposeTable, v_dw_2D  &dk_SingleTable, KMX_DWORD dk) {
```
ermshiperete, line 57: Why is `p_dk_ComposeTable` a pointer instead of a reference similar to `dk_SingleTable`? That would simplify things in this method.

linux/mcompile/keymap/deadkey.cpp
```diff
@@ -0,0 +1,468 @@
...
+		}
+	}
+
+	if( dk_SingleTable.size()>0)
+		return true;
+	else
+		return false;
+}
+
+KMX_DWORD KMX_changeKeynameToCapital(KMX_DWORD KVal, KMX_DWORD &shift, GdkKeymap* keymap) {
```
ermshiperete, line 76: The method name with the `KMX_` prefix doesn't fit the other method names in this file. Maybe rename to `change_keyname_to_capital`? And parameter and variable names should start with lowercase letter, so `kVal`, `keyval` and `cap` (or probably better `capitalKeyval`).

linux/mcompile/keymap/deadkey.cpp
```diff
@@ -0,0 +1,468 @@
...
+void create_DKTable(v_dw_2D & dk_ComposeTable) {
+  //create a 2D-Vector which contains data for ALL existing deadkey combinations on a Linux Keyboard:
+  //dk_ComposeTable[i][0] : First    (e.g. dead_circumflex)
+  //dk_ComposeTable[i][1] : Second   (e.g. a)
+  //dk_ComposeTable[i][3] : Unicode-Value   (e.g. 0x00E2)
+
+  //values taken from: https://help.ubuntu.com/community/GtkDeadKeyTable#Accents
+	// _S2 Do we want to use GTK instead of this function?
+
+  v_dw_1D line;
```
ermshiperete, line 103: Why do you use a 1D vector here instead of creating a new struct, something like:

```
typedef struct {
  KMX_DWORD diacritic;
  KMX_DWORD base_char;
  KMX_DWORD unicode_value;
} DeadKeyMapping;
```

(There might be a better name for the struct...)

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+//#define _DEBUG_FOPEN  1
+
+#ifdef __EMSCRIPTEN__
+
+const std::string get_wasm_file_path(const std::string& filename) {
+  // Verify that we are passing a fully-qualified path
+  // TODO: we need to support relative paths based on CWD
+#if _DEBUG_FOPEN
+  std::cout << "get_wasm_file_path ENTER (" << filename << ")" << std::endl;
```
ermshiperete, line 26: You could use the `__FUNCTION__` macro instead of hard coding the method name. That way it'll output correctly even if the method gets renamed:


```suggestion
  std::cout << __FUNCTION__ << " ENTER (" << filename << ")" << std::endl;
```

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+#endif
+
+  assert(
+    (filename.length() > 0 && filename.at(0) == '/') ||
+    (filename.length() > 1 && filename.at(1) == ':')
+  );
+
+#if _DEBUG_FOPEN
+  std::cout << "get_wasm_file_path assert passed " << std::endl;
+#endif
```
ermshiperete, line 36: This seems a bit unnecessary since the assert succeeded if we come here.

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    //console.log('EM_ASM path = '+path);
+    let root = '/nodefs-mount';
+    if(!FS.analyzePath(root).exists) {
+      //console.log('EM_ASM mkdir '+root);
+      FS.mkdir(root);
+      if(!isWin32Path) {
+        //console.log('EM_ASM mount '+root);
+        FS.mount(NODEFS, {root : '/'}, root);
+      }
```
ermshiperete, line 51: I think it would be better to do this in the `else` branch of `if(isWin32Path)` below since both are dealing with mounting.

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+    const isWin32Path = path.match(/^[a-zA-Z]:/);
+
+    //console.log('EM_ASM path = '+path);
+    let root = '/nodefs-mount';
+    if(!FS.analyzePath(root).exists) {
+      //console.log('EM_ASM mkdir '+root);
+      FS.mkdir(root);
+      if(!isWin32Path) {
+        //console.log('EM_ASM mount '+root);
+        FS.mount(NODEFS, {root : '/'}, root);
```
ermshiperete, line 50: In comparing this with the `if(isWin32Path)` block below, I'm wondering: don't we have to pass `path` to `FS.mount` somehow?

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+    if(isWin32Path) {
+      // Win32 path, one mount per drive
+      root += "/" + path.charAt(0);
+      if(!FS.analyzePath(root).exists) {
+        //console.log('EM_ASM mkdir '+root);
+        FS.mkdir(root);
+        //console.log('EM_ASM mount '+root);
+        FS.mount(NODEFS, {root : path.substr(0,3) }, root);
+      }
+    }
```
ermshiperete, line 63:
```suggestion
    if(isWin32Path) {
      // Win32 path, one mount per drive
      root += "/" + path.charAt(0);
      if(!FS.analyzePath(root).exists) {
        //console.log('EM_ASM mkdir '+root);
        FS.mkdir(root);
        //console.log('EM_ASM mount '+root);
        FS.mount(NODEFS, {root : path.substr(0,3) }, root);
      }
    } else {
		//console.log('EM_ASM mount '+root);
        FS.mount(NODEFS, {root : '/'}, root);		
	}
```

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+#ifdef _MSC_VER
+  std::string cpath = Filename; //, cmode = mode;
+  std::replace(cpath.begin(), cpath.end(), '/', '\\');
+  return fopen(cpath.c_str(), (const KMX_CHAR*) mode);
+#else
+  return fopen_wrapper(Filename, mode);
+  std::string cpath, cmode;
+  cpath = (const KMX_CHAR*) Filename;
+  cmode = (const KMX_CHAR*) mode;
+  return fopen_wrapper(cpath.c_str(), cmode.c_str());
```
ermshiperete, line 116: This seems to be unreachable code since we return in the line above.

linux/mcompile/keymap/filesystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+#ifdef _MSC_VER
+#include <io.h>
+#else
+#include <unistd.h>
+#endif
+
+//#define _DEBUG_FOPEN  1
+
+#ifdef __EMSCRIPTEN__
```
mcdurdin, line 20: We don't need any emscripten support for mcompile, because it will only run native -- so we can delete all emscriptn/wasm-related code blocks.

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
-  extract_difference(All_Vector);
-  //test_in_out(All_Vector);
+std::u16string CodePointToU16String(unsigned int codepoint) {
+  std::u16string str;
 
-  //print_simple_map_US(All_Vector,1);    // 1 = non-shift
-  //print_simple_map_Other(All_Vector,1); // 1 = non-shift
-  test_specific_Characters(All_Vector);
-  gdk_display_close(display);
+  if constexpr (sizeof(wchar_t) > 2) {
```
ermshiperete, line 787: I don't understand why we check the size of `wchar_t` here. Besides, it is wrong: if it is 32 bit (`sizeof(wchar_t) > 2`), as it is on Linux, we can't cast to `char16_t` but have to check if we need a surrogate pair.

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
+      str = static_cast<char16_t>(codepoint);
+  }
+  else if (codepoint <= 0xFFFF) {
+      str = static_cast<char16_t>(codepoint);
+  }
+  else {
+      codepoint -= 0x10000;
```
ermshiperete, line 794: We might want to assert that `codepoint < 0x10FFFF` so that it catches our attention if we get an invalid codepoint. We won't get an invalid surrogate pair since we drop the illegal bits below in line 796, but I think it would still be good to get the developer's attention if this should happen. Also we might want to assert that this is a little endian system.


```suggestion
...
inline bool isLittleEndianSystem() {
  char16_t test = 0x0102;
  return (reinterpret_cast<char *>(&test))[0] == 0x02;
}
...
      assert(codepoint < 0x10FFFF);
      assert(isLittleEndianSystem());
      codepoint -= 0x10000;
```

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
+      str.resize(2);
+      str[0] = static_cast<wchar_t>(0xD800 + ((codepoint >> 10) & 0x3FF));
+      str[1] = static_cast<wchar_t>(0xDC00 + (codepoint & 0x3FF));
```
ermshiperete, line 797: `std::u16string` uses `char16_t` as character type, and I think the order is wrong for little endian, so this should be:

```suggestion
      str[0] = static_cast<char16_t>(0xDC00 + (codepoint & 0x3FF));
      str[1] = static_cast<char16_t>(0xD800 + ((codepoint >> 10) & 0x3FF));
```

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
+
+  if (in == 0 )
+    return u"\0";
+
+  std::string long_name((const char*) gdk_keyval_name (in));                      // e.g. "dead_circumflex" , "U+017F" , "t"
+
+  if ( long_name.substr (0,2) == "U+" )                                           // U+... Unicode value
+    return  CodePointToU16String(in-0x1000000);
+
+  if (in < (int) deadkey_min) {                                                   // no deadkey; no Unicode
```
ermshiperete, line 589: For performance you might want to do this check before you call `gdk_keyval_name`.

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
+KMX_DWORD KMX_get_KeyValUnderlying_From_KeyCodeUnderlying(GdkKeymap *keymap, guint keycode, int shift_state_pos) {
+  GdkKeymapKey *maps;
+  guint *keyvals;
+  gint count;
+  KMX_DWORD KVal;
+
+  if (!gdk_keymap_get_entries_for_keycode(keymap, keycode, &maps, &keyvals, &count))
+    return 0;
+
+  if (!(shift_state_pos <= count))
```
ermshiperete, line 698: Isn't this off by one, i.e. shouldn't it be `if (!(shift_state_pos < count))` ? And it's easier to read if you remove the negation:

```suggestion
  if (shift_state_pos >= count)
```

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
+    return 0;
+
+   if (!(keycode <= keycode_max))
```
ermshiperete, line 701:
```suggestion
   if (keycode > keycode_max)
```

linux/mcompile/keymap/keymap.cpp
```diff
@@ -109,355 +472,330 @@ void Split_US_To_3D_Vector(v_str_3D &all_US,v_str_1D completeList) {
...
+  PKMX_WCHAR dky=NULL;
+
+
+  if (!gdk_keymap_get_entries_for_keycode(keymap, KC_underlying, &maps, &keyvals, &count))
+    return 0;
+
+  if (!(map_VKShiftState_to_LinModifier(VKShiftState) <= count))
+  return 0;
+
+  if (!(KC_underlying <= keycode_max))
```
ermshiperete, line 726: See comments/suggestions in previous method

linux/mcompile/keymap/keymap.h
```diff
@@ -1,67 +1,532 @@
...
+typedef std::vector<std::vector<KMX_DWORD> > v_dw_2D;
+typedef std::vector<std::vector<std::vector<KMX_DWORD> > > v_dw_3D;
 
-// read configuration file, split and write to 3D-Vector (Data for US on [0][ ][ ]  )
-void write_US_ToVector(v_str_3D &vec, std::string language, const char *text);
+static KMX_DWORD returnIfCharInvalid = 0;
+static KMX_DWORD keycode_max = 94;
+static KMX_DWORD deadkey_min = 0xfe50;
+static KMX_DWORD deadkey_max = 0xfe93;
+//static KMX_DWORD deadkey_max = 0xfe52;  // _S2 TOP_6 TODO This has to go! my test: to only return 3 dk
```
ermshiperete, line 70:
```suggestion
```

linux/mcompile/keymap/keymap.h
```diff
@@ -1,67 +1,532 @@
...
+	0x24, // L"K_?B2",				// &HB2
+	0x22, // L"K_?B3",				// &HB3
+	0x6C, // L"K_?B4",				// &HB4
+	0x6D, // L"K_?B5",				// &HB5
+	0x6B, // L"K_?B6",				// &HB6
+	0x21, // L"K_?B7",				// &HB7
+	0x00, // L"K_?B8",				// &HB8
+	0x00, // L"K_?B9",				// &HB9
+	0x27, // L"K_COLON",			// &HBA
+	0x0D, // L"K_EQUA0x00, // L",	// &HBB
```
ermshiperete, line 295: Looks like this ocmment had an accident :smiley:

ermshiperete (commented):
Finished review of `keymap.cpp` and `keymap.h`. More to follow...

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+      *p++ = 2;
+      *p = 0;
+      kkp++;
+    }
+  }
+
+return true;
+}
+
+const int CODE__SIZE[] = {
```
ermshiperete, line 625: Is this used anywhere?

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+    // fills m_vk with the VK of the US keyboard
+    // ( mcompile win uses MapVirtualKeyEx() to fill m_vk with the VK of the Underlying keyboard)
+    // Linux can get a VK for the US Keyboard using USVirtualKeyToScanCode/ScanCodeToUSVirtualKey
+    // Linux cannot get a VK for the underling Keyboard
+    // this "connection" is possible only when using All_Vector
+
+    KMX_VirtualKey *key = new KMX_VirtualKey(sc);
+
+    if((key->VK() != 0) ) {
+        rgKey[key->VK()] = key;
```
ermshiperete, line 388: This leaks memory - we don't delete key before rgKey gets out of scope.

linux/mcompile/keymap/mcompile.cpp
```diff
@@ -22,77 +22,68 @@
...
+#if defined(_WIN32) || defined(_WIN64)
+  int wmain(int argc, wchar_t* argv[]) {
+    std::vector<std::u16string> str_argv_16 = convert_argvW_to_Vector_u16str( argc, argv);
+    run(argc, str_argv_16);
 
-    LPKEYBOARD kmxfile;
+#else  // LINUX
+  int main(int argc, char* argv[]) {
+    std::vector<std::u16string> str_argv_16 = convert_argv_to_Vector_u16str(argc, argv);
+    run(argc, str_argv_16, argv);
```
ermshiperete, line 53: On Linux there is no need to pass UTF-16 strings for filenames. Linux uses UTF-8 and so we can directly use `char *` without converting back and forth from/to UTF-16. This will simplify the code quite a bit.

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+#define CERR_None                                          0x00000000
+#define CERR_CannotAllocateMemory                          0x00008004
+#define CERR_UnableToWriteFully                            0x00008007
+#define CERR_SomewhereIGotItWrong                          0x00008009
+
+KMX_BOOL KMX_VerifyKeyboard(LPKMX_BYTE filebase, KMX_DWORD sz);
+
+LPKMX_KEYBOARD KMX_FixupKeyboard(PKMX_BYTE bufp, PKMX_BYTE base, KMX_DWORD dwFileSize);
+
+KMX_BOOL KMX_SaveKeyboard(LPKMX_KEYBOARD kbd, PKMX_WCHAR filename) {
```
ermshiperete, line 13: We can use `char*` (or `KMX_CHAR*`) as data type for `filename`.

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+			fk->cxStoreArray * sizeof(KMX_COMP_STORE) +
+      //wcslen(fk->szName)*2 + 2 +
+			//wcslen(fk->szCopyright)*2 + 2 +
+			//wcslen(fk->szLanguageName)*2 + 2 +
+			//wcslen(fk->szMessage)*2 + 2 +
+      fk->dwBitmapSize;
+
+	for(i = 0, fgp = fk->dpGroupArray; i < fk->cxGroupArray; i++, fgp++) {
+    if(fgp->dpName)
+	    size += u16len(fgp->dpName)*2 + 2;
```
ermshiperete, line 66: Easier to understand and better portable is:

```suggestion
	    size += (u16len(fgp->dpName) + 1) * sizeof(KMX_WCHAR);
```

same below for the other strings

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
 
-	if(!buf) {
-		Err(L"Not allocmem");
-		CloseHandle(hFile);
-		return FALSE;
-	}
+	offset += sizeof(KMX_COMP_GROUP) * ck->cxGroupArray;
 
-  filebase = buf;
+	for(i = 0; i < ck->cxGroupArray; i++, gp++, fgp++) {
```
ermshiperete, line 147:
```suggestion
	offset += sizeof(KMX_COMP_GROUP) * ck->cxGroupArray;

	for(i = 0, fgp = fk->dpGroupArray; i < ck->cxGroupArray; i++, gp++, fgp++) {
```

ermshiperete (commented):
Reviewed mc_import_rules and parts of mc_kmxfile

linux/mcompile/keymap/u16.cpp
```diff
@@ -0,0 +1,329 @@
...
+    while (*q == ch) q++;
+    *ctx = q;
+  }
+  else {
+    *ctx = NULL;
+  }
+  return p;
+}
+
+KMX_WCHAR * u16tok(KMX_WCHAR* p,  KMX_WCHAR* delim, KMX_WCHAR** ctx) {
```
ermshiperete, line 282: I'm not sure what this method is supposed to do exactly. If `delim` is supposed to be a string then it doesn't work as expected. If it is supposed to be an array of delimiters then the variable should be named differently, e.g. `delimiters` or `tokens`. Again, unit tests would clarify this.

linux/mcompile/keymap/u16.h
```diff
@@ -0,0 +1,47 @@
...
+const KMX_WCHAR * u16ncat(KMX_WCHAR *dst, const KMX_WCHAR *src, size_t max);
+KMX_WCHAR * u16tok(KMX_WCHAR *p,  KMX_WCHAR ch,  KMX_WCHAR **ctx);
+KMX_WCHAR * u16tok(KMX_WCHAR* p,  KMX_WCHAR* ch, KMX_WCHAR** ctx);
+long int u16tol(const KMX_WCHAR* str, KMX_WCHAR** endptr, int base);
+double u16tof( KMX_WCHAR* str);
+
+KMX_CHAR* strrchr_slash(KMX_CHAR* Name);
+const KMX_WCHAR* u16rchr_slash(KMX_WCHAR const* Name);
+
+std::string toHex(int num1);
```
ermshiperete, line 44: `toHex` is not implemented, so it'd be good to remove it from here.

linux/mcompile/keymap/u16.cpp
```diff
@@ -0,0 +1,329 @@
...
+long int u16tol(const KMX_WCHAR* str, KMX_WCHAR** endptr, int base)
+{
+  auto s = string_from_u16string(str);
+  char* t;
+  long int result = strtol(s.c_str(), &t, base);
+  if (endptr != nullptr) *endptr = (KMX_WCHAR*)str + (t - s.c_str());
+  return result;
+}
+/*
+std::string toHex(int num1) {
```
ermshiperete, line 119: Please remove commented code

ermshiperete (changes requested):
All files reviewed now. Sorry for the many comments :smile:

SabineSIL (commented):
Hi Eberhard, here is the first round of changes addressing your comments. I created a PR for that: PR #11761. This PR is only about changes in comments and symbol names throughout all mcompile-linux. There might still be other unaddressed comments in functions or modules that have already had symbols renamed.

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+int KMX_DeadKeyMap(int index, std::vector<DeadKey *> *deadkeys, int deadkeyBase, std::vector<KMX_DeadkeyMapping> *deadkeyMappings) {   // I4327   // I4353
+  for(size_t i = 0; i < deadkeyMappings->size(); i++) {
+    if((*deadkeyMappings)[i].deadkey == index) {
+      return (*deadkeyMappings)[i].dkid;
+    }
+  }
+
+  for(size_t i = 0; i < deadkeys->size(); i++) {
+    if((*deadkeys)[i]->KMX_DeadCharacter() == index) {
+      return deadkeyBase + i;
```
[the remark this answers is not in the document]
ermshiperete: Sounds like a question for @mcdurdin then :smile:

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+  std::u16string KMX_GetShiftState(ShiftState shiftState, bool capsLock) {
+    return this->m_rgss[(UINT)shiftState][(capsLock ? 1 : 0)];
+  }
+
+  void KMX_SetShiftState(ShiftState shiftState, std::u16string value, bool isDeadKey, bool capsLock) {
+    this->m_rgfDeadKey[(UINT)shiftState][(capsLock ? 1 : 0)] = isDeadKey;
+    this->m_rgss[(UINT)shiftState][(capsLock ? 1 : 0)] = value;
+  }
+
+  bool KMX_IsSGCAPS() {
```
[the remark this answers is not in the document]
ermshiperete: ok. Should rename to `KMX_IsSwissGermanCAPS` then.

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+        }
+      }
+    }
+    return true;
+  }
+};
+
+class KMX_Loader {
+private:
+  KMX_BYTE lpKeyStateNull[256];
```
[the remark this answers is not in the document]
ermshiperete: In the c'tor we only initialize it with 0s, but otherwise it's not used on Linux. On Windows it's used in a call to `ToUnicodeEx` in `ClearKeyboardBuffer`, but we don't have that on Linux, so I think this can safely go.

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+            key++;
+          }
+        }
+      }
+    }
+    return true;
+  }
+};
+
+class KMX_Loader {
```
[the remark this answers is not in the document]
ermshiperete: Still thinking we don't need this class on Linux. The only method we call from outside the class is `KMX_MaxShiftState`, but this will always return `ShftMenuCtrl` since we never call `Set_XxxxVk`, so I think we can replace the calls to `KMX_MaxShiftState()` with the `ShftMenuCtrl` const and remove this class.

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+  }
+  return n;
+}
+
+bool KMX_ImportRules(LPKMX_KEYBOARD kp,v_dw_3D  &All_Vector, GdkKeymap **keymap, std::vector<KMX_DeadkeyMapping> *FDeadkeys, KMX_BOOL bDeadkeyConversion) {   // I4353   // I4552
+  KMX_Loader loader;
+
+  std::vector<KMX_VirtualKey*> rgKey; //= new VirtualKey[256];
+  std::vector<DeadKey*> alDead;
+  std::vector<DeadKey*> alDead_cpl = create_alDead();
```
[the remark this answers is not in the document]
ermshiperete: The method name is good now, but the variable name is still not clear. Can we rename that also?

linux/mcompile/keymap/mc_import_rules.cpp
ermshiperete: ok, but that comment is missing in the Linux version.

linux/mcompile/keymap/mc_import_rules.cpp
ermshiperete: Might be better to move it next to the implementation of `KMX_incxstr` in `mc_kmxfile.cpp` then. That would also allow to remove the declaration from `kmx_file.h`.

ermshiperete (commented):
Reviewed #11761 and replied to a few comments in this PR.

linux/mcompile/keymap/filesystem.cpp
SabineSIL: This is used in mcompile-linux and mcompile-mac. 
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue  [( # 11802)](https://github.com/keymanapp/keyman/issues/11802)

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac. 
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue https://github.com/keymanapp/keyman/issues/11802

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+
+        PKMX_WCHAR p;
+
+        if (st.size() == 0) {
+          // No character assigned here
+        }
+        else if (this->m_rgfDeadKey[(int)ss][caps]) {
+          // It's a dead key, append an @ sign.
+          key->dpContext = new KMX_WCHAR[1];
+          *key->dpContext = 0;
```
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac.
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue #11802

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+            *p = 0;
+          } else {
+
+            p = key->dpOutput = new KMX_WCHAR[4];
+            *p++ = UC_SENTINEL;
+            *p++ = CODE_DEADKEY;
+            *p++ = KMX_DeadKeyMap(st[0], deadkeys, deadkeyBase, &KMX_FDeadkeys);   // I4353
+            *p = 0;
+          }
+          key++;
```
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac.
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue #11802

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+            // key->dpOutput stores character Underlying
+
+            KMX_DWORD SC_Underlying = KMX_get_KeyCodeUnderlying_From_KeyCodeUS(keymap, All_Vector, this->SC(), (ShiftState) ss, caps);
+            key->Key = KMX_get_VKUS_From_KeyCodeUnderlying( SC_Underlying);
+
+            key->Line = 0;
+            key->ShiftFlags = this->KMX_GetShiftStateValue(capslock, caps, (ShiftState) ss);
+
+            key->dpContext = new KMX_WCHAR;
+            *key->dpContext = 0;
```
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac.
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue #11802

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+            p = key->dpOutput = new KMX_WCHAR[st.size() + 1];
+            for(size_t ich = 0; ich < st.size(); ich++) {
+              *p++ = st[ich];
+            }
```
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac.
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue #11802

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+KMX_BOOL KMX_VerifyKeyboard(LPKMX_BYTE filebase, KMX_DWORD sz){
+  KMX_DWORD i;
+  PKMX_COMP_KEYBOARD ckbp = (PKMX_COMP_KEYBOARD)filebase;
+  PKMX_COMP_STORE csp;
+
+  // Check file version //
+
+  if (ckbp->dwFileVersion < VERSION_MIN || ckbp->dwFileVersion > VERSION_MAX) {
+    // Old or new version -- identify the desired program version //
+    for (csp = (PKMX_COMP_STORE)(filebase + ckbp->dpStoreArray), i = 0; i < ckbp->cxStoreArray; i++, csp++) {
```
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac.
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue #11802

linux/mcompile/keymap/mcompile.cpp
```diff
@@ -22,77 +22,68 @@
...
-BOOL SaveKeyboard(LPKEYBOARD kbd, PWSTR filename);
-bool ImportRules(WCHAR *kbid, LPKEYBOARD kp, std::vector<DeadkeyMapping> *FDeadkeys, BOOL bDeadkeyConversion);   // I4353   // I4327
-BOOL ConvertKeyboardToUnicode(LPKEYBOARD kbd);   // I4273
-int run(int argc, wchar_t * argv[]);
 
-std::vector<DeadkeyMapping> FDeadkeys;   // I4353
+KMX_BOOL KMX_DoConvert(LPKMX_KEYBOARD kbd, KMX_BOOL bDeadkeyConversion, gint argc, gchar *argv[]);
 
-#define KEYMAN_SENTRY_LOGGER_DESKTOP_ENGINE_MCOMPILE KEYMAN_SENTRY_LOGGER_DESKTOP_ENGINE ".mcompile"
+bool KMX_ImportRules( LPKMX_KEYBOARD kp,v_dw_3D &All_Vector, GdkKeymap **keymap,std::vector<KMX_DeadkeyMapping> *KMX_FDeadkeys, KMX_BOOL bDeadkeyConversion); // I4353   // I4327
```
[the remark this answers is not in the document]
SabineSIL: This is used in mcompile-windows, mcompile-linux and mcompile-mac.
In order to have the same code of mcompile throughout all OS´s I moved this to a new issue #11802

linux/mcompile/keymap/deadkey.cpp
```diff
@@ -0,0 +1,468 @@
+#include "keymap.h"
+#include "deadkey.h"
+
+v_dw_1D createLine(std::string  first, std::string second, KMX_DWORD number, std::string nameresult) {
```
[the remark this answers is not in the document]
SabineSIL: OK-changed- see PR [#11813](https://github.com/keymanapp/keyman/pull/11813/files#diff-9b898a02febfbf82b27f5dc5a338d96011f1714324a2722477b04ddfee63a8b9)

linux/mcompile/keymap/deadkey.cpp
SabineSIL: OK-changed- see PR [#11813](https://github.com/keymanapp/keyman/pull/11813/files#diff-9b898a02febfbf82b27f5dc5a338d96011f1714324a2722477b04ddfee63a8b9)

linux/mcompile/keymap/deadkey.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+bool found_dk_inVector(KMX_WCHAR dk, std::vector<DeadKey*> &dkVec) {
+	int i=0;
+	if( dkVec.size() > 0) {
+		do {
+			if( dk == dkVec[i]->KMX_GetDeadCharacter())
+				return true;
+			i++;
+		} while (i < (int) dkVec.size());
+	}
```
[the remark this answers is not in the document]
SabineSIL: OK-changed- see PR [#11813](https://github.com/keymanapp/keyman/pull/11813/files#diff-9b898a02febfbf82b27f5dc5a338d96011f1714324a2722477b04ddfee63a8b9)

linux/mcompile/keymap/mc_import_rules.cpp
SabineSIL: OK-changed- see PR #11816;
also #included kmx_file.h from common instead of having a copy in mcompile

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
SabineSIL: That`s right. 
Nevertheless I would rather not change this because I´d like to keep things the same for mcompile-windows, mcompile-linux, mcompile-mac as much as possible. (of course only if things work fine!)
If we change things we should do this on all platforms simultaneously in order for mcompile to be more maintainable.
Therefore I moved this to Issue #11802

SabineSIL (commented):
Hi Eberhard, here are the next changes 😄

linux/mcompile/keymap/mc_import_rules.h
```diff
@@ -0,0 +1,38 @@
+
+#pragma once
+#ifndef MC_IMPORT_RULES_H
+#define MC_IMPORT_RULES_H
+
+class DeadKey {
```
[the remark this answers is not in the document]
SabineSIL: That`s right. 
Nevertheless I would rather not change this because I´d like to keep things the same for mcompile-windows, mcompile-linux, mcompile-mac as much as possible. (of course only if things work fine!)
If we change things we should do this on all platforms simultaneously in order for mcompile to be more maintainable.
Therefore I moved this to Issue #11802

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+  if (!gdk_keymap_get_entries_for_keycode(keymap, keycode, &maps, &keyvals, &count))
+    return 0;
+
+  if (!(shift_state_pos <= count))
+   return 0;
+
+  if (!(keycode <= keycode_max))
+    return 0;
+
+  KMX_DWORD KeyVal= (KMX_DWORD) KMX_get_KeyVal_From_KeyCode(keymap, keycode, ShiftState(shift_state_pos), caps);
```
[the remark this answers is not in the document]
SabineSIL: OK-changed- see PR [#11816](https://github.com/keymanapp/keyman/pull/11816)

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+  std::vector<KMX_WCHAR>::iterator it;
+  for(it=this->m_rgbasechar.begin(); it<m_rgbasechar.end(); it++) {
+    if(*it == baseCharacter) {
+      return true;
+    }
+  }
+  return false;
+}
+
+int KMX_ToUnicodeEx(guint keycode, PKMX_WCHAR pwszBuff, int shift_state_pos, int caps,GdkKeymap *keymap) {
```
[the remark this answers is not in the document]
ermshiperete: Check that the buffer is big enough for what we want to put in it. But see below.

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+          if(bDeadkeyConversion) {   // I4552
+            p = key->dpOutput = new KMX_WCHAR[2];
+            *p++ = st[0];
+            *p = 0;
+          } else {
+
+            p = key->dpOutput = new KMX_WCHAR[4];
+            *p++ = UC_SENTINEL;
+            *p++ = CODE_DEADKEY;
+            *p++ = KMX_DeadKeyMap(st[0], deadkeys, deadkeyBase, &KMX_FDeadkeys);   // I4353
```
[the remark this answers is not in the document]
ermshiperete: I think it would be better to let `KMX_DeadKeyMap` a `KMX_WCHAR` instead of an int.

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
SabineSIL: OK-done

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+
+  if (!(shift_state_pos <= count))
+   return 0;
+
+  if (!(keycode <= keycode_max))
+    return 0;
+
+  KMX_DWORD KeyVal= (KMX_DWORD) KMX_get_KeyVal_From_KeyCode(keymap, keycode, ShiftState(shift_state_pos), caps);
+  std::u16string str = convert_DeadkeyValues_To_U16str(KeyVal);
+  pwszBuff[0]= * (PKMX_WCHAR)  str.c_str();
```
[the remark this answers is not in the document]
SabineSIL: - I use a buffer here because on windows we use a buffer at this place and I am hesitant to change the code too much in order to keep things as similar as possible on all platforms
- We will not have to worry about surrogate pairs since mcompile is only used for latin scripts. 
- On windows we use ToUnicodeEx to obtain a character and a returnvalue rc that specifies if we found a dk/character/no translation. Again for reasons of similarity I called the equivalent Linux-function  KMX_ToUnicodeEx (even though it is not exactly the same as on windows) because on Linux we too need a character and the return value rc.

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
SabineSIL: OK-done

SabineSIL (commented):
Here is the second round of mc_import_rules:

I think we need to talk with @mcdurdin Marc on how much similarity we want between mcompile on the different platforms. What do we do about things that make the code nicer but are not really neccesary like moving class definitions to other modules, using a pointer instead of a buffer, renaming functions, variables etc.

Do we change things on Linux only? 
Now, later or not at all? 
Do we want to change for all 3 platforms simultaniously later? 
Something else?

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
SabineSIL: see PR #11816

linux/mcompile/keymap/mc_import_rules.cpp
```diff
@@ -0,0 +1,651 @@
...
+          for(int caps = 0; caps <= 1; caps++) {
+            int rc = KMX_ToUnicodeEx(KC_US, sbBuffer, ss, caps, *keymap);
+
+            if(rc > 0) {
+              if(*sbBuffer == 0) {
+                rgKey[iKey]->KMX_SetShiftState(ss, u"", false, (caps));       // different to windows since behavior on Linux is different
+              }
+              else {
+                if( (ss == Ctrl || ss == ShftCtrl) ) {
+                continue;
```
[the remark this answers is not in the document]
SabineSIL: OK-done. (Did not allow all changes though)

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
ermshiperete: Yes, I understand. However, a name implies a certain functionality, and so if the name contains Unicode the reader expects it to properly deal with Unicode which includes surrogate pairs. Without careful inspection of where this gets called from and what's passed in it's impossible to know that surrogate pairs are out of scope. I see three possible options to solve this:
- rename the method
- assert that str.c_str()[0] is not a surrogate, i.e. is not in the range 0xD800-0xDFFF (with a comment that those are surrogates).
- or at least a comment in the method that we intentionally don't deal with surrogate pairs since we only deal with latin scripts

linux/mcompile/keymap/keymap.h
SabineSIL: ..and 2 others as well:
```
windows/src/global/cpp/VKScanCodes.cpp
windows/src/engine/keyman32/VKScanCodes.cpp
```

linux/mcompile/keymap/keymap.cpp
```diff
@@ -1,101 +1,464 @@
...
 
-/*
-static void PrintKeymapForCode(GdkKeymap *keymap, guint keycode)
-{
-  GdkKeymapKey *maps;
-  guint *keyvals;
-  gint count;
+KMX_DWORD convertNamesTo_DWORD_Value(std::string tok_str) {
+  // more on https://manpages.ubuntu.com/manpages/jammy/man3/keysyms.3tk.html
+  std::map<std::string, KMX_DWORD > first;
```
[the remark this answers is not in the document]
SabineSIL: OK-changed- see PR #11846

linux/mcompile/keymap/keymap.cpp
SabineSIL: ok done

SabineSIL (commented):
Hi Eberhard, some more changes. Thanks for reviewing

linux/mcompile/keymap/README.md
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/keymap.h
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/km_types.h
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/mc_kmxfile.cpp
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/mc_kmxfile.h
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/mcompile.cpp
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/mcompile.h
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

linux/mcompile/keymap/meson.build
ermshiperete, line 1: File permissions should stay 644 (i.e. shouldn't be executable)!

.gitignore
```diff
@@ -199,3 +199,10 @@ lcov.info
 # /developer/src/test/auto/kmcomp/*.kvk*
 # /developer/src/test/auto/kmcomp/*.txt
 /linux/mcompile/keymap/X_bak
+/linux/mcompile/keymap/anii.*
+
+/linux/mcompile/keymap/*.kmx
+
+
+
+/GTK_PLAYGROUND
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11880

windows/src/global/cpp/VKScanCodes.cpp
```diff
@@ -169,7 +169,7 @@ const UINT USVirtualKeyToScanCode[256] =
 	0x00, // L"K_?8F",				// &H8F 
 
 	0x45, // L"K_NUMLOCK",			// &H90
-	0x46, // L"K_SCROL0x00, // L",			// &H91 
+	0x46, // L"K_SCROL",			// &H91
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/meson.build
```diff
@@ -1,11 +1,27 @@
-project('keymap', 'c', 'cpp',
+project('mcompile', 'c', 'cpp',
         license: 'MIT',
         meson_version: '>=1.0')
 
-gtk = dependency('gtk+-3.0', version: '>= 2.4')
+#do I need a Version of dependency?
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/meson.build
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/README.md
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/keymap.cpp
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/keymap.h
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/km_types.h
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/mc_kmxfile.cpp
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/mc_kmxfile.h
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/mcompile.cpp
SabineSIL: OK done: see PR #11880

linux/mcompile/keymap/mcompile.h
SabineSIL: OK done: see PR #11880

SabineSIL (commented):
... this is for comments made on several files

linux/mcompile/keymap/mc_kmxfile.cpp
SabineSIL: OK done: see PR https://github.com/keymanapp/keyman/pull/11879

linux/mcompile/keymap/mc_kmxfile.cpp
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+			offset += u16len(fgp->dpName)*2 + 2;
+		}	else {
+      gp->dpName = 0;
+    }
+
+    gp->dpKeyArray = offset;
+		kp = (PKMX_COMP_KEY) (buf + offset);
+		fkp = fgp->dpKeyArray;
+		offset += gp->cxKeyArray * sizeof(KMX_COMP_KEY);
+		for(j = 0; j < gp->cxKeyArray; j++, kp++, fkp++) {
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
 
-	if(kbp->dwIdentifier != FILEID_COMPILED) {
-    Err(L"errNotFileID");
+  fwrite(buf, size,1,hOutfile);
+	if(offset != size)
+  {
     delete[] buf;
-    return FALSE;
+    return CERR_UnableToWriteFully;
   }
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+    fclose(fp);
+    KMX_LogError(L"Could not fseek file\n" );
+    return FALSE;
+  }
 
-	if(!fileName || !lpKeyboard) {
-		Err(L"Bad Filename");
-		return FALSE;
-	}
+  auto sz = ftell(fp);
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+  if (sz < 0) {
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+  //  the file.
+  //  We save the original data at the end of buf; we don't copy strings, so
+  //  those will remain in the location at the end of the buffer.
+    buf = new KMX_BYTE[sz * 3];
+  #else
+    buf = new KMX_BYTE[sz];
+  #endif
+
+  if (!buf) {
+    fclose(fp);
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+  #endif
 
-	if(!buf) {
-		Err(L"Not allocmem");
-		CloseHandle(hFile);
-		return FALSE;
-	}
+  if (fread(filebase, 1, sz, fp) < (size_t)sz) {
+    KMX_LogError(L"Could not read file\n" );
+    fclose(fp);
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
     return FALSE;
   }
-	CloseHandle(hFile);
 
-	if(!VerifyKeyboard(filebase, sz)) {
-    Err(L"errVerifyKeyboard");
-    delete[] buf;
+  if (!KMX_VerifyKeyboard(filebase, sz)) {
+    KMX_LogError(L"errVerifyKeyboard\n" );
     return FALSE;
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
+#ifdef KMX_64BIT
+  kbp = KMX_CopyKeyboard(buf, filebase);
+#else
+  kbp = KMX_FixupKeyboard(buf, filebase, sz);
+#endif
+
+
+  if (!kbp) {
+    KMX_LogError(L"errFixupKeyboard\n" );
     return FALSE;
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

linux/mcompile/keymap/mc_kmxfile.cpp
```diff
@@ -1,297 +1,512 @@
...
-  LPKEY kp;
-
-	kbp->dpStoreArray = (LPSTORE) (base + ckbp->dpStoreArray);
-	kbp->dpGroupArray = (LPGROUP) (base + ckbp->dpGroupArray);
-
-	for(sp = kbp->dpStoreArray, csp = (PCOMP_STORE) sp, i = 0; i < kbp->cxStoreArray; i++, sp++, csp++)	{
-    sp->dpName = StringOffset(base, csp->dpName);
-		sp->dpString = StringOffset(base, csp->dpString);
-	}
+  if (*p == 0)
```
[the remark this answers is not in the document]
SabineSIL: OK done: see PR #11879

SabineSIL (commented):
... this is for comments made on mc_kmxfile

linux/mcompile/keymap/keymap.cpp
```diff
@@ -302,192 +316,235 @@ int createOneVectorFromBothKeyboards(v_dw_3D &All_Vector,GdkKeymap *keymap) {
...
+  // add contents of other keyboard to all_vector
+  if (append_underlying_ToVector(all_vector, keymap)) {
+    printf("ERROR: can't append underlying ToVector \n");
     return 2;
   }
   return 0;
 }
 
-int write_US_ToVector( v_dw_3D &vec,std::string language, const char* text) {
+int write_US_ToVector(vec_dword_3D& vec, std::string language, const char* section) {
```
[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

linux/mcompile/keymap/keymap.cpp
```diff
@@ -302,192 +316,235 @@ int createOneVectorFromBothKeyboards(v_dw_3D &All_Vector,GdkKeymap *keymap) {
...
+  std::string fullPathName = "/usr/share/X11/xkb/symbols/" + language;
+  const char* path = fullPathName.c_str();
+  FILE* fp         = fopen((path), "r");
```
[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

linux/mcompile/keymap/keymap.cpp
```diff
@@ -302,192 +316,235 @@ int createOneVectorFromBothKeyboards(v_dw_3D &All_Vector,GdkKeymap *keymap) {
...
+  vec_string_1D vector_completeUS;
+  if (createCompleteVector_US(fullPathName, section, vector_completeUS)) {
+    printf("ERROR: can't Create complete row US \n");
     return 1;
   }
 
   // split contents of 1D Vector to 3D vector
-  if( split_US_To_3D_Vector( vec,Vector_completeUS)) {
+  if (split_US_To_3D_Vector(vec, vector_completeUS)) {
     return 1;
```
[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

linux/mcompile/keymap/keymap.cpp
```diff
@@ -1,463 +1,861 @@
...
+KMX_DWORD KMX_get_KeyValUnderlying_From_KeyCodeUnderlying(GdkKeymap* keymap, UINT vk_ShiftState, UINT kc_underlying, PKMX_WCHAR deadkey) {
+  GdkKeymapKey* maps;
+  guint* keyvals;
+  gint count;
+  PKMX_WCHAR dky = NULL;
+
+  if (!gdk_keymap_get_entries_for_keycode(keymap, kc_underlying, &maps, &keyvals, &count))
+    return 0;
+
+  if (!(ensureValidInputForKeyboardTranslation(
```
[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

linux/mcompile/keymap/keymap.cpp
```diff
@@ -1,463 +1,861 @@
...
+        create_row = true;
 
       // as long as we are in the same xkb_symbol layout block and find "key <" we push the whole line into a 1D-vector
-      if ((print_OK) && (std::string(str_buf).find(key) != std::string::npos)) {
-        printf("%s", buffer);
-        complete_List.push_back(buffer);
-        KeyboardFile << buffer;
+      else if (create_row && (std::string(str_buf).find(key) != std::string::npos)) {
+        complete_List.push_back(line);
       }
```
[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

linux/mcompile/keymap/keymap.cpp
```diff
@@ -1,463 +1,861 @@
...
-  else if ( in == "key<BKSL>")    out = 62;   //correct ???
-  else if ( in == "key<LSGT>")    out = 51;   //correct ???
-  return out;
+vec_dword_2D create_empty_2D_Vector(int dim_rows, int dim_ss) {
+  vec_dword_1D shifts;
+  vec_dword_2D vector_2D;
+
+    for (int j = 0; j < dim_ss; j++) {
+      shifts.push_back(INVALID_NAME);
+    }
```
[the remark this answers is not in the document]
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886)

linux/mcompile/keymap/keymap.cpp
SabineSIL: OK-done -> PR [#11886](https://github.com/keymanapp/keyman/pull/11886):
off by one: you were right

shift_state_pos ( now renamed to gdk_level) is the level for GDK : 
Level 0 = Base; 
Level 1 = Shift; 
Level 2 = ALT; 
Level 3 = ALT+ Shift; 

Before we use the function gdk_translate_keyboard_state() we make sure that the level exists e.g. if for a keyboard count = 16, level can maximal be 15. This is what !(shift_state_pos < count) ( now moved to ensureValidInputForKeyboardTranslation() ) checks.

linux/mcompile/keymap/mcompile.cpp
SabineSIL: OK-done #11887

linux/mcompile/keymap/mcompile.cpp
```diff
@@ -300,38 +274,38 @@ void AddDeadkeyRule(LPKEYBOARD kbd, WCHAR deadkey, WORD vk, UINT shift) {
...
-  WCHAR dkid = 0;
+KMX_WCHAR KMX_ScanXStringForMaxDeadkeyID(PKMX_WCHAR str) {
+  KMX_WCHAR dkid = 0;
   while(str && *str) {
     if(*str == UC_SENTINEL) {
       switch(*(str+1)) {
       case CODE_DEADKEY:
         dkid = max(dkid, *(str+2));
       }
     }
```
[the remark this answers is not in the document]
SabineSIL: OK-done #11887

linux/mcompile/keymap/mcompile.h
```diff
@@ -17,35 +17,27 @@
...
+struct KMX_DeadkeyMapping {   // I4353
+  KMX_WCHAR deadkey, dkid;
+  UINT shift;
   KMX_WORD vk;
 };
 
-extern std::vector<DeadkeyMapping> FDeadkeys;   // I4353
+extern std::vector<KMX_DeadkeyMapping> KMX_FDeadkeys;   // I4353
 
+int run(int argc, std::vector<std::u16string>  str_argv, char* argv[]);
```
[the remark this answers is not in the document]
SabineSIL: OK-done #11887

linux/mcompile/keymap/filesystem.cpp
SabineSIL: OK-addressed  -> #11887
filesystem is removed from mcompile-linux 
function Open_File is moved to the module mc_kmxfile

linux/mcompile/keymap/filesystem.cpp
SabineSIL: OK-addressed  -> #11887
filesystem is removed from mcompile-linux 
function Open_File is moved to the module mc_kmxfile

SabineSIL (commented):
here is filesystem and mcompile

mcdurdin:
> I think we need to talk with @mcdurdin Marc on how much similarity we want between mcompile on the different platforms. What do we do about things that make the code nicer but are not really neccesary like moving class definitions to other modules, using a pointer instead of a buffer, renaming functions, variables etc.
> 
> Do we change things on Linux only? Now, later or not at all? Do we want to change for all 3 platforms simultaniously later? Something else?

Refactoring and cleanup should be separate to this work -- otherwise the changes get too entangled.

linux/mcompile/keymap/keymap.cpp
```diff
@@ -1,463 +1,861 @@
...
 
-  if (!(shift_state_pos < count))
+  if (!(ensureValidInputForKeyboardTranslation(convert_WinShiftstate_to_LinuxShiftstate(ss), (int)count, (int)keycode))) {
+    g_free(keyvals);
+    g_free(maps);
     return 0;
+  }
 
-  out = keyvals[shift_state_pos];
+  // BASE (shiftstate: 0)
```
ermshiperete, line 677: Hmm, I'm wondering if we have to use `convert_WinShiftstate_to_LinuxShiftstate(ss)` instead of directly `ss`. Otherwise calling `convert_WinShiftstate_to_LinuxShiftstate(ss)` when calling `ensureValidInputForKeyboardTranslation` doesn't make much sense to me.

linux/mcompile/keymap/keymap.cpp
```diff
@@ -1,463 +1,861 @@
...
+  if ((ss == Base) && (caps == 0)) {
+    GdkModifierType MOD_base = (GdkModifierType)(~GDK_MODIFIER_MASK);
```
ermshiperete, line 679: Probably better:

```suggestion
    GdkModifierType MOD_base = (GdkModifierType)0;
```

linux/mcompile/keymap/u16.cpp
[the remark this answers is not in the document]
SabineSIL: moved to another issue [#11922](https://github.com/keymanapp/keyman/issues/11922)
we need to change **std::codecvt** as well since this will be deprecated by C++17 
So we might do this together with tests

SabineSIL (commented):
finished comments on u16 and created a new issue [#11922](https://github.com/keymanapp/keyman/issues/11922) for tests and std::codecvt

linux/mcompile/keymap/deadkey.cpp
ermshiperete: (Should create an issue for this for a future refactoring/improvement)

SabineSIL: ok-see #11802

ermshiperete (approved):

ermshiperete:
You'll have to update the branch (aka merge master) to get a successful build. CI build related things have changed on `master` which causes the failures.

linux/mcompile/keymap/meson.build
```diff
@@ -1,11 +1,30 @@
...
+    'keymap.cpp',
+    'deadkey.cpp',
+    'mcompile.cpp',
+    'mc_kmxfile.cpp',
+    'mc_import_rules.cpp',
+    'u16.cpp',)
+
+comon_include_dir = [
+  include_directories('../../../common/include'),
+  include_directories('../../../core/src'),
```
mcdurdin, line 22: We shouldn't have a dependency on core here

linux/mcompile/keymap/meson.build
```diff
@@ -1,11 +1,30 @@
...
+deps = [gtk, xkb,libxklavier]
+
+
+cpp_files = files(
+    'keymap.cpp',
+    'deadkey.cpp',
+    'mcompile.cpp',
+    'mc_kmxfile.cpp',
+    'mc_import_rules.cpp',
+    'u16.cpp',)
```
mcdurdin, line 18: we should be using a shared kmx_u16.cpp -- rather than copying from kmcmplib.

linux/mcompile/keymap/mc_import_rules.cpp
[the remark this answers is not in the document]
mcdurdin: This is because there may be existing deadkeys in the keyboard, so we need to make sure we add any new deadkeys with unique indices after all the existing deadkeys.

PR #9387, opened 2023-08-02
chore(linux): remove change to trigger build

mcdurdin:
@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

mcdurdin:
This PR did the job I needed for the status reports on GHA builds. Ideally the subject line would show the PR but it's visible now in the body of the message which is a whole lot less confusing than before. This'll do. @ermshiperete FYI.

image

Issue #9398, opened 2023-08-02
chore(linux): Enable building for Ubuntu 23.10 Mantic in packaging GHA


ermshiperete:
Currently the Ubuntu/Debian maintainers work on the ibus 1.5.29 package which results in a lot of new packages being released. Once the ibus package stabilized in Mantic we can update our patched ibus version.

At that point we should enable Mantic in the packaging GHA.

ermshiperete:
Depends on #9520

PR #9399, opened 2023-08-02
chore(linux): Remove Kinetic from GHA

ermshiperete:
Don't build Ubuntu 22.10 Kinetic with packaging GHA.

Also prepare for Ubuntu 23.10 Mantic - we can't currently build that because they are trying to get ibus 1.5.29-beta to work on so there are often new packages. We wait until that stabilized before we're providing a new patched ibus version at which point we can enable building for mantic.
See also #9398.

@keymanapp-test-bot skip

### DIFF
--- a/.github/workflows/deb-packaging.yml
+++ b/.github/workflows/deb-packaging.yml
@@ -103,7 +103,10 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):

keyman-server:

PR #9400, opened 2023-08-02
chore(linux): Properly treat test builds with packaging GHA

ermshiperete:
Previously GHA builds always used the "local" environment. This change now detects test builds as well as release builds and properly sets the environment.

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/build-utils.sh
+++ b/resources/build/build-utils.sh
@@ -74,24 +74,31 @@ function findVersion() {
         VERSION_TAG=
     fi
 
-    if [ -z "${TEAMCITY_VERSION-}" ]; then
-        # Local dev machine, not TeamCity
+    if [ -z "${TEAMCITY_VERSION-}" ] && [ -z "${GITHUB_ACTIONS-}" ]; then
+        # Local dev machine, not TeamCity or GitHub Action
         VERSION_TAG="$VERSION_TAG-local"
         VERSION_ENVIRONMENT=local
-    else
+    elif [ -n "${TEAMCITY_PR_NUMBER-}" ]; then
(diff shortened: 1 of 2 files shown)


resources/build/build-utils.sh
```diff
@@ -74,24 +74,31 @@ function findVersion() {
...
     fi
 
-    if [ -z "${TEAMCITY_VERSION-}" ]; then
-        # Local dev machine, not TeamCity
+    if [ -z "${TEAMCITY_VERSION-}" ] && [ -z "${GITHUB_ACTIONS-}" ]; then
+        # Local dev machine, not TeamCity or GitHub Action
         VERSION_TAG="$VERSION_TAG-local"
         VERSION_ENVIRONMENT=local
-    else
+    elif [ -n "${TEAMCITY_PR_NUMBER-}" ]; then
```
mcdurdin, line 81: Why change from -z to -n?

mcdurdin (approved):

resources/build/build-utils.sh
ermshiperete: Seems easier than the double negation `! -z`. https://www.shellcheck.net/wiki/SC2236

keyman-server:

resources/build/build-utils.sh
mcdurdin: Gotcha, makes sense, I wasn't sure if support was the same for both `-n` and `-z`. Apparently `-n` has a little bit of a gotcha around quotation marks though: https://unix.stackexchange.com/questions/109625/shell-scripting-z-and-n-options-with-if#comment1152278_109631

PR #9405, opened 2023-08-02
feat(core): marker implementation 🙀

srl295:
- basic test of markers

For: #9119

@keymanapp-test-bot skip


srl295:
seems to pass tests now … cleanup and optimizing

core/src/ldml/ldml_processor.cpp
```diff
@@ -231,75 +233,89 @@ ldml_processor::process_event(
...
-
         // Look up the key
         const std::u16string str = keys.lookup(vk, modifier_state);
+
         if (str.empty()) {
-          // not found, so pass the keystroke on to the Engine
+          // no key was found, so pass the keystroke on to the Engine
           state->actions().push_invalidate_context();
           state->actions().push_emit_keystroke();
           break; // ----- commit and exit
```
mcdurdin, line 245: Big question about what happens with transforms if no key matches. Do we have a set of default keys (i.e. QWERTY)? And if so, would they trigger transforms? Or are any undefined keys no-ops?

core/src/ldml/ldml_processor.cpp
```diff
@@ -231,75 +233,89 @@ ldml_processor::process_event(
...
-        }
-        // Now process transforms
-        // Process the transforms
+        emit_text(state, str32);
+
+        // Now, process the actual transforms
         if (!!transforms) {
-          // add the newly added char to ctxt
-          ctxt.push_back(str32);
+          // add the newly added char to ctxtstr
```
mcdurdin, line 266:
```suggestion
          // add the newly added key output to ctxtstr
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -336,5 +352,76 @@ km_kbp_status ldml_processor::validate() const {
...
+void
+ldml_processor::emit_text(km_kbp_state *state, km_kbp_usv ch) {
+  assert(ch != LDML_UC_SENTINEL);
+  state->context().push_character(ch);
+  state->actions().push_character(ch);
+}
+
+void ldml_processor::emit_marker(km_kbp_state *state, KMX_DWORD marker_no) {
+    // OK, push the marker
+    state->actions().push_marker(marker_no);
```
mcdurdin, line 393: Let's assert on invalid markers

core/src/ldml/ldml_processor.cpp
```diff
@@ -336,5 +352,76 @@ km_kbp_status ldml_processor::validate() const {
...
+    state->context().push_marker(marker_no);
+}
+
+size_t
+ldml_processor::context_to_string(km_kbp_state *state, std::u32string &str) {
```
mcdurdin, line 398: Should we empty str before we start?

core/src/ldml/ldml_processor.cpp
```diff
@@ -336,5 +352,76 @@ km_kbp_status ldml_processor::validate() const {
...
+      } else if (last_type == KM_KBP_BT_MARKER) {
+        prepend_marker(str, c->marker);
+      } else {
+        break;
+      }
+    }
+    return ctxlen; // consumed the entire context buffer.
+}
+
+void ldml_processor::prepend_marker(std::u32string &str, KMX_DWORD marker) {
```
mcdurdin, line 416: This probably could be inline if we are only using it in the one place

mcdurdin (approved):
LGTM. Mostly nits and QA. Stumbled on one potential disaster for keyboards with tens of thousands of markers which shouldn't happen in reality but could be fuzzed into being...

core/src/ldml/ldml_processor.cpp
srl295: fyi @rc-swag — _my understanding_ was that unknown keys get kicked back to the OS. That seems to be what the other processor does. this is how frame keys, enter, etc get handled also.  

I think that this breaks up the context then, doesn't it? ( a point of discontinuity)

core/src/ldml/ldml_processor.cpp
```diff
@@ -336,5 +352,76 @@ km_kbp_status ldml_processor::validate() const {
   return _valid ? KM_KBP_STATUS_OK : KM_KBP_STATUS_INVALID_KEYBOARD;
 }
 
+void
+ldml_processor::emit_text(km_kbp_state *state, const std::u16string &str) {
+  const std::u32string str32 = kmx::u16string_to_u32string(str);
```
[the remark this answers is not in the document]
srl295: so maybe 0x0001 - 0xD7FE + 0xD7FF for 'any'?

(I'm not usually _emitting_ any, but i want to reserve its place in the representation)

core/src/ldml/ldml_processor.cpp
```diff
@@ -336,5 +352,76 @@ km_kbp_status ldml_processor::validate() const {
...
+      emit_marker(state, marker_no);
+    } else {
+      emit_text(state, ch);
+    }
+  }
+}
+
+void
+ldml_processor::emit_text(km_kbp_state *state, km_kbp_usv ch) {
+  assert(ch != LDML_UC_SENTINEL);
```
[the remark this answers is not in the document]
srl295: This is more of an internal assert to make sure that the logic isn't misfiring as if someone made an API error and called `emit_text(keys[n].to.value)` to emit a key directly.

we could separately have:
- a pre-scan of all key=to and transform=to strings and make sure there's no illegal unicode - this would be a validation error, not an assert. This would be at runtime.   kmc should also do the same so we catch it at compile time 
- we could retain an assert on no illegal unicode _here_, that would make sure the upcoming regex machinery doesn't emit issues. But this wouldn't be at production runtime. 

Seems like a separate ticket to me? it's drifted from the mark(er) consideration.

core/src/ldml/ldml_processor.cpp
srl295: yes with a macro to check such

core/src/ldml/ldml_processor.cpp
srl295: we could, but i thought of it as an append.  might be safer as you propose

core/src/ldml/ldml_processor.cpp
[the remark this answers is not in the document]
srl295: filed https://github.com/keymanapp/keyman/issues/9446

srl295:
ok @rc-swag let me know if you want to review this. I think i've implemented everything in the notes, except for 1. unicode invalids, split out, and 2. moving the marker range, which i will make a separate PR

rc-swag:
> ok @rc-swag let me know if you want to review this.

I will review this today but don't wait for my approval tick as you have Marc's and I'm sure anything I have would be minor edits.

srl295:
@rc-swag  i'll wait

core/src/ldml/ldml_processor.cpp
mcdurdin: Not quite. Some frame keys cause a context invalidation. Frame keys and modifier keys etc then get kicked back to the OS, but any US English character-generating key gets added as an action:

https://github.com/keymanapp/keyman/blob/c0ec8ba76c623be503e9050ff289f7db094ccf07/core/src/kmx/kmx_processevent.cpp#L323-L327

So we don't lose context on those.

core/src/ldml/ldml_processor.cpp
mcdurdin: That feels too side-effecty, should be `append_context_to_string` in that case.

mcdurdin (approved):

core/src/ldml/ldml_processor.cpp
```diff
@@ -231,75 +233,89 @@ ldml_processor::process_event(
...
+        // we convert it here instead of using the emit_text() overload
+        // so that we don't have to reconvert it inside the transform code.
         const std::u32string str32 = kmx::u16string_to_u32string(str);
-        for (const auto &ch : str32) {
-          state->context().push_character(ch);
-          state->actions().push_character(ch);
-        }
-        // Now process transforms
-        // Process the transforms
+        emit_text(state, str32);
```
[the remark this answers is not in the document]
rc-swag: By actually calling emit_text here it will cause the character to be added to the actions, this will bubble all the way up and the platform will output the character only for it to be then deleted in the case of a transform that requires deletion. It will work out ok but we will be sending the application a character then deleting it.
Instead, we can just let the transform have the new character, if there is no match at this point we add the new character. If there is a match and it wants to delete we don't have delete the first character as we had not yet added it. 

Also transform::apply states `@return match length: number of chars at end of input string to modify.  0 if no match.`
Is this correct that there are no transforms that simply result in two code points being emitted?
for example    `q -> qu`
If you can have this then in this case we will also want to add the new char and then apply the outputString

core/src/ldml/ldml_processor.cpp
```diff
@@ -231,75 +233,89 @@ ldml_processor::process_event(
...
+
+        /**
+         * a copy of the current/changed context, for transform use.
+         *
+         */
+        std::u32string ctxtstr;
+
+        if (!!transforms) {
+          (void)context_to_string(state, ctxtstr);
+        } // else, don't bother populating
```
rc-swag, line 256:
```suggestion

```

rc-swag (commented):
@srl295  I ended up having a few comments. Can discuss on a call if need be.

core/src/ldml/ldml_processor.cpp
```diff
@@ -193,6 +193,7 @@ ldml_processor::process_event(
     state->actions().clear();
 
     switch (vk) {
+    // Special handling for backspace VK
     case KM_KBP_VKEY_BKSP:
```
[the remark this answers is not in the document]
srl295: thanks for checking! I'm not sure what backspace+modifier should be doing. Maybe this should be a followup issue?

core/src/ldml/ldml_processor.cpp
srl295: I'm wondering if this could be a separate ticket? I'm definitely not clear on the implications here.

core/src/ldml/ldml_processor.cpp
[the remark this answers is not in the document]
srl295: hived off to https://github.com/keymanapp/keyman/issues/9450

core/src/ldml/ldml_processor.cpp
srl295: hived off to https://github.com/keymanapp/keyman/issues/9451

core/src/ldml/ldml_processor.cpp
srl295: +1

core/src/ldml/ldml_processor.cpp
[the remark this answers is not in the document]
srl295: the return from apply is 'the number of chars in the _matched_ text to change'

so if the input text was `pdq` and you apply `q -> qu` then the match length will be `1` but the output string `qu`, meaning to delete the `q` (size 1) and replace it with `qu` (size 2)

srl295:
@rc-swag think i understood your comments! please take a look

I restructured so that the 'common' / 'no-match' cases were at the top (inverted the sense of the tests), so that the two emit_text(state, str32); are closer together.

core/src/ldml/ldml_processor.cpp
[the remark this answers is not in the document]
rc-swag: Thanks for clarifying. It turns out the change I suggested and you made also works well then for this ` q -> qu` case as now it will just emit the `qu` without first adding and deleting it and then adding it again.

core/src/ldml/ldml_processor.cpp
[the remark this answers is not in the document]
rc-swag: I have some homework to do here also.

rc-swag:
> I restructured so that the 'common' / 'no-match' cases were at the top (inverted the sense of the tests), so that the two emit_text(state, str32); are closer together.

Yes nice the inversion is a good idea.

rc-swag:
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.159-alpha](https://keyman.com/downloads/releases/alpha/17.0.159)

Issue #8519, opened 2023-03-27
feat(touch): select default long-press key when no movement detected


MakaraSok:
### Describe the solution you'd like

Currently in version 16.0.138 of Keyman for Android, long-pressing a particular key shows key(s) as pop-ups to select one.

After focusing the pop-up key using touch, releasing the touch will trigger that key.

If we can make focus to the first key among the pop-up keys set by simply long-pressing only, it will be useful to users to type using keys that have one option under long-press.

Kind attention is invited to this matter to add to a future milestone.

### Keyman apps

- [X] Keyman for Android
- [X] Keyman for iPhone and iPad
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

16.0.138

mcdurdin:
Closely related to #1025

darcywong00:
> If we can make focus to the first key among the pop-up keys set by simply long-pressing only, it will be useful to users to type using keys that have one option under long-press.

My concern is if we default the focus to the first key among the pop-up keys, how would the user "cancel" the long-press? I can see a common case where a user is discovering what the longpress on each key is, and likely won't want that output by default.

mcdurdin:
* Default will be selectable by keyboard designer -- _not_ necessarily the first, and may be _none_.
* Where there is a default will be indicated by highlighting it in the longpress -- the normal selection highlight.
* Given that cancel is _not_ the preferred default action for a longpress, the user can cancel the longpress default by sliding outside the 'acceptance' area -- e.g. south of the key cap, which would then unhighlight the default and indicate visually that no output would occur when they release their finger.

mcdurdin:
This is a duplicate of #9416.

mcdurdin:
Will track in #9416.

PR #9432, opened 2023-08-08
feat(developer): add 'default' property for longpress keys

### LINKED ISSUE #9430: feat(developer): define long-press default key
Split from #9416 for developer functionality and schema updates.

The keyboard designer needs to be able to select the default longpress key, or to have no default. (The fallback behaviour for existing keyboards should probably be no default?)

Requires:

* Update to .keyman-touch-layout schema to allow selection of a default longpress
* Update to Developer IDE to allow keyboard designer to select a default, or no default

mcdurdin:
Fixes #9430.

Adds a 'default' property to subkeys, only used by longpress-type subkeys:

* Added property to schema and readers/writers
* Added property to touch layout validator
* Added editor for property to layout builder

# User Testing

In Keyman Developer, open the touch layout tab in a keyboard. Select a key and create a set of longpress keys.

* **TEST_DEFAULT_PROPERTY**: Select a subkey, and tick the 'Is Default' property. The border of the subkey should be bolded to indicate that it is default. Go to Code view and verify that the property `"default":true` is correctly set for the subkey.
* **TEST_RESET_DEFAULT**: Select the same subkey, and untick the 'Is Default' property on that key. The subkey should lose its bold border. Go to Code view and verify that the property `"default"` is correctly removed for the subkey.
* **TEST_RESET_ALL**: Tick the 'Is Default' property for one subkey. Then, select another subkey under the same key, and tick the 'Is Default' property on that key. The original 'default' subkey should lose its bold border, and the new subkey should have a bolded border. Go to Code view and verify that the property `"default":true` is correctly set for the new subkey, and removed for the old subkey.

### DIFF
--- a/common/schemas/keyman-touch-layout/README.md
+++ b/common/schemas/keyman-touch-layout/README.md
@@ -399,6 +399,13 @@ and `width`.
 : Number. Defines the width of the key. Not currently supported for longpress
   keys. See `key` object for details.
 
+* `default`
+
+: Boolean. Defaults to `false`. If true, this will be the default key when the
+  user longpresses the base key without moving their finger. Only one subkey
+  should be default; if multiple subkeys have the default field set, then
+  behaviour is undefined.
+
 ## `flick` object
 
 A object with a set of properties that define flick keys for given directions.
@@ -444,6 +451,9 @@ output .js file, and transforms any legacy formats (such as "number in a
 string") into the appropriate spec format.
 
 # .keyman-touch-layout version history
+
+## 2023-08-08 2.1 stable
+* Add support for default on longpress (sk) keys. Aligns with Keyman 17.0.
 
 ## 2022-06-30 2.0 stable
 * Add support for flick, multitap, hint, defaultHint. Aligns with Keyman 16.0.
--- a/common/schemas/keyman-touch-layout/keyman-touch-layout.spec.json
+++ b/common/schemas/keyman-touch-layout/keyman-touch-layout.spec.json
@@ -131,7 +131,8 @@
         "width": { "$ref" : "#/definitions/numeric" },
         "fontsize": { "$ref": "#/definitions/fontsize-spec" },
         "font": { "$ref": "#/definitions/font-spec" },
-        "dk": { "type": "string" }
+        "dk": { "type": "string" },
+        "default": { "type": "boolean" }
       },
       "required": ["id"],
       "additionalProperties": false
--- a/developer/src/tike/xml/layoutbuilder/builder.xsl
+++ b/developer/src/tike/xml/layoutbuilder/builder.xsl
@@ -247,6 +247,10 @@
           <label for='selSubKeyNextLayer'>Next Layer:</label>
           <select id='selSubKeyNextLayer'></select>
         </div>
+        <div class='toolbar-item'>
+          <label for='chkSubKeyIsDefault'>Is default:</label>
+          <input id='chkSubKeyIsDefault' type='checkbox' />
+        </div>
       </div>
 
       <!-- Subkey Controls -->
--- a/common/web/types/src/keyman-touch-layout/keyman-touch-layout-file-writer.ts
+++ b/common/web/types/src/keyman-touch-layout/keyman-touch-layout-file-writer.ts
@@ -77,6 +77,7 @@ export class TouchLayoutFileWriter {
       if(Object.hasOwn(key, 'text') && key.text === '') delete key.text;
       if(Object.hasOwn(key, 'id') && <string>key.id === '') delete key.id;
       if(Object.hasOwn(key, 'hint') && (<any>key).hint === '') delete (<any>key).hint;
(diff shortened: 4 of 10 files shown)


bharanidharanj:
# Test Results

* **TEST_DEFAULT_PROPERTY (PASSED):** Tested with the attached PR build (keyman 17.0.157-alpha-test-9432) in Windows 11 OS (VM) and here is my observation: 1. Verified that after enabling 'Is Default' property for a subkey, the border of the subkey was bold. Moved to the Code view and verified that the property 'default:true' is correctly set for the subkey. 

    


* **TEST_RESET_DEFAULT (PASSED):** 1. Verified that after disabling 'Is Default' property for the same subkey, the subkey lost its bold border. 2. Moved to the Code view and verified that the property was removed from the subkey. 


   


* **TEST_RESET_ALL (PASSED):** 1. Verified that the original 'default' subkey lost its bold border and the new subkey is having the bold border. 2. Moved to the code view and verified that the property 'default:true' is correctly set to the new subkey and it has been removed for the old subkey.

common/schemas/keyman-touch-layout/README.md
```diff
@@ -399,6 +399,13 @@ and `width`.
 : Number. Defines the width of the key. Not currently supported for longpress
   keys. See `key` object for details.
 
+* `default`
+
+: Boolean. Defaults to `false`. If true, this will be the default key when the
+  user longpresses the base key without moving their finger. Only one subkey
+  should be default; if multiple subkeys have the default field set, then
+  behaviour is undefined.
```
darcywong00, line 407: It might be nice to have the "cancel" behavior you wrote in #8519 documented somewhere

> * Given that cancel is _not_ the preferred default action for a longpress, the user can cancel the longpress default by sliding outside the 'acceptance' area -- e.g. south of the key cap, which would then unhighlight the default and indicate visually that no output would occur when they release their finger.

developer/src/tike/xml/layoutbuilder/builder.xsl
```diff
@@ -247,6 +247,10 @@
           <label for='selSubKeyNextLayer'>Next Layer:</label>
           <select id='selSubKeyNextLayer'></select>
         </div>
+        <div class='toolbar-item'>
+          <label for='chkSubKeyIsDefault'>Is default:</label>
```
darcywong00, line 251: Is the UX clear what "Is default" does? Possible confusion in the editor if "Key Type" dropdown has "Default".

"Default longpress key" might be too long, but perhaps something similar?

common/schemas/keyman-touch-layout/keyman-touch-layout.spec.json
```diff
@@ -131,7 +131,8 @@
         "width": { "$ref" : "#/definitions/numeric" },
         "fontsize": { "$ref": "#/definitions/fontsize-spec" },
         "font": { "$ref": "#/definitions/font-spec" },
-        "dk": { "type": "string" }
+        "dk": { "type": "string" },
+        "default": { "type": "boolean" }
```
darcywong00, line 135: What kind of error happens if the user manually edits "default" to a string?

developer/src/tike/xml/layoutbuilder/builder.xsl
jahorton: Perhaps "Auto-select"?

jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.158-alpha](https://keyman.com/downloads/releases/alpha/17.0.158)

MayuraVerma:
I have 17.0.161 and a keyman with longpress default is set true. Still longspress is not pickup the default key.
I am using store(&VERSION) '16.0'

store(&VERSION) '17.0' developer throws error

Am I missing something

darcywong00:
@MayuraVerma - This has only been implemented in Keyman Developer so far. The mobile apps will still need to implement handling default subkeys (likely post-17.0)

jahorton:
> store(&VERSION) '17.0' developer throws error

I'm a bit curious about that point.

> The mobile apps will still need to implement handling default subkeys ~~(likely post-17.0)~~

Likely before 17.0 release.

MayuraVerma:
> @MayuraVerma - This has only been implemented in Keyman Developer so far. The mobile apps will still need to implement handling default subkeys (likely post-17.0)

While gestures are implemented in v17, if this gesture, longpress+release is also implemented, it would be great.

mcdurdin:
> store(&VERSION) '17.0' developer throws error

The `&VERSION` statement is really related to the minimum version required to support functionality in the .kmn keyboard (without reference to the visual keyboard, touch keyboard). It is usually optional -- Keyman Developer can figure out the minimum version required. So far, there's no functionality in the .kmn format that requires 17.0, so the maximum version number available is 16.0.

But I plan to add 17.0 to the VERSION statement because flick and multitap gestures are only going to be available in 17.0+. And because that information is stored in the touch keyboard, I need to check and see if Keyman Developer is able to determine it automatically.

Issue #9121, opened 2023-06-27
feat(core): ldml proper regex implementation 🙀🙀


srl295:
- Originally added for #7375 

Remaining items handled in https://github.com/keymanapp/keyman/issues/10319

PR #9440, opened 2023-08-08
feat(core) actual regex 🙀

srl295:
- test for transform cases that require regex
- enable icui18n
- hook up icu regex engine
- try not to break existing ICU use

for #9121
 
@keymanapp-test-bot skip


srl295:
just a weather balloon 🎈 to see how the build fares…

srl295:
build succeeds! 🎉

srl295:
looks like keyman32.vcxproj probably needs an explicit linkage to the ICU libraries?

srl295:
@mcdurdin @rc-swag ready for some initial eyes.. not ready to merge

core/src/ldml/ldml_transforms.cpp
```diff
@@ -11,6 +11,19 @@
...
+#define U_FALLTHROUGH
+#include "unicode/uniset.h"
+#include "unicode/usetiter.h"
+#include "unicode/unistr.h"
+#include "unicode/regex.h"
+#include "unicode/utext.h"
+#include "unicode/utypes.h"
+#else
+#error icu4c is required for this code
+#endif
```
mcdurdin, line 25: can we precondition this instead of postcondition?
```suggestion
#if !defined(HAVE_ICU4C)
#error icu4c is required for this code
#endif

// TODO-LDML: Needed this for some compiler warnings
#define U_FALLTHROUGH
#include "unicode/uniset.h"
#include "unicode/usetiter.h"
#include "unicode/unistr.h"
#include "unicode/regex.h"
#include "unicode/utext.h"
#include "unicode/utypes.h"
```

core/src/ldml/ldml_transforms.cpp
```diff
@@ -398,26 +411,90 @@ transform_entry::transform_entry(const std::u32string &from, const std::u32strin
 
 size_t
 transform_entry::match(const std::u32string &input) const {
-  if (input.length() < fFrom.length()) {
-    // TODO-LDML: regex
-    // Too small, can't match.
-    return 0;
+  // TODO-LDML: simple approach, new regex every time
```
mcdurdin, line 414: I presume by this you mean you'll later precompile all regexes at keyboard load time?
```suggestion
  // TODO-LDML: simple approach, new regex every time -- need to precompile regexes at keyboard load
```

core/src/ldml/ldml_transforms.cpp
```diff
@@ -398,26 +411,90 @@ transform_entry::transform_entry(const std::u32string &from, const std::u32strin
...
+  return matchLen;
 }
 
 std::u32string
-transform_entry::apply(const std::u32string & /*input*/, size_t /*matchLen*/) const {
-  // TODO-LDML: regex
-  // For now, we just return the 'to' string literally.
-  return fTo;
+transform_entry::apply(const std::u32string &input, size_t matchLen) const {
+  // TODO-LDML: and if you thought the previous function was suboptimal,
```
mcdurdin, line 449: :sob:

core/src/meson.build
```diff
@@ -21,6 +21,27 @@ if cpp_compiler.get_id() == 'emscripten'
...
+# ICU4C is used for repertoire tests and core implementation
+
+if target_machine.system() == 'linux'
+  # use pkg-config when targetting linux
+  pkgconfig = 'pkg-config'
+  icu_uc = dependency('icu-uc', required: true)
+  icu_i18n = dependency('icu-i18n', required: true)
+else
+  # load ICU from wrap
+  # Requires meson of about 0.57+
```
mcdurdin, line 33:
```suggestion
```

(We req. meson 1.0 now)

mcdurdin (commented):
looking "good" so far, conceptually. The ICU libraries are static-linked, right (essential on Windows)?

srl295:
ah, i had forgotten a push before asking for review…

core/src/ldml/ldml_transforms.cpp
srl295: yes. In fact,  already precompiled. The Matcher object could be cached, but that will be more of a memory rather than time tradeoff

core/src/ldml/ldml_transforms.cpp
srl295: I actually had done exactly this just hadn't pushed it…great minds!

srl295:
Okay. I'm going to call this one ready to review as far as the implementation goes.  Then i'm working on a sub-PR focused on the build size.

srl295:
okay, given https://github.com/keymanapp/keyman/issues/9467 I think the path for this one should be:
- pass tests
- merge it

It will make wasm bigger for now, but that won't affect KMW yet. But if 9467 is a higher priority before merge, that can be done too. Without KMW calling into wasm, i don't have a way to test the JS callback unless I make a fake test harness.

mcdurdin:
> okay, given #9467 I think the path for this one should be:
> 
> * pass tests
> * merge it

SGTM.

mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.161-alpha](https://keyman.com/downloads/releases/alpha/17.0.161)

Issue #9351, opened 2023-07-26
feat(developer): Support blank or missing .model_info/.keyboard_info 🎺


mcdurdin:
With license from kps:Options/LicenseFile == LICENSE.md, and all other data from package/keyboard sources and binaries

## Generated fields that don't need change.

The following fields are reliably automatically generated, and we should avoid modifying the code for these:
* `id`: from folder name
* `name`: from .kmp (or from .js?)
* `authorName`: from .kmp
* `authorEmail`: from .kmp
* `lastModifiedDate`: when .keyboard_info is generated
* `packageFilename`: from .kmp
* `packageFileSize`: from .kmp
* `jsFilename`: from .js
* `jsFileSize`: from .js
* `isRTL`: from .js or .kmp
* `encodings`: 'unicode'
* `packageIncludes`: from .kmp
* `version`: from .kmp, .js
* `minKeymanVersion`: from .kmp, .js
* `helpLink`: from folder data (id/source/help/id.php existence)
* `platformSupport`: from .kmp, .js
* `sourcePath`: generated from the path in keyboards repo
* `deprecated`: always generated by deprecation references in `links`

Have verified that all release/ and experimental/ keyboards have a .kmp file. This means we can deprecate or even remove the .js parsing, I think, as long as all fields are represented in the .kmp.

## General tasks

- [x] Cleanup the kmc-keyboard-info interface to reduce number of required parameters ➡ keymanapp/keyman#9479
- [x] Verify that all fields are read correctly from .kmp data
- [x] Should we translate `\r\n` to `\n` in kmp.json Description field? ➡ keymanapp/keyman#9476
- [x] `lastModifiedDate` should be date of last commit in folder, which is possible to achieve during release build: `TZ=UTC0 git log -1 --no-merges --date=format:%Y-%m-%dT%H:%M:%SZ --format=%ad` ➡ keymanapp/keyman#9480
- [x] Remove metadata source file references from kmc, e.g. BuildKeyboardInfo.ts, and work from .kpj always. ➡ keymanapp/keyman#9535
- [x] https://github.com/keymanapp/keyboards/issues/2342
- [x] #9620
- [x] help: context/package-editor ➡ keymanapp/help.keyman.com#815
- [x] help: standard file layout for keyboards and models ➡ keymanapp/help.keyman.com#815
[...]

PR #9452, opened 2023-08-11
chore(common): remove deprecated fields from keyboard_info 🎺

mcdurdin:
This is the first PR in a chain that addresses #9351.

Removes the following fields:
* legacyId
* documentationFilename
* documentationFileSize
* links
* related[].note

None of these fields are needed any longer; see #9351 for steps to remove data.

Eliminates difference between keyboard_info source and distribution, renaming to keyboard_info.schema.json, and updating the required members of the json accordingly.

This has corresponding PRs against keyman.com, api.keyman.com, help.keyman.com, and keyboards repositories:

- [ ] keymanapp/api.keyman.com#204
- [ ] keymanapp/keyman.com#382
- [ ] keymanapp/help.keyman.com#798
- [x] keymanapp/keyboards#2336
- [x] keymanapp/keyboards#2337

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/test/auto/keyboard-info/test-invalid-language-code.keyboard_info
+++ b/developer/src/test/auto/keyboard-info/test-invalid-language-code.keyboard_info
@@ -1,6 +1,17 @@
 {
+    "id": "test",
+    "name": "test",
     "license": "mit",
     "languages": [
         "foooooooo=amh"
-    ]
+    ],
+    "platformSupport": {
+        "windows": "full",
+        "macos": "full",
+        "desktopWeb": "full",
+        "mobileWeb": "full",
+        "ios": "full",
+        "android": "full"
+    },
+    "lastModifiedDate": "2017-09-06"
 }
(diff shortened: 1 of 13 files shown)


developer/src/test/auto/keyboard-info/test-invalid-language-code.keyboard_info
```diff
@@ -1,6 +1,17 @@
 {
+    "id": "test",
```
darcywong00, line 2: Should this id be `test_invalid_language_code`? (and use underscore instead of dashes)

mcdurdin: The id doesn't matter for this test.

developer/src/test/auto/keyboard-info/test-valid.keyboard_info
```diff
@@ -1,4 +1,5 @@
 {
+    "id": "test-valid",
```
darcywong00, line 2: Should this use underscores instead of dashes?

darcywong00 (approved):
Some nit questions about id

developer/src/test/auto/keyboard-info/test-valid.keyboard_info
mcdurdin: ID doesn't matter for this test.

Issue #9436, opened 2023-08-08
bug(windows): Text is appearing outside of QR code box after changing the UI language for some Keyboards


bharanidharanj:
### Describe the bug

I noticed that the text goes outside of QR code box after changing the UI for some keyboards in Keyman Configuration dialog. 

### Reproduce the bug

1. Install Keyman 17.0.156-alpha build. 
2. Open Keyman Configuration dialog. 
3. Install a keyboard. (eg., EuroLatin(SIL) 
4. Open the Keyboard Layout View. 
5. Change the UI into Kannada. 
6. Click Share Keyboard button. 

Here, I noticed that the text is appearing outside of the QR code box. Seems to be a cosmetic issue. 

Note: I am able to reproduce this issue with some other languages like Portuguese, Russian, Mandara, Italian, Ukrainian, Deutsch (German). 

I have attached the Screenshot for reference. 

   <img width="300" src="https://github.com/keymanapp/keyman/assets/19683143/23367a56-8881-4091-92e3-81d9cfb95b16">

### Expected behavior

Text should appear inside the QR code box. 

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for iPhone and iPad
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [X] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

17.0.156-alpha

### Operating system

Windows 10 OS, Windows 11 OS (VM)

PR #9472, opened 2023-08-16
fix(windows): allow QR share box to grow to edge of default configuration parent window

rc-swag:
Fixes: #9436
Allow QR box to grow if the caption box is bigger than the English text. Limit of 130px is set if extends further than this then add a vertical scroll for the caption.

In testing no language actually causes the scroll bar to be displayed. However, using dummy text to test the overflow shows that it will handle this should the translation ever reach this length.
image

Here is just a screen shot to show the vertical scroll when there is a huge amount of extra text which I don't think we will reach but am handling that use case anyway.
image

# User Testing

# TEST_QR_CAPTION

1. Install Keyman for Windows Build artfifact
2. Open Keyman Configuration dialog. 
3. Install a keyboard. (eg., EuroLatin(SIL) 
4. Open the Keyboard Layout View. 
5. Change the UI into Kannada. 
6. Click Share Keyboard button. 
The QR share box should be large enough to fit the text caption below the QR. 
7. Test for Portuguese, Russian, Mandara, Italian, Ukrainian, Deutsch

### DIFF
--- a/windows/src/desktop/kmshell/xml/config.css
+++ b/windows/src/desktop/kmshell/xml/config.css
@@ -1136,7 +1136,7 @@ th
   position: fixed;
   display: none;
   width: 256px;
-  height: 324px;
+  height: auto;
   margin: -180px 0 0 -128px;
   left: 50%;
   top: 50%;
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.162-alpha](https://keyman.com/downloads/releases/alpha/17.0.162)

PR #9474, opened 2023-08-16
feat(developer): support building .keyboard_info without source version 🎺

mcdurdin:
Relates to #9351.

Adds support for building a .keyboard_info file without having source .keyboard_info file:

* Constructs a default source .keyboard_info in memory
* Hints if license is missing
* If LICENSE.md is present, verifies it matches the MIT license text
* Adds project option to turn on or off metadata generation. This will default to False for version 1.0 projects, and to True for version 2.0 projects. This means that the keyboard repository will need a PR to enable metadata generation for existing projects in the repository, but this is important to avoid breaking builds for existing projects that are not in the repository.
* Turns on additional c8 coverage for kmc projects
* Also fixes a file reference issue in kmpJson.options, where paths were stripped too early in the process.
* Removes a now-irrelevant unit test for keyboard_info.

TODO:
- [x] markdown to html during the keyboard_info build (currently just copies straight)
- [ ] font family considerations -- currently a generated name

@keymanapp-test-bot skip


developer/src/kmc-keyboard-info/src/index.ts
```diff
@@ -210,20 +198,8 @@ export class KeyboardInfoCompiler {
     }
     keyboard_info.packageIncludes = [...includes];
 
-    // Always overwrite source data
-    if(sources.kmpFileName) {
-      this.setField(keyboard_info, 'version', sources.kmpJsonData.info.version.description);
-    } else {
-      const m = jsFile?.match(/this\.KBVER\s*=\s*(['"])([^'"]+)(\1)/);
```
darcywong00, line 217: Do we not need to fallback to KBVER anymore?

mcdurdin: I verified that all keyboards in the repo have a package -- and going forward we won't accept keyboards that _don't_ have a package, so the version info is always readable from there.

There are still keyboards that don't have a .kmx, and those ones need us to parse the .js for other fields such as `isRTL`, so we still load the .js for those.

common/web/types/src/kpj/keyman-developer-project.ts
```diff
@@ -108,6 +108,11 @@ export class KeymanDeveloperProjectOptions {
   compilerWarningsAsErrors: boolean = false;
   warnDeprecatedCode: boolean = true;
   checkFilenameConventions: boolean = false;  // missing option defaults to False
+  /**
+   * Skip building .keyboard_info and .model_info files, for example in
+   * unit tests or for legacy keyboards
```
darcywong00, line 113: So will unit tests need to specify skipMetadataFiles=true?

mcdurdin: The ones that need it will do it. Currently, it's not causing trouble.

developer/src/kmc-keyboard-info/src/validate-mit-license.ts
```diff
@@ -0,0 +1,69 @@
...
+export function validateMITLicense(license: string) {
+  // NOTE: this function returns error messages which will not be easily
+  // localizable. Given this is only for use with the keyboards and
+  // lexical-models repositories, suggest we don't worry about this
+
+  // split input text into paragraphs, trimming whitespace
+  const text = license.split('\n').map(line => line.trim()).join('\n').split(/\n\s*\n/);
+
+  // MIT license, cleanup whitespace
+  const clauses = [
```
darcywong00, line 15: Is this sourced from https://opensource.org/license/mit/ ? (so we know where to update)
I see that has 
```
Copyright <YEAR> <COPYRIGHT HOLDER>
```

though copyright is tested separately

darcywong00: hmm, the source also doesnt' start with
```
The MIT License (MIT)
```

mcdurdin: The text of the MIT license is widely shared and static, but we could reference the link you give in the comments at the start of this file?

mcdurdin: See also https://spdx.org/licenses/MIT.html

darcywong00 (approved):
lgtm

PR #9476, opened 2023-08-17
feat(developer): convert markdown description to html 🎺

mcdurdin:
Next part of kmc work, converts the Markdown description field in the .kps into HTML during the .kmp build. This also filters through to the .keyboard_info file.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc-keyboard-info/package.json
+++ b/developer/src/kmc-keyboard-info/package.json
@@ -28,7 +28,7 @@
     "@keymanapp/kmc-package": "*"
   },
   "devDependencies": {
-    "@types/chai": "^4.1.7",
+    "@types/chai": "^4.3.5",
     "@types/mocha": "^5.2.7",
     "@types/node": "^20.4.1",
     "c8": "^7.12.0",
--- a/developer/src/kmc-keyboard-info/src/index.ts
+++ b/developer/src/kmc-keyboard-info/src/index.ts
@@ -152,7 +152,7 @@ export class KeyboardInfoCompiler {
     // description
 
     if(sources.kmpJsonData.info.description?.description) {
-      keyboard_info.description = markDownToHTML(sources.kmpJsonData.info.description?.description);
+      keyboard_info.description = sources.kmpJsonData.info.description?.description.trim();
     }
 
     // extract the language identifiers from the language metadata arrays for
@@ -438,10 +438,5 @@ export class KeyboardInfoCompiler {
       );
     }
   }
-
 }
 
-function markDownToHTML(markdown: string): string {
-  // TODO
-  return markdown;
-}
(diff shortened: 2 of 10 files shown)


developer/src/kmc-keyboard-info/package.json
```diff
@@ -28,7 +28,7 @@
     "@keymanapp/kmc-package": "*"
   },
   "devDependencies": {
-    "@types/chai": "^4.1.7",
+    "@types/chai": "^4.3.5",
```
mcdurdin, line 31: Just stops typescript complaint in vscode

developer/src/kmc-keyboard-info/src/index.ts
```diff
@@ -152,7 +152,7 @@ export class KeyboardInfoCompiler {
     // description
 
     if(sources.kmpJsonData.info.description?.description) {
-      keyboard_info.description = markDownToHTML(sources.kmpJsonData.info.description?.description);
+      keyboard_info.description = sources.kmpJsonData.info.description?.description.trim();
```
mcdurdin, line 155: markdown conversion moved to kmc-package

jahorton (approved):
I find it a _bit_ interesting that we're using `marked` instead of `pandoc`, given that we already use it for some of the other platforms.  Granted, that _would_ be during their builds and not distributed.

I'm guessing `marked` is a bit simpler to include and package for distribution, in which case... yeah, makes sense.  (And it almost-surely is - pandoc aims to cover _tons_ of different formats - not just Markdown.)

Also a mite surprised that we aren't linking any CSS to it.  Granted, that impression comes from my experiences maintaining the `pandoc`-based help rendering from the mobile apps.  It's not like we can't address it down the line if necessary, so I'm not going to hold this up on account of no styling.

mcdurdin:
> I find it a _bit_ interesting that we're using `marked` instead of `pandoc`, given that we already use it for some of the other platforms. Granted, that _would_ be during their builds and not distributed.
>
> I'm guessing `marked` is a bit simpler to include and package for distribution, in which case... yeah, makes sense. (And it almost-surely is - pandoc aims to cover _tons_ of different formats - not just Markdown.)

Yes, pandoc is a binary tool, not written in Javascript, so not as portable or easy to distribute, also it's _HUGE_ and we really don't need 99% of what it does. pandoc is not a great match for this.

> Also a mite surprised that we aren't linking any CSS to it.

CSS does not belong in the description blobs. The description blobs are essentially content, so they should be nice tidy HTML, and styling can be left to the app or website that consumes the description and displays it. Markdown is primarily intended for supporting links, structure, and minor emphasis, noting that we block embedded HTML in the description for similar reasons. (The CSS requirements for a future welcome.md may be different.)

PR #9479, opened 2023-08-17
chore(developer): cleanup kmc-keyboard-info interface 🎺

mcdurdin:
Relates to #9351.

Moves responsibility for loading .kps into kmc-keyboard-info, away from the caller, and removes other fields with kmc-keyboard-info can calculate by itself.

Have not moved project parsing into kmc-keyboard-info, because that's a bigger job, as currently that is mostly happening within kmc itself. A project for a future version I think.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #9489, opened 2023-08-18
fix(common): builder_parse broke on option parameters

mcdurdin:
If you had an option with a parameter, e.g. "--outfile c,b", then the "c,b" was being treated as a parameter for expansion, which caused it to expand to e.g. "configure,build". Oops.

@keymanapp-test-bot skip

### DIFF
--- a/resources/builder.inc.sh
+++ b/resources/builder.inc.sh
@@ -1183,7 +1183,21 @@ builder_parse() {
     fi
 
     if [[ $key =~ ^- ]]; then
+
+      # Expand short -o to --option in options lookup
+      if [[ ! -z ${_builder_options_short[$key]+x} ]]; then
+        key=${_builder_options_short[$key]}
+      fi
+
       exp+=($key)
+      if [[ ! -z ${_builder_options_var[$key]+x} ]]; then
+        shift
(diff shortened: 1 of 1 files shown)


jahorton (commented):
I see the code... but I'm not entirely clear on the exact nature of the problem or the solution.

Also... given that this is about a bug we've seen occur... might be nice to have an automated test for it.

darcywong00 (approved):
lgtm

mcdurdin:
> Also... given that this is about a bug we've seen occur... might be nice to have an automated test for it.

Is it worth the extra effort? Cost/benefit for this edge case?

mcdurdin:
> I see the code... but I'm not entirely clear on the exact nature of the problem or the solution. (Granted, I never use the shorthand syntax, so that may be part of the confusion.)

The fix is to move the test for _type_ of parameter into the expansion code (which happens in a separate pass before the general parameter parsing) so that the expansion code only expands actions and targets.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.170-alpha](https://keyman.com/downloads/releases/alpha/17.0.170)

PR #9504, opened 2023-08-23
feat(core): implement mapped set mapping 🙀

srl295:
- restructured element loading a bit
- added some utilities on transform_element for string list lookup
- added more validation of transforms in the C++ side
- added tests of utility functions

Review Hint: https://unicode-org.github.io/icu/userguide/strings/regexp.html has the API docs for the ICU regex function used.

For: #9121

@keymanapp-test-bot skip

```xml
  <variables>
    <set id="upper" value="A E I O U YY" />
    <set id="lower" value="a e i o u yy" />
  </variables>

  <transforms type="simple">
    <transformGroup>
      <transform from="([abc])`([def])" to="$2`$1"/> <!-- flip the order:   a`e -> e`a -->
      <transform from="([abc])`([def])" to="$2`$1"/> <!-- flip the order:   a`e » e`a -->
      <transform from="($[upper])`" to="$[1:lower]"/> <!-- Lowercase (and drop grave): A` » a -->
    </transformGroup>
  </transforms>
```


core/src/kmx/kmx_plus.cpp
```diff
@@ -386,11 +387,35 @@ COMP_KMXPLUS_ELEM::getElementList(KMX_DWORD elementNumber, KMX_DWORD &length) co
...
+COMP_KMXPLUS_ELEM_ELEMENT::loadAsStringList(KMX_DWORD length, const COMP_KMXPLUS_STRS &strs) const {
+  std::deque<std::u32string> list;
+  for (KMX_DWORD i = 0; i<length; i++) {
+    const auto &o = this[i];
+    std::u32string str;
+    if (o.type() == LDML_ELEM_FLAGS_TYPE_STR) {
+      // fetch the string
+      const auto str16 = strs.get(o.element);
+      str = km::kbp::kmx::u16string_to_u32string(str16);
+    } else {
```
srl295, line 404: Realized this should assert that the type is `LDML_ELEM_FLAGS_TYPE_UCHR` in case kmc is sending us something random (for a followon)

rc-swag (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.167-alpha](https://keyman.com/downloads/releases/alpha/17.0.167)

PR #9508, opened 2023-08-24
feat(resources): add pcm keyboard 🙀

srl295:
- simplify, drop normalization entries
- update test case
- also fix test runner to support escaped \u{…} format

- CLDR side is https://github.com/unicode-org/cldr/pull/3123

I'm adding this directly under resources in parallel with the CLDR PR, it will be identical when the actual pull is done at a later time.

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.167-alpha](https://keyman.com/downloads/releases/alpha/17.0.167)

Issue #4999, opened 2021-04-29
bug(linux): compile warnings to review


mcdurdin:
**Describe the bug**

https://jenkins.lsdev.sil.org/job/pipeline-keyman-packaging/job/PR-4992/3/console

```
16:29:27  keymanutil.c: In function ‘keyman_put_options_todconf’:
16:29:27  keymanutil.c:440:71: warning: passing argument 3 of ‘g_settings_set_strv’ from incompatible pointer type [-Wincompatible-pointer-types]
16:29:27    440 |         g_settings_set_strv(child_settings, KEYMAN_DCONF_OPTIONS_KEY, options);
16:29:27        |                                                                       ^~~~~~~
16:29:27        |                                                                       |
16:29:27        |                                                                       gchar ** {aka char **}
16:29:27  In file included from /usr/include/glib-2.0/gio/gio.h:132,
16:29:27                   from /usr/include/ibus-1.0/ibusproxy.h:46,
16:29:27                   from /usr/include/ibus-1.0/ibus.h:36,
16:29:27                   from keymanutil.h:55,
16:29:27                   from keymanutil.c:60:
16:29:27  /usr/include/glib-2.0/gio/gsettings.h:189:94: note: expected ‘const gchar * const*’ {aka ‘const char * const*’} but argument is of type ‘gchar **’ {aka ‘char **’}
16:29:27    189 |                                                                          const gchar *const *value);
16:29:27        |                                                                          ~~~~~~~~~~~~~~~~~~~~^~~~~
```

```
16:29:27  engine.c: In function ‘ibus_keyman_engine_process_key_event’:
16:29:27  engine.c:831:55: warning: passing argument 3 of ‘keyman_put_options_todconf’ from incompatible pointer type [-Wincompatible-pointer-types]
16:29:27    831 |                                 action_items[i].option->key, action_items[i].option->value);
16:29:27        |                                 ~~~~~~~~~~~~~~~~~~~~~~^~~~~
16:29:27        |                                                       |
16:29:27        |                                                       const km_kbp_cp * const {aka const short unsigned int * const}
16:29:27  In file included from engine.c:32:
16:29:27  keymanutil.h:109:53: note: expected ‘gchar *’ {aka ‘char *’} but argument is of type ‘const km_kbp_cp * const’ {aka ‘const short unsigned int * const’}
16:29:27    109 |                                              gchar *option_key,
16:29:27        |                                              ~~~~~~~^~~~~~~~~~
[...]

PR #9513, opened 2023-08-28
refactor(linux): Fix warning

ermshiperete:
Follow-up of #4999 (which mentions different warnings, but this change fixes warnings that we were still getting).

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.168-alpha](https://keyman.com/downloads/releases/alpha/17.0.168)

PR #9519, opened 2023-08-29
chore(linux): Remove workaround for mips64el 🏠

ermshiperete:
Now that Meson bug [#12017](https://github.com/mesonbuild/meson/issues/12017) is fixed we can revert the workaround.

@keymanapp-test-bot skip

### DIFF
--- a/linux/debian/rules
+++ b/linux/debian/rules
@@ -14,19 +14,6 @@ export LC_ALL=C.UTF-8
 
 include /usr/share/dpkg/pkg-info.mk
 
-# Workaround for https://bugs.debian.org/1041499 and
-# https://github.com/mesonbuild/meson/issues/12017
-# --native-file got added in Meson 0.49, so we check for that version here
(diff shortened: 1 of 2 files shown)


mcdurdin (approved):

linux/debian/rules
```diff
@@ -14,19 +14,6 @@ export LC_ALL=C.UTF-8
...
-
-configure_options :=
-
-ifeq ($(DEB_HOST_ARCH),$(DEB_BUILD_ARCH))
-ifeq ($(MESON_049_OR_HIGHER), 1)
-configure_options += --native-file=$(CURDIR)/debian/meson/no-exe-wrapper.ini
-endif
-endif
-
 # Unfortunately dh-python 3.20180325 (bionic) doesn't provide the virtual dh-sequence-python3
```
darcywong00, line 17: Are there intermediary versions of Meson where the workaround is still needed?

ermshiperete: There might be if a user manually installed or compiled Meson, but I don't think we have to deal with that. 

To my knowledge there are no released mainstream-distros that ship Meson 1.2.0 (the only version affected by the bug), so I think we can remove the workaround.

keyman-server:

PR #9523, opened 2023-08-29
fix(linux): Fix exception in km-kvk2ldml

### LINKED ISSUE #9522: bug(linux): Exception if output file specified in non-existing directory with km-kvk2ldml
Exception is thrown if the directory doesn't exist, e.g. `-o /tmp/test/ldml`.

ermshiperete:
This change fixes an exception that occurs if the user specifies an outputfile in a non-existing directory. Also catch several other related exceptions.

Fixes #9522.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/km-kvk2ldml
+++ b/linux/keyman-config/km-kvk2ldml
@@ -42,14 +42,25 @@ def main():
     if args.print:
         print_kvk(kvkData, args.keys)
 
-    if args.output:
-        outputfile = args.output
-    else:
-        outputfile = name + ".ldml"
+    outputfile = args.output if args.output else f'{name}.ldml'
 
-    with open(outputfile, 'wb') as ldmlfile:
-        ldml = convert_ldml(kvkData)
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

mcdurdin (commented):
Some of these errors may have underlying causes from other code of ours? Don't we call km-kvk2ldml when installing a keyboard? So catching the exception may not help resolve the root cause?

ermshiperete:
> Some of these errors may have underlying causes from other code of ours? Don't we call km-kvk2ldml when installing a keyboard? So catching the exception may not help resolve the root cause?

`km-kvk2ldml` is just the wrapper script around `keyman_config/kvk2ldml.py` which is what we call when installing a keyboard.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.169-alpha](https://keyman.com/downloads/releases/alpha/17.0.169)

PR #9527, opened 2023-08-30
docs(linux): Explicitly specify emscripten version

ermshiperete:
emscripten 3.1.44 works, but 3.1.45 fails, so this change modifies the documentation to explicitly request `3.1.44` instead of `latest`.

@keymanapp-test-bot skip

### DIFF
--- a/docs/build/linux-ubuntu.md
+++ b/docs/build/linux-ubuntu.md
@@ -69,8 +69,8 @@ You'll also have to install `emscripten` (version 3.1.44 is known to work):
 ```shell
 git clone https://github.com/emscripten-core/emsdk.git
 cd emsdk
-./emsdk install latest
-./emsdk activate latest
+./emsdk install 3.1.44
+./emsdk activate 3.1.44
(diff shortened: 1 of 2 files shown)


mcdurdin:
> emscripten 3.1.44 works, but 3.1.45 fails, so this change modifies the documentation to explicitly request `3.1.44` instead of `latest`.

Can we open a separate issue to address the 3.1.45 failure? (Why are we getting failures with patch level releases of emscripten -- do they not follow semantic versioning?)

mcdurdin (approved):

mcdurdin:
This is good anyway, because I prefer to be explicit about tooling and dependency versions anyway because it's fewer surprises down the line.

ermshiperete:
> Can we open a separate issue to address the 3.1.45 failure? (Why are we getting failures with patch level releases of emscripten -- do they not follow semantic versioning?)

#9529

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.170-alpha](https://keyman.com/downloads/releases/alpha/17.0.170)

PR #9530, opened 2023-08-31
chore(linux): Workaround: Don't run Wayland tests 🏠

ermshiperete:
This is a workaround for running the tests on Ubuntu 22.04 Jammy. For 16.0 we don't have everything in place to successfully run the tests with Wayland. Additionally the Wayland support in Keyman is still incomplete. So we disable the Wayland tests in order to be able to get successful builds of 16.0.

This is not a problem with older Ubuntu versions because they don't have Mutter 40.x, but with the new Jammy build agent on TC we're running into this issue. Things work in `master` branch.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/tests/run-tests.sh
+++ b/linux/ibus-keyman/tests/run-tests.sh
@@ -165,7 +165,9 @@ function run_tests() {
   cleanup
 }
 
-USE_WAYLAND=1
+# WORKAROUND for 16.0: disable Wayland tests
+USE_WAYLAND=0
+# USE_WAYLAND=1
 USE_X11=1
 
 while (( $# )); do


linux/ibus-keyman/tests/run-tests.sh
```diff
@@ -165,7 +165,9 @@ function run_tests() {
   cleanup
 }
 
-USE_WAYLAND=1
+# WORKAROUND for 16.0: disable Wayland tests
```
darcywong00, line 168:
```suggestion
# WORKAROUND for 16.0: disable Wayland tests - PR #9530
```

darcywong00 (approved):
lgtm

keyman-server:

PR #9535, opened 2023-09-01
feat(developer): eliminate source .model_info and .keyboard_info files 🎺

mcdurdin:
Removes all references to source .keyboard_info and .model_info files in the kmc compiler and types. Updates the model info compiler to support building purely from model sources.

Adds support for calculating `isRTL` and `license` fields in .model_info compiler.

Also:
* establishes @keymanapp/developer-utils shared module
* moves license validation into @keymanapp/developer-utils
* refactors kmc-model-info to a class and general cleanup.

@keymanapp-test-bot skip

PR #9536, opened 2023-09-01
chore: workaround npm/cli#3466 when bundling internal deps

mcdurdin:
Works around npm/cli#3466 when bundling internal dependencies using the bundleDependencies package.json property.

This change works in tandem with the npm pack/publish process -- when we run `developer/src/kmc/build.sh publish` (or `pack`), we end up with `npm version` stomping on all our package.json files, so the repo is dirty after this. We need a copy of the top-level package.json before this stomping happens, in order to get a simple map of the location of each of our internal dependencies, from the `dependencies` property (it would be possible to figure this out with a lot more parsing of our package.json files, but this is simpler).

This means, in future, we should avoid publishing our internal dependencies such as those under common/ to npm, as they serve no practical purpose there.

I will cherry-pick this commit onto my epic-package-metadata branch in order to move it forward, but felt like this belonged separately.

Note: there are not packages making use of this functionality in this branch; see #9535 for this in use in kmc-keyboard-info, kmc-model-info.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/build.sh
+++ b/developer/src/kmc/build.sh
@@ -133,6 +133,10 @@ fi
 if builder_start_action publish; then
   . "$KEYMAN_ROOT/resources/build/build-utils-ci.inc.sh"
 
+  # To ensure that we cache the top-level package.json, we must call this before
+  # the global publish
+  builder_publish_cleanup
+
   # For now, kmc will have responsibility for publishing keyman-version and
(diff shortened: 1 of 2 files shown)


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.181-alpha](https://keyman.com/downloads/releases/alpha/17.0.181)

PR #9543, opened 2023-09-05
fix(linux): Prevent exception if neither USER, LOGNAME nor SUDO_USER set

### LINKED ISSUE #9542: bug(linux): Exception in `verify_ibus_daemon` if neither USER nor SUDO_USER set
On one system we got an exception in `verify_ibus_daemon`. It looks like neither USER, LOGNAME nor SUDO_USER were set to a non-empty string.

ermshiperete:
Closes #9542.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/ibus_util.py
+++ b/linux/keyman-config/keyman_config/ibus_util.py
@@ -84,6 +84,10 @@ def restart_ibus_subp():
 def verify_ibus_daemon(start):
     realuser = os.environ.get('SUDO_USER')
     user = realuser or getpass.getuser()
+
+    if not user:
+        return
+
     try:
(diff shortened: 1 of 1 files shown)


linux/keyman-config/keyman_config/ibus_util.py
```diff
@@ -84,6 +84,10 @@ def restart_ibus_subp():
 def verify_ibus_daemon(start):
     realuser = os.environ.get('SUDO_USER')
     user = realuser or getpass.getuser()
+
+    if not user:
+        return
```
darcywong00, line 89: Do we need to log?

ermshiperete: Good idea. Done.

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.172-alpha](https://keyman.com/downloads/releases/alpha/17.0.172)

PR #9469, opened 2023-08-15
chore(web): Update keymanweb-osk.ttf to v. 3.0

darcywong00:
From WS Tech's release notes of [SymChar v3.000](https://github.com/silnrsi/font-symchar/releases/tag/v3.000)
> This is the first release of a new set of fonts that can be used by applications to show symbols that represent 'invisible' characters (such as NBSP). There are also symbols for many common keyboard keys and a few social media sources.
>
>There are two variants, released in separate packages:
>
> * SymChar - The package for general use, with all symbols in the PUA
> * SymCharK - A version that adds some basic Latin letters and symbols, specifically for use by Keyman. This incorporates all the symbols from the KeymanWeb-OSK font.

On Slack, @jvgaultney also adds:
> The only difference from KeymanWeb-OSK is that the default line metrics are a tiny bit larger (by 1%) - although the glyphs themselves are unchanged in size. This could impact CSS alignment, but only by a trivial amount.

---------

This PR takes SymCharK and updates the keymanweb-osk.ttf font used in KeyamnWeb

User Testing updated to use a custom test9469 keyboard which uses special characters on the touch keyboard keycaps and longpresses.

## User Testing
Note: the test9469 keyboard only displays the special characters on the keycaps. None of the keys have output defined.

* **TEST_WEB** - Verifies special characters render on the OSK keycaps and longpress keys
Use the "Test special characters rendering with keymanweb-osk.ttf (#9469)" test page link on a mobile form-factor device in landscape orientation.
1. Load the test page and display the OSK. The two backspace keys that look identical are `*LTRBkSp*` and the regular `*BkSp*`.
Note: `*Tab*` and `*TabLeft*` appear different sizes, but I haven't investigated to see if the previous font did that too.
3. Verify the keycaps on the default layer display
4. Longpress on the <kbd>Shift</kbd> button to display longpress options for other "layers".
5. Verify the shift longpresses render fine (Should not see `*` characters)
6. Longpress on the <kbd>Return</kbd> button to display longpress options for Return.
7. Verify the return longpresses render fine. (The one that looks like normal return is actually `*LTREnter*`)

kmw-osk shift layers
Long-presses on shift layer

* **TEST_ANDROID** - Verifies special characters render on the OSK keycaps and longpress keys
Setup - Install the PR build of KeyboardHarness on an Android 5.0 Lollipop device/emulator
1. Launch the KeyboardHarness app
2. Use the Globe key to select "test9469" keyboard. (There's several test keyboards to cycle through
[...]

PR #9547, opened 2023-09-11
chore(web): Add non-printing characters to the OSK

darcywong00:
Follows #9469

Addresses #6167 for KeymanWeb by defining PUA points per https://github.com/silnrsi/font-symchar/blob/v4.000/documentation/encoding.md. Developer to be addressed on separate PR

KeymanWeb test9469 keyboard updated to render the non-printing characters on the default layer of the OSK.

Symbols in the SymChar font, but **not** added on this PR:
* THREE-PER-EM SPACE
* FOUR-PER-EM SPACE
* SIX-PER-EM SPACE
* FIGURE SPACE
* size reference (all)

I note the SymChar font has a ZWNJ character distinct from Android and iOS, but KeymanWeb only uses the symbol that matches the platform.
image

https://github.com/keymanapp/keyman/blob/a47f7230a1aa62f7ff47c17c53dd2043ecfe2a84/web/src/engine/osk/src/keyboard-layout/oskKey.ts#L292-L300

(edited)
I note WS Tech also recently updated [SymChar](https://github.com/silnrsi/font-symchar/releases/tag/v4.000) to 4.0, which uses heaver strokes on the markers. 

Commit 301cd687e8fa8104ccf75503670bd398f94ac39c updates keymanweb-osk.ttf to 4.0.
I note the glyph for "zero width joiner" changed between 3.0 and 4.0.

--------

Symbols requested in silnrsi/font-keymanweb-osk#3 that aren't in the SymChar font:
* Mongolian vowel separator

-----

## User Testing
Based on tests from #9469

Note: the test9469 keyboard only displays the special characters on the keycaps. Most of the keys don't have output defined.

* **TEST_WEB** - Verifies special characters render on the OSK keycaps and longpress keys
Use the "Test special characters rendering with keymanweb-osk.ttf (#9469)" test page link on a mobile form-factor device in landscape orientation. Select an Android emulator (like Samsung Galaxy S20 Ultra)
1. Load the test page and display the OSK. 
3. Verify the keycaps on the default layer display match the screenshot below (updated with ZWNJGeneric on the right of left-shift)
web generic

* **TEST_ANDROID** - Verifies special characters render on the OSK keycaps and longpress keys
Setup - Install the PR build of KeyboardHarness on an Android device/emulator (newer than Android 5.0)
1. Launch the KeyboardHarness app
2. Use the Globe key to select "test9469" keyboard. (There's several test keyboards to cycle through
3. Verify the keycaps on the default layer display and match the screenshot below (updated with ZWNJGeneric on the right of left-shift)
image

* **TEST_IOS** - Verifies special characters render on the OSK keycaps and longpress keys
Setup - Install the PR build of Keyman for iPhone and iPad
1. Launch Keyman app
3. Verify the OSK is fine


web/src/engine/osk/src/specialCharacters.ts
```diff
@@ -40,9 +40,32 @@ let specialCharacters = {
   '*RTLBkSp*':        0x72,
   '*ShiftLock*':      0x73,
   '*ShiftedLock*':    0x74,
-  '*ZWNJ*':           0x75, // If this one is specified, auto-detection will kick in.
+  '*ZWNJ*':           0x79, // If this one is specified, auto-detection will kick in.
```
darcywong00, line 43: @jahorton - I updated this to match the codepoint from the font, but left in the KeymanWeb code that overrides to use the iOS/Android version.

jahorton: At the very least, as is, we'd get a desync with the following:

https://github.com/keymanapp/keyman/blob/5455aa741b8394ec0b63531b7c6fe8fe3b072c0c/developer/src/tike/xml/layoutbuilder/constants.js#L594

@mcdurdin - would this cause impacts on the Developer side?  Looking at `CSpecialText10` and `CSpecialText14`, I see something that could be interpreted as a pseudoarray in the defined strings... and if that's used to compute an "index" of sorts to reproduce the code, the change here would also cause a breakage there.

mcdurdin: I think we need to stick with 0x75 to match existing behaviour. We could add a new `'*ZWNJGeneric*': 0x79` (which matches the naming style of `*ZWNJiOS*` and `*ZWNJAndroid*`) to specifically use the generic one.

We do need to update the Developer side of this as well so the keyboard devs can pick the new names.

darcywong00: > We do need to update the Developer side of this as well...

Started in another branch
https://github.com/keymanapp/keyman/tree/chore/developer/non-printing-chars

Both of these are more back-burner tasks that I started during the planning week.

darcywong00: Reverted ZWNJ
Added ZWNJGeneric
And rebuilt the test keyboard

mcdurdin (approved):

bharanidharanj:
* **TEST_WEB (FAILED):** Tested with the given test page in chrome browser and here is my observation: 1. Changed the browser view into mobile form-factor device in landscape orientation. 2. Noticed that the key which appear right to the shift key seems to be different when compared with the given screenshot. Seems to be an issue.

bharanidharanj:
* **TEST_ANDROID (FAILED):** Tested with the attached PR build (Keyboard Harness - test9469)  on an Android 5.0 Lollipop emulator and here is my observation:  1. Noticed that some keys are missing and it does not look similar to the given screenshot.

bharanidharanj:
# Test Results

* **TEST_IOS (PASSED):** Tested with the attached PR build (Keyman 17.0.178-alpha-test-9547) in iOS 17.0 / iPhone 15 Pro Max and here is my observation: 1. Verified that the default keyboard is working fine.

darcywong00:
@bharanidharanj - I forgot to update screenshots after changing the test keyboard to use *ZWNJGeneric* (right of the left-shift key). I've updated the PR description with new screenshots.

@keymanapp-test-bot retest TEST_WEB TEST_ANDROID

bharanidharanj:
> @bharanidharanj - I forgot to update screenshots after changing the test keyboard to use _ZWNJGeneric_ (right of the left-shift key). I've updated the PR description with new screenshots.
> 
> @keymanapp-test-bot retest TEST_WEB TEST_ANDROID

@darcywong00 Okay, I will test it.

bharanidharanj:
* **TEST_WEB (PASSED):** Retested with the updated PR build and here is my observation:  Opened the test page on a mobile from-factor device in landscape orientation. Verified that the keycaps on the default layer displayed the same as it mentioned in the given screenshot. Seems to be working fine.

bharanidharanj:
# Test Results

* **TEST_ANDROID (FAILED):** Retested with the attached PR build (Keyboard Harness - test9469) on an Android 5.0 Lollipop emulator and here is my observation: 1. Noticed that some keys are missing and it does not look similar to the given screenshot.

darcywong00:
@jahorton - I ran the Chrome inspector on the Android 5.0 emulator to investigate the blank keys.
Any clues why the computation would be 0px?

**Blank Key**
blank key

**Visible Key**
visible key

web/src/engine/osk/src/specialCharacters.ts
jahorton: Since the IDs are different now...

https://github.com/keymanapp/keyman/blob/4d2dd477a655ef14bac48d01d4c7759fa312b7aa/web/src/engine/osk/src/keyboard-layout/oskKey.ts#L285-L301

Might want to add `case '*ZWNJGeneric*':` as an alternate case matching the same block (as `case '*ZWNJ*':`) of the switch-case statement.

mcdurdin: > I suppose that keeping them separate would make sense.

Yes, that's why I suggested the separate identifier.

darcywong00:
@bharanidharanj - I spun-off the blank keys on Android 5.0 as a separate issue #9650. can you retest with a newer Android emulator?

@keymanapp-test-bot retest TEST_ANDROID

bharanidharanj:
> @bharanidharanj - I spun-off the blank keys on Android 5.0 as a separate issue #9650. can you retest with a newer Android emulator?
> 
> @keymanapp-test-bot retest TEST_ANDROID

@darcywong00 Sure, I will do it.

bharanidharanj:
# Test Results

* **TEST_ANDROID (PASSED):** Tested with the updated PR build (Keyboard Harness - test9469 keyborad) in Android 9.0 / API 28 emulator and here is my observation: 1. Verified that the keycaps on the default layer display and match with the above mentioned screenshot (for android). Seems to be working fine.

keyman-server:

PR #9553, opened 2023-09-12
chore(linux): Exclude environment.sh from build

ermshiperete:
`environment.sh` is used only on Mac. This change modifies `build.sh` to generate this file only when building on Mac. It also excludes `environment.sh` from the Linux source tarball.

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.172-alpha](https://keyman.com/downloads/releases/alpha/17.0.172)

PR #9556, opened 2023-09-13
chore(linux): Ignore tests that fail on ba-jammy-64-ta 🏠

ermshiperete:
`k_031___caps_lock` and `k_032___caps_control` tests from `stable-16.0` branch fail when running on ba-jammy-64-ta for unknown reasons. They pass on the `master` branch, so instead of spending a lot of time trying to find the reason we ignore them now.

@keymanapp-test-bot skip


mcdurdin (approved):
Can we add an issue tracking this?

ermshiperete:
> Can we add an issue tracking this?

Why do we need an issue? This will disappear when we release 17.0.

mcdurdin:
My concern was around:

> unknown reasons

It'd be good to understand _why_ the tests are failing -- in case there is a deeper issue. If it's a bug that's fixed already in 17.0, then ideally we should link to the fix here.

keyman-server:

Issue #9053, opened 2023-06-21
chore(linux): failing tests due to ibus connection error?


mcdurdin:
See:
* https://build.palaso.org/buildConfiguration/KeymanLinux_TestPullRequests/390373?buildTab=log&linesState=585&logView=flowAware&focusLine=0

This is intermittent I guess but would be good to know what is causing it.

ermshiperete:
Maybe fixed by #9570. I'll keep this open for a while to see if we still get failing tests.

ermshiperete:
No, still not fixed: https://build.palaso.org/buildConfiguration/Keyman_Linux_Test_Integration/410343?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true

mcdurdin:
😭

ermshiperete:
Probably fixed by #9648. Closing this issue optimistically. If it's still not fixed we can re-open.

PR #9570, opened 2023-09-15
chore(linux): Split startup process

ermshiperete:
This change improves the startup method so that it will also work if ibus hasn't started yet. Also fixes two memory leaks.

This is in response to sporadically failing integration tests on ba-jammy-64-ta.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/main.c
+++ b/linux/ibus-keyman/src/main.c
@@ -59,29 +59,33 @@ ibus_disconnected_cb(IBusBus *unused_bus, gpointer unused_data) {
 }
 
 static void
-start_component(void) {
-  GList *engines, *p;
+add_single_keyboard(gpointer data, gpointer user_data) {
+  IBusEngineDesc *engine = IBUS_ENGINE_DESC(data);
+#if IBUS_CHECK_VERSION(1, 3, 99)
+  const gchar *engine_name = ibus_engine_desc_get_name(engine);
+#else
+  const gchar *engine_name = engine->name;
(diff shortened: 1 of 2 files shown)


linux/ibus-keyman/src/main.c
```diff
@@ -91,6 +95,22 @@ start_component(void) {
...
+
+  ibus_init();
+
+  bus = ibus_bus_new();
+
+  if (ibus_bus_is_connected(bus)) {
+    add_keyboards(bus, NULL);
+  } else {
+    g_message("Waiting for ibus-daemon to start up...");
+    g_signal_connect(bus, "connected", G_CALLBACK(add_keyboards), NULL);
```
mcdurdin, line 112: Is it necessary to disconnect this signal at any time, e.g. when shutting down?

mcdurdin (approved):
LGTM assuming my query is a negative :)

ermshiperete:
This PR might fix #9053.

linux/ibus-keyman/src/main.c
ermshiperete: Good question.
Since we're passing `NULL` for the `user_data` parameter, we don't have to explicitly disconnect. The signal handlers are automatically disconnected when `bus` gets finalized. 
See [Memory management of signal handlers](https://developer-old.gnome.org/gobject/stable/gobject-Signals.html#signal-memory-management).

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.175-alpha](https://keyman.com/downloads/releases/alpha/17.0.175)

PR #9583, opened 2023-09-18
chore(linux): Add coverage action to `ibus-keyman/build.sh`

ermshiperete:
@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/build.sh
+++ b/linux/ibus-keyman/build.sh
@@ -26,9 +26,6 @@ builder_describe \
 
 builder_parse "$@"
 
-builder_describe_internal_dependency \
-  report:engine test:engine
-
 if builder_is_debug_build; then
   MESON_TARGET=debug
   export CPPFLAGS=-DG_MESSAGES_DEBUG
--- a/linux/ibus-keyman/build.sh
+++ b/linux/ibus-keyman/build.sh
@@ -19,10 +19,15 @@ builder_describe \
   "test" \
   "install                   install artifacts" \
   "uninstall                 uninstall artifacts" \
+  "report                    create coverage report" \
   "@/core:arch" \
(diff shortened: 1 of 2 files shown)


linux/ibus-keyman/build.sh
```diff
@@ -19,11 +19,16 @@ builder_describe \
   "test" \
   "install                   install artifacts" \
   "uninstall                 uninstall artifacts" \
+  "report                    create coverage report" \
```
darcywong00, line 22: Rather than making a new action for `report`, could it be an option `--report` inside the `test` action?
(The coverage report depends on test being run, right?)

ermshiperete: Done. Seems a bit weird to call `ninja` in response to an option, but it works.

linux/ibus-keyman/build.sh
```diff
@@ -20,10 +20,15 @@ builder_describe \
...
-  "--no-integration          don't run integration tests"
+  "--no-integration          don't run integration tests" \
+  "--report                  create coverage report" \
+  "--coverage                capture test coverage"
 
 builder_parse "$@"
 
+builder_describe_internal_dependency \
+  report:engine test:engine
+
```
darcywong00, line 31:
```suggestion
```

Now that `--coverage` is an option checked inside the test action, this can go away, right?

darcywong00 (approved):
lgtm

linux/ibus-keyman/build.sh
ermshiperete: Oops. Thanks!

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.177-alpha](https://keyman.com/downloads/releases/alpha/17.0.177)

PR #9595, opened 2023-09-19
refactor(linux): Add more tests for `keymanutil.c` 🏘️

ermshiperete:
@keymanapp-test-bot skip


linux/ibus-keyman/src/keymanutil.c
```diff
@@ -416,47 +445,37 @@ keyman_put_options_todconf(gchar *package_id,
 
     // Obtain keyboard options from DConf
     gchar **options = keyman_get_options_fromdconf(package_id, keyboard_id);
```
darcywong00, line 447: Do you want to match l. 397?
```suggestion
    g_auto(GStrv) options = keyman_get_options_fromdconf(package_id, keyboard_id);
```

linux/ibus-keyman/src/test/keymanutil_tests.c
```diff
@@ -1,59 +1,139 @@
...
+_get_keyboard_details(gchar * description, gchar * license) {
+  keyboard_details* details = g_new0(keyboard_details, 1);
+  details->id               = g_strdup("tst");
+  details->description      = g_strdup(description);
+  details->license          = g_strdup(license);
+  return details;
+}
+
+add_keyboard_data*
+_get_keyboard_data() {
```
darcywong00, line 72: Ah, initially I wondered why _get_kmp_foo() and _get_keyboard_foo() weren't in kmpdetails.c, but I see these generate test stubs. Since the keyboard ID is `tst`, I suggest `_get_tst_` prefixes.

To me, that would make it a little clearer some of the info isn't coming from kmp.json. (ll 36-39)

linux/ibus-keyman/src/keymanutil.c
```diff
@@ -465,8 +484,6 @@ keyman_put_options_todconf(gchar *package_id,
     }
 
     g_object_unref(G_OBJECT(child_settings));
-    g_free(path);
-    g_free(needle);
     g_strfreev(options);
```
ermshiperete, line 487:
```suggestion
```

linux/ibus-keyman/src/test/keymanutil_tests.c
ermshiperete: I couldn't think of a name that would make it clearer, but using `tst` is a good suggestion.

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.183-alpha](https://keyman.com/downloads/releases/alpha/17.0.183)

PR #9606, opened 2023-09-21
fix(linux): Fix detection of unit tests

### LINKED ISSUE #9577: bug(linux): Ubuntu 20.04 Focal reports running user tests when starting km-config
When running km-config (17.0.x) on Ubuntu 20.04 Focal the message "Running unit tests, not reporting to Sentry" is output on the terminal. Cf #9567.

ermshiperete:
When running on Ubuntu 20.04 Focal, one of the Python dependencies loads the `unittest` module so that running `km-config` output the message that Sentry would be disabled because unit tests are running.

This change now uses the call stack to check if we're running unit tests.

Fixes #9577.

# User Testing

- The test should be run on Ubuntu 20.04 Focal with Gnome Shell and X11

**TEST_VERIFY**: start `km-config` from the command line and verify that no output is shown

### DIFF
--- a/linux/keyman-config/keyman_config/sentry_handling.py
+++ b/linux/keyman-config/keyman_config/sentry_handling.py
@@ -5,6 +5,7 @@ import logging
 import os
 import platform
 import sys
+import traceback
 from keyman_config.version import (
   __version__,
   __versionwithtag__,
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
It feels a little squirmy (especially given the slow case for the search is the production case), and I wonder how often `is_sentry_enabled` is called. Should we be caching the result; is `traceback.format_stack()` expensive?

But I guess LGTM otherwise :grin:

bharanidharanj:
# Test Results

* **TEST_VERIFY (PASSED):** Tested with the attached PR build (Keyman 17.0.178-alpha-local (package version 17.0.178-1~PR-9606-1288.1+focal1) in Ubuntu 20.04 Focal Linux OS and here is my observation: 1. Run km-config command line on the terminal window. 2. The Keyman Configuration dialog opens on the screen. 3. No message appears on the terminal window.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.180-alpha](https://keyman.com/downloads/releases/alpha/17.0.180)

PR #9607, opened 2023-09-21
feat(developer): Windows package installer compiler 🎺

### LINKED ISSUE #9538: feat(developer): support building package installers with kmc 🎺
CompilePackageInstaller needs to be supported by kmc, currently only in Delphi code. I think this is needed even though package installers are essentially deprecated in 17.0.

mcdurdin:
Fixes #9538.

Replaces kmcomp's package installer build infrastructure. While this is mostly legacy, package installers are still used in adhoc distribution scenarios, so we need to continue to support with the new kmc.

The package installer is a Windows self-extracting zip archive. The user is expected to find setup-redist.exe and keymandesktop.msi themselves, from the Keyman Developer release files. They are not included with Keyman Developer, but are available as standalone downloads from https://downloads.keyman.com/windows/.

`kmc build` has been updated to include two subcommands:
* `kmc build ldml-test-data`
* `kmc build windows-package-installer`

Both of these subcommands are a little long, but the use cases for them are fairly narrow, so I believe this is okay. `kmc build-test-data` has been removed, as `kmc build ldml-test-data` replaces it.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #9608, opened 2023-09-21
chore(linux): Check and restart background processes

ermshiperete:
Before we run a test we now check that ibus and the other processes we need are still running. If not we run test setup again.

This will likely fix #9570.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/tests/scripts/run-single-test.sh
+++ b/linux/ibus-keyman/tests/scripts/run-single-test.sh
@@ -21,7 +21,7 @@ fi
 
 function help() {
   echo "Usage:"
-  echo "  $0 [--env <envfile>] [-k] [--tap] [--surrounding-text] [--no-surrounding-text] [--wayland|--x11] [--] TEST"
+  echo "  $0 [--env <envfile>] [-k] [--tap] [--surrounding-text] [--no-surrounding-text] [--wayland|--x11] [--check <pidfile> --cleanup <file>] [--] TEST"
   echo
   echo "Arguments:"
   echo "  --help, -h, -?          Display this help"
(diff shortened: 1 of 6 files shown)


linux/ibus-keyman/tests/scripts/run-single-test.sh
```diff
@@ -34,6 +34,8 @@ function help() {
   echo "  --wayland               run tests with Wayland"
   echo "  --x11                   run tests with X11"
   echo "  --env <envfile>         Name of the file containing environment variables to use"
+  echo "  --check <pidfile>       Name of the file containing pids to check are running"
+  echo "  --cleanup <file>        Name of the file containing cleanup of processes"
```
darcywong00, line 38: Does the usage on l.24 get longer with these additional args? :)

linux/ibus-keyman/tests/scripts/test-helper.inc.sh
```diff
@@ -176,25 +181,31 @@ function _setup_display_server() {
...
     echo "export DISPLAY=\"$DISPLAY\"" >> "$ENV_FILE"
   fi
 }
 
 function _setup_schema_and_gsettings() {
-  local ENV_FILE PID_FILE
+  local ENV_FILE CLEANUP_FILE
   ENV_FILE=$1
-  PID_FILE=$2
+  CLEANUP_FILE=$2
```
darcywong00, line 208: CLEANUP_FILE looks unused. Maybe it can be removed?

darcywong00 (approved):
lgtm

linux/ibus-keyman/tests/scripts/run-single-test.sh
ermshiperete: Thanks!

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.181-alpha](https://keyman.com/downloads/releases/alpha/17.0.181)

Issue #9403, opened 2023-08-02
feat(developer): ldml scancodes 🙀 


srl295:
Background: see [CLDR-16604](https://unicode-org.atlassian.net/browse/CLDR-16604) and https://github.com/unicode-org/cldr/pull/2929 and https://github.com/unicode-org/cldr/pull/3276

- The proposal is for the CLDR repo to not allow embedded scanCodes in a keyboard file even though the spec will support it 
- kmc should do likewise
- also we should have unit tests that sanity check that the _stock_ CLDR scancodes match the keyman expectations. (won't change between cldr drops)

mcdurdin:
warn or error? I think it's a bad idea for kmc to not accept valid ldml .xml files, isn't it?

srl295:
> warn or error? I think it's a bad idea for kmc to not accept valid ldml .xml files, isn't it?

For kmc i think these should be errs (both the kmc and the unit test).  We wouldn't expect such a keyboard to work properly in keyman.

mcdurdin:
It bothers me that we are building the spec in such a way that keyboards generated under it cease to be cross-platform (because if Keyman can't work it then other platforms will also be in the same boat). It seems like a backwards step to me and leads inevitably to -windows -mac -linux -... variants of keyboards, and then even worse, possibly to -windows-102, -windows-101 variants etc. Which benefits nobody.

srl295:
> It bothers me that we are building the spec in such a way that keyboards generated under it cease to be cross-platform (because if Keyman can't work it then other platforms will also be in the same boat). It seems like a backwards step to me and leads inevitably to -windows -mac -linux -... variants of keyboards, and then even worse, possibly to -windows-102, -windows-101 variants etc. Which benefits nobody.

help me out here… how are they not cross-platform?

srl295:
My reading is that Keyman, like CLDR's repo, would only support keyboards defined in terms of a few standard layouts.  If you had a [triple keyboard](https://www.youtube.com/watch?v=AEkzWP6lTBo) with 306 keys on it, you would need to wait until that layout was added to the standard list.  You could use the triple keyboard (or something else) but you would not be able to define keys not part of us/iso/abnt2/jis.

But yet, platforms are allowed to have customized layouts in their own implementations.

mcdurdin:
OK, so what you are saying is, the spec allows for esoteric scan code layouts which we won't support in Keyman. I guess I can live with that.

mcdurdin:
[...]

Issue #7135, opened 2022-08-26
feat(core): ldml vkey support 🙀


srl295:
- validate vkey table
- apply vkey lookup when hardware (non-touch) event
- revisit missing-vkey-as-error #9236 ?

#5015

srl295:
- keyman core doesn't currently emit vkeys
- future feature

srl295:
closed the PR For now, this will need to be a future feature

srl295:
@mcdurdin close this 'now' that vkeys are dropped? Leaving a pointer here in case we need to get the code back.

mcdurdin:
Yes will close.

PR #9615, opened 2023-09-22
feat(developer): ldml scan codes support 🙀 

srl295:
- pivot:  this now dynamically generates the vkey map based on CLDR scan codes
- add infrastructure to detect imported items
- warn (hint) if user adds a custom scancode
- also includes an LDML resource update  - including removal of the `vkeys`, see #7135 
- also minor, use source maps when launching `kmc` from tests.

Fixes: #9403 
@keymanapp-test-bot skip


common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -174,13 +187,13 @@ export const ISOVirtualKeyMap: KeyMap = [
 
 export const JISVirtualKeyMap: KeyMap = [
   // [Hankaku/Zenkaku] 1 2 3 4 5 6 7 8 9 0 - ^ ¥ [bksp]
-  [ k.K_1, k.K_2, k.K_3, k.K_4, k.K_5, k.K_6, k.K_7, k.K_8, k.K_9, k.K_0, k.K_HYPHEN, k.K_EQUAL, k.K_BKSLASH /* YEN */ ],
+  [ k.K_BKQUOTE, k.K_1, k.K_2, k.K_3, k.K_4, k.K_5, k.K_6, k.K_7, k.K_8, k.K_9, k.K_0, k.K_HYPHEN, k.K_EQUAL, k.K_BKSLASH /* YEN */ ],
   // [tab] Q W E R T Y U I O P @«`» [ [enter]
```
mcdurdin, line 191:
```suggestion
  // [tab] Q W E R T Y U I O P @ [ [enter]
```

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -174,13 +187,13 @@ export const ISOVirtualKeyMap: KeyMap = [
...
   [ k.K_Q, k.K_W, k.K_E, k.K_R, k.K_T, k.K_Y, k.K_U, k.K_I, k.K_O, k.K_P, k.K_BKQUOTE, k.K_LBRKT  ],
   // [caps] A S D F G H J K L ; : ] [enter]
   [ k.K_A, k.K_S, k.K_D, k.K_F, k.K_G, k.K_H, k.K_J, k.K_K, k.K_L, k.K_COLON, k.K_QUOTE, k.K_RBRKT ],
   // [shift] Z X C V B N M , . / _ [shift]
-  [ k.K_Z, k.K_X, k.K_C, k.K_V, k.K_B, k.K_N, k.K_M, k.K_COMMA, k.K_PERIOD, k.K_SLASH, k.K_oE2 /* ろ */ ],
+  [ k.K_oE2, k.K_Z, k.K_X, k.K_C, k.K_V, k.K_B, k.K_N, k.K_M, k.K_COMMA, k.K_PERIOD, k.K_SLASH, k.k_oC1 /* ろ */ ],
```
mcdurdin, line 196: My JIS keyboard doesn't have a key between Shift and Z:



FWIW, in Windows, the virtual key code at far-right (scan code 0x73) is `K_oE2` when using Japanese base keyboard, or `K_oC1` when using US base keyboard. Yay VK!

We should probably use `K_oC1` internally for scan code 0x73 in Keyman Engine for Windows, which we need to track as a separate issue. See also the KS discussion above.

common/web/types/src/ldml-keyboard/ldml-keyboard-xml.ts
```diff
@@ -193,3 +194,36 @@ export interface LKDisplays {
...
+};
+
+/**
+ * Utilities for determining the import status of items
+ */
+export class ImportStatus {
+  /** item came in via implied (spec based) import, such as keys-Latn-implied.xml */
+  static impliedImport = Symbol.for('@keymanapp:implied_import');
+  /** item came in via import */
+  static import = Symbol.for('@keymanapp:import');
```
mcdurdin, line 218: This introduces a new namespace that we need to be careful with -- the `@keymanapp` Symbol namespace. Conceptually I think it is fine but I am just a little wary about future trouble with it. Might be good to use a longer prefix, for disambiguation e.g.:

```suggestion
  static impliedImport = Symbol.for('@keymanapp:common/ldml/implied_import');
  /** item came in via import */
  static import = Symbol.for('@keymanapp:common/ldml/import');
```

Also, just verifying is this just a way of tagging arbitrary structures with an "imported" flag? And you wanted to use this to avoid potential stumbling blocks in the way we use those structures, because it won't be an interable key on an object?

developer/src/kmc-ldml/src/compiler/messages.ts
```diff
@@ -149,5 +149,9 @@ export class CompilerMessages {
   m(this.ERROR_DisplayNeedsToOrId, `display ${CompilerMessages.toOrId(o)} needs to= or id=, but not both`);
   static ERROR_DisplayNeedsToOrId = SevError | 0x0022;
 
+  static Hint_UnsupportedCustomForm = (o:{id: string}) =>
+  m(this.HINT_UnsupportedCustomForm, `Custom <form id="${o.id}"> element is not supported.`);
+  static HINT_UnsupportedCustomForm = SevHint | 0x0023;
```
mcdurdin, line 154: Should this be a hint or a warning? Does the custom form cause breaking issues down the line?

developer/src/kmc-ldml/test/fixtures/sections/layr/hint-custom-form.xml
```diff
@@ -0,0 +1,25 @@
...
+    <key id="grave" to="`" />
+    <key id="one" to="1" />
+  </keys>
+  <forms>
+    <!-- warning, custom scancodes -->
+    <form id="us">
+      <scanCodes codes="01 02 03 04" />
+    </form>
+  </forms>
+  <layers form="us"> <!-- INVALID: 'holographic' is not a valid form. -->
```
mcdurdin, line 19: I don't understand the comment here

mcdurdin (changes requested):
Not quite ready to approve this, sorry!

common/web/types/src/ldml-keyboard/ldml-keyboard-xml.ts
srl295: On reflection, I think this doesn't need a named symbol at all. this symbol should be used only within one run of kmc.  it doesn't need to be independently detected. I will drop the namespace.

yes, it's just a way to tag the structure as imported - it's not iterable on the object, it doesn't affect the API at all at any level. It's metadata answering the question "how did this data object come in?"

developer/src/kmc-ldml/test/fixtures/sections/layr/hint-custom-form.xml
srl295: copy and paste, fixed

developer/src/kmc-ldml/src/compiler/messages.ts
srl295: maybe a warning is better than a hint. Basically, it's not processed by kmc.

common/web/types/src/consts/virtual-key-constants.ts
srl295: OK, mine didn't have one there either. Looks like a copy and paste. thanks

srl295: This is broken in the CLDR file, so opened a linked PR there.

resources/standards-data/ldml-keyboards/techpreview/import/scanCodes-implied.xml
srl295, line 1: (This is a file comment. Somehow, I haven't seen that feature before.)

i'm updating this prospectively from the parallel CLDR PR.

srl295:
@mcdurdin ptal

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -159,6 +159,19 @@ export const USVirtualKeyMap: KeyMap = [
   [ k.K_SPACE ],
 ];
 
+export const KSVirtualKeyMap: KeyMap = [
+  // ` 1 2 3 4 5 6 7 8 9 0 - = [bksp]
+  [ k.K_BKQUOTE, k.K_1, k.K_2, k.K_3, k.K_4, k.K_5, k.K_6, k.K_7, k.K_8, k.K_9, k.K_0, k.K_HYPHEN, k.K_EQUAL, k.K_BKSLASH /* YEN */  ],
```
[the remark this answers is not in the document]
mcdurdin: > On a Japanese keyboard, 0x7D produces `K_BKSLASH`, and 0x2B produces `K_oC1`. That may be the safest pattern for us to match.

And then I realised that that won't work. 0x2B is `K_BKSLASH` on kbdus. So we'll need to use something else for 0x7D (`K_oC1`?) in order to differentiate, because we can't change the already assigned 0x2B. The actual code doesn't really matter all that much so long as we are consistent internally.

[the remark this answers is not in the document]
mcdurdin: So I am wondering about the advisability of these layout maps, as they form a separate source of overlapping truth -- we already have scanCodes-implied.xml. I think we should just a have single scancode:vk map somewhere, and construct layout maps from that+scanCodes-implied.xml.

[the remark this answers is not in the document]
srl295: could it be `k.K_oE2` ? which is sometimes backslash-like things in ISO and ABNT2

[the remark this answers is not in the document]
srl295: @mcdurdin  scancodes to vkeys… seems to make sense, should I just build up a table from the data we already have here? (maybe merge this first as a baseline)

On the maybe-plus side, that means we actually could support custom scancodes! Did I say that out loud?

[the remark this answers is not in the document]
mcdurdin: `K_oE2` is mapped to scan code 0x56. I don't think we should use the same VK for different scan codes.

Internally we probably could have been using scan codes directly rather than VKs. But that boat has sailed.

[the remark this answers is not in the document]
srl295: okay.  I'll keep looking.

it sounds like there should also be a test,  a sort of virtual-`@mcdurdin`, which verifies 1:1 mapping from scan code to vkey here.

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -159,6 +159,19 @@ export const USVirtualKeyMap: KeyMap = [
   [ k.K_SPACE ],
 ];
 
+export const KSVirtualKeyMap: KeyMap = [
+  // ` 1 2 3 4 5 6 7 8 9 0 - = YEN [bksp]
+  [ k.K_BKQUOTE, k.K_1, k.K_2, k.K_3, k.K_4, k.K_5, k.K_6, k.K_7, k.K_8, k.K_9, k.K_0, k.K_HYPHEN, k.K_EQUAL, k.K_oE2 /* YEN */  ],
+  // [tab] Q W E R T Y U I O P [ ] \
+  [ k.K_Q, k.K_W, k.K_E, k.K_R, k.K_T, k.K_Y, k.K_U, k.K_I, k.K_O, k.K_P, k.K_LBRKT, k.K_RBRKT, k.k_oC1 ],
```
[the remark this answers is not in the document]
srl295: Given the upstream PR. 

Think I may attempt the dynamic approach. Or at least generated (for perspicuity).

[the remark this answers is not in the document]
mcdurdin: Dynamic sounds like a great idea. More front-end work but longer term much cleaner.

srl295:
i'm temporarily skipping the scancode test— to show that the LDML updates are otherwise passing.

srl295:
temporary file output => https://gist.github.com/srl295/48685c96bb23ee19ce1f14428ed5d91c

```json
 {
  "scan": "0A",
  "vname": "K_9",
  "vcode": 57
 },
 {
  "scan": "0B",
  "vname": "K_0",
  "vcode": 48
 },
 {
  "scan": "0C",
  "vname": "K_HYPHEN",
  "vcode": 189
 },
```

mcdurdin:
That works, a bit verbose, but certainly easy to work with. It will be a separate generated file?

Evil alternative:

```javascript
const scancodes = [
  ["0A","K_9",57],
  ["0B","K_0",48],
  ["0C","K_HYPHEN",189]
].map(x => ({scan:x[0], vname:x[1], vcode:x[2]}));
```

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -174,13 +187,13 @@ export const ISOVirtualKeyMap: KeyMap = [
 
 export const JISVirtualKeyMap: KeyMap = [
```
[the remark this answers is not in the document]
srl295: ptal at the new approach…

common/web/types/src/ldml-keyboard/ldml-keyboard-xml.ts
srl295: OK here?

common/web/types/src/consts/virtual-key-constants.ts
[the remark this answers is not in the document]
srl295: that's now given by definition…

srl295:
@mcdurdin PTAL at progress but TODO:
- [ ] unit test: verify that scan code map is 1:1
- [ ] kmc: error if CLDR scancode results in no vkey

Is there any other checking needed?

srl295:
```
  0x56: k.K_oE2, // << Dup
  0x73: k.k_oC1,
  0x7D: k.K_oE2, // << Dup
```

except we already have dups.

srl295:
Okay, PTAL.

- see if the scancode map needs updating. But it's in one place! It was already not 1:1

mcdurdin:
> ```
>   0x56: k.K_oE2, // << Dup
>   0x73: k.k_oC1,
>   0x7D: k.K_oE2, // << Dup
> ```
> 
> except we already have dups.

This would only be a problem if both `0x56` and `0x7D` scan codes appear on the same hardware keyboard, which they do not: `0x56` is on ISO and ABNT2, and `0x7D` is on JIS.

common/web/types/src/consts/virtual-key-constants.ts
```diff
@@ -144,72 +144,92 @@ export const USVirtualKeyCodes = {
...
-  // [caps] A S D F G H J K L ç ~ ] [enter]
-  [ k.K_A, k.K_S, k.K_D, k.K_F, k.K_G, k.K_H, k.K_J, k.K_K, k.K_L, k.K_COLON, k.K_QUOTE, k.K_BKSLASH ],
-  // [shift] \ Z X C V B N M , . ; / [shift]
-  [ k.K_oE2, k.K_Z, k.K_X, k.K_C, k.K_V, k.K_B, k.K_N, k.K_M, k.K_COMMA, k.K_PERIOD, k.K_SLASH, k.k_oC1 /* ABNT2 */ ],
-  // space
-  [ k.K_SPACE ],
-];
+  0x56: k.K_oE2, // << Dup
+  0x73: k.k_oC1,
+  0x7D: k.K_oE2, // << Dup
```
mcdurdin, line 204:
```suggestion
  0x56: k.K_oE2, // << Same as 0x7D; found on iso, abnt2
  0x73: k.k_oC1,
  0x7D: k.K_oE2, // << Same as 0x56; found on jis
```

We need to fix the casing of `k.k_oC1` to `k.K_oC1` but that can be a separate PR I guess.

developer/src/kmc-ldml/test/fixtures/basic.txt
mcdurdin, line 1: This is a little painful to edit but it's still far better than a hex blob -- I think it is worth the pain!

mcdurdin (approved):
LGTM

srl295:
> > ```
> >   0x56: k.K_oE2, // << Dup
> >   0x73: k.k_oC1,
> >   0x7D: k.K_oE2, // << Dup
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > except we already have dups.
> 
> 
> This would only be a problem if both `0x56` and `0x7D` scan codes appear on the same hardware keyboard, which they do not: `0x56` is on ISO and ABNT2, and `0x7D` is on JIS.

Maybe I should exclude this specific dup.

common/web/types/src/consts/virtual-key-constants.ts
srl295: someother test failure so i'll pull that in…

mcdurdin:
> Maybe I should exclude this specific dup.

Sounds sensible -- then we can track if future dups appear.

common/web/types/test/ldml-keyboard/test-ldml-keyboard-xml-reader.ts
```diff
@@ -129,3 +142,32 @@ describe('ldml keyboard xml reader tests', function () {
...
+    assert.sameDeepMembers(ckeys, [[49, undefined]]);
+    assert.equal(badScans.size, 1);
+    assert.sameDeepMembers(Array.from(badScans.values()), [0xFF]);
+  });
+  it('CLDRScanToUSVirtualKeyCodes should be 1:1', () => {
+    const vkeyToScan = new Map<number, number>();
+    // check all scancodes
+    for (let scan = 0; scan <= 0xFF; scan++) {
+      const vkey = CLDRScanToVkey(scan);
+      if (vkey !== undefined && vkey !== 226) {
```
mcdurdin, line 166: Can we use the defined constant here and add a comment to explain why we ignore K_oE2? And in fact, can we be more explicit about the test -- scan should be either 0x56 or 0x7D:

```suggestion
      if ((scan === 0x56 || scan === 0x7D) && vkey === USVirtualKeyCodes.K_oE2) {
        // We allow dup of K_oE2 / 0x56 iso,abnt2 / 0x7D - jis
        continue;
      }
      if (vkey !== undefined) {
```

srl295: sorry i missed your suggestion.  in any event, my code catches if some other scan code duplicates K_oE2

mcdurdin (approved):
LGTM.

common/web/types/test/ldml-keyboard/test-ldml-keyboard-xml-reader.ts
mcdurdin: Your changes look good.

keyman-server:

PR #9629, opened 2023-09-26
chore(linux): Dynamically choose display number

ermshiperete:
This improves reliability if some processes are still running from a previous run, which previously caused the tests to fail.

Also in the case of failed processes, list which ones are missing.

Additionally fixes `run-tests.sh` which I forgot to change in a previous PR.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

PR #9652, opened 2023-10-02
feat(developer): show an INFO message when warnings have failed a build

### LINKED ISSUE #9610: feat(developer): give info message when only warnings have failed a build
> >    "You have chosen 'treat warnings as errors' for this project. Due to one or more warnings (or errors), compile has failed."
> 
> Not a bad idea, the compiler could be more friendly here. An INFO message about warnings failing the build would help. 

_Originally posted by @MattGyverLee in https://github.com/keymanapp/keyman/issues/9589#issuecomment-1727244581_

mcdurdin:
Fixes #9610.

If a build fails when "treat warnings as errors" is on, but there are only warnings and no error messages, it can be confusing for the developer. Adds the following `INFO_WarningsHaveFailedBuild` message:

`The build failed because option "treat warnings as errors" is enabled and there are one or more warnings.`

Unit test also added to verify behaviour.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/src/commands/build.ts
+++ b/developer/src/kmc/src/commands/build.ts
@@ -129,6 +129,9 @@ async function build(filename: string, parentCallbacks: NodeCompilerCallbacks, o
         : InfrastructureMessages.Info_FileBuiltSuccessfully({filename:buildFilename})
       );
     } else {
+      if(!callbacks.hasFailureMessage(false)) { // false == check only for error+fatal messages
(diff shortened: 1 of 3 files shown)


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.186-alpha](https://keyman.com/downloads/releases/alpha/17.0.186)

PR #9656, opened 2023-10-02
feat(developer): support `store(&version) '17.0'`

### LINKED ISSUE #9541: feat(developer): Add `17.0` to `&version` statement
> store(&VERSION) '17.0' developer throws error

The `&VERSION` statement is really related to the minimum version required to support functionality in the .kmn keyboard (without reference to the visual keyboard, touch keyboard). It is usually optional -- Keyman Developer can figure out the minimum version required. So far, there's no functionality in the .kmn format that requires 17.0, so the maximum version number available is 16.0.

But I plan to add 17.0 to the VERSION statement because flick and multitap gestures are only going to be available in 17.0+. And because that information is stored in the touch keyboard, I need to check and see if Keyman Developer is able to determine it automatically.

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/issues/9432#issuecomment-1704531674_

mcdurdin:
Fixes #9541.

Also adds version 16.0 support to the kmcmplib compiler constants, and unit tests for both versions. Does not add any support for automatic version feature detection, because that forces an inverted dependency on the touch layout compilation phase (done in kmc-kmw), which would be a significant refactor. This may be something we need to support in the future.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

keyman-server:

PR #9667, opened 2023-10-03
feat(developer): warn if .kps includes a .js which is not touch-capable

### LINKED ISSUE #9324: feat(developer): Warn if a .kps includes a .js which is not touch-capable
Note: also fix the .keyboard_info metadata generation for this scenario.

mcdurdin:
Fixes #9324.

As .js keyboards in packages are only used by the iOS and Android apps, it's kinda helpful if the keyboard itself includes a touch layout. While Keyman will fall back to a desktop vk layout if a touch layout is not present, this is less than ideal -- but that's why this is setup as a hint and not a warning.

@keymanapp-test-bot skip


jahorton (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.186-alpha](https://keyman.com/downloads/releases/alpha/17.0.186)

PR #9687, opened 2023-10-04
feat(core): match any marker 🙀

srl295:
- add a regex form that matches any marker

For: #9119

@keymanapp-test-bot skip

### DIFF
--- a/common/web/types/src/ldml-keyboard/pattern-parser.ts
+++ b/common/web/types/src/ldml-keyboard/pattern-parser.ts
@@ -64,9 +64,15 @@ export class MarkerParser {
   public static readonly MAX_MARKER_INDEX = constants.marker_max_index;
   /** Max count of markers */
   public static readonly MAX_MARKER_COUNT = constants.marker_max_count;
+
+  private static anyMarkerMatch() : string {
+    const start = (`0000` + (this.MIN_MARKER_INDEX).toString(16)).slice(-4);
+    const end   = (`0000` + (this.MAX_MARKER_INDEX).toString(16)).slice(-4);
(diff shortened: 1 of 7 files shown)


common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -64,6 +64,9 @@ export class MarkerParser {
   public static readonly MAX_MARKER_INDEX = constants.marker_max_index;
   /** Max count of markers */
   public static readonly MAX_MARKER_COUNT = constants.marker_max_count;
+  /** Expression that matches any marker */
+  public static readonly ANY_MARKER_MATCH =
+    this.SENTINEL + this.MARKER_CODE + `[\\u0001-\\u${this.MAX_MARKER_INDEX.toString(16)}]`;
```
srl295, line 69: At least the end of the range is dynamic. Or is it worth importing left-pad here for the beginning?

mcdurdin: I'm not sure I understand the comments here?

mcdurdin (approved):

common/web/types/src/ldml-keyboard/pattern-parser.ts
srl295: @mcdurdin sorry, it was obscure.

I was joking about whether I should `require('left-pad')`  (of infamy https://www.theregister.com/2016/03/23/npm_left_pad_chaos/ ) in order to convert `1` into `0001`.

mcdurdin: Surely we should define `MIN_MARKER_INDEX` as 1. Then you might get your chance to left pad. But with `padStart` please 🤣

keyman-server:

PR #9688, opened 2023-10-05
feat(developer): ldml fix all remaining TODOs around markers and variables 🙀 

srl295:
- finish all of the developer LDML-TODOs around markers and variables

For: #9121
For: #9119

@keymanapp-test-bot skip


common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -513,6 +514,18 @@ export class List extends Section {
     }
     return this.allocList(strs, s.split(' ').map(unescapeString));
   }
+  /** perform string variable, marker, and unescaping */
+  allocListFromSubstitutedSpaces(s: string, sections: DependencySections): ListItem {
+    if(s === undefined || s === null) {
+      s = '';
+    }
```
mcdurdin, line 521:
```suggestion
    s = s ?? '';
```

Seems shorter :grin:

Ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -513,6 +514,18 @@ export class List extends Section {
...
+    return this.allocList(sections.strs, s.split(' ').map(s => {
+      s = sections.vars.substituteStrings(s, sections);
+      s = sections.vars.substituteMarkerString(s);
+      s = unescapeString(s);
+      return s;
+    }));
```
mcdurdin, line 527:
```suggestion
    return this.allocList(sections.strs, s.split(' ').map(t => {
      t = sections.vars.substituteStrings(t, sections);
      t = sections.vars.substituteMarkerString(t);
      t = unescapeString(t);
      return t;
    }));
```

Reusing `s` on inner scope is not ideal for readability

developer/src/kmc-ldml/src/compiler/keys.ts
```diff
@@ -130,7 +130,11 @@ export class KeysCompiler extends SectionCompiler {
 
       for (let lkflick of lkflicks.flick) {
         let flags = 0;
-        const to = sections.strs.allocAndUnescapeString(lkflick.to, true);
+        let cookedTo = lkflick.to;
+        // pull in string variables and markers
+        cookedTo = sections.vars.substituteStrings(cookedTo, sections);
+        cookedTo = sections.vars.substituteMarkerString(cookedTo);
+        const to = sections.strs.allocAndUnescapeString(cookedTo, true);
```
mcdurdin, line 137: This pattern is repeated, refactor? It is also similar in disp.ts but written nested for added confusions!

developer/src/kmc-ldml/src/compiler/keys.ts
```diff
@@ -168,24 +171,23 @@ export class KeysCompiler extends SectionCompiler {
...
-        sections.vars.substituteMarkerString(key.longPressDefault),
-      );
-      const multiTap: ListItem = sections.list.allocListFromEscapedSpaces(
-        sections.strs,
-        // TODO-LDML: markers,variables
-        key.multiTap
+      let cookedLongPressDefault = key.longPressDefault;
+      cookedLongPressDefault = sections.vars.substituteStrings(cookedLongPressDefault, sections);
+      cookedLongPressDefault = sections.vars.substituteMarkerString(cookedLongPressDefault)
+      const longPressDefault = sections.strs.allocAndUnescapeString(cookedLongPressDefault);
```
mcdurdin, line 181: Here's the dup

developer/src/kmc-ldml/src/compiler/disp.ts
```diff
@@ -59,13 +59,14 @@ export class DispCompiler extends SectionCompiler {
...
     result.baseCharacter = sections.strs.allocAndUnescapeString(this.keyboard3.displays?.displayOptions?.baseCharacter);
 
-    // TODO-LDML: substitute variables!
-
     // displays
     result.disps = this.keyboard3.displays?.display.map(display => ({
-      to: sections.strs.allocAndUnescapeString(sections.vars.substituteMarkerString(display.to)),
+      to: sections.strs.allocAndUnescapeString(
+        sections.vars.substituteMarkerString(
+          sections.vars.substituteStrings(display.to, sections))),
```
mcdurdin, line 66: This is similar to the code in keys.ts

developer/src/kmc-ldml/src/compiler/disp.ts
```diff
@@ -59,13 +59,14 @@ export class DispCompiler extends SectionCompiler {
...
       id: sections.strs.allocString(display.id), // not escaped, not substituted
-      display: sections.strs.allocAndUnescapeString(display.display),
+      display: sections.strs.allocAndUnescapeString(
+        sections.vars.substituteStrings(display.display, sections)),
```
mcdurdin, line 69: This is nearly similar

developer/src/kmc-ldml/src/compiler/keys.ts
```diff
@@ -168,24 +171,23 @@ export class KeysCompiler extends SectionCompiler {
...
+        key.multiTap,
+        sections,
       );
       const keySwitch = sections.strs.allocString(key.switch); // 'switch' is a reserved word
       const toRaw = key.to;
-      // TODO-LDML: variables
-      let toCooked = sections.vars.substituteMarkerString(toRaw);
+      let toCooked = sections.vars.substituteStrings(toRaw, sections);
+      toCooked = sections.vars.substituteMarkerString(toCooked);
       const to = sections.strs.allocAndUnescapeString(toCooked, true);
```
mcdurdin, line 191: And dupity dup dup again

mcdurdin (approved):
LGTM but would be good to refactor away the potentially gnarly string replacement snippet

srl295:
@mcdurdin trying to DRY this out a little bit…

```js
      const id = sections.strs.allocString(key.id);
      const longPress: ListItem = sections.list.allocListFromSpaces(
        key.longPress, {
          substituteStrs: true,
          markers: true,
          unescape: true,
        },
        sections);

      const longPressDefault = sections.strs.allocString(key.longPressDefault,
        {
          stringVariables: true,
          markers: true,
          unescape: true,
        },
        sections);
```

srl295:
Work in progress. tests fail.

mcdurdin:
> ```js
> ... {
>           substituteStrs: true,
>           markers: true,
>           unescape: true,
>         }, ...
> ```

I like. These objects then lend themselves to being constants defined as part of the structural definitions, which is then easy to compare against the spec.

srl295:
VERY strange

```
  1) vars
       markers
         should compile: sections/vars/markers-maximal.xml:
     TypeError: source is not iterable
      at new ListItem (file:///Users/srl295/src/keymanapp/keyman/common/web/types/src/kmx/string-list.ts:39:23)
      at ListItem.map (<anonymous>)
      at ListItem.toStringArray (file:///Users/srl295/src/keymanapp/keyman/common/web/types/src/kmx/string-list.ts:88:17)
      at Object.callback (file:///Users/srl295/src/keymanapp/keyman/developer/src/kmc-ldml/test/test-vars.ts:206:54)
      at Context.<anonymous> (file:///Users/srl295/src/keymanapp/keyman/developer/src/kmc-ldml/test/helpers/index.ts:230:18)
```

basically `this /* ListItem */ .map(v => v.toString())` seemed to end up allocating a new ListItem (see stack!), but allocated it with a number such as 2, 3, 4.   Not iterable.  This code was working before.  Anyway ListItem.toString() is only used for tests/debugging, so I just changed it into a for loop which worked fine.

srl295:
```js
export class ListItem extends Array<ListIndex> implements OrderedStringList {

… 
```

but yet, inside of ListItem.toStringArray():

```shell
> this.map(q => q);
Uncaught TypeError TypeError: source is not iterable
    at ListItem (/Users/srl295/src/keymanapp/keyman/common/web/types/src/kmx/string-list.ts:41:23)
    at eval (repl:1:6)
```

If I'm in a  member function, why is it _allocating_ a ListItem?  is it looking for a copy constructor?

this works, though

```shell
> Array.from(this.values())
(3) [ListIndex, ListIndex, ListIndex]
Array.from(this.values())[0]
á
Array.from(this.values())[1]
é
Array.from(this.values()).map(q => q)
(3) [ListIndex, ListIndex, ListIndex]
Array.from(this.values()).map(q => q.toString())
(3) ['á', 'é', 'í']
```

so somehow TypeScript's `Array` does something different when calling `map` …

srl295:
OK I think i undertstand now.

- i've subclassed `Array` , and changed its constructor
-  however, `Array.map()` probably assumes it can call `new Array(2)` to create an empty 2-element array
- … so it ends up calling `new ListItem(2)` which doesn't make sense (to ListItem)
- Perhaps I should instead have a factory static `ListItem.from(source: Array<string>, opts: StrsOptions, sections: DependencySections)` with the fancy init, and leave the array c'tors alone
- This is a classic is-a vs has-a problem,  if I HAD an array instead of being one, TS wouldn't get confused.

srl295:
I'll do another check w/ base branch and then merge this.

keyman-server:

PR #9591, opened 2023-09-19
change(android,web) Use web-based popup key longpresses

darcywong00:
Addresses part of #9573 and fixes #9499

This updates Keyman Engine for Android to match iOS using KeymanWeb for handling longpresses.
The longpress keys will be limited by the top of the banner/OSK.

### KeymanWeb changes ###
* Remove Android-specific portions of `buildEmbeddedGestureConfig` and `pendingLongpress`

### Keyman Engine for Android changes ###
* Remove native Java code for popup, preview, and longpresses keys, as it's handled in KeymanWeb.

> Note: this means banner will always be visible on Android, no option to switch off banner any more.

This portion of #9573 will be addressed on a separate PR. Android will possibly need some image banners to display when predictions are disabled.

## User Testing
**Setup** - Install the PR build of Keyman for Android on an Android device emulator. In the Keyman app, also install the IPA (SIL) keyboard

* **TEST_IPA_RENDERING** - Verifies #9499 is fixed
1. Install the PR build of Keyman for Android on an Android device/emulator
2. Open Keyman In-App and dismiss the "Get Started" menu
3. Switch to IPA (SIL) keyboard in the Keyman App
4. Click the Number (123) key
5. Long press <kbd>[</kbd> key
6. From the longpress options, select <kbd><</kbd> special character
7. Verify the `<` symbol appears on the screen.

* **TEST_ANDROID_INAPP** - Verifies longpresses and globe key work in the Keyman app
1. Install the PR build of Keyman for Android on an Android device/emulator
2. Open Keyman In-App and select sil_eurolatin keyboard
3. On the default sil_eurolatin keyboard, briefly press a top row key and release
4. Verify the key preview briefly appears, and the selected key is output
5. Longpress on a top row key and hold
6. Verify the longpress keys appear (position limited to the top of the banner)
7. Select a longpress key and release
8. Verify the selected key is output
9. Short-press the globe key
10. Verify the Keyman keyboard switches to IPA (SIL) keyboard
11. Long-press the globe key and release
12. Verify the keyboard picker menu appears.
13. Select a keyboard
14. Verify Keyman switches to the selected keyboard

* **TEST_ANDROID_SYSTEM_LONGPRESSES** - Verifies longpresses and globe key work as a system keyboard
1. Install the PR build of Keyman for Android on an Android device/emulator
2. Open Keyman In-App and select the "Get Started" menu
3. Enable Keyman as a system keyboard and set as default keyboard
[...]

PR #9696, opened 2023-10-06
fix(android): Always display HTML banner when suggestions aren't available

darcywong00:
Follows #9591 in fixing #9573

With the move to web-based popup key longpresses, Keyman for Android needs to display an ~image~ HTML banner for keyboards w/o suggestions. This allows the popup keys to then be displayed in the banner.

Change to KeymanWeb
~* On Android device, default to image banner with a black banner "theme".~

Change to Keyman Engine for Android
* Specify a fallback black/gray banner so Keyboard apps will always have a region for popup key longpresses.
* `KMManager.setHTMLBanner()` allows developers to pass an HTML blob to use for the banner.

Change to Keyman for Android
* Use the Keyman theme (with tri-color band) for both in-app and system keyboard
* Add BannerController helper to apply banner theming

Change to FV for Android
* Use the FV theme (with gold band) for system keyboard

TODO
- [X] I'm having issues getting KeymanWeb to display the specified image banner theme on first load. (In-app keyboard banner specified a white banner in MainActivity and system keyboard banner specified a Keyman banner in SystemKeyboard). Thanks to @jahorton for helping me resolve this
- [x] Apply Android tablet theme
- [x] Apply FV theme

Current limitations
~I'm using the Keyman landscape png as generated by https://elmah.io/tools/base64-image-encoder/
There's notificable distortion as the Android app doesn't do computations like the iOS image banner.
Perhaps this can all be moved to KeymanWeb?~

This is now resolved using svg within the HTML blob.

### Original Screenshots
<details>
<summary>(edited) These are all obsolete with the svg/html refactor</summary>

#### In-app keyboard
white image banner when suggestions not available
inapp white banner after toggles

#### System keyboard
Keyman banner when suggestions not available
system keyman after toggles

Keyman banner in landscape
system keyman landscape

#### KMSample2
Test app with Tamil dictionary temporarily disabled to show custom light-blue theme

image

#### FV for Android

image
</details>

## User Testing
**Setup** - Install the PR build of the corresponding Android app on an Android device/emulator of SDK 21

* **TEST_KEYMAN_PHONE** Verifies theme shows on in-app and system keyboard (when suggestion banner isn't used)
1. Install Keyman for Android
2. With the phone in Portrait orientation from the Keyman app, install a keyboard that doesn't use a dictionary (e.g. sil_ipa). Note, the shift key sometimes registers as a camera control in the emulator, so maybe search for ipa instead of sil_ipa.
3. When the sil_ipa keyboard is installed,  verify the suggestion banner is replaced with a white white banner
4. Verify you can long-press keys on top row and they appear in the image banner
5. From Keyman settings, enable Keyman as the default system keyboard.
6. Rotate the device to landscape orientation
7. Verify the Keyman banner on the keyboard fills the keyboard width.
8. Rotate the device back to portrait orientation
9. Launch a separate app (e.g. Google Chrome)
10. In Google Chrome, select a text field and switch to Keyman sil_ipa keyboard
11. Verify the sil_ipa keyboard shows a Keyman image banner
12. Verify you can long-press keys on the top row and they appear in the image banner.

flex style

* **TEST_FV** - Verifies theme on FirstVoices system keyboard (when suggestion banner isn't used)
1. Install FirstVoices for Android
2. With the phone in Portrait orientation from the FV app, install:
  * BC  Coast --> Haisla (a keyboard that doesn't use a dictionary)
  * BC  Coast --> SENCOTEN (a keyboard that does use a dictionary)
3. From the FV app, set FV as the default system keyboard
4. Launch a separate app (e.g. Google Chrome)
5. In Google Chrome, select a text field and switch to the FV Haisla keyboard
6. Verify the FV keyboard shows with a red FirstVoices banner
7. Verify you can long-press keys on the top row and they appear in the image banner.
8. Use the globe key to switch to the FV Sencoten keyboard
9. Verify the FV keyboard shows suggestions

fv flex


mcdurdin:
The screenshots above show a distorted Keyman logo. Can we get those fixed before we proceed?

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -175,6 +201,21 @@ public String toString()  {
 
   private static KMManager.SpacebarText spacebarText = KMManager.SpacebarText.LANGUAGE_KEYBOARD; // must match default given in kmwbase.ts
 
+  // Tiling colored background generated from https://elmah.io/tools/base64-image-encoder/
+  // TODO: Replace with image/css?
```
mcdurdin, line 205: Yes, let's use .png and CSS rather than embedding base64 into the code

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -175,6 +201,21 @@ public String toString()  {
...
+
+  // Green banner for testing as a contrast to the Android styling.
+  // TODO: remove this after testing.
+  public static final String KM_BANNER_THEME_GREEN =
+    "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABoAAAARCAYAAADDjbwNAAAABHNCSVQICAgIfAhkiAAAACNJREFUOI1j9PnP8J+BDoCJHpaMWjRq0ahFoxaNWjRqEQEAAG4mAmxMRRQ1AAAAAElFTkSuQmCCBT8AAcEGbxwAAAAASUVORK5CYII=";
```
mcdurdin, line 216: See TODO before merge

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -1370,15 +1411,83 @@ public static void deleteLexicalModel(Context context, int position, boolean sil
...
-    public static boolean setBannerOptions(boolean mayPredict) {
-    String url = KMString.format("setBannerOptions(%s)", mayPredict);
+  /**
+   * setBannerOptions - Update KMW whether to generate predictions.
+   *                    For now, also display banner
+   * @param mayPredict - boolean whether KMW should generate predictions
+   * @return boolean - Success
+   */
+  public static boolean setBannerOptions(boolean mayPredict) {
+    String url = KMString.format("setBannerOptions(%b)", mayPredict); // TODO?
```
mcdurdin, line 1421: TODO?

android/KMEA/app/src/main/res/drawable-land-mdpi/banner.png
mcdurdin, line 1: These images are too small for hi-res devices

android/KMEA/app/src/main/res/drawable/banner.png
mcdurdin, line 1: These images are too small for hi-res devices

oem/firstvoices/android/app/src/main/java/com/firstvoices/keyboards/SystemKeyboard.java
```diff
@@ -33,6 +33,8 @@
 import io.sentry.Sentry;
 
 public class SystemKeyboard extends InputMethodService implements OnKeyboardEventListener {
+    public static final String FV_BANNER_THEME_GOLD =
```
mcdurdin, line 36: What does the iOS FV app do for the banner?

web/src/engine/osk/src/banner/bannerView.ts
```diff
@@ -178,6 +178,18 @@ export class BannerController {
...
+  public static readonly DEFAULT_ANDROID_PHONE_OPTIONS: BannerOptions = {
+    alwaysShow: true,
+    imagePath: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABoAAAARCAIAAABM7ytaAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAAEnQAABJ0Ad5mH3gAAAAVSURBVDhPYxgFo2AUjIJRQDpgYAAABT8AAcEGbxwAAAAASUVORK5CYII="
+  }
+
+  // Default to gray image banner for Android tablet
+  public static readonly DEFAULT_ANDROID_TABLET_OPTIONS: BannerOptions = {
+    alwaysShow: true,
+    imagePath: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAB8AAAAXCAYAAADz/ZRUAAAABHNCSVQICAgIfAhkiAAAABl0RVh0U29mdHdhcmUAZ25vbWUtc2NyZWVuc2hvdO8Dvz4AAAAtdEVYdENyZWF0aW9uIFRpbWUARnJpIDA2IE9jdCAyMDIzIDAyOjUxOjQ0IFBNICswN4WPPMwAAABRSURBVEiJ7c1BCoAwDAXRaeP9b6PQs9kkbt02CEX4sx9eu86RCWQmK7k7c95ExNL37qAAA5gZZlaGAXoU4K/qbRsNfaMtXLhw4cKFCxf+Y/wBEz4WP2Yx4QAAAAAASUVORK5CYII=="
+  }
```
mcdurdin, line 191: I don't think we should be embedding PNGs as base64

web/src/engine/osk/src/banner/bannerView.ts
```diff
@@ -186,7 +198,17 @@ export class BannerController {
 
     // Initialize with the default options - any 'manually set' options come post-construction.
     // This will also automatically set the default banner in place.
+    console.log(this.hostDevice);
```
mcdurdin, line 201: debugging

mcdurdin (changes requested):
This is heading in the right direction but we need to work on the aspect ratio and resolution of the images. There's a few other bits and pieces that are more minor.

web/src/engine/osk/src/banner/bannerView.ts
darcywong00: I was hoping all the images could be handled within KeymanWeb

bharanidharanj:
# Test Results

* **TEST_KEYMAN (PASSED):** Installed Keyman 17.0.199-alpha-test-9696 build in an Android mobile (Samsung Galaxy A23 5g) and here is my observation: 1. Verified that the suggestion banner was replaced with a white banner. 2. Also, verified that when I use long-press keys on the top row, they appear in the image banner. 3. Enabled Keyman as the default system keyboard. 4. In the chrome browser, I clicked the text field and then switched to the Keyman sil_ipa keyboard. Verified that the sil_ipa keyboard showed a (stretched) Keyman image banner. 5. Also, verified that when I use long-press keys on the top row they appear in the image banner. 

..white banner

   


..long-press key options appear in the image banner

   


..Stretched Keyman image banner 

   


..long-press key options appear in the image banner

bharanidharanj:
# Test Results

* **TEST_FV (PASSED):** Installed FirstVoices build in an Android mobile (Samsung Galaxy A23 5g) and here is my observation: 1. Verified that the FirstVoices keyboard shows with a gold image banner. 2. Also, verified that when I press the long-press keys on the top row, they appear in the image banner.

..FirstVoices keyboard appears with a gold image banner

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
jahorton: The reason it's done (base64) on iOS is because the image is dynamically built; there actually isn't a single backing file for the image.  If you're just loading static images in and forwarding those to the banner... yeah, there's little reason to not go the route suggested above.

web/src/engine/osk/src/banner/bannerView.ts
jahorton: `imagePath` can take a static URL.

iOS uses the dataURL approach because it _dynamically builds_ the image at run time; the image exists first and foremost in memory.  `base64` becomes the best approach because of that, as we can then bypass saving the image to an actual file (and having to remember to delete it later).

darcywong00: So I'm not currently using the .png files (yet) on this PR. I didn't know how to scale them for all the formfactors and orientations.

Even  in the Keyman for Android app, the "banner" at the top of the app is a composition of the logo with code generating the tri-color theme.
https://github.com/keymanapp/keyman/blob/e5ef4e9ee6fabab8f1b1b992bec12da25128e95d/android/KMAPro/kMAPro/src/main/java/com/tavultesoft/kmapro/MainActivity.java#L922-L955

jahorton: > So I'm not currently using the .png files (yet) on this PR. I didn't know how to scale them for all the formfactors and orientations.

I bet!  That's a chore even for iOS, which has a far more restricted range of resolution options.  Hence why iOS does the same for _both_ banners - both the app banner _and_ the keyboard banner.  (The latter of which is rendered to an image, then passed to Web.)

mcdurdin: Why don't we use .svg? We have a Keyman logo in SVG and that would mean just 1 image (perhaps 2, one for super-low res and one for everything else, but we have that too).

keyman_logo_DETAIL (2)
keyman_logo_DETAIL

Both from Keyman Share Drive: Resources/Images/Keyman 9.0 Futago/Logo Pak (this folder could do with some tidyup, particularly we should make a well organised set of logo files in various formats!)

jahorton: Hmm.  That might work.  I'd have to double-check the banner's exact styling, but it shouldn't be hard to add a simple option to prevent stretching and/or tiling of the banner's image.

darcywong00: While @mcdurdin and I were :duck: , he brought up the idea to see if @ermshiperete would be keen on updating KeymanWeb to render SVG's on the image banner. And keeping the Keyman logo within web/ :smile:

mcdurdin: Logo needs to be a Keyman for Android asset but should be handled entirely within the js/html with no Java interaction AFAICT. Remember  that other consumers will have their own logos for the banner.

darcywong00:
Back to draft to handle .svg files for the banner

oem/firstvoices/android/app/src/main/java/com/firstvoices/keyboards/SystemKeyboard.java
darcywong00: @jahorton or @sgschantz ?

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboardWebViewClient.java
```diff
@@ -173,8 +173,7 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
       if (KMManager.currentLexicalModel != null) {
         modelPredictionPref = prefs.getBoolean(KMManager.getLanguagePredictionPreferenceKey(KMManager.currentLexicalModel.get(KMManager.KMKey_LanguageID)), true);
       }
-      kmKeyboard.setCurrentBanner((isModelActive && modelPredictionPref) ?
-        KMKeyboard.KM_BANNER_STATE_SUGGESTION : KMKeyboard.KM_BANNER_STATE_BLANK);
+      KMManager.setBannerOptions(isModelActive && modelPredictionPref);
```
darcywong00, line 176: This tells KMW whether predictions are enabled or not

darcywong00:
alright. I believe the svg in html banner themes are working now.

mcdurdin:
Are the screenshots in the OP still up-to-date? Because the logos are too big in those.

android/KMEA/app/src/main/assets/android-host.js
```diff
@@ -168,18 +205,13 @@ function enableSuggestions(model, mayPredict, mayCorrect) {
...
 
 function setBannerOptions(mayPredict) {
   keyman.core.languageProcessor.mayPredict = mayPredict;
 }
 
-function registerModel(model) {
-  //window.console.log('registerModel: ' + model);
-  keyman.addModel(model);
-}
-
```
mcdurdin, line 182: why is this change in this PR?

android/KMEA/app/src/main/assets/android-host.js
```diff
@@ -65,13 +93,21 @@ function notifyHost(event, params) {
...
+
+    if (keyman.osk) {
+      keyman.osk.bannerView.activeBannerHeight = bannerHeight;
+    }  
   }
-  // Refresh KMW OSK
+
+  // Refresh KMW's OSK
   keyman.refreshOskLayout();
+  doResetContext();
```
mcdurdin, line 110: Why do we need to reset the context when adjusting the banner?

android/KMEA/app/src/main/assets/android-host.js
```diff
@@ -82,6 +118,7 @@ function setOskHeight(h) {
     keyman.core.activeKeyboard.refreshLayouts();
   }
   keyman.refreshOskLayout();
+  doResetContext();
```
mcdurdin, line 121: Why do we need to reset the context when adjusting the keyboard height?

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
```diff
@@ -400,6 +381,9 @@ public void onConfigurationChanged(Configuration newConfig) {
 
     int bannerHeight = KMManager.getBannerHeight(context);
     int oskHeight = KMManager.getKeyboardHeight(context);
+    if (this.htmlBannerString != null && !this.htmlBannerString.isEmpty()) {
+      loadJavascript(KMString.format("setBannerHTML('%s')", this.htmlBannerString));
```
mcdurdin, line 385: Will this work with arbitrary HTML, even HTML that contains `'` characters? i.e. are they escaped?

darcywong00:
> Are the screenshots in the OP still up-to-date? Because the logos are too big in those.

The screenshots in the User Testing section are up to-to-date.

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
mcdurdin: And shouldn't this read:

```suggestion
      setHTMLBanner(this.htmlBannerString);
```

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
```diff
@@ -653,6 +630,29 @@ public boolean setKeyboard(String packageID, String keyboardID, String languageI
...
+    currentBanner = bannerType;
+  }
+
+  public String getHTMLBanner() {
+    return this.htmlBannerString;
+  }
+
+  public void setHTMLBanner(String contents) {
+    this.htmlBannerString = contents;
+    String jsString = KMString.format("setBannerHTML('%s')", contents);
```
mcdurdin, line 652: Will this work with arbitrary HTML, even HTML that contains `'` characters? i.e. are they escaped?

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -305,6 +326,10 @@ public static String getCloudDir() {
     return getResourceRoot() + KMDefault_UndefinedPackageID + File.separator;
   }
 
+  public static String getSVGDir() {
+    return getResourceRoot() + KMDefault_AssetSVG + File.separator;
+  }
```
mcdurdin, line 331: Why SVG? KMEA shouldn't know about SVG files. It should be a 'banner' folder if anything and should probably be a parameter passed into Engine rather than a hardcoded folder name

mcdurdin (changes requested):
This is definitely looking good. A few questions and things that I think we probably need to sort out before it lands 😄

mcdurdin:
> > Are the screenshots in the OP still up-to-date? Because the logos are too big in those.
> 
> The screenshots in the User Testing section are up to-to-date.

Okay, so then the banner images are definitely too large. The Keyman logo should be similar in scale to the application banner, with whitespace:

image

(Note, the positioning and scale of that top-banner logo is not ideal in my mind either -- it could be a fix pixels higher and/or larger!)

But currently the keyboard banner logo is much bigger:

image

Same applies for FV.

android/KMEA/app/src/main/assets/android-host.js
darcywong00: I originally started banner management by keeping similarity with ios-host.js

https://github.com/keymanapp/keyman/blob/8ebb7257b3dcef7344a6d9b753ce3d53fb8ca4f7/ios/engine/KMEI/KeymanEngine/resources/Keyman.bundle/Contents/Resources/ios-host.js#L323-L330

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
darcywong00: An early iteration planned to just display an svg file (before we decided to use html blobs. I think we still need a known folder name for the assets to be copied in ll.852-860.

I can change the folder name from "svg" to "banner".

mcdurdin: We should make that a configurable path -- because the app dev may want to put keyman resources under an arbitrary subpath, e.g. 3rdparty/keyman/banner/ or something like that. Can do it in the same function as setting the HTML banner -- and if it is null/"" then there are no additional assets to copy.

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
jahorton: This is setting up a WebView operation.  Don't forget about that JS queue we set up a while back; this adds an entry to that queue (when necessary).  We want the equivalent of what you've typed when within the WebView; the trick is making sure the values marshal across the WebView boundary properly in order to do so.

We'd probably be best off by using `JSON.stringify` encoding directly on `this.htmlBannerString` here, then `JSON.parse()` within the inner `setBannerHTML` function from the WebView-internal receiver.

mcdurdin: Not sure if you are responding to my suggestion here @jahorton?

> This is setting up a WebView operation. Don't forget about that JS queue we set up a while back; this adds an entry to that queue (when necessary). 

See ll.387-389. those are also all calling `loadJavascript` too

> (We _are_ on the Java / Android side here; your suggestion would be a JS-side call.)

See ll.650-654.

mcdurdin: FWIW, all of these should be setup as a single `loadJavascript()` call, appending each of the functions, so we are not doing multiple callbacks on keyboard load. That should improve performance and probably reduce flicker too.

jahorton: See #8534 - what has been noted here is a very likely contributor.  If those calls are kept separately, each call is a _separate async call_ into the WebView, and those delays could **very easily** contribute to layout reflow / flashing... especially since each affects the keyboard's layout.

android/KMEA/app/src/main/assets/android-host.js
darcywong00: I got the pattern from ios-host.js

android/KMAPro/kMAPro/src/main/assets/svg/banner.html
```diff
@@ -0,0 +1,10 @@
+<!-- Custom banner theme -->
+<div style="background: white; width: 100%; height: 100%; position: absolute; left: 0; top: 0">
+  <!--<div style="height: 1px; left: 0; position: absolute; top: 0; width: 100%; background: black"></div>-->
+  <img src="$BANNER" style="height:100%; background: white; display: block">
```
jahorton, line 4: I'm no CSS wizard, but...

```suggestion
<div style="background: white; width: 100%; height: 100%; position: absolute; left: 0; top: 0; box-sizing: border-box">
  <!--<div style="height: 1px; left: 0; position: absolute; top: 0; width: 100%; background: black"></div>-->
  <div style="width: 100%; height: 93%; position: absolute; left: 0; top: 0; box-sizing: border-box; padding: 5px>
    <img src="$BANNER" style="height:100%; background: white; display: block">
  </div>
```

Something like this should help us shrink the image.

`box-sizing: border-box`:  ensures that addition of borders and padding counts as part of the width & height's "100%" spec - they thus subtract from the space allotted to nested contents - that SVG.

Height: 93%:  so that the SVG stays entirely above the tri-color line.  100% might actually be fine, though.

Just spitballing an idea, but you may wish to replace `padding: 5px` with `padding: $PADDING` and do a replacement that's based upon the device resolution or similar, if it turns out we need different padding amounts for different devices.

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
mcdurdin: For proper escaping, here's the basic pattern:

https://github.com/keymanapp/keyman/blob/5a0c1a38a09047ff7e45a4bfc0b3b35dcd5b1378/android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java#L618-L636

Something like:

```suggestion
      loadJavascript(KMString.format("setBannerHTML(%s)",
        JSONObject.quote(this.htmlBannerString)));
```

android/KMAPro/kMAPro/src/main/assets/svg/banner.html
jahorton: Also, a small warning before you consider getting fancy: https://css-tricks.com/oh-hey-padding-percentage-is-based-on-the-parent-elements-width/

Padding percentage doesn't work ideally for us here.

android/KMEA/app/src/main/assets/android-host.js
darcywong00: reverted

android/KMEA/app/src/main/assets/android-host.js
darcywong00: reverted this and the other call too

android/KMAPro/kMAPro/src/main/assets/svg/banner.html
mcdurdin, line 1: 1. banner.html will go to android/KMAPro/kMAPro/src/main/assets/banner.html
2. keyman_banner.svg will go to android/KMAPro/kMAPro/src/main/assets/banner/keyman_banner.svg
3. the img src will be `<img src="banner/keyman_banner.svg" ...` in banner.html
4. string replacement requirement goes away
5. `KMManager.copyHTMLBannerAssets(path)` will call `copyAsset()` for each file in the path to copy it to the resources 'banner/' folder: `copyAsset(context, filename, "banner/", true)`
6. KMAPro will call `KMManager.setHTMLBannerAssetsPath(getAssetsPath() + '/banner')

android/KMAPro/kMAPro/src/main/assets/svg/banner.html
darcywong00: I'll update the user-testing screenshots with some updated styling (`display: flex;` and `margin: auto 2%`)

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
darcywong00: Handled now with JSONObject.quote()

android/KMEA/app/src/main/java/com/keyman/engine/KMKeyboard.java
darcywong00: > FWIW, all of these should be setup as a single loadJavascript() call, appending each of the functions, so we are not doing multiple callbacks on keyboard load. That should improve performance and probably reduce flicker too. (That would undo my suggestion of calling setBannerHTML but I would be okay with that)

This will be addressed on separate PR

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
darcywong00: :+1: 
As long as keyboard developers keep their banner assets in a folder, it'll now get copied inside the KMManager.copyHTMLBannerAssets() call.

darcywong00:
Lots of changes so @keymanapp-test-bot retest
Though @bharanidharanj - please wait till all the review comments are addressed. (I'll post when ready)

android/KMAPro/kMAPro/src/main/assets/banner/banner.html
mcdurdin, line 1: This could be in the parent folder, as it doesn't need to be copied with other banner assets

android/KMAPro/kMAPro/src/main/java/com/keyman/android/BannerController.java
```diff
@@ -0,0 +1,29 @@
...
+import com.keyman.engine.KMManager;
+import com.keyman.engine.util.FileUtils;
+
+import java.io.File;
+
+public class BannerController {
+
+  // Paths relative to assets folder for banner themes
+  public static final String KM_BANNER_DIR = "banner";
+  public static final String KM_BANNER_THEME_KEYMAN = KM_BANNER_DIR + "/banner.html";
```
mcdurdin, line 14: suggest this should just be:

```suggestion
  public static final String KM_BANNER_THEME_KEYMAN = "banner.html";
```

android/KMEA/app/src/main/assets/android-host.js
```diff
@@ -269,6 +304,10 @@ function hideKeyboard() {
   window.location.hash = 'hideKeyboard' + fragmentToggle;
 }
 
+function doResetContext() {
+  keyman.resetContext();
+}
+
```
mcdurdin, line 310:
```suggestion
```

oem/firstvoices/android/app/src/main/assets/banner/banner.html
```diff
@@ -0,0 +1,11 @@
+<!-- Custom banner theme -->
+<div style="background: #aa1225; width: 100%; height: 100%; position: absolute; left: 0; top: 0; display: flex">
+  <!--<div style="height: 1px; left: 0; position: absolute; top: 0; width: 100%; background: black"></div>-->
```
mcdurdin, line 3:
```suggestion
```

android/KMAPro/kMAPro/src/main/assets/banner/banner.html
```diff
@@ -0,0 +1,12 @@
+<!-- Custom banner theme -->
+<div style="background: white; width: 100%; height: 100%; position: absolute; left: 0; top: 0; display: flex">
+  <!--<div style="height: 1px; left: 0; position: absolute; top: 0; width: 100%; background: black"></div>-->
```
mcdurdin, line 3:
```suggestion
```

oem/firstvoices/android/app/src/main/assets/banner/banner.html
mcdurdin, line 1: This could go in parent folder also

oem/firstvoices/android/app/src/main/java/com/firstvoices/android/BannerController.java
```diff
@@ -0,0 +1,29 @@
...
+import com.keyman.engine.KMManager;
+import com.keyman.engine.util.FileUtils;
+
+import java.io.File;
+
+public class BannerController {
+
+  // Paths relative to assets folder for banner themes
+  public static final String FV_BANNER_DIR = "banner";
+  public static final String FV_BANNER_THEME = FV_BANNER_DIR + "/banner.html";
```
mcdurdin, line 14:
```suggestion
  public static final String FV_BANNER_THEME = "banner.html";
```

mcdurdin (approved):
Changes look good to me, a few comments but all minor!

darcywong00:
ok for @bharanidharanj to test now.

bharanidharanj:
# Test Results

* **TEST_KEYMAN_PHONE (FAILED):** Tested with the attached PR build Keyman 17.0.212-alpha-test-9696 in an Android (ver 5.0) emulator (API 21) and here is my observation: 1. Installed sil_ipa keyboard. Verified that the suggestion banner is replace with a white banner. Also, verified that the long-press keys on to row and they appeared in the image banner. 2. Changed to Landscape orientation and verified that the suggestion banner appeared as expected. 3. Set the Keyman as the system wide and default keyboard. 4. Opened a browser window. Typed in the Search bar. Able to see the Keyman banner appears in the sil_ipa keyboard. 5. Switch to default (EuroLatin (SIL) keyboard. Again revert to sil_ipa keyboard. And this time, I noticd that the Suggestion banner is gone. Seems to be an issue. FYI, I tested the same in the Android mobile device (Android ver 13) and I am able to reproduce this issue. 

..banner visible in Keyman app

   

..banner is missing in the browser window

darcywong00:
@bharanidharanj caught that the banner doesn't refresh when KeymanWeb reloads. I've added a commit to fix Keyman and FV apps.

@keymanapp-test-bot retest TEST_KEYMAN_PHONE

bharanidharanj:
> @bharanidharanj caught that the banner doesn't refresh when KeymanWeb reloads. I've added a commit to fix Keyman and FV apps.
> 
> @keymanapp-test-bot retest TEST_KEYMAN_PHONE

@darcywong00 Okay, Thanks.

bharanidharanj:
# Test Results

* **TEST_KEYMAN_PHONE (FAILED):** Retested with the updated build (Keyman 17.0.212-alpha-test-9696) in an Android 5.0 (API 21) emulator and I can reproduce this issue. (ie., Keyman banner is missing after swapping the keyboard using the globe key in the browser view)

   

* **TEST_FV (FAILED):** Tested with the attached PR build FirstVoices 17.0.212-alpha-test-9696 in an Android (ver 5.0) emulator (API 21) and here is my observation: 1. After setting FirstVoices keyboard as the system keyboard, I opened the browser window. 2. Set Haisla FirstVoices keyboard as the system keyboard and noticed that the banner is missing on the OSK.

darcywong00:
@bharanidharanj - I cannot repro the blank banners you saw. Did find a typo and pushed an update.
Did you uninstall the previous apps and delete the previous apk's when you retest?

@keymanapp-test-bot retest

bharanidharanj:
> @bharanidharanj - I cannot repro the blank banners you saw. Did find a typo and pushed an update. Did you uninstall the previous apps and delete the previous apk's when you retest?
> 
> @keymanapp-test-bot retest

@darcywong00 Yes, Darcy. I have uninstalled the previous apps and deleted the previous apk files before the retest. I am getting this error (black banner) when I opened the keyman keyboard (as a system keyboard) in the browser view. Anyhow, I will retest it. Thanks.

bharanidharanj:
# Test Results

* **TEST_KEYMAN_PHONE (PASSED):**  Retested with the attached PR build Keyman 17.0.212-alpha-test-9696 in an Android (ver 5.0) emulator (API 21) and here is my observation: 1. Installed sil_ipa keyboard. Verified that the suggestion banner is replaced with a white banner. Also, verified that the long-press keys on to row and they appeared in the image banner. 2. Changed to Landscape orientation and verified that the suggestion banner appeared as expected. 3. Set the Keyman as the system-wide and default keyboard. 4. Opened a browser window. Typed in the Search bar. Able to see the Keyman banner appears in the sil_ipa keyboard.


keymanlongpress



* **TEST_FV (PASSED):** Tested with the attached PR build FirstVoices 17.0.212-alpha-test-9696 in an Android (ver 5.0) emulator (API 21) and here is my observation: 1. After setting FirstVoices keyboard as the system keyboard. 2. Opened the browser window. Clicked the Google Search bar. 3. Able to see the FirstVoices banner on the Haisla keyboard. 4. Switch to the Sencoten keyboard. 5. Verified the Suggestion bar appears on the top of the keyboard. 

..Haisla keyboard

   


..Sencoten keyboard

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.213-alpha](https://keyman.com/downloads/releases/alpha/17.0.213)

PR #9706, opened 2023-10-06
fix(linux): Explicitly initialize GTK

### LINKED ISSUE #9705: bug(linux): Unhandled RuntimeError because of missing GTK initialization
On some systems we get a runtime error in `km-config` because GTK isn't explicitly initialized: `RuntimeError: Gtk couldn't be initialized. Use Gtk.init_check() if you want to handle this case.`

km-config at line 54:
```
        download_and_install_package(args.install)
    elif args.url:
        download_and_install_package(args.url)
    else:
        from keyman_config.view_installed import ViewInstalledWindow
        w = ViewInstalledWindow() # <--
        try:
            w.run()
        except KeyboardInterrupt:
            logging.debug('User cancelled the app')
        w.destroy()
```

Sentry [KEYMAN-LINUX-59](https://keyman.sentry.io/issues/4370267320/events/2532493db6cc424b8459f0f73612a1d5/).

ermshiperete:
Usually this is done automatically. However, if e.g. no `DISPLAY` variable is set it'll fail. When we explicitly initialize GTK it still fails, but shows a more helpful warning instead of throwing a runtime error.

Fixes #9705.

# User Testing

## Preparations

- The test should be run on any Linux platform with X11:

- Reboot

## Tests

**TEST_WARNING**: if `DISPLAY` is not set we should get a warning instead of a runtime error

- in a terminal window, run `unset DISPLAY`
- in the same terminal, start Keyman Configuration by typing `km-config`
- Verify that instead of opening the Keyman Configuration window the terminal displays a warning similar to `Gtk-WARNING **: 16:51:19.867: cannot open display:`. There should be no lines containing the text `RuntimeError`.

### DIFF
--- a/linux/keyman-config/km-config
+++ b/linux/keyman-config/km-config
@@ -2,6 +2,12 @@
 
 import argparse
 import logging
+import sys
+import gi
+
+gi.require_version('Gtk', '3.0')
+
+from gi.repository import Gtk
 
 from keyman_config import __versionwithtag__, __pkgversion__, add_standard_arguments, initialize_logging, initialize_sentry
 from keyman_config.handle_install import download_and_install_package
@@ -29,6 +35,8 @@ if __name__ == '__main__':
     add_standard_arguments(parser)
 
     args = parser.parse_args()
+
+    Gtk.init(sys.argv[1:])
 
     initialize_logging(args)
     initialize_sentry()


mcdurdin (approved):

bharanidharanj:
# Test Results

* **TEST_WARNING (FAILED):** Tested with the attached PR build (keyman 17.0.187-alpha-local (package version 17.0.187-1~PR-9706-1427.1+lunar1) in Lunar 23.04 Linux OS and here is my observation: 1. Run the above mentioned commands in the terminal window. Noticed that it opens the Keyman Configuration window on the screen. 2. The expected result is not displayed on the terminal. Seems to be an issue.

ermshiperete:
@bharanidharanj I tried and it works for me on Lunar. Can you please try again? And if it fails again run `dpkg -l keyman` in the terminal. Thanks!

@keymanapp-test-bot retest

bharanidharanj:
# Test Results

* **TEST_WARNING (FAILED):** Retested as per your suggestion in Lunar Linux OS and here is my observation: 1. I run the commands in the terminal window as per your instructions. 2. Still, I am seeing the Keyman Configuration window open after running the km-config command in the terminal window.  

..Keyman Configuration window still opens

ermshiperete:
ok, so the problem is that `DISPLAY` is still set (`echo $DISPLAY` should output an empty line). Does `DISPLAY= km-config` give the desired results (i.e. `DISPLAY=` followed by `km-config` in the same line)?

@keymanapp-test-bot retest

ermshiperete:
Test failed because of incomplete test instructions. It can't be tested with Wayland but has to be tested with X11.

bharanidharanj:
# Test Results

* **TEST_WARNING (PASSED):** Retested as per eberhard's new instructions in Ubuntu Lunar Linux OS (VM) and here is my observation: 1. Tested this in Ubuntu on Xorg (X11) Lunar Linux OS. 2. Verified that instead of opening keyman configuration dialog it shows an error message on the terminal window. 3. Also, verified that it does not show 'Runtime error' in the terminal window. Seems to be working as expected. 

..Warning Message in Terminal Window

keyman-server:

Issue #9477, opened 2023-08-17
feat(developer): package compiler should convert .md files to .htm at compile time


mcdurdin:
If any of the metadata fields ReadmeFile, LicenseFile, WelcomeFile (TBD) are listed in a package, and the file it refers to is a .md file, then it should be converted to .htm during the package compile, and the kmp.json file should refer to the .htm, and only the .htm should be included.

Note: don't forget to check for filename conflict when renaming -- if there is also a license.htm in the file, that will cause grief. (Case insensitive comparison reqd)

Question: header and footer for the .htm, as the .md will only be the content.

This relates to the kmp info.description field, which is also .md, although inline and without header and footer. This particular field also disables inline html, but the isolated .md files should allow inline html.

mcdurdin:
Note for @keymanapp/developers: please do not attempt to consume LICENSE.md when found in a .kmp, even if we have .kmp files in the wild that have LICENSE.md files. We'll only start to present license metadata from packages once we implement this feature in the package compiler -- it is not appropriate for us to add Markdown parsers to the Keyman apps.

mcdurdin:
Additional design thought: we should make it possible to specify a single custom CSS for the .md files in the packages, and include a standard one in Keyman Developer. (If we tag the body of the generated files with readme, welcome, and license class names, the single .css could cover all three files easily enough.)

mcdurdin:
Examples of documentation automation emerging from 7th Keyman Conference in Chiang Mai, Nov 2024:

```
# This is markdown

## Keyboard Layout

Base Layer

Shift Layer
<keyboard layers="shift" highlight="x y z" form="touch 300mm" />

One Picture To Bind Them All
<keyboard layers="shift ctrl alt shift+ctrl" form="iso" />

Here is a table of Gestures:
<gestures some-parameter-here />

Hear is how to type: <keys id="ធ  ោ" result="ធោ" />

The a-acute is available with: <keys id="a" longpress="a-acute" result="a'" />

The a-acute is available by typing a and holding it down and then sliding to the longpress.
```

Suggest that the `<keys>` elements can be used to run automatic tests.

Filenames should be localizable (e.g. welcome.en.md). This may require more work in the doc fields.

See also:
* #11792
* #10965
* #12143
* #2765
* #186
* #15223
* #11050

PR #9709, opened 2023-10-08
feat(developer): introduce WelcomeFile property to packages 🎺

### LINKED ISSUE #9478: feat(developer): package file format should include Options.WelcomeFile  🎺
Currently we assume the presence of welcome.htm by name in packages. In the future, this should be a formal reference in the Options object of kps/kmp.json. See also #9477 for how this allows for other file formats such as Markdown.

The Keyman apps would look first at the WelcomeFile property of kmp.json, and if that is missing, fallback to the old welcome.htm fixed filename.

Consider also localization, which applies to any of these metadata files.

mcdurdin:
Fixes #9478.

This adds a property WelcomeFile to .kps and kmp.json, which allows us to move away from the hardcoded welcome.htm filename in the future, and makes transform from Markdown (#9477) a simpler operation, and just generally starts the cleanup of the messiness of hard-coded filenames.

The package compiler will fallback to injecting welcome.htm into this property if (a) welcome.htm is present, and (b) the property does not already have a value. This doesn't buy us much because we still need to support welcome.htm for existing legacy packages, but does mean that our .kmp package metadata will be more consistent for packages compiled with 17.0+ compilers.

# User Testing

**Note:** Keyman itself on all platforms still only supports the welcome.htm filename at this time, so do not expect documentation with arbitrary filenames to be visible in the client apps.

TEST_WELCOME_FILE: In the package editor in Keyman Developer, select a welcome.htm file in the new Documentation file field. Compile the package and verify that there are no errors. Add and remove the reference to the documentation file and verify that the field behaves as you would expect.

### DIFF
--- a/common/windows/delphi/packages/PackageInfo.pas
+++ b/common/windows/delphi/packages/PackageInfo.pas
@@ -150,6 +150,7 @@ type
     FReadmeFile: TPackageContentFile;
     FGraphicFile: TPackageContentFile;
     FLicenseFile: TPackageContentFile;
+    FWelcomeFile: TPackageContentFile;
     FLoadLegacy: Boolean;
     procedure SetReadmeFile(const Value: TPackageContentFile);
     procedure SetExecuteProgram(Value: WideString);
@@ -161,6 +162,9 @@ type
       EventType: TPackageNotifyEventType; var FAllow: Boolean);
     procedure SetLicenseFile(const Value: TPackageContentFile);
     procedure LicenseRemoved(Sender: TObject;
+      EventType: TPackageNotifyEventType; var FAllow: Boolean);
+    procedure SetWelcomeFile(const Value: TPackageContentFile);
+    procedure WelcomeRemoved(Sender: TObject;
       EventType: TPackageNotifyEventType; var FAllow: Boolean);
   public
     constructor Create(APackage: TPackage); override;
@@ -178,6 +182,7 @@ type
     property ReadmeFile: TPackageContentFile read FReadmeFile write SetReadmeFile;
     property GraphicFile: TPackageContentFile read FGraphicFile write SetGraphicFile;
     property LicenseFile: TPackageContentFile read FLicenseFile write SetLicenseFile;
+    property WelcomeFile: TPackageContentFile read FWelcomeFile write SetWelcomeFile;
   end;
 
   { Package Information }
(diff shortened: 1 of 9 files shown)


common/windows/delphi/packages/PackageInfo.pas
```diff
@@ -697,16 +708,23 @@ procedure TPackageOptions.LicenseRemoved(Sender: TObject; EventType: TPackageNot
   FLicenseFile := nil;
 end;
 
+procedure TPackageOptions.WelcomeRemoved(Sender: TObject; EventType: TPackageNotifyEventType; var FAllow: Boolean);
```
rc-swag, line 711: How doe the FAllow boolean fit in with these procedures?

mcdurdin: The `FAllow` var parameter is an output parameter, allowing the function to prevent the removal of the item, which we never need to do, so it's safely ignored.

bharanidharanj:
# Test Results

* **TEST_WELCOME_FILE (PASSED):** Tested with the attached Keyman Developer 17.0.187-alpha-test-9709 in Windows 10 OS and here is my observation: 1. Opened khmer_angkor keyboard project. 2. Opened the welcome.htm file in the documentation file field. In the package editor, compiled the package and verified that there were no errors. 3. Removed the reference to the documentation file and verified that there were no errors after the compilation. Seems to be working as expected.

..with reference file in welcome.htm after compiling the package

   


..without reference file in welcome.htm after compiling the package

mcdurdin:
@bharanidharanj I think perhaps my test steps were not very clear, sorry!

There is a new field in the Package Editor, called "Documentation File", in the Details tab, which we need to test. I don't have an image handy sorry as my machine is on a different branch, but it should be at the top of the "Optional Information" section of the tab, above the "Readme file" entry.

Would you mind re-testing? Thanks so much!

@keymanapp-test-bot retest

bharanidharanj:
> @bharanidharanj I think perhaps my test steps were not very clear, sorry!
> 
> There is a new field in the Package Editor, called "Documentation File", in the Details tab, which we need to test. I don't have an image handy sorry as my machine is on a different branch, but it should be at the top of the "Optional Information" section of the tab, above the "Readme file" entry.
> 
> Would you mind re-testing? Thanks so much!
> 
> @keymanapp-test-bot retest

@mcdurdin Thank you so much for your clarification! I will retest and add my post my results.

rc-swag (approved):
lgtm

PR #9723, opened 2023-10-09
chore(core): rename keyboardprocessor.h to keyman_core_api.h

### LINKED ISSUE #9721: chore(core): rename keyboardprocessor.h to keyman_core_api.h
The original name 'keyboardprocessor.h' was selected before we chose the name Keyman Core, and now is confusing, because a keyboardprocessor is an internal component of Core.

mcdurdin:
Fixes #9721.

Also tweaks some documentation references to keyboardprocessor, but does not touch those relating to debian/control.

@keymanapp-test-bot skip


srl295 (approved):

rc-swag:
Looks fine. I don't quite understand what is required but do we do need to make changes to the api_version.in or the name of the built library to match this comment https://github.com/fcitx/fcitx5-keyman/issues/3#issuecomment-1754032283.

mcdurdin:
> but do we do need to make changes to the api_version.in or the name of the built library to match this comment [fcitx/fcitx5-keyman#3 (comment)](https://github.com/fcitx/fcitx5-keyman/issues/3#issuecomment-1754032283).

@ermshiperete thoughts?

ermshiperete:
The names in `linux/debian/libkmnkbp0-0.symbols` should not have been renamed - that's the safety net that should trigger when we modify the API in ways that require a version change. Also, that file shows that we missed one: `kbp_state_get_intermediate_context`.

We should rename the library to `libkmncore` and probably also change the version.

mcdurdin:
oops okay! Do you want to revert that specific change and put in place what should be done instead?

Also, `kbp_state_get_intermediate_context` was missed because it had the wrong name -- it was missing `km_` prefix! So that's a double-oops.

ermshiperete:
Ok, I'll take care of the .symbols file when I do #9733.

keyman-server:

PR #9725, opened 2023-10-09
chore(linux): Improve Sentry reports

ermshiperete:
- fix the calculation of the user hash so that the same user now always gets the same hash value. Previously we got a different value after restarting the app because the Python `hash` function calculates the hash based on the memory location.
- add additional details to Sentry events so that the "All Events" tab on keyman.sentry.io will show helpful values in the "device" and "os" columns instead of emptry strings.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/sentry_handling.py
+++ b/linux/keyman-config/keyman_config/sentry_handling.py
@@ -134,6 +134,15 @@ class SentryErrorHandling:
             scope.set_tag("platform", platform.platform())
             scope.set_tag("system", platform.system())
             scope.set_tag("tier", __tier__)
+            scope.set_tag("device", platform.node())
+            try:
+                os_release = platform.freedesktop_os_release()
+                scope.set_tag('os', os_release['PRETTY_NAME'])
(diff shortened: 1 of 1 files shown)


linux/keyman-config/keyman_config/sentry_handling.py
```diff
@@ -124,13 +125,24 @@ def _sentry_sdk_initialize(self):
           integrations=[sentry_logging],
           before_send=self._before_send
         )
-        set_user({'id': hash(getpass.getuser())})
+        hash = hashlib.md5()
+        hash.update(getpass.getuser().encode())
+        set_user({'id': hash.hexdigest()})
```
mcdurdin, line 130: I'm not sure about this -- md5 hashes of short strings are pretty easy to reverse with rainbow tables. It's low risk but it is a bit of an overshare.

ermshiperete: So better use sha256?

keyman-server:

Issue #9468, opened 2023-08-15
feat(core): normalization per spec for transforms/etc 🙀 


srl295:
[CLDR-16943](https://unicode-org.atlassian.net/browse/CLDR-16943) details (or will detail) SC consensus about the role of Unicode normalization. Implement it.

srl295:
- Cached Context needs to stay in NFD
- Engine needs to be able to do a normalization-insensitive compare
- Normalization should be available to KMN but not switched on now - KeyboardProcessor

mcdurdin:
NFD marker tricks:

1. `U+0061 \m{marker1} U+0300` gives NFC of `à`. So what happens to the marker in the cached context? Is cached context NFD but app context is NFC? How do we sync?
2. `U+0061 \m{marker1} U+0300 \m{marker2} U+0320` goes to NFD `U+0061 U+0320 U+0300` which breaks markers, because we no longer know where they go. Should this be a compiler error? Or is there some clever way we can work around this (e.g. normalize both app context and cached context before comparison for equality?)

srl295:
upstream CLDR normalization ticket was merged, but basically, we don't need the ticket, we need the behavior. So this is shovel ready.

srl295:
So, I'm kind of thinking at this moment about *not* trying to normalize in kmc at all. The reason is, because the core side will already need to be able to normalize not just all strings in the compiled data, but  also the context. Secondly, it gets us out of having to even consider what version of node (or browser!) kmc is running under. This could even lead to a class of non-determinism in the compiler, where two runs of kmc give different kmx depending on the node version.  By a 'leave it alone' approach, we just write into kmx exactly whatever is in the xml.

mcdurdin:
Hmm. I hear you on non-determinism. It should only impact unencoded scripts though, right? Given stability rules?

But... how will you do regex matching? It seems to me it would be difficult to normalize regexes post-construction.

srl295:
> Hmm. I hear you on non-determinism. It should only impact unencoded scripts though, right? Given stability rules?

previously unencoded, yes. 

> But... how will you do regex matching? It seems to me it would be difficult to normalize regexes post-construction.

The regex pattern is just a string and can be normalized before constructing the matcher, I'd think?  (He says, confidently)

mcdurdin:
> The regex pattern is just a string and can be normalized before constructing the matcher, I'd think? (He says, confidently)

`\uxxxx` says otherwise.

[...]

PR #9728, opened 2023-10-09
feat(core): initial normalization 🙀

srl295:
- [X] failing tests
- [x] fix the tests

For: #9468

@keymanapp-test-bot skip


mcdurdin:
Markers: `1e\u{0300}\m{problem_marker}\u{0320}` is the troublesome child.

core/subprojects/packagefiles/icu/meson.build
```diff
@@ -35,7 +35,7 @@ uconfig.set('U_ENABLE_DYLOAD', 0) # no DLL
 uconfig.set('U_CHECK_DYLOAD', 0) # no DLL
 uconfig.set('UCONFIG_NO_FILE_IO', 1)
 uconfig.set('UCONFIG_NO_LEGACY_CONVERSION', 1) # turn off file based codepage conversion
-uconfig.set('UCONFIG_NO_NORMALIZATION', 1) # TODO-LDML: may want this
+uconfig.set('UCONFIG_NO_NORMALIZATION', 0)
```
srl295, line 38: Yeah. you think we might want this at zero? 

I do wonder why my build locally (mac) passed with this at 1.

core/src/ldml/ldml_transforms.cpp
```diff
@@ -831,6 +844,99 @@ transforms::load(
   return transforms;
 }
 
+// string
+
+// TODO-LDML: copypasta -> refactor
+std::u32string &normalize_nfd(std::u32string &str, UErrorCode &status) {
+  const icu::Normalizer2 *nfd = icu::Normalizer2::getNFDInstance(status);
```
rc-swag, line 851: Could this function just `const std::u16string rstr = km::kbp::kmx::u32string_to_u16string(str);` 
then call the u16 version of normalize_nfd with rstr 
then have lines 864 to get the result back into the u32 str. ?

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,6 +259,104 @@ ldml_processor::process_event(
...
+  std::u32string new_ctxtstr_changed(ctxt_prefix.second,ctxtstr_cleanedup.end());
+
+  // drop the old suffix. Note: this mutates old_ctxtstr_changed.
+  remove_text(state, old_ctxtstr_changed, old_ctxtstr_changed.length());
+  assert(old_ctxtstr_changed.length() == 0);
+  emit_text(state, new_ctxtstr_changed);
+}
+
+void
+ldml_processor::remove_text(km_core_state *state, std::u32string &str, size_t length) {
```
rc-swag, line 326: Having this `remove_text` is good because with the planned action object change remove text will be able to more efficiently, update that action object.

rc-swag (commented):
Looking good I will review the tests. and the transform change asap next week.

core/src/ldml/ldml_transforms.cpp
srl295: yes, that's a good point. I was hoping it could do more of an in-place change, but  not for now.

rc-swag:
Is this PR getting the tests ready for #9468. 
I can see that `Steven's note: moving this into the issue.` has not been implemented yet.

core/src/ldml/ldml_transforms.cpp
srl295: PTAL

srl295:
> Is this PR getting the tests ready for #9468. I can see that `Steven's note: moving this into the issue.` has not been implemented yet.

It actually does normalization for the first time. So it is part of the implementation. I moved the notes out because all of those notes haven't been implemented, so they'd be on the issue not the PR

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,6 +259,104 @@ ldml_processor::process_event(
...
+  // so that we don't have to reconvert it inside the transform code.
+  std::u32string key_str32 = kmx::u16string_to_u32string(key_str);
+  // normalize the keystroke to NFD
+  ldml::normalize_nfd(key_str32, status);
+
+  // extract context string, in NFC
+  std::u32string old_ctxtstr_nfc;
+  (void)context_to_string(state, old_ctxtstr_nfc, false);
+  ldml::normalize_nfc(old_ctxtstr_nfc, status);
+  assert(U_SUCCESS(status));
```
mcdurdin, line 277: Do we need to `return` if `!U_SUCCESS(status)` when in release builds, because `old_ctxtstr_nfc` will be in an uncertain state?

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,6 +259,104 @@ ldml_processor::process_event(
   return KM_CORE_STATUS_OK;
 }
 
+void
+ldml_processor::process_key_string(km_core_state *state, const std::u16string &key_str) const {
+  UErrorCode status = U_ZERO_ERROR;
+  // We know that key_str is not empty per the caller.
```
mcdurdin, line 265:
```suggestion
  // We know that key_str is not empty per the caller.
  assert(!key_str.empty());
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,6 +259,104 @@ ldml_processor::process_event(
...
+
+  // we convert the keys str to UTF-32 here instead of using the emit_text() overload
+  // so that we don't have to reconvert it inside the transform code.
+  std::u32string key_str32 = kmx::u16string_to_u32string(key_str);
+  // normalize the keystroke to NFD
+  ldml::normalize_nfd(key_str32, status);
```
mcdurdin, line 271:
```suggestion
  ldml::normalize_nfd(key_str32, status);
  assert(U_SUCCESS(status));
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,6 +259,104 @@ ldml_processor::process_event(
...
+  ctxtstr.append(outputString); // TODO-LDML: should be able to do a normalization-safe append here.
+  ldml::normalize_nfd(ctxtstr, status);
+  assert(U_SUCCESS(status));
+
+  // Ok. We've done all the happy manipulations.
+
+  /** NFC and no markers */
+  std::u32string ctxtstr_cleanedup = ctxtstr;
+  // TODO-LDML: remove markers!
+  ldml::normalize_nfc(ctxtstr_cleanedup, status);
```
mcdurdin, line 311:
```suggestion
  ldml::normalize_nfc(ctxtstr_cleanedup, status);
  assert(U_SUCCESS(status));
```

mcdurdin: I see a pattern emerging with these assertions which we should potentially cleanup, e.g.

```
if(!ldml::normalize_nfc_asserted(ctxtstr_cleanedup)) {
  return;
}
```

and something like:

```
bool ldml::normalize_nfc_asserted(std::u32string &str) {
  UErrorCode status = U_ZERO_ERROR;  
  ldml::normalize_nfc(str, status);
  assert(U_SUCCESS(status));
  return U_SUCCESS(status);
}
```

The function name I've come up with could be better!

The fact that two assertions are missing in your PR highlights the need for the refactor :grin:

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,6 +259,104 @@ ldml_processor::process_event(
...
+      state->actions().push_backspace(KM_CORE_BT_CHAR, c->character);  // Cause prior char to be removed
+    } else if (type == KM_CORE_BT_MARKER) {
+      // it's a marker, 'worth' 3 uchars
+      assert(length >= 3);
+      assert(lastCtx == c->marker);  // end of list
+      length -= 3;
+      // pop off the three-part sentinel string
+      str.pop_back();
+      str.pop_back();
+      str.pop_back();
```
mcdurdin, line 348: We should consider asserting the values of the UC_SENTINEL and CODE_DEADKEY?

core/src/ldml/ldml_transforms.cpp
```diff
@@ -462,9 +462,13 @@ transform_entry::init() {
...
-    /* const */ icu::UnicodeString patustr = icu::UnicodeString(patstr.data(), (int32_t)patstr.length());
+    /* const */ icu::UnicodeString patustr_raw = icu::UnicodeString(patstr.data(), (int32_t)patstr.length());
     // add '$' to match to end
-    patustr.append(u'$');
+    patustr_raw.append(u'$');
+    icu::UnicodeString patustr;
+    const icu::Normalizer2 *nfd = icu::Normalizer2::getNFDInstance(status);
+    // NFD normalize on pattern creation
+    nfd->normalize(patustr_raw, patustr, status);
     fFromPattern.reset(icu::RegexPattern::compile(patustr, 0, status));
```
mcdurdin, line 472: Do we need to assert status after each call?
* Yes, because we want to know exactly which call unravelled.
* Yes, because the input state to the following functions will be invalid -- or does icu check value of status variable?

Same thing applies for release builds we want to do _something_ if `!U_SUCCESS(status)`?

core/src/ldml/ldml_transforms.cpp
```diff
@@ -831,6 +844,67 @@ transforms::load(
...
+// TODO-LDML: copypasta -> refactor
+std::u32string &normalize_nfd(std::u32string &str, UErrorCode &status) {
+  std::u16string rstr = km::kbp::kmx::u32string_to_u16string(str);
+
+  normalize_nfd(rstr, status);
+  if (U_SUCCESS(status)) {
+    // if failure, leave it alone
+    str = km::kbp::kmx::u16string_to_u32string(rstr);
+  }
+  return str;
```
mcdurdin, line 858: I think I'd prefer the failure cases in each function to be tested and `return str` just like the following functions do, for consistency.

Also, is there a good reason to return str as well as modify the input parameter? This seems likely to lead to unintentional wrong uses of the functions. Why not return bool success value instead (as suggested earlier in my review)?

core/src/ldml/ldml_transforms.hpp
```diff
@@ -270,6 +271,17 @@ class transforms {
...
+// string routines
+
+/** Normalize a u32string inplace. Returns a reference to the same string. */
+std::u32string &normalize_nfd(std::u32string &str, UErrorCode &status);
+/** Normalize a u16string inplace. Returns a reference to the same string. */
+std::u16string &normalize_nfd(std::u16string &str, UErrorCode &status);
+/** Normalize a u32string inplace. Returns a reference to the same string. */
+std::u32string &normalize_nfc(std::u32string &str, UErrorCode &status);
+/** Normalize a u16string inplace. Returns a reference to the same string. */
+std::u16string &normalize_nfc(std::u16string &str, UErrorCode &status);
```
mcdurdin, line 283: The comment should say what the normalization is otherwise it has less information than the function name!

```suggestion
/** Normalize a u32string inplace to NFD. Returns a reference to the same string. */
std::u32string &normalize_nfd(std::u32string &str, UErrorCode &status);
/** Normalize a u16string inplace to NFD. Returns a reference to the same string. */
std::u16string &normalize_nfd(std::u16string &str, UErrorCode &status);
/** Normalize a u32string inplace to NFC. Returns a reference to the same string. */
std::u32string &normalize_nfc(std::u32string &str, UErrorCode &status);
/** Normalize a u16string inplace to NFC. Returns a reference to the same string. */
std::u16string &normalize_nfc(std::u16string &str, UErrorCode &status);
```

mcdurdin (changes requested):
I like where this is going, but hope we can simplify the use of the normalization functions a little.

core/src/ldml/ldml_processor.cpp
srl295: It'll be in an unchanged state but yes.

core/src/ldml/ldml_processor.cpp
srl295: makes sense- thanks

core/src/ldml/ldml_transforms.hpp
srl295: originally it was all nfd!

core/src/ldml/ldml_transforms.cpp
srl295: Actually, when calling multiple ICU functions, we can actually not assert between calls, because each icu call is guaranteed to do nothing `if U_FAILURE(status)`.   All of these I would expect to fail under catastrophic situations.

srl295: yes, it needs to propagate an error.

mcdurdin: OK, but even so it's probably beneficial to know which call failed?

mcdurdin: I think logging the failure code would be helpful also. And then should probably not expose the status to any of the consumers of the functions

srl295: discussed DebugLog probably with a convenience macro to call `u_errorName(status)`

srl295:
here's what happens when `UASSERT_SUCCESS()` detects failure ( `U_CE_NOT_FOUND_ERROR` was a bogus error I set for demonstration purposes)

```
context   :  []
found vkey 65:0
- key action: 0x41/modifier 0x0
579665555	../../../src/ldml/ldml_transforms.cpp:863	normalize_nfd	U_FAILURE(U_CE_NOT_FOUND_ERROR)
----------------------------------- stderr -----------------------------------
Assertion failed: (ldml::normalize_nfd(key_str32)), function process_key_string, file ldml_processor.cpp, line 270.
```

srl295:
PTAL

core/src/ldml/ldml_transforms.hpp
```diff
@@ -25,11 +26,19 @@
...
 namespace km {
 namespace kbp {
 namespace ldml {
 
+// macro for working with ICU calls
+#define UASSERT_SUCCESS(status) \
+  if (U_FAILURE(status)) { \
+    DebugLog("U_FAILURE(%s)", u_errorName(status)); \
+    assert(U_SUCCESS(status)); \
+  }
```
mcdurdin, line 40: What do we do in failure cases in release builds, where the assert won't happen? We'll get different behaviour because it will fall through and continue. Do we need to return false or something in the function?

Also might be better to do this as an inline function because it can be optimized more cleanly and avoids accidental misuse with function calls as the status parameter etc.

mcdurdin (approved):
This is pretty close now. LGTM

srl295:
In-line functions won’t get the assert line number

mcdurdin:
```
#define my_assert(x) my_assert_((x),__LINE__)
```

mcdurdin:
```c++
#define my_assert(x) my_assert_((x),__LINE__)
```

image

pffft

core/src/ldml/ldml_transforms.hpp
srl295: changed to return false. 

Note that the call site _already_ kept the string unchanged if the assert failed.

srl295: I also improved the transform loading code to handle assert failures as validation failures

mcdurdin (approved):

core/src/ldml/ldml_transforms.cpp
```diff
@@ -456,18 +458,23 @@ transform_entry::transform_entry(
   }
 }
 
-void
+bool
 transform_entry::init() {
   if (!fFrom.empty()) {
```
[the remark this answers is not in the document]
srl295: nah, found another … `valid` shouldn't be overwritten without check.  PTAL

core/src/ldml/ldml_transforms.cpp
```diff
@@ -798,7 +815,15 @@ transforms::load(
         const std::u32string toStr                      = kmx::u16string_to_u32string(kplus.strs->get(element->to));
         KMX_DWORD mapFrom                               = element->mapFrom; // copy, because of alignment
         KMX_DWORD mapTo                                 = element->mapTo;   // copy, because of alignment
-        newGroup.emplace_back(fromStr, toStr, mapFrom, mapTo, kplus);  // creating a transform_entry
+        assert(!fromStr.empty());
```
srl295, line 818: so we get an earlier assert

core/src/ldml/ldml_transforms.cpp
```diff
@@ -403,25 +403,29 @@ transform_entry::transform_entry(const transform_entry &other)
...
     const std::u32string &from,
     const std::u32string &to,
     KMX_DWORD mapFrom,
     KMX_DWORD mapTo,
-    const kmx::kmx_plus &kplus)
+    const kmx::kmx_plus &kplus,
+    bool &valid)
     : fFrom(from), fTo(to), fFromPattern(nullptr), fMapFromStrId(mapFrom), fMapToStrId(mapTo) {
+  if (!valid)
+    return; // exit early
```
srl295, line 420: so, if valid came in false we leave with it false.

Only change it from true->false on err, don't set it to true here.

core/src/ldml/ldml_transforms.cpp
```diff
@@ -417,13 +416,16 @@ transform_entry::transform_entry(
     const kmx::kmx_plus &kplus,
     bool &valid)
     : fFrom(from), fTo(to), fFromPattern(nullptr), fMapFromStrId(mapFrom), fMapToStrId(mapTo) {
-  valid = false;
+  if (!valid)
+    return; // exit early
```
mcdurdin, line 420: This is similar to the status pattern from ICU. It means more work on the caller side but I guess I am okay with it!

mcdurdin (approved):

core/src/ldml/ldml_transforms.cpp
srl295: it makes sure the c'tor can return failure.  Could be a factory model, but the caller is using `emplace_back()` so the idea is a stack based construction anyway. on failure we're going to bail out of the entire load.

keyman-server:

PR #9731, opened 2023-10-10
chore: clean up a few minor discrepancies in builder.inc.sh

mcdurdin:
- Some uses of `KEYMAN_ROOT` (not cross-repo compatible)
- `set -eu` and `SHLVL` now set there as well

Other Updates:
* https://github.com/keymanapp/shared-sites/pull/3
* https://github.com/keymanapp/keyboards/pull/2393 (https://github.com/keymanapp/keyboards/pull/2393/commits/3e1368dd649631ab1b4e91d5c41b286ef949e669)
* https://github.com/keymanapp/lexical-models/pull/226 (https://github.com/keymanapp/lexical-models/pull/226/commits/28c7506b83074d5b6ff287647265eba5b19d01b8)

@keymanapp-test-bot skip

### DIFF
--- a/resources/build/build-utils.sh
+++ b/resources/build/build-utils.sh
@@ -29,6 +29,11 @@
 # Note: keep changes to version, tier and tag determination in sync with mkver (windows/src/buildutils/mkver)
 #
 
+# Note: set -eu and SHLVL are deliberately set both here and in builder.inc.sh,
+#       because we want them set as early as possible, and because
+#       builder.inc.sh is shared to other repos, this keeps the usage consistent
+#       there as well.
+
 # Exit on command failure and when using unset variables:
 set -eu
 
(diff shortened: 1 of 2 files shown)


ermshiperete (approved):

ermshiperete:
I guess we don't use the shared site scripts in the keyman repo yet, that's why we have a `builder.inc.sh` in both places?

mcdurdin:
> I guess we don't use the shared site scripts in the keyman repo yet, that's why we have a `builder.inc.sh` in both places?

Good idea -- I hadn't really considered sharing back to the keyman repo. Sharing the script back to the keyman repo could be considered in the future, although most of the files in _common are really only site-relevant at present (so we may want to consider this carefully). But for now, just sharing the web site infrastructure is a good step forward.

keyman-server:

PR #9734, opened 2023-10-10
refactor(linux): Simplify adding keyboard

ermshiperete:
Before we add a keyboard we check if we already added this keyboard (in a different area). If we did (with the same or newer version) we skip this keyboard. This change now immediately returns whereas previously we set a flag and then didn't execute the rest of the method.

@keymanapp-test-bot skip


mcdurdin (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.189-alpha](https://keyman.com/downloads/releases/alpha/17.0.189)

PR #9758, opened 2023-10-13
chore(linux): Add code coverage index page

ermshiperete:
This change adds an index page that can show the reports for the child projects.

Also fix a path problem in the `keyman-system-service/build.sh` script.

@keymanapp-test-bot skip

### DIFF
--- a/docs/linux/README.md
+++ b/docs/linux/README.md
@@ -87,6 +87,21 @@ scripts/run-tests
 The `run-tests` script accepts different arguments which can be seen with
 `scripts/run-tests --help`.
 
+### Code Coverage Report
+
+All three projects (ibus-keyman, keyman-config, and keyman-system-service)
+can produce code coverage reports.
+
+Run `./build.sh --debug --coverage --report test` to create the report.
+
+**Note:** You might have to run `./build.sh clean` first.
+
+There's also an index page (`linux/CodecoverageReports.html`) that links to all
+three reports. You can run
+`linux/build.sh --debug --coverage --report --no-integration test`
+to create the coverage reports for all three Linux projects and open the
+index page in the browser.
+
 ## Running Keyman for Linux
 
 ### Setting up Ibus
--- a/docs/linux/README.md
+++ b/docs/linux/README.md
@@ -98,7 +98,7 @@ Run `./build.sh --debug --coverage --report test` to create the report.
 
 There's also an index page (`linux/CodecoverageReports.html`) that links to all
 three reports. You can run
-`linux/build.sh --debug --coverage --report --no-integration test`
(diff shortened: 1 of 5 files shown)


linux/build.sh
```diff
@@ -30,3 +30,12 @@ builder_describe \
 builder_parse "$@"
 
 builder_run_child_actions clean configure build test install uninstall
+
+test_action() {
+  if builder_has_option --report; then
+    builder_echo "Opening coverage reports in browser..."
+    xdg-open "file://${THIS_SCRIPT_PATH}/CodeCoverageReports.html"
+  fi
```
mcdurdin, line 38: Does this create a coverage report or open it in the browser or both? It would be better for the creation of the report to be separated from the opening of it for use in CI.

linux/CodeCoverageReports.html
```diff
@@ -0,0 +1,37 @@
...
+
+  <script>
+    function openPage(elmnt, filename) {
+      document.getElementById('iframe_a').src = filename;
+      tablinks = document.getElementsByClassName("tablink");
+      for (i = 0; i < tablinks.length; i++) {
+        tablinks[i].style.backgroundColor = "";
+      }
+      elmnt.style.backgroundColor = "#777";
+    }
```
mcdurdin, line 31: I wonder if it would be better for the reports to just be three plain links in this file, `<li><a href='.../index.html'>ibus-keyman</a></li>`

That's less to maintain and easier to share a specific report

docs/linux/README.md
```diff
@@ -87,6 +87,21 @@ scripts/run-tests
...
+
+Run `./build.sh --debug --coverage --report test` to create the report.
+
+**Note:** You might have to run `./build.sh clean` first.
+
+There's also an index page (`linux/CodecoverageReports.html`) that links to all
+three reports. You can run
+`linux/build.sh --debug --coverage --report --no-integration test`
+to create the coverage reports for all three Linux projects and open the
+index page in the browser.
```
mcdurdin, line 103: The coverage reports should be included as build artifacts in the TeamCity build, along with the index.html file. They'll still be a little buried for now in the TC artifacts but at least we can get a canonical URL for them.

mcdurdin (changes requested):
I like; some tweaks for you?

linux/build.sh
ermshiperete: I added a `--open` option to open the report.

mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.193-alpha](https://keyman.com/downloads/releases/alpha/17.0.193)

PR #9759, opened 2023-10-13
chore(core): dx: ldml test improvement, backspace test 🙀

srl295:
- add a filter option to run just one subtest. I don't try to do anything with the "embedded" (`@@keys …`) tests but skip them if any filter is set.  This is very helpful for running within the debugger.
- support backspace event, revamp test JSON format
- backspace transforms not implemented, but a test is ready for them: #9760

For: #9468

@keymanapp-test-bot skip


core/tests/unit/ldml/keyboards/k_006_backspace-test.xml
```diff
@@ -0,0 +1,15 @@
...
+<keyboardTest3 conformsTo="techpreview">
+  <info keyboard="k_006_backspace.xml" author="Team Keyboard" name="backspace" />
+  <tests name="backspace">
+    <test name="regex-test-0-a">
+      <startContext to="" />
+      <keystroke key="t" />
+      <keystroke key="a" />
+      <keystroke key="b" />
+      <backspace />
+      <check result="ta" /> <!-- TODO-LDML: #9760 Should be 't' due to backspace transform -->
```
srl295, line 12: This test verifies that `<backspace />` works at all, #9760 to fix it for transforms

srl295:
OK this is a little concerning… https://build.palaso.org/buildConfiguration/Keyman_Test_Common_Mac/414588?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true&showLog=414588_726_114&logView=flowAware — it's acting as if common/web/types is out of date. (see the .ts change in this PR)

mcdurdin:
> it's acting as if common/web/types is out of date.

I'm not really seeing the problem you are? Line 401 returns an object which looks like it has a similar structure to what the previous version of the reader returned, so that would fail the test because it doesn't match the actual fixture.

common/web/types/src/ldml-keyboard/ldml-keyboard-xml-reader.ts
```diff
@@ -393,15 +393,12 @@ export class LDMLKeyboardXMLSourceFileReader {
...
-            case 'check':     return { check:     subv };
-            case 'emit':      return { emit:      subv };
-            case 'startContext': return null; // handled above
-            default: this.callbacks.reportMessage(CommonTypesMessages.Error_TestDataUnexpectedAction({ subtag })); return null;
+          const type = v['#name']; // element name
+          if (type === 'startContext') {
+            return null; // handled above
           }
+          const subv = LDMLKeyboardXMLSourceFileReader.defaultMapper(v, r);
+          return Object.assign({ type }, subv);
```
srl295, line 401: This will return an object similar to `{keystroke: subv}` ? Seems like it should be different. , `{ type: "keystroke", (other props from subv) }`

srl295: The old format was one level deeper in objects, wasn't it?

mcdurdin: yes. does the test fail locally?

srl295:
OK. User error. There are two test cases which test this data. I fixed one of them, and misread the output of the other one.  I'll update the test cases to add a FYI about both of them.

common/web/types/src/ldml-keyboard/ldml-keyboard-xml-reader.ts
srl295: I was running the wrong test.

mcdurdin (approved):
LGTM

keyman-server:

PR #9781, opened 2023-10-17
chore(linux): Fix build scripts

ermshiperete:
Creating the coverage option to ninja previously didn't work reliably. This change re-orders the parameters passed to `meson setup` and seems to work better. Additionally this change refactors the scripts and makes use of `builder_run_action`.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-system-service/build.sh
+++ b/linux/keyman-system-service/build.sh
@@ -35,15 +35,22 @@ else
 fi
 MESON_PATH="build/$(uname -m)/$MESON_TARGET"
 
-builder_describe_outputs \
-  configure "${MESON_PATH}/build.ninja" \
-  build "${MESON_PATH}/src/keyman-system-service"
+clean_action() {
+  rm -rf "$THIS_SCRIPT_PATH/build/"
+}
 
-if builder_has_option --coverage; then
-  MESON_COVERAGE=-Db_coverage=true
-else
-  MESON_COVERAGE=
-fi
+check_missing_coverage_configuration() {
+  if builder_has_option --coverage && [ -d "$MESON_PATH" ]; then
+    # It's possible that we got configured because we're a dependency of ibus-keyman
+    # in which case the `--coverage` option wasn't passed along.
+    cd "$MESON_PATH"
+    if ! ninja -t targets | grep -q coverage-html ; then
+      cd "$THIS_SCRIPT_PATH"
+      clean_action
+    fi
+    cd "$THIS_SCRIPT_PATH"
+  fi
+}
 
 configure_action() {
   # shellcheck disable=SC2086,SC2154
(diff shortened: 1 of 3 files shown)


mcdurdin (approved):
The `[ -d "$MESON_PATH" ] && cd "$MESON_PATH"` line is a little sneaky and perhaps will cause confusion if ninja then attempts to run in the wrong folder because `configure` hasn't been run, but I understand what you are doing and it's concise.

LGTM

linux/keyman-system-service/build.sh
```diff
@@ -35,6 +35,39 @@ else
...
+  if builder_has_option --coverage && [ -d "$MESON_PATH" ]; then
+    # It's possible that we got configured because we're a dependency of ibus-keyman
+    # in which case the `--coverage` option wasn't passed along.
+    cd "$MESON_PATH"
+    if ! ninja -t targets | grep -q coverage-html ; then
+      cd "$THIS_SCRIPT_PATH"
+      clean_action
+    fi
+    cd "$THIS_SCRIPT_PATH"
+  fi
```
ermshiperete, line 52: Is there an easier solution? I don't think our `build.sh` system allows to pass the `--coverage` option to dependencies, or does it?

ermshiperete (commented):
I realized coverage report creation still wasn't reliably working. `keyman-system-service` gets build as a dependency of `ibus-keyman` in which case it doesn't get passed the `--coverage` option.

linux/keyman-system-service/build.sh
mcdurdin: keyman-system-service should be a child build rather than a dependency. A dependency is explicitly something that is required for the project but that you only want to build and use. A child build is a component owned by the parent build.

https://github.com/keymanapp/keyman/blob/1e734a47f5337515b58cfdc55145bc5dd4de2660/resources/build/build-utils.md#L447-L451

ermshiperete: It's both. You can't run ibus-keyman tests without first building keyman-system-service.

Or do you mean keyman-system-service should be a child of ibus-keyman? But that doesn't make completely sense either because it's not really a component of ibus-keyman.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.194-alpha](https://keyman.com/downloads/releases/alpha/17.0.194)

linux/keyman-system-service/build.sh
mcdurdin: I'm not really sure what you are trying to solve here. It seems like the top-level linux/build.sh is already building all the modules as child builds, and the `--coverage` option already has `+` so it should be passed through.

If things are building in the wrong order, you could try putting the keyman-system-service first in linux/build.sh.

Issue #9733, opened 2023-10-10
chore(linux): Rename `libkmnkbp` to `libkeymancore`


ermshiperete:
As a follow-up of #9721 and #9722 we should rename the library to `libkeymancore` and also bump the version number. This also requires changes in the `debian` directory.

- new binary will be `libkeymancore.so.1.0`
- package names will be `libkeymancore` and `libkeymancore-dev`

_Edit: Updated library name following our discussion in standup meeting. `libkeymancore` is less cryptic than `libkmncore`._

mcdurdin:
See also notes in https://github.com/keymanapp/keyman/pull/9723#issuecomment-1755512073

ermshiperete:
`core/meson.build` specifies `lib_version` which has to be updated.

Also noticed that `libkmnkbp(.a)` is also used on the other platforms.

PR #9793, opened 2023-10-18
chore(linux): Rename (lib)kmnkbp to (lib)keymancore 〽️

ermshiperete:
Part of #9733.

@keymanapp-test-bot skip


core/src/version.rc
```diff
@@ -23,7 +23,7 @@
         VALUE "InternalName", "Keyman Core"
         VALUE "LegalCopyright", "© SIL International"
         VALUE "LegalTrademarks", ""
-        VALUE "OriginalFilename", "KMNCORE0-0.DLL"
+        VALUE "OriginalFilename", "KEYMANCORE1-0.DLL"
```
mcdurdin, line 26: Here we have KEYMANCORE1-0.DLL, but in kmdev.wxs and elsewhere we have keymancore1.dll.

```suggestion
        VALUE "OriginalFilename", "KEYMANCORE1.DLL"
```

linux/debian/libkmnkbp0-0.symbols
```diff
@@ -1,43 +1,43 @@
-libkmnkbp0.so.0 libkmnkbp0-0 #MINVER#
+libkeymancore.so.1 libkmnkbp0-0 #MINVER#
```
mcdurdin, line 1: Is this still supposed to be libkmnkbp0-0 here? And noting the filename is still given as libkmnkbp0-0.symbols.

mcdurdin (approved):
LGTM but now I am wondering why we have both keyman-engine.vcxproj which appears to link keymancore.lib and keyman32.vcxproj which has links libkeymancore.a? (@rc-swag?)

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.197-alpha](https://keyman.com/downloads/releases/alpha/17.0.197)

PR #9794, opened 2023-10-18
chore(linux): Rename KBP to CORE 〽️

ermshiperete:
Part of #9733.

@keymanapp-test-bot skip

### DIFF
--- a/common/include/kmx_file.h
+++ b/common/include/kmx_file.h
@@ -25,7 +25,7 @@ typedef KMX_WORD  __attribute__((aligned(1))) KMX_WORD_unaligned;
 #define KMX_WORD_unaligned  KMX_WORD
 #endif
 
-#ifdef KMN_KBP
+#ifdef KMN_CORE
 // TODO: move this to a common namespace keyman::common::kmx_file or similar in the future
 namespace km {
 namespace kbp {
@@ -402,7 +402,7 @@ static_assert(sizeof(COMP_KEY) == KEYBOARDFILEKEY_SIZE, "COMP_KEY must be KEYBOA
 static_assert(sizeof(COMP_GROUP) == KEYBOARDFILEGROUP_SIZE, "COMP_GROUP must be KEYBOARDFILEGROUP_SIZE bytes");
 static_assert(sizeof(COMP_KEYBOARD) == KEYBOARDFILEHEADER_SIZE, "COMP_KEYBOARD must be KEYBOARDFILEHEADER_SIZE bytes");
 
-#ifdef KMN_KBP
+#ifdef KMN_CORE
 } // namespace kmx
 } // namespace kbp
 } // namespace km
--- a/common/include/kmx_file.h
+++ b/common/include/kmx_file.h
@@ -25,7 +25,7 @@ typedef KMX_WORD  __attribute__((aligned(1))) KMX_WORD_unaligned;
 #define KMX_WORD_unaligned  KMX_WORD
 #endif
 
-#ifdef KMN_CORE
+#ifdef KM_CORE_LIBRARY
 // TODO: move this to a common namespace keyman::common::kmx_file or similar in the future
 namespace km {
 namespace kbp {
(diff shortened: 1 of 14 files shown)


common/include/kmx_file.h
```diff
@@ -25,7 +25,7 @@ typedef KMX_WORD  __attribute__((aligned(1))) KMX_WORD_unaligned;
 #define KMX_WORD_unaligned  KMX_WORD
 #endif
 
-#ifdef KMN_KBP
+#ifdef KMN_CORE
```
mcdurdin, line 28: Everywhere else the prefix is KM_CORE, do we want to use the same, and some thing after to make it clear it's a preprocessor control, e.g. KM_CORE_LIBRARY?

core/include/keyman/keyman_core_api_bits.h
```diff
@@ -44,23 +44,23 @@
...
 #else
   typedef char const * km_core_path_name;
-  #define _KM_KBP_PATH_SEPARATOR ('/')
-  #define _KM_KBP_EXT_SEPARATOR ('.')
+  #define _KM_CORE_PATH_SEPARATOR ('/')
+  #define _KM_CORE_EXT_SEPARATOR ('.')
 #endif
 
-#if defined KMN_KBP_STATIC
+#if defined KMN_CORE_STATIC
```
mcdurdin, line 51: KM_CORE_LIBRARY_STATIC?

core/include/keyman/keyman_core_api_bits.h
```diff
@@ -44,23 +44,23 @@
...
   #define KMN_API             _kmn_tag_fn(_kmn_static_flag)
   #define KMN_DEPRECATED_API  _kmn_tag_fn(_kmn_deprecated_flag _kmn_and _kmn_static_flag)
-#elif defined KMN_KBP_EXPORTING
+#elif defined KMN_CORE_EXPORTING
```
mcdurdin, line 54: KM_CORE_LIBRARY_EXPORTING?

core/include/keyman/keyman_core_api_bits.h
```diff
@@ -44,23 +44,23 @@
...
   #define KMN_DEPRECATED_API  _kmn_tag_fn(_kmn_deprecated_flag _kmn_and _kmn_export_flag)
 #else
   #define KMN_API             _kmn_tag_fn(_kmn_import_flag)
   #define KMN_DEPRECATED_API  _kmn_tag_fn(_kmn_deprecated_flag _kmn_and _kmn_import_flag)
 #endif
 
-#ifndef KMN_KBP
-#define KMN_KBP
+#ifndef KMN_CORE
+#define KMN_CORE
```
mcdurdin, line 63: KM_CORE_LIBRARY?

core/src/processor.hpp
```diff
@@ -75,7 +75,7 @@ namespace kbp
      * updateing the state as required.
      *
      * @param   state  An opaque pointer to a state object
-     * @return  km_core_status  `KM_CORE_STATUS_OK`: On success. Else KB_KBP_ error code
+     * @return  km_core_status  `KM_CORE_STATUS_OK`: On success. Else KB_CORE_ error code
```
mcdurdin, line 78:
```suggestion
     * @return  km_core_status  `KM_CORE_STATUS_OK`: On success. Else KM_CORE_ error code
```

core/src/version.rc
```diff
@@ -23,7 +23,7 @@
         VALUE "InternalName", "Keyman Core"
         VALUE "LegalCopyright", "© SIL International"
         VALUE "LegalTrademarks", ""
-        VALUE "OriginalFilename", "KMNKBP0-0.DLL"
+        VALUE "OriginalFilename", "KMNCORE0-0.DLL"
```
mcdurdin, line 26:
```suggestion
        VALUE "OriginalFilename", "KEYMANCORE0-0.DLL"
```

I also wonder about the need to append 0-0, surely we should be using the actual version?

mcdurdin (changes requested):
Good cleanups; I think we should tweak a few of the names slightly more though

core/src/version.rc
ermshiperete: The next PR (#9793) updates the version number. But I don't think we want to use 17 as version number since this is the Core API version which usually doesn't change in breaking ways between Keyman versions.

mcdurdin: True. We just need to have a documented version number for the API then

ermshiperete: Good point. Created #9801.

mcdurdin (approved):

keyman-server:

PR #9828, opened 2023-10-23
feat(core): new actions APIs

### LINKED ISSUE #9720: refactor(core): rework action and context APIs to simplify Engine implementations
Look at a redesign of the actions list that comes out of Keyman Core, to:

* normalize the chars + backspaces so that we don't interleave them
* convert into a structure rather than an arbitrary list of actions

Things to consider:

* implementation coordination - how to work with all different consumer engines
* backspace types - char vs marker vs unknown
* eradication of Engine-ownership of cached context (Windows, Linux?)

## Proposal

1. Keep the existing interface `km_core_state_action_items()`, but add a new function `km_core_state_get_actions()` which returns a single structure, and in which Keyman Engine no longer needs to think about markers at all, because they are maintained within Keyman Core's cached context. When you have finished with the returned structure, free it with `km_core_actions_dispose()`.

    ```c++
    typedef struct {
      // number of codepoints (not codeunits!) to delete from app context, 0+
      int delete_back;
    
      // null-term string of characters to insert into document
      km_core_usv* output;
    
      // list of options to persist, terminated with KM_CORE_OPTIONS_END
      km_core_option_item* persist_options;
    
      // issue a beep
      bool do_alert;
    
      // emit the input keystroke to the application, unmodified?
      bool emit_keystroke;
    
      // -1=unchanged, 0=off, 1=on
      int new_caps_lock_state;
    } km_core_actions;
    ```

2. Add a new function `km_core_state_context_validate()` which does a comparison of the passed-in context from a compliant app to the state's cached context, allowing Core to know if it needs to reset the context. This should replace all the existing context API functions. This should only be used with compliant apps, on every key event, before `km_core_process_event()` is called.

3. If a mouse click is received on a non-compliant app, then Engine can call `km_core_state_context_invalidate()` to wipe out any cached context.

## Design discussion document

* https://docs.google.com/document/d/13kRnMm_5b2h4ijWzsOfqq963nwnM1RED2WAHmknzOhQ/edit

## Benefits

* Simplification of APIs reduces complexity of Engine implementations and centralizes the key context management algorithms 
* One source of truth for cached context
* Backspace types - char vs marker vs unknown becomes an internal distinction only in Core

## Costs

* Implementation time and coordination of Core changes vs Engine changes. Mitigated by maintenance of existing functions until all consumers have rolled over to the new design.

### LINKED ISSUE #9832: bug(core): `marker` has `uint32_t` type and `uintptr_t` type confusion
```c++
typedef struct {
  ...
    uint32_t    marker;
  ...
} km_core_context_item;
```

vs

```c++
typedef struct {
  ...
    uintptr_t             marker;          // MARKER type
  ...
} km_core_action_item;
```

Suggest converting both to `uint32_t` to be consistent about sizing. In practice `uintptr_t` is never going to be smaller than `uint32_t`, but I don't think we need a 64-bit value here. Note that we have an implicit assumption about markers being no larger than `uint16_t` within KMX processor (and perhaps elsewhere too).

mcdurdin:
Fixes #9720.
Fixes #9832.

Note: this should make a number of Keyman Core context and action APIs unnecessary, except for use by IMX and debugger. Once we move over to these APIs with existing Engines, then we should audit the remaining usage and remove unnecessary APIs to reduce the maintenance burden.

@keymanapp-test-bot skip


core/tests/unit/kmnkbd/action_api.cpp
```diff
@@ -0,0 +1,432 @@
+/*
+  Copyright:    © 2018 SIL International.
+  Description:  Tests for the context API family of functions.
+  Create Date:  19 Oct 2018
+  Authors:      Tim Eves (TSE)
+  History:      19 Oct 2018 - TSE - Initial implementation.
+                22 Oct 2018 - TSE - Refactor to add and use try_status macro
+                                    for improved readability.
+                                  - Add more tests to cover corner cases and
+                                    mutation functions.
```
mcdurdin, line 10:
```suggestion
  Authors:      Marc Durdin (MCD)
  History:      23 Oct 2023 - MCD - Initial implementation.
```

core/include/keyman/keyman_core_api.h
```diff
@@ -310,6 +310,38 @@ km_core_context_items_to_utf8(km_core_context_item const *item,
...
+  `buf_size` will contain the space required. The contents of the buffer are
+  undefined.
+##### Parameters:
+- __context_items__: A pointer to the start of an array `km_core_context_item`.
+    Must be terminated with a type of `KM_CORE_CT_END`.
+- __buf__: A pointer to the buffer to place the UTF-32 string into.
+    May be null to request size calculation.
+- __buf_size__: a pointer to the result variable:
+    The size of the supplied buffer in codepoints if `buf` is given.
+    On return will be the size required if `buf` is null.
```
sgschantz, line 335: I realize that this if following the pattern of the UTF8 function, but I think it would be good to have a separate function to get the required size rather than using a null buf to request that. Without it, that existence of that behavior may be hidden in this comment.

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+
+/*
+```
+### `km_core_state_get_actions`
+##### Description:
+Returns a pointer to an actions object which details all the actions
+that the Platform layer must take after a keystroke. The `delete_back`
+action must be performed before the `output` action, but the other
+actions may be performed in any order.
+##### Return:
```
sgschantz, line 602: I'm wondering if we can name the delete_back differently to include the two important comments about it and make it more self-documenting: 1) it is number of code points, not code units and 2) it must be executed before the output.
The name for that could be long, but it should only be consumed once in the platform code and would help to not have to search the docs to understand the proper way to apply it.

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+);
+
+/*
+```
+### `km_core_state_context_validate`
+##### Description:
+Determines if the input context has changed, and if so, resets
+the internal cached context (including markers), to the new
+context string.
+
```
sgschantz, line 640: I'm a little confused by this description: the idea of the context changing but the API name is validate. You say that it 'resets', but I think of a reset as clearing the context not setting it to a new value. The part about including markers seems a little ambiguous, as if the markers could also be provided in the input, but it should never have any.

Maybe it would be clearer if it said that it was *replacing* the internal cached context *with* the input context string. (If I'm understanding that correctly.)

An API named validate doesn't sound like it is going to change state. So if it has the possibility of replace the current context in the core, then I think it needs to reflect that.

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+This and `km_core_state_context_invalidate` will replace existing Core context
+APIs.
+
+##### Parameters:
+- __state__: An opaque pointer to a state object.
+- __application_context__: A pointer to an null-terminated `km_core_cp` string
+    representing the current context from the application.
+##### Return status:
+- `KM_CORE_STATUS_OK`: On success.
+- `KM_CORE_STATUS_INVALID_ARGUMENT`: If non-optional parameters are null.
```
sgschantz, line 650: I guess 'KM_CORE_STATUS_OK' means that the context input was consistent with that in the core, do it was not and successfully replaced it? We don't really care about which happened?

What non-optional parameters are not allowed to be null? Just the state or the context also?

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+- __state__: An opaque pointer to a state object.
+##### Return status:
+- `KM_CORE_STATUS_OK`: On success.
+- `KM_CORE_STATUS_INVALID_ARGUMENT`: If non-optional parameters are null.
+
+```c
+*/
+KMN_API
+km_core_status
+km_core_state_context_invalidate(
```
sgschantz, line 680: I wonder if it would be better to call this clear or reset instead of invalidate. This might clash with our existing naming convention, but invalidate makes it sound like it was flagged rather than obliterated.

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+km_core_state_context_validate(
+  km_core_state *state,
+  km_core_cp const *application_context
+);
+
+/*
+```
+### `km_core_state_context_invalidate`
+##### Description:
+Flushes the internal cached context for the state.
```
sgschantz, line 665: Flush makes me think of flushing a buffer out to disk rather than just erasing it.

sgschantz (commented):
It will be a nice improvement, but I have a few comments on naming and documentation of the APIs.

core/include/keyman/keyman_core_api.h
```diff
@@ -310,6 +310,38 @@ km_core_context_items_to_utf8(km_core_context_item const *item,
...
 
+/*
+```
+### `km_core_context_items_to_utf32`
+##### Description:
+Convert a context item array into a UTF-32 encoded string placing it into
+the supplied buffer of specified size, and return the number of codepoints
+actually used in the conversion. If null is passed as the buffer the
+number codepoints required is returned. This will strip markers from the
+context during the conversion.
```
srl295, line 321:
```suggestion
Convert a context item array into a UTF-32 encoded string placing it into
the supplied buffer of specified size, and return the number of codepoints
actually used in the conversion. If null is passed as the buffer, the
number of codepoints required is returned.
Any markers in the context will not be included in the output buffer.
```

I assume it doesn't actually mutate the context _array_, just the output. Also, typo.

core/include/keyman/keyman_core_api.h
srl295: so ICU's idiom is to return the equivalent of `KM_CORE_STATUS_INSUFFICENT_BUFFER` if there isn't enough space, even if a buffer size was originally given.  And then, if you pass in 0 for the length ("pure preflight") then only calculation is done.

The 'mixed' use case could be where you preallocate a buffer of a certain size that should work for most cases, and then for a bad case you (re)allocate and try again.

srl295: What is the return status when `nullptr` is passed in? 

I think `buf_size` should be 0 if it's for size calculation.

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+services framework to transform the text store in the Client Application, among
+other actions.
+
+This API replaces the Action items APIs, which is now deprecated and will be
+removed in the future.
+```c
+*/
+
+typedef struct {
+  // number of codepoints (not codeunits!) to delete from app context, 0+
```
srl295, line 575:
```suggestion
  // number of codepoints (not codeunits!) to delete from app context. Must be >=0
```

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+  km_core_option_item* persist_options;
+
+  // issue a beep
+  bool do_alert;
+
+  // emit the input keystroke to the application, unmodified?
+  bool emit_keystroke;
+
+  // -1=unchanged, 0=off, 1=on
+  int new_caps_lock_state;
```
srl295, line 591: should be an enum

core/include/keyman/keyman_core_api.h
srl295: `code_points_to_delete` ? seems like a good idea

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,128 @@ enum km_core_action_type {
...
+*/
+KMN_API
+km_core_actions*
+km_core_state_get_actions(
+  km_core_state const *state
+);
+
+/*
+```
+### `km_core_actions_dispose`
```
srl295, line 618: OK if `dispose` is the local term for it, I would think `free` would be more reminiscent of  malloc/free

core/include/keyman/keyman_core_api.h
srl295: The question is, when should the caller call this? Should be named something such as `update_state_with_new_context` ?

core/include/keyman/keyman_core_api.h
srl295: clear or reset makes sense.  And say, "called to notify that the internal cached context is no longer valid and should be reset…"

core/src/action.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <algorithm>
+#include <sstream>
+
+#include <keyman/keyman_core_api.h>
+
+#include "action.hpp"
+#include "state.hpp"
+#include "option.hpp"
+
+km_core_actions * km::kbp::action_items_to_actions(
```
srl295, line 19:
```suggestion
km_core_actions * km::kbp::action_item_list_to_actions_object(
```

Too bad we can't use markdown in the function name here! 

clarifying name

srl295: thought about 'flatten' in the name also

core/src/action.cpp
```diff
@@ -0,0 +1,140 @@
...
+        assert(false);
+    }
+  }
+
+  output.push_back({KM_CORE_CT_END});
+
+  // Strip the markers from the output to be passed to the app
+
+  size_t buf_size;
+
```
srl295, line 117:
```suggestion
     // measure number of codepoints needed for buffer
```

core/src/action.cpp
```diff
@@ -0,0 +1,140 @@
...
+  km_core_action_item const *action_items
+) {
+  assert(action_items != nullptr);
+  if(action_items == nullptr) {
+    return nullptr;
+  }
+
+  km_core_status status = KM_CORE_STATUS_OK;
+
+  km_core_actions* actions = new km_core_actions;
```
srl295, line 29:
```suggestion
  std::unique_ptr<km_core_actions> actions(new km_core_actions);
```

core/src/action.cpp
srl295: could perhaps use a std::unique_ptr for this buffer too

core/src/action.hpp
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+#pragma once
+
+#include <keyman/keyman_core_api.h>
+
+namespace km {
+namespace kbp
+{
+  km_core_actions* action_items_to_actions(
```
srl295, line 16: name change suggestion above

core/include/keyman/keyman_core_api.h
rc-swag: @srl the caller would now call this on every keystroke instead of `km_core_context_set`.  Before when the platform was managing its own version of the local cache (with markers) it could strip out the markers and see if the context was still the same. For example, the user had not moved the cursor to a new position.  
I agree that `validate` is not a good name as it doesn't suggest it is doing any mutation. 
Either the api has a call that just does a validation/match only and then if returns `false` we call `km_core_context_set`.
Or we have a single function that provides that functionality? 
`km_km_core_context_confirm_match_or_set`   -- a bit long maybe @sgschantz  will have a better name.

core/include/keyman/keyman_core_api.h
rc-swag: > clear or reset makes sense. And say, "called to notify that the internal cached context is no longer valid and should be reset…"

"called to notify the state context is no longer valid and should be reset"

core/include/keyman/keyman_core_api.h
rc-swag: The using null to request size is a pretty established pattern (not just this core), I am happy to add a `get_buffer_size` call but would suggest leaving this null_buf to behaviour with this function also.  In fact the the `get_buffer_size` could just call this function with the null_buff.

mcdurdin:
Note: failed build due to #9832.

core/include/keyman/keyman_core_api.h
mcdurdin: Given we already have two other functions that follow this pattern, I think it makes more sense to be consistent -- if we think this buffer sizing pattern is bad, we should fix all three functions rather than introduce a new pattern for just one of them. But that should be a separate change and not in this PR.

core/include/keyman/keyman_core_api.h
mcdurdin: I'm good with `delete_back_codepoints` but I think reading the API docs is essential for using an API correctly, because you can't really embed the _reasoning_ in the name without causing additional grief. I'm not that keen on adding `_first` because the entire state machine pattern needs to be understood anyway and `_first` actually confuses things IMO. What does `_first` refer to -- is it order of operations, or is it delete from first item in buffer, or ...? We'd have to be too wordy to make it helpful.

core/include/keyman/keyman_core_api.h
mcdurdin: Agree that the function names are somewhat difficult. Am going to refactor this into two functions I think, although that makes usage on Engine side slightly more difficult.

core/include/keyman/keyman_core_api.h
mcdurdin: The return value should be resolved when I split into two functions.

Both opts are required, will update comment

core/include/keyman/keyman_core_api.h
mcdurdin: ok with clear or reset. I'll probably go with clear.

core/include/keyman/keyman_core_api.h
mcdurdin: will change

core/include/keyman/keyman_core_api.h
mcdurdin: Thank you will apply this

core/include/keyman/keyman_core_api.h
mcdurdin: ok

core/include/keyman/keyman_core_api.h
mcdurdin: `dispose` matches existing api so for consistency will keep

core/src/action.cpp
mcdurdin: this is internal, not an api, so I am less worried about it, will accept this suggestion

core/src/action.cpp
mcdurdin: Hmm, how do we return this object then? std::move?

core/include/keyman/keyman_core_api.h
sgschantz: Ok, I was wondering whether that was a common pattern. I really dislike it when null is a magic value, but it doesn't make sense to change the rules, especially for an internal API.

core/include/keyman/keyman_core_api.h
sgschantz: I think code_points_to_delete is an improvement. Or you could try code_points_to_delete_before_inserting_text, but I have a feeling you would consider that too wordy. Isn't it friendly and readable though? :)

mcdurdin: `code_points_to_delete_before_inserting_text` (44 chars) would probably run into some maximum symbol size issue on one of our platforms 😉 

[For example](https://stackoverflow.com/a/57440831/1836776):
> Rule 5.1 (required): Identifiers (internal and external) shall not rely on the significance of more than 31 characters

sgschantz (approved):
thanks for making these changes -- I will be calling this code very soon!

rc-swag (approved):
lgtm

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,152 @@ enum km_core_action_type {
...
+typedef enum { KM_CORE_CAPS_UNCHANGED = -1, KM_CORE_CAPS_OFF = 0, KM_CORE_CAPS_ON = 1 } km_core_caps_state;
+
+typedef struct {
+  // number of codepoints (not codeunits!) to delete from app context.
+  unsigned int code_points_to_delete;
+
+  // null-term string of characters to insert into document
+  km_core_usv* output;
+
+  // list of options to persist, terminated with KM_CORE_OPTIONS_END
```
ermshiperete, line 584:
```suggestion
  // list of options to persist, terminated with KM_CORE_OPTIONS_END, or null
```

ermshiperete (changes requested):

core/src/action.cpp
mcdurdin: Thanks @ermshiperete ... guess I am having a bad C++ day 😆

core/include/keyman/keyman_core_api.h
```diff
@@ -526,6 +558,152 @@ enum km_core_action_type {
...
+  km_core_option_item* persist_options;
+
+  // issue a beep, 0 = no, 1 = yes
+  km_core_bool do_alert;
+
+  // emit the input keystroke to the application, unmodified? 0 = no, 1 = yes
```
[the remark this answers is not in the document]
mcdurdin: unmodified just means 'no modification'. We don't support modification of the keystroke.

core/include/keyman/keyman_core_api.h
mcdurdin: Hmm, I want them to use KM_CORE_OPTIONS_END so won't encourage use of null :grin:

core/include/keyman/keyman_core_api.h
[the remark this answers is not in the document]
ermshiperete: Ok, now I understand what's meant. The `unmodified` refers mainly to `input keystroke`, not `no/yes`. And it's not a left-over question from API design. 

This might be easier to understand if the word order is different or the question mark omitted:

```suggestion
  // emit the unmodified input keystroke to the application? 0 = no, 1 = yes
```

or

```suggestion
  // emit the (unmodified) input keystroke to the application? 0 = no, 1 = yes
```

or

```suggestion
  // emit the input keystroke to the application, unmodified. 0 = no, 1 = yes
```

core/include/keyman/keyman_core_api.h
ermshiperete: The way it's documented I'd expect to have an array with just one element, KM_CORE_OPTIONS_END, if there are no options, but that doesn't match the implementation. So I think we should document that this can be null if there are no options to persist. Especially since `output` above will be an empty string (instead of null) if there is no output.

ermshiperete (changes requested):

core/include/keyman/keyman_core_api.h
mcdurdin: Gotcha. I think I will change `options` to always point to a valid array ending in `KM_CORE_OPTIONS_END`. That's one less test for the consumer.

ermshiperete (approved):
Thanks for the changes! I'm happy now :smile:

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.198-alpha](https://keyman.com/downloads/releases/alpha/17.0.198)

PR #9859, opened 2023-10-26
feat(developer): extract font family from .ttf in kmc-keyboard-info

### LINKED ISSUE #9858: feat(developer): kmc-keyboard-info needs to read font family from .ttf rather than use a generated name, where possible

mcdurdin:
Fixes #9858.

.keyboard_info files have a section for web fonts associated with the keyboard. Currently, the web fonts use a generated font family name based on the name of the keyboard, because all we have in the .kps source (which is where get the information from) is a reference to a .ttf or .woff file, and we don't currently read the font file to know anything about it.

The web font data is used to write a `@font-face {` CSS rule by keymanweb.com, with the font family as provided in the .keyboard_info.

Instead of generating a name for the web fonts, we should be using the font family from the .ttf, as we have it available. This allows keymanweb.com to avoid downloading the font if the user already has it installed on their system.

Uses ttfmeta npm package, which is a lightweight .ttf reader. Updates the honking big unit test to do some basic tests on this functionality.

@keymanapp-test-bot skip


developer/src/kmc-keyboard-info/src/font-family.ts
```diff
@@ -0,0 +1,17 @@
+import ttfMeta from 'ttfmeta';
+
+export async function getFontFamily(source: Uint8Array) {
```
mcdurdin, line 3:
```suggestion
/**
 * Extracts the font-family from an in-memory TTF or WOFF blob in `source`
 * parameter
 */
export async function getFontFamily(source: Uint8Array) {
```

developer/src/kmc-keyboard-info/src/index.ts
```diff
@@ -463,6 +466,39 @@ export class KeyboardInfoCompiler {
...
+      return null;
+    }
+
+    const result = {
+      family: await getFontFamily(this.callbacks.loadFile(this.callbacks.resolveFilename(kpsFilename, sourcePath))),
+      source
+    };
+
+    if(!result.family) {
+      this.callbacks.reportMessage(KeyboardInfoCompilerMessages.Error_FileDoesNotExist({filename: sourcePath}));
```
mcdurdin, line 497: This needs to be a different error message -- file does exist but can't be parsed

rc-swag (approved):
lgtm

keyman-server:

PR #9861, opened 2023-10-26
fix(developer): handle displayMap correctly with 'fill from layout'

### LINKED ISSUE #9851: bug(developer): Fill From Layout gets confused with paths
Per @LornaSIL:

> I opened the `sil_hebrew` keyboard and added this line:
> 
> ```
> store(&displayMap) '../../../shared/fonts/kbd/KbdHebr.json'
> ```
> 
> And it compiles. But am I supposed to select the font somewhere? I installed the font and tried to select it in Keyboard / Fonts. But, it didn't use the font, because it didn't seem to have changed the mapping from Hebrew to PUA. Do I need to reimport the OSK?
> 
> I did try to reimport the OSK with Fill from layout and I got these errors:
> 
> ```
> kmn5DDE.kmn: 5002 Building kmn5DDE.kmn
> kmn5DDE.kmn: Error: 290A Error encountered parsing display map C:\Users\evansl\shared\fonts\kbd\KbdHebr.json: TypeError: The "list" argument must be an instance of SharedArrayBuffer, ArrayBuffer or ArrayBufferView.
> kmn5DDE.kmn: Error: 2908 Error encountered parsing sil_hebrew.kvks: TypeError: Cannot read properties of null (reading 'byteLength')
> kmn5DDE.kmn: 5007 kmn5DDE.kmn failed to build.
> ```
> 
> What is kmn5DDE.kmn?
> 
> I see the path for my map is totally wrong. It should have this path: `"C:\0WSTech\working\keyboards\release\shared\fonts\kbd\KbdHebr.json"`

Issue here is that currently Fill From Layout makes a temp file in another folder to do a build in order to ascertain the characters to go onto the OSK. The problem with paths is not specific to `&displayMap`; I expect it would occur with any system store that references files. It's probably a regression with the 17.0-alpha compiler.

Version: 17.0.19x-alpha

mcdurdin:
Fixes #9851.

Two fixes:

* errors from loading the displayMap from disk should be handled before passing result to next function
* temp compile of the .kmn for the 'fill from layout' function should be in the same path as the original .kmn, so that referenced files can be located

@keymanapp-test-bot skip

(User test involves setting up a `&displayMap` which is a bit too involved to be feasible right now...  I have tested locally.)

### DIFF
--- a/developer/src/kmc-kmn/src/compiler/compiler.ts
+++ b/developer/src/kmc-kmn/src/compiler/compiler.ts
@@ -280,6 +280,9 @@ export class KmnCompiler implements UnicodeSetParser {
 
       if(result.extra.displayMapFilename) {
         result.displayMap = this.loadDisplayMapping(infile, result.extra.displayMapFilename)
+        if(!result.displayMap) {
+          return null;
+        }
       }
 
       if(result.extra.kvksFilename) {
@@ -406,6 +409,10 @@ export class KmnCompiler implements UnicodeSetParser {
     try {
       // Expected file format: displaymap.schema.json
       const data = this.callbacks.loadFile(displayMapFilename);
+      if(!data) {
+        this.callbacks.reportMessage(CompilerMessages.Error_FileNotFound({filename: displayMapFilename}));
+        return null;
+      }
       const mapping = JSON.parse(new TextDecoder().decode(data));
       return Osk.parseMapping(mapping);
     } catch(e) {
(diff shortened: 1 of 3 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.201-alpha](https://keyman.com/downloads/releases/alpha/17.0.201)

PR #9865, opened 2023-10-26
chore(linux): Improve repository_dispatch 🎬

### LINKED ISSUE #9548: chore(linux): Allow manual runs aka workflow dispatch for Linux packages
See [cldr/.github/workflows/maven.yml at main · unicode-org/cldr](https://github.com/unicode-org/cldr/blob/main/.github/workflows/maven.yml#L6-L31) for an example of how to allow 'manual' runs of an action.

From [discussion at our planning meeting](https://docs.google.com/document/d/1gCwivHBhYtp_efmMC0wM9VD22p6K2eWQF5rJlY3sRv8/edit?skip_itp2_check=true&pli=1#heading=h.4hftbckfqkny).

ermshiperete:
Some improvements to repository_dispatch, and documentation how to trigger GHA build manually for testing purposes.

Closes #9548.

@keymanapp-test-bot skip

### DIFF
--- a/.github/workflows/deb-packaging.yml
+++ b/.github/workflows/deb-packaging.yml
@@ -1,32 +1,16 @@
 name: "Ubuntu packaging"
-run-name: "Ubuntu packaging - ${{ github.event.inputs.branch }} (branch ${{ github.head_ref }}), by @${{ github.actor }}"
+run-name: "Ubuntu packaging - ${{ github.event.client_payload.branch }} (branch ${{ github.head_ref }}), by @${{ github.actor }}"
 on:
-  workflow_dispatch:
-    inputs:
-      ref:
-        description: 'For a PR `refs/pull/<PR-NUMBER>/merge`'
-        required: true
-      headRef:
-        description: 'The ref of HEAD. For a PR `refs/pull/<PR-NUMBER>/head`.'
-        required: true
-      branch:
-        description: 'The branch. For a PR `PR-<PR-NUMBER>`.'
-        required: true
-        default: 'master'
-      baseBranch:
-        description: 'For a PR the base branch, otherwise the same as `branch`.'
-        required: true
-        default: 'master'
-      baseRef:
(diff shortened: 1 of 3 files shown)


.github/workflows/deb-packaging.yml
```diff
@@ -1,8 +1,33 @@
...
+        description: 'The ref of HEAD. For a PR `refs/pull/<PR-NUMBER>/head`.'
+        required: true
+      branch:
+        description: 'The branch. For a PR `PR-<PR-NUMBER>`.'
+        required: true
+        default: 'master'
+      baseBranch:
+        description: 'For a PR the base branch, otherwise the same as `branch`.'
+        required: true
+        default: 'master'
```
darcywong00, line 19: Would branch and baseBranch change on tier? 
alpha = master
beta = beta
stable = stable-xy ?

resources/build/trigger-builds.inc.sh
```diff
@@ -69,30 +69,31 @@ function triggerGitHubActionsBuild() {
...
-      \"client_payload\": { \
+  local DATA="{
+    \"ref\": \"$GIT_REF\", \
+    \"inputs\": { \
         \"ref\": \"$GIT_REF\", \
-        \"sha\": \"$GIT_SHA\", \
+        \"headRef\": \"$GIT_HEAD_REF\", \
         \"branch\": \"$GIT_BRANCH\", \
         \"isTestBuild\": \"$IS_TEST_BUILD\" \
     }}"
```
mcdurdin, line 99: I'm not all that comfortable with this change because this means that we are no longer building a specific commit from the trigger. I was deliberately targeting a specific commit in our existing model.

ermshiperete: Turns out we don't need this variable at all - it's removed in one of the following PRs. 

We're still building a specific commit - see `ref` at the top of `DATA`.

mcdurdin (approved):

PR #9866, opened 2023-10-26
chore(linux): Refactor deb-packaging.sh script 🎬

ermshiperete:
Make use of builder_run_action.

@keymanapp-test-bot skip

### DIFF
--- a/linux/scripts/deb-packaging.sh
+++ b/linux/scripts/deb-packaging.sh
@@ -28,13 +28,11 @@ else
   END_STEP=""
 fi
 
-if builder_start_action dependencies; then
+dependencies_action() {
   sudo mk-build-deps --install --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes' debian/control
-  builder_finish_action success dependencies
-  exit 0
-fi
+}
 
-if builder_start_action source; then
+source_action() {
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

linux/scripts/deb-packaging.sh
```diff
@@ -50,25 +48,23 @@ if builder_start_action source; then
...
     echo ":heavy_check_mark: ${LIB_NAME} API didn't change" >&2
   fi
-  builder_finish_action success verify
-  exit 0
-fi
+}
+
+builder_run_action dependencies  dependencies_action
+builder_run_action source        source_action
+builder_run_action verify        verify_action
```
mcdurdin, line 70: I notice you are no longer doing `exit 0` after each action. Is this change in behaviour intentional?

ermshiperete: Yes, this brings it more in line with how the other `build.sh` scripts work. I can't see a reason why you shouldn't be able to run multiple actions.

mcdurdin: Sounds sensible to me. Just wasn't sure if there was a particular reason it was doing that before :grin:

keyman-server:

PR #9867, opened 2023-10-26
chore(core): Add additional API checks 🎬

ermshiperete:
- Add additional checks for the symbols file:
  - if a line with a method name gets changed, the package version number in that line also needs to be updated
- Add additional checks for the API:
  - for major API changes (methods renamed or removed as reflected in the .symbols file) the API version
    numbers needs to be incremented
  - the API version numbers in the .symbols file and in `CORE_API_VERSION.md` need to match

Closes #9801

@keymanapp-test-bot skip

PR #9877, opened 2023-10-27
chore(linux): Add Core API version number 🎬

### LINKED ISSUE #9801: feat(core): Add API version number
The Keyman Core API usually doesn't change in breaking ways between Keyman versions, so it doesn't make sense to change the version number in the core library (_libkmnkbp0-0 / libkeymancore1_) with each Keyman release.

But then it would be good to document the Core API version somewhere. See [this](https://github.com/keymanapp/keyman/pull/9794#discussion_r1364983057) discussion.

A good indication when the Core API version number needs to be updated is the file `linux/debian/libkeymancore.symbols`. If that file changes the Core API version needs to be bumped:
- the minor version number needs to be bumped if new lines get added
- the major version number needs to be bumped if existing lines get changed or removed

The API verification step in `.github/workflows/deb-packaging.yml` could be enhanced to check that the API version got updated.

ermshiperete:
Split out from #9867.

Closes #9801.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #9902, opened 2023-10-31
fix(developer): Project MRU now saves correctly

### LINKED ISSUE #9899: bug(developer): MRU appears to have two sources - registry and project_mru.xml
welcome page uses `project_mru.xml`:

https://github.com/keymanapp/keyman/blob/c3b2df80bf95b9a00095c901ac202e6cebeb07c1/developer/src/tike/project/Keyman.Developer.System.Project.WelcomeRenderer.pas#L27-L48

but Developer internally still uses `TMRUList` which has a registry source:

https://github.com/keymanapp/keyman/blob/d2e1a41d06ec9c80ee88eaa3e3f5f4c624831983/developer/src/tike/main/mrulist.pas#L112-L145

This means that the Welcome page links often don't work. Recommended fix -- always uses `project_mru.xml` and eliminate the registry setting.

mcdurdin:
Fixes #9899.

Several problems resolved here:
* project mru did not refresh the welcome screen on first load, so the welcome screen was often showing projects that were no longer relevant
* content saved to registry did not match expectations on load, so mru never worked between sessions (always flushed on reload)
* synchronization between simultaneous sessions of TIKE did not work, because the list was only saved when Keyman Developer exited -- meaning the last exited instance of TIKE would overwrite the MRU from earlier sessions, losing the projects opened in those sessions.

Minor tidyup:
* WideString -> string -- legacy cleanup from when Delphi was non-Unicode.
* Reordered functions to group more clearly.

Note: the issue was not that there was both registry and xml content; the source of truth is the registry for Project MRU, and project_mru.xml is copied from that for the welcome screen. So instead of taking the fix recommended in #9899, have addressed this by fixing the above points in the registry code.

# User Testing

* **TEST_WELCOME:** Open Keyman Developer, load a project. Select Project/Close Project. The project name should appear in the list in the Welcome screen. Close Keyman Developer, restart it. The project name should _still_ appear in the list in the Welcome screen, and if you click it, the project should open.

* **TEST_MULTI_SESSION:** Open three instances of Keyman Developer, and open projects in two of the instances. In the third instance, close the project (Project/Close Project). Open various projects in the first two instances. As you switch projects in the other two instances, the list of projects in the third instance should show this (you will need to focus to the third instance in order to see the list refresh).

### DIFF
--- a/developer/src/tike/main/UfrmMain.pas
+++ b/developer/src/tike/main/UfrmMain.pas
@@ -362,6 +362,8 @@ type
     function OpenModelEditor(FFileName: string): TfrmTikeEditor;
     procedure DoCloseCleanup;
 
+    procedure RefreshProjectMRU;
+
   protected
     procedure WndProc(var Message: TMessage); override;
     procedure CreateWnd; override;
@@ -512,10 +514,11 @@ begin
   modActionsMain := TmodActionsMain.Create(Self);
   modActionsModelEditor := TmodActionsModelEditor.Create(Self);
 
+  FChildWindows := TChildWindowList.Create;
+
   FProjectMRU := TMRUList.Create('Project');
   FProjectMRU.OnChange := ProjectMRUChange;
-
-  FChildWindows := TChildWindowList.Create;
+  FProjectMRU.Load;
 
   ShowDebug(False);
 
@@ -863,6 +866,9 @@ var
   i: Integer;
 begin
   FWndProcInit := False;
+
+  RefreshProjectMRU;
+
   for i := 0 to FChildWindows.Count - 1 do
     if FChildWindows[i] is TfrmTikeEditor then
       with FChildWindows[i] as TfrmTikeEditor do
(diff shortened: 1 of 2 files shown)


bharanidharanj:
# Test Results

* **TEST_WELCOME (PASSED):** Tested with the attached PR build Keyman Developer 17.0.203-alpha-test-9902 on Windows 11 OS (VM) and here is my observation: 1. Verified that the project name appears in the Welcome screen. Closed then restarted the Keyman Developer. Able to open the existing project, by clicking on the project name in the Welcome Screen.

bharanidharanj:
# Test Results

* **TEST_MULTI_SESSION (SKIPPED):**  I attempted to open Keyman Developer with multiple instances in the hope of having multiple developer windows visible on my desktop. However, this approach proved ineffective, as clicking multiple times on the Keyman Developer window only resulted in a single developer window opening. Hence, I skipped this test.

bharanidharanj:
@mcdurdin "I encountered difficulty in opening multiple instances of Keyman Developer on the desktop. Are there alternative methods to achieve this?" I need your help. Thanks!

mcdurdin:
Oh sorry! There's an option in Tools/Options to allow multiple instances

rc-swag (approved):
lgtm

keyman-server:

PR #9912, opened 2023-10-31
fix(linux): Fix environment for Debian packages

### LINKED ISSUE #9878: bug(linux): Installed Keyman package shows `-local` tag
Running `km-config --version` with Keyman packages installed incorrectly shows

```
km-config version 17.0.198-alpha-local (package version 17.0.198-1+jammy1)
```

The correct version number would be `17.0.198-alpha-test`.

ermshiperete:
When didn't set the correct environment when building Debian packages with GitHub actions. The version information wasn't included in the source file but instead was generated at binary package build time, at which time not all necessary environment variables were available. This change will create a patch file on-the-fly which gets included in the source package that contains the correct information.

Fixes #9878.

@keymanapp-test-bot skip

### DIFF
--- a/.github/workflows/deb-packaging.yml
+++ b/.github/workflows/deb-packaging.yml
@@ -16,10 +16,8 @@ jobs:
     runs-on: ubuntu-22.04
     outputs:
       VERSION: ${{ steps.version_step.outputs.VERSION }}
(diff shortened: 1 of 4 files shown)


mcdurdin (approved):
RSLGTM

PR #9917, opened 2023-10-31
fix(developer): don't use osk-always-visible on touch devices

### LINKED ISSUE #9845: bug(developer): server-trying to select a keyboard down where keyboard is shown fails
https://github.com/keymanapp/keyman/assets/4498365/422ae958-ae28-4a47-94fb-cd0d454b41d5

In Keyman Developer Server, connecting on a touch device, if you try and select a keyboard that shows below the area where the touch keyboard would be displayed, it instead shows the touch keyboard and fails to touch the menu.

Not clear if this is a Developer Server issue or a KeymanWeb issue or something falling in between the two.

17.0.196-alpha

mcdurdin:
Fixes #9845.

It seems that the logic for `osk-always-visible` is not quite right on touch devices -- the OSK disappears on blur but remains touchable -- so presses in the OSK region emit key events. For Keyman Developer Server, the simple workaround is to only use `<body class="osk-always-visible">` when on desktop devices.

We should review the logic for `osk-always-visible` in KeymanWeb, so that this issue does not arise on touch devices. This patch addresses the issue in Keyman Developer Server, and matches the behaviour we want on touch devices in any case, as we don't really want the OSK visible when blurred, unlike on desktop.

# User Testing

* **TEST_INVISIBLE_OSK_ON_TOUCH_DEVICE:** Open Keyman Developer Server on a touch device (or in touch device emulation in Chrome); ensure there is a Keyman keyboard active for testing. Touch on the touch OSK to input some text into the test box. Then touch outside the text box to hide the OSK. Then, touch where the OSK _was_ visible. This should _no longer_ cause text to be inserted into the test box.

### DIFF
--- a/developer/src/server/src/site/index.html
+++ b/developer/src/server/src/site/index.html
@@ -21,7 +21,7 @@
     <link href="inc/keyboards.css" type="text/css" rel="stylesheet" /> <!-- This script loads the dynamic keyboard fonts -->
   </head>
 
-  <body class='osk-always-visible'>
+  <body>
 
     <header class='navbar navbar-dark bg-dark text-white'>
       <div class='container-fluid'>
--- a/developer/src/server/src/site/test.js
+++ b/developer/src/server/src/site/test.js
@@ -42,6 +42,10 @@ function enableControls(enable) {
 let keymanInitialized = false;
 enableControls(false);
 
+if(!keyman.util.isTouchDevice()) {
+  document.body.className = 'osk-always-visible';
+}
+
 keyman.init({
   ui:'button',
   resources:'/resource/',


rc-swag (approved):
lgtm

bharanidharanj:
# Test Results

* **TEST_INVISIBLE_OSK_ON_TOUCH_DEVICE (FAILED):** Tested with Keyman Developer 17.0.204-alpha-test-9917 on Windows 11 OS (VM) and here is what I did: 1. Opened the Keyman Developer Server in a touch device emulation in Chrome browser. 2. Select Khmer_Angkor keyboard from the Keyboard list box. Able to see the OSK appear on the screen. 3. Typed some letters using the OSK. Then, touched outside of the the text box. But the OSK is still visible on the screen. 4. Again, typed some texts using OSK. The typed text appeared in the text box. Seems to be an issue.

bharanidharanj:
# Test Results

* **TEST_INVISIBLE_OSK_ON_TOUCH_DEVICE (PASSED):** Retested as per your instructions and here is my observation: 1. Selected a keyboard from the Keyboard list. (eg., Khmer_Angkor). 2. Typed some text using the OSK. 3. Clicked outside of the textbox. 4. Then, I clicked a letter key from the OSK. Verified that no letter appears on the text box.

mcdurdin:
@bharanidharanj I think you might have missed the step on using mobile emulation mode. But I just retested that here so I think we are good to go.

keyman-server:

bharanidharanj:
> @bharanidharanj I think you might have missed the step on using mobile emulation mode. But I just retested that here so I think we are good to go.

@mcdurdin Oops! I have done this test on the Mozilla Firefox browser. However, I retested this in Chrome browser and it seems to be working as expected.  Thanks for your guidance.

PR #9921, opened 2023-11-01
fix(developer): warn if layer switch key is missing ID

### LINKED ISSUE #8000: bug(developer): Touch layout with missing key id's
Your `Núm` and Emoji buttons don't have an `ID` in at least the postmodern_english_uk_natural keyboard. I suspect it's related to that. However, I don't know what the ID should be... I would expect `K_NUMERALS` and `K_SYMBOLS` to be acceptable.

image

_Originally posted by @LornaSIL in https://github.com/keymanapp/lexical-models/issues/180#issuecomment-1375917261_

A user was asking for help why Keyman for Android failed to switch layers on the "emoji" and "Núm" keys. @LornaSIL pointed out the key IDs are blank.

I've marked 1 of the 4 instances here
https://github.com/keymanapp/keyboards/blob/22b4b3ab4eb0b05f3407c21f57c135723495ef58/release/p/postmodern_english_us_natural/source/postmodern_english_us_natural.keyman-touch-layout#L280-L284

Shouldn't Keyman Developer flag this during compilation?

mcdurdin:
Fixes #8000.

In earlier versions of KeymanWeb, it was assumed that layer switch keys would not necessarily need an identifier, as they would not be generating a standard KeymanWeb key event. However, this assumption does not really hold, so we should be warning on missing identifiers for layer switch keys.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc-kmn/src/kmw-compiler/validate-layout-file.ts
+++ b/developer/src/kmc-kmn/src/kmw-compiler/validate-layout-file.ts
@@ -109,7 +109,7 @@ function CheckKey(
   //
 
   if(FId.trim() == '') {
-    if(!([TouchLayout.TouchLayoutKeySp.blank, TouchLayout.TouchLayoutKeySp.spacer].includes(FKeyType)) && FNextLayer == '') {
+    if(!([TouchLayout.TouchLayoutKeySp.blank, TouchLayout.TouchLayoutKeySp.spacer].includes(FKeyType))) {
       callbacks.reportMessage(KmwCompilerMessages.Warn_TouchLayoutUnidentifiedKey({layerId: layer.id}));
     }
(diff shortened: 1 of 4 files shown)


jahorton (approved):
Should I also investigate the Web side of this and see why the layer-switch wasn't occurring?

Granted... our definitions in common/web/types do tend to assert that keys have IDs, so the work done here is good to have anyway.

mcdurdin:
> Should I also investigate the Web side of this and see why the layer-switch wasn't occurring?

No, I don't think we need to update Web. I think it probably was a very early design choice which wasn't the best. I now think the assumption that the key has an id is appropriate and so we should be catching it at compile time.

keyman-server:

PR #9922, opened 2023-11-01
fix(developer): restore selection in layout builder even with duplicate ids

### LINKED ISSUE #7851: bug(developer): Layer Refresh Focuses on Wrong Key When Key ID Occurs Twice
### Describe the bug

This is a little challenging to describe but actually frequent with Ethiopic keyboards where Ethiopic letters get mapped to both upper and lowercase Latin letters.  For example, `K_P` with no modifier, and `K_P` with a shift modifier.

Lets say `K_P+shift` occurs at the top of the keyboard, and `K_P` at the bottom.  When an edit is made to the `K_P` key (such as backspacing on a padding value like `100`, it temporarily becomes `10`), the layout will refresh but the focus will now jump to the `K_P+shift` key. 

I observe this happening in many scenarios, my guess is that the focus resets to the first occurrence of the key ID which will be found on the lowest row number.

### Reproduce the bug

Described above, the scenario can be tried out on the GFF Mesobe Fidelat and GFF Harege Fidelat keyboards, GFF Amharic as well I believe.

### Expected behavior

Focus does not change, and edits continue on the original key.

### Related issues

_No response_

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [X] Keyman Developer
- [ ] KeymanWeb

### Keyman version

15.0.273

### Operating system

Windows 11

### Device

Windows Laptop

### Target application

N/A

### Browser

N/A

### Keyboard name

N/A

### Keyboard version

N/A

### Language name

N/A

### Additional context

_No response_

mcdurdin:
Fixes #7851.

Note that this fixes the selection pop issue with duplicate ids in some scenarios but not all. For example, the undo stack doesn't have enough detail at this point to be able to track this. This fix addresses the most egregious issue relating to key size adjustment and similar.

# User Testing

* **TEST_KEY_RESIZE:** Open a touch layout, and give two keys on the same layer the same ID (e.g. `K_Q`). Select each key in turn and resize it by dragging on the right hand border. When you resize, the selected key should remain the same -- previously the selection would jump to the first key with the same ID.

### DIFF
--- a/developer/src/tike/xml/layoutbuilder/builder.js
+++ b/developer/src/tike/xml/layoutbuilder/builder.js
@@ -863,10 +863,17 @@ $(function() {
 
   this.rescale = function () {
     builder.saveUndo();
-    var keyId = builder.selectedKey().data('id');
+    const k = builder.selectedKey();
+    const keyId = k.data('id');
+    const keyItems = $('#kbd .key').filter((_index,item) => $(item).data('id') === keyId);
+    const keyItemIndex = keyItems.indexOf(k.length ? k[0] : null);
     builder.prepareLayer();
-    if (keyId !== null)
-      builder.selectKey($('#kbd .key').filter(function (index) { return $(this).data('id') === keyId; }).first());
+    if (keyId !== null && keyItemIndex >= 0) {
+      const newKeyItems = $('#kbd .key').filter((_index,item) => $(item).data('id') === keyId);
+      if(keyItemIndex < newKeyItems.length) {
+        builder.selectKey(newKeyItems[keyItemIndex]);
+      }
+    }
   };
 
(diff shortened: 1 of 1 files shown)


bharanidharanj:
# Test Results

* **TEST_KEY_RESIZE (PASSED):** Tested with the attached PR build (Keyman 17.0.204-alpha-test-9922) on Windows 11 OS (Virtual Machine) and here is my observation: 1. Verified that the selected key can be resized and it does not jump to the first key with the same ID. Seems to be working as expected.

jahorton (approved):

keyman-server:

PR #9932, opened 2023-11-03
fix(developer): kmc code generation for context(n) in context

### LINKED ISSUE #9930: bug(developer): kmc generates invalid code for `context(n)` in context part of rule
common/test/resources/keyboards/test_deadkeys.kmn has a divergent build with kmc, specifically with one rule:

```
'?' any(deadQwerty) context(2) + '.' > '(' index(liveQwerty, 2) index(liveQwerty, 2) ')'
```

Difference between kmcomp and kmc:

```js
if(k.KFCM(3,t,['?',{t:'a',a:this.s_deadQwerty_12},{t:'c',c:2}])){  // kmcomp
if(k.KFCM(3,t,['?',{t:'a',a:this.s_deadQwerty_12},{t:'c',c:1}])){  // kmc
```

That is, the `index` parameter appears to be offset by 1.

kmc-17.0.205-alpha vs kmcomp 16.0.138-stable.

Discovered by @ermshiperete

mcdurdin:
Fixes #9930.

* Fixes the index offset calculation, with +1 for >=v10.0 keyboards and -1 for &lt;v10.0 keyboards (difference is because older version keyboards have context offsets calculated in the opposite direction). Note that this happened because the `CODE_CONTEXTEX` index is stored in memory as 0-based in kmc, but was 1-based in kmcomp.

* Adds unit tests for v9.0 and v10.0 keyboards for `context(n)` in context.

* Adds unit tests for v9.0 and v10.0 keyboards for `context(n)` in output. These were already correct but now we can be assured that they are being tested.

* Enables code coverage for kmw-compiler, and sets a lower threshold for pass because coverage is still just over 70%.

@keymanapp-test-bot skip


ermshiperete (approved):
LGTM

keyman-server:

PR #9940, opened 2023-11-06
fix(developer): Sentry in Server should honour reporting settings

### LINKED ISSUE #9939: bug(developer): sentry is always enabled in Server (both backend and frontend)
This needs to be controlled by command line parameter (standalone use) or by registry value (Keyman Developer use)

https://github.com/keymanapp/keyman/blob/8e23b11dfcb357e82f37d1b04dbcf14a3b9e745d/developer/src/server/src/index.ts#L3-L8

https://github.com/keymanapp/keyman/blob/067028cec2d7448de0f6253030e3d0804d574758/developer/src/server/src/site/lib/sentry/init.js.in#L10-L14

Registry value:

https://github.com/keymanapp/keyman/blob/3336539ec5f2864ed892c29976cb26a5c4b5e3d1/common/windows/delphi/general/RegistryKeys.pas#L182

mcdurdin:
Fixes #9939.

Splits KeymanSentry.ts into common functionality in developer-utils, and TestKeymanSentry.ts. Uses refactored KeymanSentry.ts in developer-utils and updates the loader in Server to run from that.

The changes to tray.ts and friends are required to deal with a top-level await issue with the shutdown call.

The top-level await issue is described in detail at https://marc.durdin.net/2023/11/dynamic-import-on-node-js-with-circular-dependencies-leads-to-interesting-failure-modes/

tl;dr: shutdown needed to do a proper Sentry shutdown. The shutdown code was in index.ts. This led to a circular dependency and caused a silent and abrupt exit code 13 in Node.js. Moved shutdown code into shutdown.ts to resolve this.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.206-alpha](https://keyman.com/downloads/releases/alpha/17.0.206)

PR #9944, opened 2023-11-06
fix(web): Fix clearing of deadkeys

### LINKED ISSUE #8568: bug(web): different output with lao_phonetic keyboard
### Describe the bug

The [lao_phonetic keyboard, v1.1.2](https://keyman.com/keyboards/lao_phonetic) can sometimes generate different output on web vs Windows (Core) -- this seems to be most visible when backspacing.

This could possibly be caused by one or more of the following:

1. differences in the compiler
2. deadkey handling
3. backspace handling

At this point, I am just documenting the difference!

### Reproduce the bug

1. Type `phaa saa laaw ko` then press <kbd>bksp</kbd>, <kbd>bksp</kbd>
2. On Windows, you'll see the following output: ພາສາລາວ
3. In the Developer debugger, you can also see the deadkeys: ພາ￼ສາ￼ລາວ￼ (all `dk(1)`)
4. However, on web, you'll see this: ພາສາລາົວ (with 0ebb inserted prior to final 0ea7 ວ).

### Expected behavior

_No response_

### Related issues

_No response_

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [ ] Keyman Developer
- [X] KeymanWeb

### Keyman version

16.0.139

### Operating system

Web

### Device

_No response_

### Target application

_No response_

### Browser

any?

### Keyboard name

lao_phonetic

### Keyboard version

1.1.2

### Language name

Lao

### Additional context

_No response_

ermshiperete:
- don't clear deadkeys after pressing backspace
- add unit test
- also some small refactorings

Fixes #8568.

# User Testing

**TEST_8568**: Verify that #8568 is fixed

1. Type `phaa saa laaw ko` then press <kbd>bksp</kbd>, <kbd>bksp</kbd>
2. Verify you get the following output: ພາສາລາວ


common/web/input-processor/tests/cases/inputProcessorTests.js
```diff
@@ -175,4 +176,51 @@ describe('InputProcessor', function() {
       });
     });
   });
-});
\ No newline at end of file
+
+  describe('Deadkeys bug #8568', function () {
```
mcdurdin, line 180:
```suggestion
  describe('Deadkeys bug #8568 - backspace should not reset all deadkeys', function () {
```

common/test/keyboards/baseline/k_020___deadkeys_and_backspace.kmn
```diff
@@ -22,3 +23,6 @@ group(main) using keys
 + '4' > dk(4a) dk(4b)
 + '5' > 'a' dk(5a) dk(5b) 'b'
 + '6' > 'a' dk(6a) dk(6b) 'b' dk(6c) dk(6d)
++ '7' > 'c' dk(7) 'd'
+'c' 'd' + [K_BKSP]       > ' fail'
+'c' dk(7) 'd' + [K_BKSP] > ' ok'
```
mcdurdin, line 28: This is a slightly different test case -- testing the matching of rules rather than the default backspace behaviour. Can we move it to a separate keyboard and document accordingly?

mcdurdin (changes requested):
This looks okay, but I have some questions around the testing of the default backspace handling with deadkeys -- are all the scenarios in k_020__deadkeys_and_backspace being covered? Are they being tested automatically at present or did you manually test?

Also can you split out the change you made to k_020 because it is a different test case?

common/test/keyboards/baseline/k_014___groups_and_virtual_keys.kmn
jahorton, line 1: While I'm guessing this test keyboard was under consideration for use, it looks like that got cancelled.  Not sure about the removal of a few comments from lines 25 and 26, so I'll leave that aspect of the review to others.

common/test/keyboards/baseline/k_020___deadkeys_and_backspace.kmn
jahorton: If the actual test uses the context `c dk(u) d f` and then processes two backspaces, I think that would be adequate for the test.  It's just that there's no special rule for the `f`... and the first backspace _would_ be default.  Keeping in mind that the actual error bug resulted from the _lack_ of a deadkey after a backspace triggering a _backspace_-based rule, I'm personally fine with the form of the check.

Thing is, this test keyboard _is_ called "deadkeys and backspace".  You sure it doesn't fit here?  The other tests handle deadkeys between the latest context char and the context char before it; this test handles a deadkey based one char earlier than _that_.

ermshiperete: yes, having two backspaces is definitely better since the deadkey removal could happen before or after processing the first one. I updated the test.

common/test/keyboards/baseline/k_020___deadkeys_and_backspace.kmn
```diff
@@ -6,8 +6,9 @@ c 3. One deadkey and one char in context      dk(3) 'a' + BKSP = nul
 c 4. Two deadkeys in a row in context         dk(4a) dk(4b) + BKSP = nul
 c 5. One char and two deadkeys in context     'a' dk(5a) dk(5b) 'b' + BKSP = 'a'
 c 6. One char and two deadkeys and one char and two deadkeys in context     'a' dk(6a) dk(6b) 'b' dk(6c) dk(6d) + BKSP = 'a'
-c keys: [K_1][K_BKSP][K_2][K_BKSP][K_3][K_BKSP][K_4][K_BKSP][K_5][K_BKSP][K_6][K_BKSP]
-c expected: 78aa
+c 7. Two chars and one deadkey in between in context: 'c' (dk7) 'd' + BKSP = 'ok'
```
mcdurdin, line 9:
```suggestion
c 7. Tests behaviour when deleting two characters with deadkey prior to first character,
c    verifying that deadkeys are preserved with the first backspace event: 
c    'c' (dk7) 'de' + BKSP + BKSP = 'ok'
```

mcdurdin (approved):

bharanidharanj:
# Test Results

* **TEST_8568 (PASSED):** Tested with the attached PR build (Keyman Developer 17.0.206-alpha-test-9944) on Windows 10 OS and here is my observation: 1. Installed Keyman 17.0.206-alpha build in the system. 2. Installed Lao Phonetic keyboard. 3. Opened the text editor. 3. Verified that it outputs the expected result. 4. Opened Lao Phonetic project in Keyman developer. Repeat step 1 as it mentioned in the User Testing. I have attached the screenshot for your reference. 5. Open the keyboard in Chrome mobile emulator mode.

..in Windows

   


..in the Debugger window

 


..in Web

keyman-server:

PR #9951, opened 2023-11-07
fix(web): Ignore `osk-always-visible` on non-desktop devices

### LINKED ISSUE #9918: bug(web): `osk-always-visible` has strange behaviours on touch devices
From #9917:

> It seems that the logic for `osk-always-visible` is not quite right on touch devices -- the OSK disappears on blur but remains touchable -- so presses in the OSK region emit key events. For Keyman Developer Server, the simple workaround is to only use <body class="osk-always-visible"> when on desktop devices.
>
> We should review the logic for `osk-always-visible` in KeymanWeb, so that this issue does not arise on touch devices.

ermshiperete:
See also #9917 for the Keyman Developer part.

Fixes #9918.

# User Testing

* **TEST_WEB:** Open Keyman Web on a touch device (or in touch device emulation in Chrome); ensure there is a Keyman keyboard active for testing. Touch on the touch OSK to input some text into the test box. Then touch outside the text box to hide the OSK. Then, touch where the OSK _was_ visible. This should _no longer_ cause text to be inserted into the test box.

### DIFF
--- a/web/src/engine/osk/src/views/oskView.ts
+++ b/web/src/engine/osk/src/views/oskView.ts
@@ -964,8 +964,10 @@ export default abstract class OSKView extends EventEmitter<EventMap> implements 
    * Performs the _actual_ logic and functionality involved in hiding the OSK.
    */
   protected finalizeHide() {
-    if(document.body.className.indexOf('osk-always-visible') >= 0) {
-      return;
+    if (document.body.className.indexOf('osk-always-visible') >= 0) {
+      if (this.hostDevice.formFactor == 'desktop') {
+        return;
+      }
     }
 
     if(this._Box) {


mcdurdin (approved):

bharanidharanj:
# Test Results

* **TEST_WEB (PASSED):** Tested with the attached PR build (Keyman 17.0.207-alpha-test-9951)on Windows 10 OS and here is my observation: 1. Opened Khmer Angkor project via Keyman Developer. 2. Opened the keyboard in touch emulation in Chrome. 3. Selected Khmer Angkor keyboard for testing. 4. Typed some text using the OSK. Then, touched outside the text box to hide the OSK. 5. Touched where the OSK was visible. Verified that touch on the space (previously the OSK was visible on that space) does not show any text on the test box. seems to be working as expected.

keyman-server:

PR #9965, opened 2023-11-08
chore(linux): Fix GHA triggering

ermshiperete:
When we build a PR on TC we use `pull/ID/head` for the build, not `pull/ID/merge`, i.e. the tip of the feature branch and not a merge with the base branch. This means that during the build `pull/ID/merge` is unknown. This change modifies the trigger to use `pull/ID/head` for GHA builds as well.

It's still a mystery why `pull/ID/merge` was known to builds in most cases but failed for some PRs.

@keymanapp-test-bot skip


ermshiperete:
Linux integration tests will fail until #9967 is merged.

mcdurdin (approved):

keyman-server:

Issue #9438, opened 2023-08-08
bug(developer): visual-keyboard-compiler throwing instead of callbacks 🙀 


srl295:
```
fr-t-k0-azerty.xml - info KM05002: Building keyboards/3.0/fr-t-k0-azerty.xml
fr-t-k0-azerty.xml - hint KM00009: Virtual key 'BKQUOTE' is not found in the CLDR VKey Enum table.
fr-t-k0-azerty.xml - hint KM00009: Virtual key 'MINUS' is not found in the CLDR VKey Enum table.
fr-t-k0-azerty.xml - hint KM00009: Virtual key 'LBRKT' is not found in the CLDR VKey Enum table.
fr-t-k0-azerty.xml - hint KM00009: Virtual key 'COLON' is not found in the CLDR VKey Enum table.
fr-t-k0-azerty.xml - hint KM00009: Virtual key 'BKQUOTE' is not found in the CLDR VKey Enum table.
fr-t-k0-azerty.xml - hint KM00009: Virtual key 'EXCLAMATION' is not found in the CLDR VKey Enum table.
fr-t-k0-azerty.xml - fatal KM05001: Unexpected exception: Error: Internal Error: could not find key id="extra" in layer "base", row "3"

Call stack:
Error: Internal Error: could not find key id="extra" in layer "base", row "3"
    at LdmlKeyboardVisualKeyboardCompiler.compileHardwareLayer (file:///usr/local/lib/node_modules/@keymanapp/kmc/node_modules/@keymanapp/kmc-ldml/build/src/compiler/visual-keyboard-compiler.js:32:27)
    at LdmlKeyboardVisualKeyboardCompiler.compile (file:///usr/local/lib/node_modules/@keymanapp/kmc/node_modules/@keymanapp/kmc-ldml/build/src/compiler/visual-keyboard-compiler.js:15:22)
    at buildLdmlKeyboardToMemory (file:///usr/local/lib/node_modules/@keymanapp/kmc/build/src/commands/buildClasses/BuildLdmlKeyboard.js:65:27)
    at async BuildLdmlKeyboard.build (file:///usr/local/lib/node_modules/@keymanapp/kmc/build/src/commands/buildClasses/BuildLdmlKeyboard.js:16:31)
    at async build (file:///usr/local/lib/node_modules/@keymanapp/kmc/build/src/commands/build.js:106:22)
[...]

PR #9979, opened 2023-11-10
fix(developer): vis kbd callbacks instead of throw  🙀

srl295:
- fix API and call sites

For: #9438

@keymanapp-test-bot skip


srl295:
rebasing hoping it will trigger…

srl295:
oh… only developer was changed, that's why

mcdurdin (approved):

keyman-server:

PR #9985, opened 2023-11-13
chore(developer): only build source files 🦕

mcdurdin:
For batch builds (e.g. project/compile all) of v2.0 projects, we only want to build files that are listed in the project's $SOURCEPATH, so that the project can contain extra folders with source file types that won't break a project-level build.

Any source file types in extra folders can still be built one-by-one, which is helpful for investigation and test for keyboard authors, but will not be built if project build is selected.

This commit also formalizes the source/ SourcePath for upgraded projects which contain a source/ folder. If the project to be upgraded does not contain the source/ folder, then no SourcePath is assigned, and any source file in any folder will be built.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

Issue #9948, opened 2023-11-06
feat(developer): Keyman Developer v2.0 Projects


mcdurdin:
## Background

This is a meta-issue collecting work items relating to support for v2.0 projects in TIKE.

* #2761

## TODO
- [ ] Folder tree file view
- [ ] Fix opening project file from project view
- [ ] Add XML LDML keyboards to Keyboards list? (Treat as keyboards, in any case)
- [ ] Open file from Explorer with no project open -- "Open as Folder" UfrmMain.pas:1160
- [ ] Open file that is not in active project -- "Open as Folder"
- [ ] Project upgrade -- source folder settings
- [ ] Project Settings -- v2.0 settings
- [ ] Clone Project / Rename Project
- [ ] Support re-loading directory-based projects (where .kpj may not exist) UfrmMain.pas:553
- [ ] Consider upgrade project and default paths?
- [ ] Linting of project file layout and file names
- [ ] Add kmc to server (for performance reasons)

## Completed in #xxxx

For all operating modes:

- [x] Remove blank projects / untitled projects mode
- [x] Add Open Project Folder buttons and links
- [x] Remove Project Save As
- [x] Render uses in-memory XML
- [x] Upgrade to v2.0 project file (along with checks)

For v2.0 projects:

- [x] Remove 'Add File to Project'
- [x] Populate Files to list all files in folder (note: consider efficiency here as currently files are parsed for subfile data)
- [x] Version 2.0 project options and defaults

PR #9986, opened 2023-11-13
chore(developer): manage SourcePath in project upgrade 🦕

mcdurdin:
Relates to #9948.

When upgrading a project to v2.0, it is important that all source files are in the same folder. This change makes the upgrade process verify that this is the case and blocks the upgrade if there are source files in multiple folders.

Source files are .kmn, .xml (ldml keyboard), .kps, and .model.ts.

Note: cleanups in other units are removing unused code.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

PR #9996, opened 2023-11-13
chore(developer): handle project version checks cleanly in kmc 🦕

mcdurdin:
Relates to #9948.

While the schema validation does check the version already, it gives a fairly obtuse message, and we can do better.

Found a secondary issue with a iterator type guard needed in kpj-file-reader when preparing unit test, which was probably masked in other tests by not verifying all messages generated, so updated all infrastructure message tests to verify the full set of messages generated.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

keyman-server:

Issue #10004, opened 2023-11-14
bug(common): truthful test results, print error logs


srl295:
- always use `std::exit(EXIT_FAILURE)` if there is an error, not other random numbers. Per POSIX, only the lower 8 bits of the result are uses. This was causing tests to show passing that were failng.
- use `--print-errorlogs` with meson so that when there are errors, the error log gets included.

PR #10005, opened 2023-11-14
fix(common): dx: don't call exit(0) on failure  🙀

srl295:
Fixes: #10004

- always use std::exit(EXIT_FAILURE) if there is an error.
- use --print-errorlogs with meson to show errors to the log
- 32 bit Debug_UnicodeString()

@keymanapp-test-bot skip


srl295:
@ermshiperete  what is this?

> [linux/scripts] ## verify starting...
> [linux/scripts] OK: libkeymancore API didn't change
> [linux/scripts] ERROR: libkeymancore.symbols file got changed without changing the package version number of the symbol

trying an update from master…

ermshiperete:
That's a bug in the GHA script 🙁. See my latest PR.

mcdurdin (approved):

keyman-server:

PR #10016, opened 2023-11-16
fix(developer): projects 2.0 internal path enumeration 🦕

mcdurdin:
Relates to #9948.

Restricts enumeration of files for the project to the project folder and the SourcePath folder. This prevents problems where a project may be in a folder with many subfolders which would take a long time to enumerate, and avoids confusion where there are source-type files in other folders.

At the same time, sorts out forward slash vs backslash in paths. While forward slash works in many scenarios, there are several filename manipulation functions, such as ExpandFileName, which would build valid but non-optimal paths when forward slashes were encountered, which cascaded into files appearing to be different and presentation issues.

For the Project view, the Distribution tab now shows file relative paths, which helps with organization. Have opted _not_ to show the relative paths in the other tabs, because that information is visible when the file details are expanded, and because those files should always be in SourcePath anyway.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tike/project/Keyman.Developer.System.Project.ProjectFile.pas
+++ b/developer/src/tike/project/Keyman.Developer.System.Project.ProjectFile.pas
@@ -134,8 +134,8 @@ const DefaultProjectOptions: array[TProjectVersion] of TProjectOptionsRecord = (
   ProjectType: ptKeyboard;
   Version: pv10
 ), ( // 2.0
-  BuildPath: '$PROJECTPATH/build';
-  SourcePath: '$PROJECTPATH/source';
+  BuildPath: '$PROJECTPATH\build';
+  SourcePath: '$PROJECTPATH\source';
   CompilerWarningsAsErrors: False;
   WarnDeprecatedCode: True;
   CheckFilenameConventions: False;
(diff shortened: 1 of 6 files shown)


developer/src/tike/project/Keyman.Developer.System.Project.ProjectFile.pas
```diff
@@ -621,8 +621,8 @@ function TProjectFile.IsSourceFile: Boolean;
     Exit(True);
 
   // Only return true if the file is directly in the ProjectOptions.SourcePath folder
-  SourcePath := ReplaceStr(IncludeTrailingPathDelimiter(FProject.ResolveProjectPath(FProject.Options.SourcePath)), '/', '\');
-  FilePath := ReplaceStr(ExtractFilePath(FFileName), '/', '\');
+  SourcePath := DosSlashes(FProject.ResolveProjectPath(FProject.Options.SourcePath));
```
darcywong00, line 624: Just checking this intentionally no longer has `IncludeTrailingPathDelimiter`?

mcdurdin: Yes, because `ResolveProjectPath` does that now. (See https://github.com/keymanapp/keyman/pull/10016/files#diff-123b0a7530c20259d4b04d281a9a8dcd1c2051f83b8298f789f516f8d4495984R1231)

darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.219-alpha](https://keyman.com/downloads/releases/alpha/17.0.219)

PR #10020, opened 2023-11-16
fix(linux): Fix packaging GHA

ermshiperete:
For some reason with some PR builds we got an incorrect base ref. This change makes better use of the information that GitHub provides and so should work around the problem.

@keymanapp-test-bot skip


srl295 (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.213-alpha](https://keyman.com/downloads/releases/alpha/17.0.213)

Issue #9961, opened 2023-11-07
bug(core): ldml don't require startContext 🙀 


srl295:
the compilation /  LDML test runner crashes if you don't have at least:

```xml
<startContext to="" />
…
```

the runner (or compiler) should just default to an empty context if none is provided.

PR #10021, opened 2023-11-16
fix(core): dx: ldml: startContext should be optional in ldml test data 🙀

srl295:
- per spec
- update tests to remove these

Fixes: #9961

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

Issue #9073, opened 2023-06-23
spec: Text Selection and Key Processing


rc-swag:
### Keyman apps

- [X] Keyman for Android
- [X] Keyman for iPhone and iPad
- [X] Keyman for Linux
- [X] Keyman for macOS
- [X] Keyman for Windows
- [X] Keyman Developer
- [X] KeymanWeb
- [ ] Other - give details at bottom of form

ermshiperete:
# Text Selection Tests

(Selection and insertion pointer marked with `⁞` throughout this document)

## General test scenarios

1. With a keyboard that has the rule `'^' + [K_A] > 'Â'`:
   `^⁞X the quick brown fox⁞A` in doc. Typing `BKSP`.
   Expected result: `^⁞A`
2. With a keyboard that has the rule `'^' + [K_A] > 'Â'`:
   `^⁞X the quick brown fox⁞` in doc. Typing `BKSP`, then `A`.
   Expected result: `Â⁞`
3. With a keyboard that has the rule `'^' + [K_A] > 'Â'`:
   `⁞^⁞` in doc, selection as indicated, typing `A`.
   Expected result: `A⁞`
4. With a keyboard that has the rule `'^' + [K_A] > 'Â'`:
   `^⁞X⁞` in doc. Typing `A`.
   Expected result: `Â⁞`
5. With a keyboard that has the rule `'^' + [K_A] > 'Â'`:
   `^⁞X the quick brown fox⁞` in doc. Typing `A`.
   Expected result: `Â⁞`
6. With a keyboard that has the rule `'^' + [K_BKSP] > 'foo'`:
   `^⁞X⁞` in doc. Typing `BKSP`.
   Expected result: `^⁞`
7. With a keyboard that has the rule `'^' + [K_BKSP] > 'foo'`:
   `^⁞X⁞` in doc. Typing `BKSP`, `BKSP`.
   Expected result: `foo⁞`
8. `^⁞X⁞` in doc. Pressing and releasing `SHIFT` key.
   Expected result: `^⁞X⁞` (no change). On Touch keyboard switch to Shift
   layer. (cf #7866)
9. `a⁞bc de⁞` in doc with selection starting at `b`. Typing `BKSP`.
   Expected result: `a⁞`. (cf #8827)
10. `a⁞bc de⁞` in doc with selection starting at `e`. Typing `BKSP`.
    Expected result: `a⁞`. (cf #8827)

## Start-of-sentence tests

[...]

PR #10026, opened 2023-11-17
chore(core): Add test keyboard for text selection tests

ermshiperete:
Part of #9073.

@keymanapp-test-bot skip


mcdurdin (approved):
We _could_ remove the unnecessary files, e.g. all the .md files, etc, because they are a bit noisy, but at the same time, they are just boilerplate so they aren't getting in the way.

keyman-server:

PR #10062, opened 2023-11-23
feat(developer): warn on usage of virtual keys in rule output

### LINKED ISSUE #10059: chore(developer): add warning or error if virtual key used in output of rule
See also #10049.

This has never been officially supported and never worked properly; Core never passed anything to Engines, so vk output has not worked at all since Core was integrated anyway. Time to let our keyboard developers know!

mcdurdin:
Fixes #10059.

Use of the unsupported and undocumented virtual key output, that doesn't work in recent Keyman versions, at all, now results in a build warning.

Only a warning, because it did kinda work in old versions of Keyman.

Sample output:

```
$ ./build.sh --compiler /c/Projects/keyman/app/developer/src/kmc/build/src/kmc.js build -k release/g/gandhari/
...
gandhari.kmn:118 - warn KM020AF: Virtual keys are not supported in output
...
```

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/common/include/kmn_compiler_errors.h
+++ b/developer/src/common/include/kmn_compiler_errors.h
@@ -240,6 +240,8 @@
 
 #define CHINT_UnreachableRule                              0x000010AE
 
+#define CWARN_VirtualKeyInOutput                           0x000020AF
+
 #define CERR_BufferOverflow                                0x000080C0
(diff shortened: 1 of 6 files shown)


rc-swag (approved):
lgtm

keyman-server:

PR #10073, opened 2023-11-23
fix(core): memory management of options in action struct

### LINKED ISSUE #10067: bug(core): when calling km_core_state_get_actions, persist_options contains unexpected values
### Describe the bug

After processing a key with Keyman Core, calling km_core_state_get_actions does not return a valid key value pair in the persist_options field as expected. 

### Reproduce the bug

1. use the Armenian Mnemonic keyboard
2. type[alt-v]
3. expected: "√"
4. type[w]
5. expected: "" -- ("√" is deleted)

This should toggle the option 'option_ligature_ew' from a value of 0 (zero) to 1 (one)

However, calling the API returns invalid values, so they will not be persisted correctly on the platform and may cause the app to crash while attempting to parse them.

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for Linux
- [X] Keyman for macOS
- [ ] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb

### Keyboard name

Armenian Mnemonic

mcdurdin:
Fixes #10067.

Management of memory for persisted options was wrong in the action struct, as the members key and value would be freed immediately after being added to the temporary vector (because the vector was of the struct rather than of the class).

Given the struct is a C struct, we need the memory management to be explicit, so we now release() each option into the vector as we create it, which means that its member values will not be freed when the option is then immediately deleted. (This allows us to use the initial copy of the members of option that option() constructor does.)

Added the release() function as that was a relatively clear way of indicating that the contents of the structure are now owned by the caller, following the pattern from std::unique_ptr.

Finally, the unit test for persisted options was in the action_api.cpp test module, but it was never called, so this was not being tested. Now it is.

@keymanapp-test-bot skip

### DIFF
--- a/core/src/option.cpp
+++ b/core/src/option.cpp
@@ -43,6 +43,14 @@ option::option(km_core_option_scope s, char16_t const *k, char16_t const *v)
   }
 }
 
+km_core_option_item
+option::release() {
+  km_core_option_item opt = *this;
+  key = nullptr;
+  value = nullptr;
+  return opt;
+}
+
 // TODO: Relocate this and fix it
 json & km::core::operator << (json &j, abstract_processor const &)
 {
--- a/core/src/action.cpp
+++ b/core/src/action.cpp
@@ -95,13 +95,15 @@ km_core_actions * km::core::action_item_list_to_actions_object(
         output.push_back({KM_CORE_CT_MARKER,{0},{action_items->marker}});
         break;
       case KM_CORE_IT_PERSIST_OPT:
+      {
         // TODO: lowpri: replace existing item if already present in options vector?
-        options.push_back(km::core::option(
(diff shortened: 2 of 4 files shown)


core/src/option.cpp
```diff
@@ -43,6 +43,14 @@ option::option(km_core_option_scope s, char16_t const *k, char16_t const *v)
   }
 }
 
+km_core_option_item
+option::release() {
```
darcywong00, line 47: nit - the code style in this file seems to have open braces on a new line.
Maybe also add a short documententation of the memory complexity😄

darcywong00: nvm I see you have it in the header

darcywong00 (approved):
lgtm

keyman-server:

PR #10089, opened 2023-11-28
fix(web): Also move source map

ermshiperete:
Previously we renamed `dom-keyboard-loader.mjs` but forgot to move the source map along.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.219-alpha](https://keyman.com/downloads/releases/alpha/17.0.219)

Issue #9451, opened 2023-08-10
bug(core): ldml key-not-found not quite right 🙀 


srl295:
From review at https://github.com/keymanapp/keyman/pull/9405/files#r1289675373

> Some frame keys cause a context invalidation. Frame keys and modifier keys etc then get kicked back to the OS, but any US English character-generating key gets added as an action

ldml processor doesn't do this yet.

Also see https://unicode-org.atlassian.net/browse/CLDR-17262 we may need some spec improvement here.

mcdurdin:
> but any US English character-generating key gets added as an action

Note that we resolved that unassigned keys do nothing on ldml keyboards.

srl295:
> > but any US English character-generating key gets added as an action
> 
> Note that we resolved that unassigned keys do nothing on ldml keyboards.

okay, but in reading the kmx code, i'm not sure how I distinguish between an unassigned key and a "frame key and modifier etc".  

Do i need to check for   `KM_CORE_VKEY_ENTER ||   KM_CORE_VKEY_SHIFT || … ||  KM_CORE_VKEY_PRTSCN` etc?

I could check for _every base vkey that could occur on the scancodes lists…_  Then I would know which vkeys were expected to be mappable by ldml.  So `K_A` if unmapped would do nothing but `K_ENTER` would be passed through to the OS?

srl295:
Comments here:
- [ ] there isn't a KMX+ representation for the 'new' `<form>` element. probably needs a new ticket for that. If I had that, I could look for 'expected scancodes'.
- [ ] If a key is a 'gap' and you press that key, right now it probably looks like a key with a 0-length string output.   Could probably short circuit this (and any other situation where a key has a 0-length output for that matter) and do a little less processing. , basically ignored.
- [ ] kmc currently only fills in keys in the rows and columns until it runs out of input from the xml.  This means that `<row keys="A gap gap B"/>` only fills in four entries, for A, gap, gap, B.   It _could_ actually complete the row, perhaps automatically assigning the 'missing keys' to `gap`, or better yet a 0-length keystring simply meaning a missing key.   If this were done, then the `kmap` subtable would have _every expected vkey_ filled in.  And a lookup for those vk ids would always succeed, but return either the 'null' key or a `gap` key.  In any event, the key could then be processed by ignoring it (no output).  Whereas any frame key would _not_ match these vkeys and could then be passed on to the OS.

srl295:
```
[...]

PR #10090, opened 2023-11-29
feat(core): ldml improve key-not-found 🙀

srl295:
- distinguish between unmapped and 0-length output strings
- don't do any processing for 0-length output strings
- no change to developer, that will be next
- remove a comment referencing transform=no

For: #9451

@keymanapp-test-bot skip


srl295:
PTAL and see if it makes sense. The NEXT step will be to change kmc to actually emit some kind of pseudo-gap ( essentially `<key gap="true" />` for all keys in a layout that aren't othewise specified.

core/src/ldml/ldml_processor.cpp
```diff
@@ -252,13 +252,15 @@ ldml_processor::process_backspace(km_core_state *state) const {
 void
 ldml_processor::process_key(km_core_state *state, km_core_virtual_key vk, uint16_t modifier_state) const {
   // Look up the key
-  const std::u16string key_str = keys.lookup(vk, modifier_state);
+  bool found = false;
+  const std::u16string key_str = keys.lookup(vk, modifier_state, found);
 
-  if (key_str.empty()) {
+  if (!found) {
```
mcdurdin, line 258: We need to exclude modifiers (incl. Caps) from this test. Press and release Ctrl or Shift should not invalidate context. Press and release Alt on Windows is a special case -- because it opens the menu. But it may not need to invalidate context because other actions will do that anyway.

However, Ctrl+A _should_ invalidate context. So we need to be aware of the modifier state of key combinations and invalidate context for missing key events. And if we define Ctrl+Q in the keyboard, does that mean that all other Ctrl+key combinations will be treated as gap? Philosophy time.

Also, I wonder how many keyboard developers will get annoyed when their spacebar doesn't work until they add a key for it?

mcdurdin (changes requested):
This is looking pretty close. There are some questions around modifier keys and key combinations that I think we need to cover

core/src/ldml/ldml_processor.cpp
srl295: - we're only on key down here, for one thing. key up events are silently ignored.  (line 185) - should that change?
- if Ctrl-A was assigned as a key, then it should not invalidate context. So I'm going to need some more details here...
- space - we could detect space here and simply send it to the OS, invalidating context.  But, since we already have an implied `space` key in LDML, kmc could simply slip that one in unless it's otherwise defined. (i.e. instead of gap).  The net effect would be that if you define _no_ keys you'll get a keyboard with one functional key- the spacebar. 
- seems like there's enough here that it's time for a CLDR ticket: https://unicode-org.atlassian.net/browse/CLDR-17262

srl295:
the other question is how in the world can i test this?

srl295:
@mcdurdin I've restructured a bit to make it easier to pass-through keys as needed pending that discussion.

core/src/ldml/ldml_processor.cpp
mcdurdin: > * we're only on key down here, for one thing. key up events are silently ignored.  (line 185) - should that change?

The key up events don't tend to harm if the key is already up (in my experience). So should be okay to go through.

> * if Ctrl-A was assigned as a key, then it should not invalidate context. So I'm going to need some more details here...

If Ctrl+A is assigned, then yes, it gets processed as normal. If it isn't assigned, however, and falls back through to the app (as a user would expect), doing a select-all on Windows for example, then we would want a context reset.

mcdurdin:
> the other question is how in the world can i best test this besides with a debugger?

the heuristic of losing markers is actually pretty good.

srl295:
> > the other question is how in the world can i best test this besides with a debugger?
> 
> 
> 
> the heuristic of losing markers is actually pretty good.

Seems a little in direct, but I'll give it a try

core/src/ldml/ldml_processor.cpp
srl295: > key up

I don't think you mean "all keyup events trigger passthrough/context reset" as that would be disruptive to markers.   so I'm not sure what the code should do different here.

> ctrl=A 

OK so far.

I'm still unclear on *which* vkeys need to be passed through to the OS, causing a context reset.   is there a list? should the list be in CLDR?

srl295:
little bit of a side excursion fixing a markers issue. #9468 

Commit is now much improved, but overproducing markers in the context:

```
- key action: ['K_BKSP' 0x8]/modifier Unmodified 0x0
context   : U+0036 U+0065  [6e]
testcontext U+0036 \m{1} \m{1} \m{1} \m{1} U+0065
- check expected
expected  : U+0059 U+0045 U+0053 U+0036  [YES6]
text store: U+0036 U+0065  [6e]
```

Still to-do is to get back to the key-not-found logic.

core/src/ldml/ldml_processor.cpp
```diff
@@ -279,7 +285,7 @@ size_t ldml_processor::process_output(km_core_state *state, const std::u32string
   assert(ldml::normalize_nfd_markers(nfd_str)); // TODO-LDML: else fail?
   // extract context string, in NFD
   std::u32string old_ctxtstr_nfd;
-  (void)context_to_string(state, old_ctxtstr_nfd, false);
+  (void)context_to_string(state, old_ctxtstr_nfd, true);
```
srl295, line 288: sure enough, it was NOT INCLUDING MARKERS until this was turned to true.

core/src/ldml/ldml_processor.cpp
```diff
@@ -316,8 +322,8 @@ size_t ldml_processor::process_output(km_core_state *state, const std::u32string
 
   // Ok. We've done all the happy manipulations.
 
-  /** NFD and no markers */
-  std::u32string ctxtstr_cleanedup = ldml::remove_markers(ctxtstr);
```
srl295, line 320: sure enough it was, surprisingly, REMOVING MARKERS

core/src/ldml/ldml_processor.cpp
```diff
@@ -331,25 +337,23 @@ size_t ldml_processor::process_output(km_core_state *state, const std::u32string
...
   assert(old_ctxtstr_changed.length() == 0);
   // old_ctxtstr_changed is now empty because it's been removed.
   // context is "aa" in the above example.
-  emit_text(state, new_ctxtstr_changed);
 
-  // TODO-LDML: need to emit marker here - need to emit text w/ markers, and handle appropriately.
-  // // TODO-LDML: 1-marker hack! need to support a string with intermixed markers.
-  if (str.length() == 3 && str[0] == LDML_UC_SENTINEL && str[1] == LDML_MARKER_CODE) {
-    emit_marker(state, str[2]);
-  }
```
srl295, line 345: @mcdurdin this is the hack I was mentioning. This PR is thus now a major boost to #9468

core/src/ldml/ldml_processor.cpp
```diff
@@ -438,6 +456,11 @@ ldml_processor::emit_marker(km_core_state *state, KMX_DWORD marker_no) {
   state->context().push_marker(marker_no);
 }
 
+void ldml_processor::emit_invalidate_passthrough_keystroke(km_core_state *state) {
+  state->actions().push_invalidate_context();
+  state->actions().push_emit_keystroke();
+}
+
```
srl295, line 463: encapsulating this in case it needs to change later. i guess I could pass the vkey here?

core/src/ldml/ldml_vkeys.cpp
```diff
@@ -27,12 +27,12 @@ static const uint16_t BOTH_ALT  = LALTFLAG  | RALTFLAG;
 static const uint16_t BOTH_CTRL = LCTRLFLAG | RCTRLFLAG;
 
 std::u16string
-vkeys::lookup(km_core_virtual_key vk, uint16_t modifier_state) const {
+vkeys::lookup(km_core_virtual_key vk, uint16_t modifier_state, bool &found) const {
```
srl295, line 30: empty string return now means "no output", distinct from "not found"

srl295:
at this point I'm not going to try to split this into two PRs, key-not-found vs markers.  If I realized the test cases were going to bump up against that so quickly, i might have done so.

srl295:
I copied the table from VKScanCodes.cpp , it seemed to be what i wanted, ptal…

mcdurdin:
> I copied the table from VKScanCodes.cpp , it seemed to be what i wanted, ptal…

I wonder if we should be trying to centralize that as a common shared file under /common/cpp/scancodes/some_file_name_here.cpp?

srl295:
> > I copied the table from VKScanCodes.cpp , it seemed to be what i wanted, ptal…
> 
> I wonder if we should be trying to centralize that as a common shared file under /common/cpp/scancodes/some_file_name_here.cpp?

maybe a good item for a future PR? ?

mcdurdin (approved):
LGTM, pending the assert just noted and if possible moving the vkscancodes stuff into a separate cpp

keyman-server:

PR #10098, opened 2023-11-30
fix(core): set_if_needed updates an empty cached context

rc-swag:
In the case where the cached context had been cleared the km_core_state_context_set_if_needed call would just compare the null terminations of both strings and not set the cached context to the application context.

I first thought I could just change the decrement of the pointers to occur before the iteration not after e.g.

``` 

for(; context_p > context && cached_context_p > cached_context; --context_p, --cached_context_p) {
    if(*context_p != *cached_context_p) {
      // The cached context doesn't match the application context, so it is
      // invalid
      return false;
    }
  }
```
  
This skips the comparing the null, however it will still also drop out and return true. It is just quicker to bail if the cached context is an "empty" string. 
I also considered renaming is_context_valid to is_context_update_needed.

@keymanapp-test-bot skip

### DIFF
--- a/core/src/km_core_state_api.cpp
+++ b/core/src/km_core_state_api.cpp
@@ -262,6 +262,10 @@ void km_core_state_imx_deregister_callback(km_core_state *state)
 }
 
 bool is_context_valid(km_core_cp const * context, km_core_cp const * cached_context) {
+  if (context == nullptr || cached_context == nullptr || *cached_context == NULL) {
+      // If the cached_context is "empty" then it needs updating
+      return false;
+  }
   km_core_cp const* context_p = context;
   while(*context_p) {
     context_p++;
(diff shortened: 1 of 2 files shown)


core/src/km_core_state_api.cpp
```diff
@@ -262,6 +262,10 @@ void km_core_state_imx_deregister_callback(km_core_state *state)
 }
 
 bool is_context_valid(km_core_cp const * context, km_core_cp const * cached_context) {
+  if (context == nullptr || cached_context == nullptr || *cached_context == NULL) {
+      // If the cached_context is "empty" then it needs updating
+      return false;
```
mcdurdin, line 267:
```suggestion
    // If the cached_context is "empty" then it needs updating
    return false;
```

sgschantz (approved):
LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.220-alpha](https://keyman.com/downloads/releases/alpha/17.0.220)

PR #10101, opened 2023-11-30
fix(core): check for null termination

rc-swag:
Determine the best way to check for null termination all compilers

@keymanapp-test-bot skip

### DIFF
--- a/core/src/km_core_state_api.cpp
+++ b/core/src/km_core_state_api.cpp
@@ -262,7 +262,7 @@ void km_core_state_imx_deregister_callback(km_core_state *state)
 }
 
 bool is_context_valid(km_core_cp const * context, km_core_cp const * cached_context) {
-  if (context == nullptr || cached_context == nullptr || *cached_context == NULL) {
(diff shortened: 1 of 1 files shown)


core/src/km_core_state_api.cpp
```diff
@@ -262,7 +262,7 @@ void km_core_state_imx_deregister_callback(km_core_state *state)
 }
 
 bool is_context_valid(km_core_cp const * context, km_core_cp const * cached_context) {
-  if (context == nullptr || cached_context == nullptr || *cached_context == NULL) {
+  if (context == nullptr || cached_context == nullptr || *cached_context == '\0') {
```
mcdurdin, line 265:
```suggestion
  if (context == nullptr || cached_context == nullptr || *cached_context == 0) {
```

I think that should work in all compilers

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.220-alpha](https://keyman.com/downloads/releases/alpha/17.0.220)

Issue #9999, opened 2023-11-14
feat(core): support normalization in Core 🌱


mcdurdin:
## Impacted modules

* Core Consumers:
  * Keyman Engine for Windows
  * Keyman Engine for macOS
  * Keyman Engine for Linux
  * Keyman Developer Debugger
* Keyman Core
* Core Keyboard Processors:
  * LDML Keyboard Processor
* Unit tests

## Steps

1. Add a Core API to convert `km_core_actions` to `km_core_action_item*` (mirrors `action_item_list_to_actions_object`)

2. Update Engines to _always_ call `km_core_state_context_set_if_needed`, and stop managing their own cached context.

3. LDML processor can then generate `km_core_actions` and existing engines can consume it with no further modification.

4. Normalization support for LDML keyboards:
  * input context (`km_core_state_set_context_if_needed`)
  * and action output (`km_core_state_get_actions` and `actions_object_to_action_item_list`)

17.0 target API usage is:

- kmn: emit queue
- ldml: emit struct
- windows: read queue
- mac: read struct
- linux: read queue
- debugger: read queue
- ldml unit tests: read struct
- Core will translate _only_ as needed, so this way, the big blocker (debugger) can wait until 18.0.

### Move to `km_core_state_context_set_if_needed`
- [x] Windows: use `km_core_state_context_set_if_needed` ➡ #10052
- [x] macOS: use `km_core_state_context_set_if_needed` ➡ #8403
- [x] #10212
- [ ] #10213
- [ ] #10214
- [ ] #10215
- [x] #10384

## Longer term

1. Move to action struct in Engine for Windows, Developer Debugger, Engine for Linux ([#10353](https://github.com/keymanapp/keyman/issues/10353)) {tech debt}
  * Removes dead wood apis
  * Possible difficulties with the Keyman Developer debugger (will need sideband data on cached context as debugger is, and must continue to be, marker-aware)

2. Move to action struct for output from kmn Keyboard Processor. Then the concept of Action Queue moves to an internal component of kmn Keyboard Processor (and the debugger will query kmn keyboard processor directly).

## Implementation notes

in `km_core_state_context_set_if_needed` we'll need to do something like this:

```
  // TODO: normalization
  km_core_cp* text;
  switch(get_keyboard_processor_normalization()) {
    case NFD: text = normalize_nfd(application_context); break;
    case NFC: text = normalize_nfc(application_context); break;
    default: /*NFW*/ text = application_context;
  }
```

[...]

PR #10115, opened 2023-12-01
feat(core): action struct to action items conversion

mcdurdin:
Relates to #9999.

Adds `state->set_actions()`. This sets the Core's action list to match the contents of the action struct. Note that markers are not supported and backspace expected_values will be empty, as this information is not available. As the intended consumer of the action struct does not need to know this information, this should be adequate.

@keymanapp-test-bot skip

### DIFF
--- a/core/src/action.cpp
+++ b/core/src/action.cpp
@@ -151,7 +151,7 @@ bool km::core::state::set_actions(
   // number of codepoints (not codeunits!) to delete from app context.
 
   for(unsigned int i = 0; i < actions.code_points_to_delete; i++) {
-    _actions.push_backspace(KM_CORE_BT_CHAR, 0); //(*end).character);
+    _actions.push_backspace(KM_CORE_BT_CHAR, 0); // expected value is not known
   }
 
   for(auto output = actions.output; *output; output++) {
--- a/core/src/action.cpp
+++ b/core/src/action.cpp
@@ -17,7 +17,7 @@
 #include "state.hpp"
 #include "option.hpp"
 
-km_core_actions * km::core::action_item_list_to_actions_object(
+km_core_actions const * km::core::action_item_list_to_actions_object(
   km_core_action_item const *action_items
 ) {
   assert(action_items != nullptr);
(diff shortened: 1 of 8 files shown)


core/src/state.hpp
```diff
@@ -152,8 +152,18 @@ class state
     void imx_deregister_callback();
 
     void imx_callback(uint32_t imx_id);
-};
 
+    // This is intended to be used to take the actions given in the actions
+    // parameter, and load them into the _actions member of this class. Used by
+    // keyboard processors to set the output actions, and is a long-term
+    // replacement for the actions()::push_*() functions. Note that the
```
rc-swag, line 159: I have a different understanding. 
We have `engines` oneside of the core and `keyboard processors` on the other side. On the `engine` side we need `km_core_state_get_actions` to obtain the `km_core_actions` struct for `engines` who consume it. On the keyboard processor side we need `set_actions` to do the reverse to handle keyboard processors that produce a `km_core_actions` struct. So that we can convert the actions struct back into a km_core_action_items` list for the `engines` that still consume this. 
The keyboard processors will do a call something like `km_core_state_set_actions` with the action struct then internally the core in that method (`km_core_state_set_actions`) will call `set_actions`. 

Current situation
1. kmx keyboard processor does not have an understanding of the `km_core_actions`. It uses the `actions()::push_*()` completed with `actions()::commit`. To create a `km_core_action_items` list.  This will be converted to `km_core_actions` struct when a core consumer (engine) calls `km_core_state_get_actions`. MacOS engine is on this path already.
2. In implementing the LDML processor transformations it was discovered it needed and beneficial to be able to produce `km_core_actions` struct and be able to set it in the core. 
3. As the Windows and Linux Engines are still consuming`km_core_action_items` list they will need this PR with `set_actions` to convert the `km_core_actions` struct back to the `km_core_action_items` 
4. If all platform engines are updated to consume `km_core_actions` struct. Then this `set_actions` method will no longer be needed, excepted to serve the deprecated API.  Even if the kmx_processor is still pushing action_items, as the `km_core_state_get_actions` will supply the struct to the engine.

mcdurdin: I think that's what my code comment says, isn't it?

rc-swag: //Used by
    // keyboard processors to set the output actions,
The keyboard processors should not be using `set_actions` to set the output actions. They shouldn't know it exists. They use the push/set call to get the `km_core_actions` to the core. The core can internally at this time call `set_actions` to satisfy the engines that need the `km_core_action_items` via the to-be-depricated `km_core_state_action_items` call.
// and is a long-term
 // replacement for the actions()::push_*() functions
`set_actions` is internal to core helper method and will only live for as long as the deprecated actions list is supported. The replacement for the `actions()::push_*() functions` will be a method that just pushes the a `km_core_actions` struct to the core. It could mirror the get `km_core_state_set_actions` or it could match the current methods `actions()::push_actions()

That is my understanding anyway.

rc-swag: Oh, I think the penny has finally dropped. I was seeing this `set_actions` just as the conversion function from actions struct to actions list. Something that the method that actually sets the actions calls/runs. That way we could keep what is the main functionality and which is assisting the deprecated code separate. 
I presume then that this `set_actions` is not complete in that in the future it will store the struct in a yet-to-be-added private member of "_actions_struct".

rc-swag (approved):
updated my comments
LGTM

core/src/action.cpp
```diff
@@ -138,3 +138,42 @@ km_core_actions * km::core::action_item_list_to_actions_object(
...
+//       of state also, so that we can move memory management into state?
+bool km::core::state::set_actions(
+  km_core_actions const &actions
+) {
+  _actions.clear();
+
+  // number of codepoints (not codeunits!) to delete from app context.
+
+  for(unsigned int i = 0; i < actions.code_points_to_delete; i++) {
+    _actions.push_backspace(KM_CORE_BT_CHAR, 0); // expected value is not known
```
srl295, line 154: so… unlike the similar function in ldml_processor, this one won't push call `_actions.push_backspace(KM_CORE_BT_MARKER, …)` if it's a marker.  Or am I getting mixed up and there are no markers in these actions anymore?

srl295: ok no markers here.  Could note that here.

core/src/state.hpp
srl295: so this is a long term function, not temporary.

srl295 (approved):

core/src/state.hpp
mcdurdin: > I presume then that this `set_actions` is not complete in that in the future it will store the struct in a yet-to-be-added private member of "_actions_struct".

Eventually all the `push_` methods will disappear, and `set_actions` will be the only way for a keyboard processor to set the actions; this will mean some changes internally for `km::core::state` and `km::core::actions` objects, yes.

core/src/action.cpp
mcdurdin: no markers in actions -- they are only internal to the cached context now.

keyman-server:

PR #10137, opened 2023-12-05
fix(developer): prevent opening .kpj in multiple processes 🦕

### LINKED ISSUE #10003: bug(developer): opening same project in two TIKE windows should not be possible
Original description:

1. Load a v2.0 project in Developer
2. In Project distribution tab, click the .kpj file to open it again.
3. Close the new TIKE window that opens
4. The original TIKE Project window refreshes, and shows an apparently v1.0 project, with no content, and clicking the Upgrade button silently fails.

Examination of the .kpj shows that `<Version>2.0</Version>` element has been removed. This is happening because the .kpj is being loaded by the stable-16.0 version of Developer, rather than a new instance of the same version.

Updated details:

* This should focus the existing window if you attempt to open the same project twice, or open a file in a folder in an already-open project.

mcdurdin:
Fixes #10003.

Multiple small fixes required here:
* Add .kpj as an editable file type
* Ensure that when OpenFile returns nil (which it always will for a .kpj), that we don't crash
* Ensure that projects open in a new process rather than replacing the current project session

Adds a TODO for New Project -- this should be opened in a new process if we already have a project open in the current process. Will push this into a separate PR (see #10138).

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tike/actions/dmActionsMain.pas
+++ b/developer/src/tike/actions/dmActionsMain.pas
@@ -576,6 +576,7 @@ end;
 
 procedure TmodActionsMain.actProjectNewExecute(Sender: TObject);
 begin
+  // TODO: new process
   if not frmKeymanDeveloper.BeforeOpenProject then
     Exit;
   ShowNewProjectForm(frmKeymanDeveloper);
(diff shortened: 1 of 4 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.225-alpha](https://keyman.com/downloads/releases/alpha/17.0.225)

PR #10141, opened 2023-12-05
fix(web): Fix null error with legacy keyboards

### LINKED ISSUE #10099: bug(web): Cannot read properties of null (reading 'baseKeyEvent')
I was troubleshooting keymanweb.com locally with Docker when I ran into the following console error.

Was able to reproduce on the live keymanweb.com site 
https://keymanweb.com/?tier=alpha#hy,Keyboard_armenian 

(legacy Armenian keyboard)
https://keyman.com/keyboards/armenian

console error (ignoring null keyboard stub)

```
TypeError: Cannot read properties of null (reading 'baseKeyEvent')
    at i.constructKeyEvent (keyboard.ts:473:21)
    at i.constructBaseKeyEvent (activeLayout.ts:287:62)
    at i.polyfill (activeLayout.ts:239:10)
    at i.polyfill (activeLayout.ts:371:17)
    at i.polyfill (activeLayout.ts:507:17)
    at i.polyfill (activeLayout.ts:763:19)
    at i.layout (keyboard.ts:426:34)
    at t._GenerateKeyboardView (oskView.ts:777:31)
    at t.loadActiveKeyboard (oskView.ts:729:58)
    at t.set (oskView.ts:521:10)
    at t.<anonymous> (keymanEngine.ts:153:18)
    at W.emit (index.js:181:35)
    at t.<anonymous> (contextManagerBase.ts:277:12)
    at g (tslib.js:142:27)
    at Object.next (tslib.js:123:57)
    at o (tslib.js:113:62)
```

In the Chrome debugger
```

  constructKeyEvent(key: ActiveKey, device: DeviceSpec, stateKeys: StateKeyMap): KeyEvent {
    // Make a deep copy of our preconstructed key event, filling it out from there.
    const Lkc = key.baseKeyEvent;
    Lkc.device = device;
```

has key = null

ermshiperete:
Fixes #10099.

# User Testing

**TEST_FIXED**: 

- Open the [Unminified Source test page](https://build.palaso.org/repository/download/Keymanweb_TestPullRequests/428137:id/src/test/manual/web/unminified.html)
- Load the Armenian keyboard by typing "armenian" in the "Add a keyboard by keyboard name" field and pressing the "Add" button
- click in the "Type here" field and switch to the Armenian keyboard
- Verify that the Armenian OSK is shown, but no error

### DIFF
--- a/common/web/keyboard-processor/src/keyboards/activeLayout.ts
+++ b/common/web/keyboard-processor/src/keyboards/activeLayout.ts
@@ -282,9 +282,7 @@ export class ActiveKey implements LayoutKey {
 
       // Support version 1.0 KeymanWeb keyboards that do not define positional vs mnemonic
       if(!keyboard.definesPositionalOrMnemonic) {
-        // Not the best pattern, but currently safe - we don't look up any properties of any of the
-        // arguments in this use case, and the object's scope is extremely limited.
-        Lkc.Lcode = KeyMapping._USKeyCodeToCharCode(keyboard.constructNullKeyEvent(null, null));
+        Lkc.Lcode = KeyMapping._USKeyCodeToCharCode(Lkc);
         Lkc.LisVirtualKey=false;
       }
     }
--- a/common/web/keyboard-processor/src/keyboards/activeLayout.ts
+++ b/common/web/keyboard-processor/src/keyboards/activeLayout.ts
@@ -284,11 +284,7 @@ export class ActiveKey implements LayoutKey {
       if(!keyboard.definesPositionalOrMnemonic) {
         // Not the best pattern, but currently safe - we don't look up any properties of any of the
         // arguments in this use case, and the object's scope is extremely limited.
-        Lkc.Lcode = KeyMapping._USKeyCodeToCharCode(keyboard.constructKeyEvent(null, null, {
-          K_CAPS: false,
-          K_NUMLOCK: false,
-          K_SCROLL: false
-        }));
+        Lkc.Lcode = KeyMapping._USKeyCodeToCharCode(keyboard.constructNullKeyEvent(null, null));
         Lkc.LisVirtualKey=false;
       }
     }
--- a/common/web/keyboard-processor/src/keyboards/keyboard.ts
+++ b/common/web/keyboard-processor/src/keyboards/keyboard.ts
@@ -502,7 +502,7 @@ export default class Keyboard {
     // Set the flags for the state keys - for desktop devices. For touch
     // devices, the only state key in use currently is Caps Lock, which is set
     // when the 'caps' layer is active in ActiveKey::constructBaseKeyEvent.
-    if(!Lkc.device?.touchable) {
+    if(!Lkc.device.touchable) {
       /*
        * For desktop-style keyboards, start from a blank slate.  They have a 'default'
        * (implicit 'NO_CAPS') layer but not a 'caps' layer.  With caps set, it just
--- a/common/web/keyboard-processor/src/keyboards/keyboard.ts
+++ b/common/web/keyboard-processor/src/keyboards/keyboard.ts
@@ -502,7 +502,7 @@ export default class Keyboard {
     // Set the flags for the state keys - for desktop devices. For touch
     // devices, the only state key in use currently is Caps Lock, which is set
     // when the 'caps' layer is active in ActiveKey::constructBaseKeyEvent.
-    if(!Lkc.device.touchable) {
+    if(!Lkc.device?.touchable) {
       /*
        * For desktop-style keyboards, start from a blank slate.  They have a 'default'
        * (implicit 'NO_CAPS') layer but not a 'caps' layer.  With caps set, it just


common/web/keyboard-processor/src/keyboards/activeLayout.ts
```diff
@@ -284,11 +284,7 @@ export class ActiveKey implements LayoutKey {
       if(!keyboard.definesPositionalOrMnemonic) {
         // Not the best pattern, but currently safe - we don't look up any properties of any of the
         // arguments in this use case, and the object's scope is extremely limited.
-        Lkc.Lcode = KeyMapping._USKeyCodeToCharCode(keyboard.constructKeyEvent(null, null, {
-          K_CAPS: false,
-          K_NUMLOCK: false,
-          K_SCROLL: false
-        }));
+        Lkc.Lcode = KeyMapping._USKeyCodeToCharCode(keyboard.constructNullKeyEvent(null, null));
```
mcdurdin, line 287: Can you explain the reason for this change? I don't understand it

common/web/keyboard-processor/src/keyboards/keyboard.ts
```diff
@@ -502,7 +502,7 @@ export default class Keyboard {
     // Set the flags for the state keys - for desktop devices. For touch
     // devices, the only state key in use currently is Caps Lock, which is set
     // when the 'caps' layer is active in ActiveKey::constructBaseKeyEvent.
-    if(!Lkc.device.touchable) {
+    if(!Lkc.device?.touchable) {
```
mcdurdin, line 505: I think one of the input assumptions here is that `device` is valid. So this may be masking a more significant bug.

common/web/keyboard-processor/src/keyboards/activeLayout.ts
jahorton: Hmm... it _is_ worth noting that the original line doesn't reference an actual key.  Personally, when running a trace on what's going on, this block is extremely illuminating:

https://github.com/keymanapp/keyman/blob/4c7b7992da6491dc5bf30245dc5851f8c1d471c1/common/web/keyboard-processor/src/keyboards/keyboard.ts#L471-L473

Note that we're within the `constructBaseKeyEvent` method (for this comment).  L473 above is relying upon the results of said method.  This is an obvious problem.

Since this hasn't exactly been a section of code I've actually dived into much, perhaps a comparison against stable-10.0 is in order.  I've moved the code a few times since then, so I'll highlight the corresponding section to move things along:

https://github.com/keymanapp/keyman/blob/f16872a99896a2fc208c68ac92a827820ab07e3f/web/source/kmwdomevents.ts#L722-L734

(Note that this is constructing the event in full from nothing, rather than the partial construction intended for `constructBaseKeyEvent`.)

So... KMW 1.0 used different key-code values, and we just need to map the key code we use now to the one that was used then.  It's not even like we're replacing the other key-event values... just `.Lcode` here and `.LisVirtualKey` on the next line.

Something far closer to the original would likely be best here.

common/web/keyboard-processor/src/keyboards/keyboard.ts
jahorton: With regard to the stack trace from the issue:

    TypeError: Cannot read properties of null (reading 'baseKeyEvent')
    at i.constructKeyEvent (keyboard.ts:473:21)
    at i.constructBaseKeyEvent (activeLayout.ts:287:62)
    at i.polyfill (activeLayout.ts:239:10)

https://github.com/keymanapp/keyman/blob/4c7b7992da6491dc5bf30245dc5851f8c1d471c1/common/web/keyboard-processor/src/keyboards/activeLayout.ts#L283-L293

It appears this code path went unused and untested until the bug was observed.  Now, I think I did do a lot of key-event preprocessing code refactoring in 17.0 during ES module work, so this is probably 17.0-only.  All in all, the bug is probably a "loose end" I accidentally didn't properly handle during that initiative.

jahorton (changes requested):

common/web/keyboard-processor/src/keyboards/activeLayout.ts
mcdurdin: > this block is extremely illuminating

Not to me! Perhaps you could explain it more clearly for me?

jahorton: > Note that we're within the constructBaseKeyEvent method (for this comment). L473 above is relying upon the results of said method. This is an obvious problem.

In other words, we have an impossible-to-meet precondition here.  We're within a method that must complete to satisfy the precondition of a method it's calling.

jahorton: Fortunately, we can simplify things greatly here... that's what the rest of the analysis is about.

bharanidharanj:
# Test Results

* **TEST_FIXED (PASSED):** Tested with the attached "Unminified Source Test" page and verified that the armenian OSK appear without showing any error message.

mcdurdin (commented):
I will leave this to @jahorton to review because there's a whole lot of detail I'm not understanding in the previous thread here

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.226-alpha](https://keyman.com/downloads/releases/alpha/17.0.226)

PR #10151, opened 2023-12-06
fix(developer): various project fixes 🦕

### LINKED ISSUE #10144: chore(developer): Fix opening various text files from project view, incl. keyman-touch-layout

### LINKED ISSUE #10145: bug(developer): Open recently used fails to load files with long paths elided with `...`

### LINKED ISSUE #10146: feat(developer): Add option to disable prompting for project upgrade

### LINKED ISSUE #10148: chore(developer): Validate sourcepath and buildpath -- should be at most 1 level deep from project path
This should be done in the Project Settings dialog for v2.0 projects. We won't add the same restrictions to v1.0 projects.

mcdurdin:
A single PR for 4 separate fixes to Keyman Developer, all relating to Projects. The fixes are separated by commit but were small enough and isolated enough that I didn't think I needed to split into separate PRs (yeah, yeah, I know).

* fe772e1008 fix(developer): Validate SourcePath and BuildPath in Project Settings

   Fixes #10148.

   We now force the SourcePath and BuildPath project properties to be a direct subfolder of the project folder, in order to avoid issues with paths in the future. This is only applied to v2.0 projects, and only enforced via the UI at this point.

* 6c918e6aa8 fix(developer): Option to disable prompting for project upgrade

   Fixes #10146.

   Some developers may wish to stick with kpj-1.0 for now, and the prompt to upgrade would annoying because it cannot be hidden. This resolves that.

   Note that I haven't at this point renamed UrlRenderer.pas, although it may be worth considering in the future.

* ccbb00db9b fix(developer): Open recent files with long paths elided with ellipsis

   Fixes #10145.

   Turns out we were clobbering out internal data because Delphi strings are copy-on-write. At the same time, moved from using the Hint property to just referencing the source filename in the array, as that is cleaner anyway.

* 149fc5e7c8 fix(developer): Support opening .keyman-touch-layout

   Fixes #10144.

   Also adds .keyboard_info, for now, although it is removed in 17.0, as there will be numerous projects which still have a .keyboard_info and it is handy to be able to load it and view its contents while upgrading.

* 4f83c9cf14 chore(developer): remove commented code

   From TODO item in #9948.

@keymanapp-test-bot skip

I have already tested these locally and do not think they warrant separate user testing at this point.

### DIFF
--- a/developer/src/tike/dialogs/UfrmOptions.dfm
+++ b/developer/src/tike/dialogs/UfrmOptions.dfm
@@ -29,7 +29,7 @@ inherited frmOptions: TfrmOptions
         Top = 185
         Width = 125
         Height = 25
-        Caption = '&Proxy Settings...'
+        Caption = 'Prox&y Settings...'
         TabOrder = 4
         OnClick = cmdProxySettingsClick
       end
@@ -38,7 +38,7 @@ inherited frmOptions: TfrmOptions
         Top = 47
         Width = 401
         Height = 45
-        Caption = '&External Editor Path'
+        Caption = 'E&xternal Editor Path'
         TabOrder = 1
         object editExternalEditorPath: TEdit
           Left = 8
@@ -61,7 +61,7 @@ inherited frmOptions: TfrmOptions
         Top = 216
         Width = 125
         Height = 25
-        Caption = 'S&MTP Settings...'
+        Caption = '&SMTP Settings...'
         TabOrder = 5
         OnClick = cmdSMTPSettingsClick
       end
@@ -101,7 +101,7 @@ inherited frmOptions: TfrmOptions
           Top = 16
           Width = 69
           Height = 21
-          Caption = '&Browse...'
+          Caption = 'Bro&wse...'
           TabOrder = 1
           OnClick = cmdBrowseDefaultProjectPathClick
         end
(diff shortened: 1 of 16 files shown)


developer/src/tike/dialogs/UfrmOptions.dfm
```diff
@@ -29,7 +29,7 @@ inherited frmOptions: TfrmOptions
         Top = 185
         Width = 125
         Height = 25
-        Caption = '&Proxy Settings...'
+        Caption = 'Prox&y Settings...'
```
mcdurdin, line 32: Note: I fixed up overlapping shortcut keys in this dialog pane.

developer/src/tike/main/mrulist.pas
```diff
@@ -231,11 +231,13 @@ destructor TMRUList.Destroy;
...
-  Result := Files[Index];
+  StrPCopy(buffer, Files[Index]);
 
-  if PathCompactPath(0, PWideChar(Result), GetSystemMetrics(SM_CXSCREEN) div 3) then   // I4697
-    Result := string(PChar(Result))  // This removes the terminating nul
+  if PathCompactPath(0, buffer, GetSystemMetrics(SM_CXSCREEN) div 3) then   // I4697
+    Result := buffer  // This removes the terminating nul
   else
     Result := ExtractFileName(Files[Index]);
 end;
```
mcdurdin, line 243: This was badly buggy previously -- `PathCompactPath` would overwrite `Result` (note that `Result` was not guaranteed to be `MAX_PATH` which is technically a violation of the API contract but was a secondary issue), but as Delphi strings are copy-on-write, it ended up overwriting `Files[Index]` as well!

darcywong00 (approved):
lgtm

keyman-server:

PR #10154, opened 2023-12-06
feat(developer): Add Window menu 🦕

### LINKED ISSUE #10149: feat(developer): multi-process model would benefit from a Window menu
We should have a Window menu listing all the open Keyman Developer processes so that it is easy to switch between projects.

mcdurdin:
Fixes #10149.

Adds a Window menu so user can easily switch between Keyman Developer processes or open a new instance.

image

# User Testing

* **TEST_NEW_WINDOW:** Open Keyman Developer. Use Window|New Window to open a new instance of Keyman Developer, which should open with no project open.
* **TEST_WINDOW_MENU:** Open several instances of Keyman Developer and open projects in each instance. Verify that the windows shown in the Window menu match up with the open projects.

### DIFF
--- a/developer/src/tike/main/Keyman.Developer.System.TikeMultiProcess.pas
+++ b/developer/src/tike/main/Keyman.Developer.System.TikeMultiProcess.pas
@@ -4,6 +4,7 @@ interface
 
 uses
   System.Generics.Collections,
+  System.Generics.Defaults,
   Keyman.Developer.System.MultiProcess;
 
 type
@@ -12,9 +13,15 @@ type
     ThreadID: Cardinal;
     ProjectFilename: string;
     SourcePath: string;
+    IsProjectOpen: Boolean;
+    IsTemporaryProject: Boolean;
     function OwnsProject(const filename: string): Boolean;
     function OpenFile(const filename: string): Boolean;
     function FocusProcess: Boolean;
+  private
+    function GetTitle: string;
+  public
+    property Title: string read GetTitle;
   end;
 
   TTikeProcessList = TObjectList<TTikeProcess>;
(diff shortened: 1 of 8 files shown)


darcywong00:
Some UX question to catch me up.
1. Is the "Window" listing replacing the previous "File" and "Project" menu items in Developer?
2. It feels like a verb/action is missing in the screenshot (Would the user know what selecting a .kpj does?)
3. Is there a menu to navigate for closing one of these open Keyman Developer processes?

darcywong00:
Gotcha. And what does selecting "1 (No project loaded)" do? I saw it's the title when `not IsProjectOpen` - So is it a placeholder when process 1 is available?

mcdurdin:
> And what does selecting "1 (No project loaded)" do? I saw it's the title when `not IsProjectOpen` - So is it a placeholder when process 1 is available?

Process 1 is a Keyman Developer process that is running but does not currently have an active project -- so no files open.

developer/src/tike/main/Keyman.Developer.System.TikeMultiProcess.pas
```diff
@@ -67,17 +74,34 @@ destructor TTikeMultiProcess.Destroy;
...
   Result := TObjectList<TTikeProcess>.Create;
   MultiProcessCoordinator.Enumerate;
   for p in MultiProcessCoordinator.Processes do
   begin
+    if p.Identifier = '*' then
+    begin
+      tp := TTikeProcess.Create;
+      tp.IsProjectOpen := False;
+      tp.IsTemporaryProject := False;
+      tp.ProjectFilename := '';
```
darcywong00, line 99: Why is 
`tp.IsTemporaryProject := False` when l. 22 compares

```
tp.IsTemporaryProject := tp.ProjectFilename = '';
```

Would it be easier to manage `IsTemporaryProject` as (psuedocode)

```
function IsTemporaryProject: Boolean;
begin
  Result := ProjectFilename = '';
end;
```

mcdurdin: That doesn't quite work -- it is a separate flag. This would work though:

```
function IsTemporaryProject: Boolean;
begin
  Result := IsProjectOpen and (ProjectFilename = '');
end;
```

bharanidharanj:
# Test Results

* **TEST_NEW_WINDOW (PASSED):** Tested with the attached PR build (Keyman 17.0.225-alpha-test-10154) on Windows 11 OS (VM) and here is my observation: 1. 1. Opened an existing project in Keyman Developer. 2. Using the Window / New Window option, open another project window. 3. Tried to open different projects using the Window / New Window option. 4. I got an error message while trying to open a custom project. Tried to reproduce this with another project. But it seems to be a random one. 

..Open a new instance using the Window / New Window option

   


..error message while trying to open a custom project (random)

mcdurdin:
> * I got an error message while trying to open a custom project. Tried to reproduce this with another project. But it seems to be a random one.

It looks like you clicked the 'Upgrade Project' button? If so, this is by design and working correctly -- the project cannot be upgraded to 17.0 format because there are several .kmn files listed in it which are in the wrong folder.

darcywong00 (approved):
lgtm

keyman-server:

PR #10157, opened 2023-12-06
fix(common): kpj loader needs to handle empty Files element

### LINKED ISSUE #10143: bug(common): kmc cannot load a v1.0 project which has no files
via email from @markcsinclair

If a v1.0 .kpj has an empty `<Files>` tag (i.e. no `<File>` tags), then kmc fails to load the file:

```
  {
    "code": 3166216,
    "message": "Project file is not valid: Error: [{\"instancePath\":\"/KeymanDeveloperProject/Files\",\"schemaPath\":\"#/type\",\"keyword\":\"type\",\"params\":{\"type\":\"object\"},\"message\":\"must be object\"}]",
    "filename": ""
  }
```

It should be able to load this file, even though it's not a useful project file!

Note: if the `<Files>` tag is removed, then an obscure error is given:

```
     TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

This should also be corrected.

mcdurdin:
Fixes #10143.

@keymanapp-test-bot skip

### DIFF
--- a/common/web/types/test/kpj/test-kpj-file-reader.ts
+++ b/common/web/types/test/kpj/test-kpj-file-reader.ts
@@ -123,4 +123,32 @@ describe('kpj-file-reader', function () {
     assert.isEmpty(f.details);
     assert.lengthOf(f.childFiles, 0);
   });
+
+  it('should load a v1.0 keyboard project with missing <File>', function() {
+    const path = makePathToFixture('kpj', 'project-missing-file', 'project_missing_file.kpj');
+    const input = fs.readFileSync(path);
+    const reader = new KPJFileReader(callbacks);
+    const kpj = reader.read(input);
+    reader.validate(kpj);
+    if(callbacks.messages.length) {
+      callbacks.printMessages();
+    }
+    assert.equal(callbacks.messages.length, 0);
+    const project = reader.transform(path, kpj);
+    assert.equal(callbacks.messages.length, 0);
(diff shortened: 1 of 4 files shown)


common/web/types/test/kpj/test-kpj-file-reader.ts
```diff
@@ -123,4 +123,32 @@ describe('kpj-file-reader', function () {
...
+  it('should load a v1.0 keyboard project with missing <File>', function() {
+    const path = makePathToFixture('kpj', 'project-missing-file', 'project_missing_file.kpj');
+    const input = fs.readFileSync(path);
+    const reader = new KPJFileReader(callbacks);
+    const kpj = reader.read(input);
+    reader.validate(kpj);
+    if(callbacks.messages.length) {
+      callbacks.printMessages();
+    }
+    assert.equal(callbacks.messages.length, 0);
```
darcywong00, line 136: Does the kpj get updated to assert one of these?

```suggestion
    assert.equal(callbacks.messages.length, 0);
    assert.lengthOf(kpj.KeymanDeveloperProject.Files.File, 0);
```
or
```suggestion
    assert.equal(callbacks.messages.length, 0);
    assert.lengthOf(kpj.KeymanDeveloperProject.Files.File, 1);
```

darcywong00 (approved):
lgtm

keyman-server:

PR #10158, opened 2023-12-06
feat(developer): kmc emit hint if project file is old version 🦕

### LINKED ISSUE #10147: feat(developer): Add lint hint to kmc when compiling v1.0 project recommending upgrade
Note: this should probably honour the setting in #10146.

mcdurdin:
Fixes #10147.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/src/commands/buildClasses/BuildProject.ts
+++ b/developer/src/kmc/src/commands/buildClasses/BuildProject.ts
@@ -41,6 +41,11 @@ class ProjectBuilder {
       return false;
     }
 
+    // Give a hint if the project is v1.0
+    if(this.project.options.version != '2.0') {
+      this.callbacks.reportMessage(InfrastructureMessages.Hint_ProjectIsVersion10());
+    }
+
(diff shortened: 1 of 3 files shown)


developer/src/kmc/src/commands/buildClasses/BuildProject.ts
```diff
@@ -41,6 +41,11 @@ class ProjectBuilder {
       return false;
     }
 
+    // Give a hint if the project is v1.0
+    if(this.project.options.version != '2.0') {
```
darcywong00, line 45: Wouldn't a version comparison (< 2.0) future-proof this check?

mcdurdin: Perhaps, but we'll need to revisit all this anyway if we want to upgrade project file versions again, and we may want to do something different for that case.

darcywong00 (approved):
lgtm

keyman-server:

Issue #10072, opened 2023-11-23
bug(windows): keyman64.dll does not build debug in the engine.sln


rc-swag:
Keyman64 debug build has issues with linked libraries that stop it from building.

PR #10176, opened 2023-12-07
fix(windows): debug includes libcmtd vs libcmt.dll

rc-swag:
Fixes: #10072 
Change the additional dependencies to include libcmtd.dll instead of limbcmt.dll for the debug build. Debug for keyman64 now builds.

@keymanapp-test-bot skip


keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.227-alpha](https://keyman.com/downloads/releases/alpha/17.0.227)

PR #10180, opened 2023-12-07
fix(linux): Do not restart Fcitx

wengxt:
1. Fcitx does not need to be restarted to load newly installed keyman keyboard
2. Restart fcitx may cause issue on kwin/weston Wayland where input method have to be spawned by compositor process. Manually start fcitx will not have permission to be functional with wayland input method protocol.

### DIFF
--- a/linux/keyman-config/keyman_config/fcitx_util.py
+++ b/linux/keyman-config/keyman_config/fcitx_util.py
@@ -12,6 +12,3 @@ def is_fcitx_running():
     except Exception:
         return False
 
-
-def restart_fcitx():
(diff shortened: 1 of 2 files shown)


keyman-server:

mcdurdin:
I will defer to @ermshiperete on review of this -- he may have more comments on why the reset is there at present for fcitx.

ermshiperete (approved):
Thanks for your contribution! LGTM

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.227-alpha](https://keyman.com/downloads/releases/alpha/17.0.227)

PR #10201, opened 2023-12-11
chore(mac): force detach of disk image during build after 5 failures

### LINKED ISSUE #10200: maint(mac): use `-force` when unmounting disk image to prevent build failures
CI stability improvement:

```
07:46:18   Detaching "/Users/mcdurdin/buildAgent/work/99b311828f4ee7c/keyman/mac/Keyman4MacIM/output/Keyman-temp-17.0.228.dmg"
07:46:19   hdiutil: couldn't unmount "disk5" - Resource busy
07:46:19   Process exited with code 16
```

This happens relatively frequently, with recent macOS versions. It's not clear what is locking the disk because trying manually has always unmounted successfully.

Per https://superuser.com/questions/714965/cant-unmount-dmg-keep-getting-resource-busy we should be able to `hdiutil unmount ... -force` followed by `hdiutil detach ...` (or possibly `hdiutil detach ... -force`, not verified)

mcdurdin:
Fixes #10200.

This is a two-part fix. First, with our normal `set -e` build environment, the existing loop never ran more than once, so the 5 second delay pattern we used to use in the dim dark past was no longer being applied. It is likely that in most cases, this will address the failure to detach.

The second part is to escalate our detach with a `-force` parameter, after 5 failed polite attempts. This is new and is addressing the suggestion in #10200, so we'll have to see if it has any impact on build stability.

@keymanapp-test-bot skip

### DIFF
--- a/mac/Keyman4MacIM/make-km-dmg.sh
+++ b/mac/Keyman4MacIM/make-km-dmg.sh
@@ -190,11 +190,17 @@ displayInfo "Detaching \"$WORKING_COPY_OF_IMAGE\""
 # because macOS may still be working in the folder in the
 # background
 DETACH_SUCCESS=0
+DETACH_FORCE=
 while (( DETACH_SUCCESS < 10 )); do
-    hdiutil detach $STAGING_DIR $VERBOSITY
-    if  [[ $? != 0 || -d "$STAGING_DIR" ]] ; then
+    DETACH_RESULT=0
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm 

DETACH_SUCCESS feels more like DETACH_FAILURE_COUNT

mcdurdin:
> DETACH_SUCCESS feels more like DETACH_FAILURE_COUNT

Indeed -- though this entire script needs to be builderified so will leave that name for now :grin:

keyman-server:

PR #10216, opened 2023-12-11
feat(developer): Move options to ~/.keymandeveloper/options.json

### LINKED ISSUE #10161: feat(developer): move from registry settings to .rc file
Because we share settings with kmc, it would make sense to move from using the Registry for storing settings to a .rc file. We will need a strategy to migrate the settings, but the settings are all read/written in KeymanDeveloperSettings.pas, so it should be reasonably straightforward to edit. Note that we will want to be careful about reader/writer file locks as we often have multiple processes reading the file at once.

File name will be `~/.keymandeveloper/options.json`.

File search strategy: look only in ~ (/Users/<name> on Windows) for now, with local overrides to come in a future update.

File format should be JSON.

Ref, allows us to eliminate costly reg.exe launch with every run of kmc and friends:

https://github.com/keymanapp/keyman/blob/8e6252c7de1d5fa424048dc196ded51acb71d31a/developer/src/common/web/utils/src/KeymanSentry.ts#L23-L38

mcdurdin:
Fixes #10161.

Includes transition of existing options from registry to options.json, and once options.json exists, will always read from that and ignore registry. Was able to eliminate some of the legacy datatype usage for several options. Reader/writer includes naive file lock retry to avoid contention, which I am hopeful will be sufficient for our needs, given the low chance of contention.

TODO:
- [x] Contention management in typescript
- [x] User Testing template

# User Testing

* **TEST_TRANSITION:** Install an existing version of Keyman Developer, set some of the options in Tools/Options. Then upgrade to the new version of Keyman Developer, start it once, exit, then start it again, and verify that the options you have set are still the same. You can also check `C:\Users\<yourname>\.keymandeveloper\options.json` to verify that the options there match what you set in the earlier Keyman Developer.
* **TEST_CLEAN:** Uninstall all versions of Keyman Developer. In Registry Editor, delete the `HKCU\Software\Keyman\Keyman Developer` key, if it exists. Delete the folder `C:\Users\<yourname>\.keymandeveloper` if it exists. Install this version of Keyman Developer, and verify that it starts correctly.


ermshiperete (approved):

bharanidharanj:
# Test Results

* **TEST_TRANSITION (PASSED):** Tested with Keyman Developer on Windows 11 OS and here is my observation: 1. Intsalled KeymanDeveloper 17.0.190 in the system. 2. Opened an existing project. 3. Clicked Tools / Options menu. 4. Set some options in the Options menu. 5. Upgraded to 17.0.228-alpha. 6. Opened Tools / Options menu. 7. Verified that the options I have set in the previous build are the same in the updated build. However, I was not able to check with the path that was given in the link to verify.

bharanidharanj:
# Test Results

* **TEST_CLEAN (PASSED):** Tested with Keyman Developer on Windows 11 OS and here is my observation: 1. Uninstalled Keyman Developer from the system. 2. Delete the reg key from HKCU\Software\Keyman\Keyman Developer path. 3. Successfully installed the attached PR build (Keyman 17.0.228-alpha-test-10216) in the system. 4. Verified that the installed version opened correctly.

keyman-server:

Issue #10220, opened 2023-12-11
feat: add reference to back_1 from keymanweb-osk.ttf


MayuraVerma:
### Is your feature request related to a problem? Please describe.

Need reference for special character
back_1 : 0x21
in
web/src/engine/osk/src/specialCharacters.ts

![image](https://github.com/keymanapp/keyman/assets/14062593/6e6b9f35-ba45-4564-a536-5e5669b9923d)

This glyph would be needed in creating iOS stock-keyboard like keyman-keyboards.

![image0](https://github.com/keymanapp/keyman/assets/14062593/c0eb4755-e274-4cee-8346-9aa76c52f2a8)

### Describe the solution you'd like

add in 
'*BkSp1*':     0x21,
in
web/src/engine/osk/src/specialCharacters.ts

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for iPhone and iPad
- [ ] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [X] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

mcdurdin:
Please see discussion in https://github.com/keymanapp/keyman/pull/10221#issuecomment-1851244289 and ref also #10226. This would be addressed with future theming support.

PR #9846, opened 2023-10-25
feat(developer): Add more non-printing characters

darcywong00:
Continues #9547

This adds special characters to Developer that were added in the new KeymanWeb font.

## User Testing
**Setup** - Install the PR build of Keyman Developer

* **TEST_SPECIAL_CHARACTERS** - verifies IDE can assign new characters
1. Launch Keyman Developer
2. Create a new keyboard that includes a touch layout keyboard
3. Go to the touch layout keyboard editor tab and select a key
4. In the Key properties, select the text dropdown and check for new strings
(will be in this list) https://github.com/keymanapp/keyman/blob/a4e081a46b305569a2c54e93c6d64ae9db061b33/web/src/engine/osk/src/specialCharacters.ts#L46-L67

Issue #2655, opened 2020-02-14
[iOS] keyboard shift/delete icons and text colors


MayuraVerma:
**Is your feature request related to a problem? Please describe.**

1. shift/delete key
in iOS keyboard, keyman keyboard shift and delete button icons look different than the system keyboard.

2. text colors
The text color for menu's "123"/ "abc" "space/keyboard name", globe icon and return key should be black similar to iOS system keyboard.

3. return key
Instead of the return key icon, we should have "return" in black color.
This word "return" should also change with context, "search", "tweet", "next", etc similar to system keyboard.

**Describe the solution you'd like**
![image2](https://user-images.githubusercontent.com/14062593/74569926-2aef0f00-4f41-11ea-9167-ff0b9d936651.png)
![image1](https://user-images.githubusercontent.com/14062593/74569928-2b87a580-4f41-11ea-95cb-5b3bc2bb55c0.png)

**Product context**
 - OS: iOS 13.3

mcdurdin:
Thank you for the suggestions. We'll consider these for a future release. It can be tricky to maintain consistency with different platforms, especially as they change over time.

We'd welcome a pull request which made these updates!

I note that the implementation of the dynamic label of the Return key will be significantly more complex than the styling changes.

MayuraVerma:
Here is an open source project. They have implemented shift, delete exactly same as the native keyboard.
Please review this.
Image+button:
https://github.com/KeyboardKit/KeyboardKit/blob/92bcb1940911060a227aafa1257ff291e3caa5b9/Sources/KeyboardKit/Appearance/Image%2BButton.swift
https://github.com/KeyboardKit/KeyboardKit/blob/92bcb1940911060a227aafa1257ff291e3caa5b9/Sources/KeyboardKit/Appearance/UIImage%2BButton.swift

Also there is the gesture implementation.
Gestures:
https://github.com/KeyboardKit/KeyboardKit/tree/master/Sources/KeyboardKit/Gestures

MayuraVerma:
@jahorton Please review this

MayuraVerma:
@mcdurdin @darcywong00 Can you guys please close this issue and 
https://github.com/keymanapp/keyman/issues/5493 in version 15. 
This will make iOS version very authentic.

mcdurdin:
We won't be able to do this in v15 -- we just don't have the people to work on it, and the team is fully committed on v15 functionality. Hopefully v16.

We welcome pull requests to improve this type of thing.

[...]

PR #10221, opened 2023-12-11
fix(web): Add BkSp1 special characer

MayuraVerma:
added 
'*BkSp1*': 0x21
to resolve https://github.com/keymanapp/keyman/issues/10220

### DIFF
--- a/web/src/engine/osk/src/specialCharacters.ts
+++ b/web/src/engine/osk/src/specialCharacters.ts
@@ -4,7 +4,7 @@ let specialCharacters = {
   '*Shift*':    8,
   '*Enter*':    5,
   '*Tab*':      6,
-  '*BkSp*':     4,
+  '*BkSp*':     4,   //filled shape denoting backspace, *BkSp1* is alternate outline shape
   '*Menu*':     11,
   '*Hide*':     10,
   '*Alt*':      25,
@@ -18,6 +18,7 @@ let specialCharacters = {
   '*Shifted*':  9,
   '*AltGr*':    2,
   '*TabLeft*':  7,
+  '*BkSp1*':    0x21, //outline shape denoting backspace, *BkSp* is alternate filled shape
   '*LAlt*':     0x56,
   '*RAlt*':     0x57,
   '*LCtrl*':    0x58,


keyman-server:

web/src/engine/osk/src/specialCharacters.ts
```diff
@@ -18,6 +18,7 @@ let specialCharacters = {
   '*Shifted*':  9,
   '*AltGr*':    2,
   '*TabLeft*':  7,
+  '*BkSp1*':    0x21, //outline shape denoting backspace, *BkSp* is alternate filled shape
```
darcywong00, line 21: Thanks for your submission! Could you move this block between lines 48-49?
That's the section of new characters added in Keyman 17.0.

At some point, we'd also want to update Developer to add this character. (Similar to #9846)

darcywong00: It would be also nice to update the test keyboard
web/src/test/manual/web/issue9469/test9469/source/test9469.keyman-touch-layout

and add a key with text `*BkSp1*` to test the new backspace keycap.

jahorton (changes requested):
Thanks for submitting a pull request!

Unfortunately, these changes only cover a _part_ of the problem you're seeking to address, rather than all of it.  Fortunately, the hardest part appears to already be in place - I can confirm that the symbol you're seeking to use does, in fact, exist already within the relevant font.

(For other reviewers: refer to https://github.com/silnrsi/font-symchar/blob/main/documentation/encoding.md, CTRL-F `U+E021`.  I believe it's a private repository, so apologies to external contributors.)

That said, note that this change does not selectively choose between `'*BkSp*'` and `'*BkSp1*'` when relevant - we'd want to pick the 'right' BkSp for the device, rather than make things mismatch worse on some.  That said, it does appear that modern iOS and Android devices (I quick-tested using an API 32 emulated device) both use something notably closer to your `'*BkSp1*'` than the existing `'*BkSp*'`.

Checking against Android API 21, that Android version's Gboard _does_ use something closer to `'*BkSp*'`, though.  I'm not sure where the transition point for this is... and/or how high priority we consider keeping the old style in place for older devices such as this one.

As this submission stands, current keyboards would see no change.  They'd have to explicitly specify use of the alternate backspace style within their touch-layout to see any difference.  We'd need to replace `'*BkSp*'`'s value directly to see the difference.  I'm not the final judge to convince on this, but I think it'd actually be reasonable to make `0x21` the default BkSp glyph and `0x04` a 'legacy' version.

From that, we'd _ideally_ have a way to condition on the Android version and auto-select the best-matching version on older Android devices... something that would require extra code.  Not sure if it's something we'd highly-prioritize, though.

@mcdurdin likes to pitch in on design / styling decisions like this, so I'm pinging him on this in hopes of getting a more definitive answer and/or direction to take here.  There's a chance that his decision might depend upon when the Gboard transition-point is - on how long ago Android's change to the BKSP glyph occurred, and that may take a bit of extra research to determine.

mcdurdin:
Thank you for your contribution.

I think this is not something we want to address in this way. It needs to be addressed with user-side theming support:

* On Android, the 'stock' keyboard is dependent on the vendor, and there is a wide range of different styles. On iOS, of course the 'stock' keyboard is always Apple. However, as I noted in #10226, we don't actually have a specific goal to precisely mimic keyboard styling from the stock platform.

* This should also not be a keyboard designer's styling decision, but rather something that is styling specific to the platform. So adding `*BkSp1*` is actually heading in the wrong direction.

* Furthermore, we don't currently have a glyph for the RTL backspace, outlined. So we'd need to add that before we could merge this in any case.

So for those reasons, I don't think we can move ahead with this, sorry.

MayuraVerma:
@mcdurdin Please let me know if I could work on any specific simple tasks. I can contribute to this project in little ways.

mcdurdin:
Hi @MayuraVerma, thank you for your contributions. They are always appreciated. One big area where we could really use help is in the documentation. Even looking at individual topics in the Developer documentation and improving the wording is helpful.

We have a few issues listed in this repo as ['help wanted'](https://github.com/keymanapp/keyman/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) and I know that you looked at for example #2655 relating to this specific PR, and we said we welcome pull requests in that issue, so this I appreciate may be a little frustrating.

We're currently holding off as far as possible on minor tweaks and improvements to the on screen keyboard, because each tweak represents another thing we have to factor into the upcoming theming rewrite. It's already tricky because a number of keyboards have their own styling and I don't know how we are going to integrate that with a more general theming system!

We'll keep brainstorming here for other "low hanging fruit" issues that you could take on.

MayuraVerma:
> Thank you for your contribution.
> 
> I think this is not something we want to address in this way. It needs to be addressed with user-side theming support:
> 
> * On Android, the 'stock' keyboard is dependent on the vendor, and there is a wide range of different styles. On iOS, of course the 'stock' keyboard is always Apple. However, as I noted in [feat(ios): How to achieve iOS stock keyboard backspace behavior #10226](https://github.com/keymanapp/keyman/issues/10226), we don't actually have a specific goal to precisely mimic keyboard styling from the stock platform.
> * This should also not be a keyboard designer's styling decision, but rather something that is styling specific to the platform. So adding `*BkSp1*` is actually heading in the wrong direction.
> * Furthermore, we don't currently have a glyph for the RTL backspace, outlined. So we'd need to add that before we could merge this in any case.
> 
> So for those reasons, I don't think we can move ahead with this, sorry.

@mcdurdin 
Although the symbol font has glyph for U+E021, keyman is not displaying the text, but a TOFU.

There is no documentation of embedded css/js to make this work.

Please help

Screenshot 2023-12-16 at 4 40 44 PM

image


OR

1. I will update the font with right hand version of glyph. I will close this issue. 
https://github.com/silnrsi/font-symchar/issues/2

2. I will update this documentation
https://help.keyman.com/developer/16.0/guides/develop/creating-a-touch-keyboard-layout-for-amharic-the-nitty-gritty

image

with a row indicating alternate glyph.

If it helps accepting this pull request.

Thank you.

mcdurdin:
> Although the symbol font has glyph for U+E021, keyman is not displaying the text, but a TOFU.
> 
> There is no documentation of embedded css/js to make this work.

The E021 codepoint is internal and subject to change, so keyboards should not make use of it. The codepoints can and may well change between Keyman versions. Thus, this is undocumented.

Issue #10182, opened 2023-12-07
refactor(core): KM_CORE_IT_INVALIDATE_CONTEXT shouldn't be no-op, or there should be other api to handle it


wengxt:
### Describe the bug

I'm trying to implement with new get_actions API, and noticed that there's no equivalent for KM_CORE_IT_INVALIDATE_CONTEXT in the km_core_actions struct.

I think it's still necessary have a flag to indicate whether KM_CORE_IT_INVALIDATE_CONTEXT appear in actions. Especially for application that doesn't support surrounding text on linux.

### Reproduce the bug

km_core_actions doesn't have a bool invalidate_context field.

### Keyman apps

- [ ] Keyman for Android
- [ ] Keyman for iPhone and iPad
- [X] Keyman for Linux
- [ ] Keyman for macOS
- [ ] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

17.0.225

### Operating system

Linux

### Keyboard name

telex

mcdurdin:
The <kbd>Enter</kbd> key triggers a context reset internally in Core, because it is a non-character key (even if in some contexts it emits a `\n`).

mcdurdin:
(But, this would be a good example of a unit test to verify that things are working as we intend!)

wengxt:
at least I didn't see that from happening in the log


D2023-12-07 18:02:02.526529 instance.cpp:905] KeyEvent: Key(Return states=0) rawKey: Key(Return states=0) origKey: Key(Return states=0) Release:0 keycode: 36
D2023-12-07 18:02:02.526722 engine.cpp:182] Set context from application: thuang
D2023-12-07 18:02:02.526756 engine.cpp:446] before process key event context: thuang
D2023-12-07 18:02:02.526790 engine.cpp:448] km_mod_state=0
364359405        ../../../src/kmx/kmx_context.cpp:108        Set        KMX_Context::Set(): ENTER [6]: (6) U+0074 U+0068 U+0075 U+0061 U+006E U+0067
364359405        ../../../src/kmx/kmx_context.cpp:109        Set                                        (6) U+0074 U+0068 U+0075 U+0061 U+006E U+0067
364359405        ../../../src/kmx/kmx_processevent.cpp:220        ProcessGroup        m_state.vkey: ['K_ENTER' 0xd] shiftFlags: 0; charCode: 0
[...]

PR #10230, opened 2023-12-12
fix(core): clear core context on invalidate cache

rc-swag:
Fixes: #10182 
The core context cache belonging to the state object is now cleared when the keyboard processor determines the context needs to be invalidated. 

Test k_000___null_keyboard has been updated as the [RALT K_B] should
cause the context to be reset.

Test k_008___vkey_input__ctrl_alt_2_ has been updated as the [LCTRL LALT
K_B] does not match a keyboard rule the context is deemed invalid and should be
reset.

It could be a consideration to modify the text_store object to better
reflect the state of an application context however, this would mean
we would need two expected results one for the application and one
for the context.

@keymanapp-test-bot skip
Only skipping because the Windows and Linux builds can't be used to test as they currently still use the actions list and clear the core context cache. MacOS is probably is probably there?


common/test/keyboards/baseline/k_000___null_keyboard.kmn
```diff
@@ -1,8 +1,8 @@
 store(&NAME) '000 - null keyboard'
-c Description: Tests null keyboard 
+c Description: Tests null keyboard
 c keys: [K_A][RALT K_B][SHIFT K_C]
-c expected: aC
-c context: 
+c expected: C
+c context:
```
mcdurdin, line 5: `expected` should be testing the output that is expected from the keyboard -- not the context. So the `expected` output would still be `aC` here, because the `a` is never deleted. We need something else to verify the cached context (which should also verify markers); perhaps something like this?

```
c expected: aC
c expectedContext: C
c context:
```

core/src/kmx/kmx_processor.cpp
```diff
@@ -247,6 +247,7 @@ kmx_processor::internal_process_queued_actions(km_core_state *state) {
       }
       break;
     case QIT_INVALIDATECONTEXT:
+      state->context().clear();
```
mcdurdin, line 250: :+1:

But ... I worry that it may be possible to clear the context at the wrong time during the kmx processing? Is that possible?

mcdurdin (changes requested):
I think the unit tests may need extra work for this to get to a resolution -- because they should be comparing the received output and not the cached context.

common/test/keyboards/baseline/k_000___null_keyboard.kmn
rc-swag: Yes we would need to add that. Did you see what I wrote above. 

> It could be a consideration to modify the text_store object to better
reflect the state of an application context however, this would mean
we would need two expected results one for the application and one
for the context.

Currently, the kmx.cpp believes expected === text_store == context;
Also note that these tests all assume non-context aware apps. Otherwise when the cache is invalidated on the next key stroke a Context aware app would just set the context again.

mcdurdin: Gotcha, I did miss that sorry. So we are on the same page. I think we should update the unit test accordingly, because of the ownership change of context.

core/src/kmx/kmx_processor.cpp
rc-swag: Great question I will do brief code inspection and see if It sounds correct.
internal_process_queued_actions is called 3 times

- [ x ] 1. kmx_processor::process_event *after* returning from _kmx.ProcessEvent. It is the final thing before returning

This should be absolutely fine it is not during the kmx processing.

- [ x ] 2. `kmx_processor::external_eventext`

This one is fine for now because the only event utilised for KM_CORE_EVENT_KEYBOARD_ACTIVATED. I think any future
external events that want do an action that causes the context to be cleared will be making that a deliberate action.

- [ ? ] 3. km_core_process_queued_actions->process_queued_actions->internal_process_queued_actions

This for the IMX keyboards...
Called in `kmhook_getmessage.cpp` line 285 
This case it can occur during kmx_processing of a keystroke, however the kmx_keyboard processors cache context will not be cleared I don't believe.

mcdurdin: Good analysis, thanks!

rc-swag:
The unit tests needed to be updated to handle the cases
where the context is invalidated by a keystroke. However, they
still need to be able to verify the final string the application
should receive. To do this an expected context token has been added
to the unit test format.

Question for @ermshiperete  and @sgschantz  I updated this so that the existing unit tests did not need the `expect context` field added.  The majority of tests are going to see expected and expected_context be the same. If `expected context` is not supplied it defaults to be the same as `expected`.
The alternative is that every test will need to set `expected` and `expected context`.  Thoughts?

ermshiperete:
I think making `expected` the default if `expected context` is not set makes sense.

core/tests/unit/kmx/kmx.cpp
```diff
@@ -241,13 +254,12 @@ run_test(const km::core::path &source, const km::core::path &compiled) {
       assert(ci->type != KM_CORE_CT_END && test_ci != test_context.end()); // Verify that both lists are same length
       assert(test_ci->type == ci->type && test_ci->marker == ci->marker);
     }
-
     km_core_context_items_dispose(citems);
-    if (text_store != buf) {
-      std::cerr << "text store has diverged from buf" << std::endl;
+    if ((!context_invalidated) && (text_store != core_context_str)) {
+      std::cerr << "text store has unexpectedly diverged from core_context" << std::endl;
       std::cerr << "text store: " << string_to_hex(text_store) << " [" << text_store << "]" << std::endl;
```
ermshiperete, line 260:
```suggestion
      std::cerr << "text store  : " << string_to_hex(text_store) << " [" << text_store << "]" << std::endl;
```

core/tests/unit/kmx/kmx.cpp
```diff
@@ -241,13 +254,12 @@ run_test(const km::core::path &source, const km::core::path &compiled) {
...
-      std::cerr << "context   : " << string_to_hex(buf) << " [" << buf << "]" << std::endl;
-      assert(false);
+      std::cerr << "core context   : " << string_to_hex(core_context_str) << " [" << core_context_str << "]" << std::endl;
```
ermshiperete, line 261:
```suggestion
      std::cerr << "core context: " << string_to_hex(core_context_str) << " [" << core_context_str << "]" << std::endl;
```
...to align the output

core/tests/unit/kmx/kmx.cpp
```diff
@@ -274,10 +286,11 @@ run_test(const km::core::path &source, const km::core::path &compiled) {
 
   std::cout << "expected  : " << string_to_hex(expected) << " [" << expected << "]" << std::endl;
   std::cout << "text store: " << string_to_hex(text_store) << " [" << text_store << "]" << std::endl;
-  std::cout << "context   : " << string_to_hex(buf) << " [" << buf << "]" << std::endl;
+  std::cout << "expected context  : " << string_to_hex(expected_context) << " [" << expected_context << "]" << std::endl;
+  std::cout << "context   : " << string_to_hex(core_context_str) << " [" << core_context_str << "]" << std::endl;
```
ermshiperete, line 290:
```suggestion
  std::cout << "expected context: " << string_to_hex(expected_context) << " [" << expected_context << "]" << std::endl;
  std::cout << "context         : " << string_to_hex(core_context_str) << " [" << core_context_str << "]" << std::endl;
```

core/tests/unit/kmx/kmx.cpp
```diff
@@ -274,10 +286,11 @@ run_test(const km::core::path &source, const km::core::path &compiled) {
 
   std::cout << "expected  : " << string_to_hex(expected) << " [" << expected << "]" << std::endl;
   std::cout << "text store: " << string_to_hex(text_store) << " [" << text_store << "]" << std::endl;
```
ermshiperete, line 288:
```suggestion
  std::cout << "expected        : " << string_to_hex(expected) << " [" << expected << "]" << std::endl;
  std::cout << "text store      : " << string_to_hex(text_store) << " [" << text_store << "]" << std::endl;
```

ermshiperete (approved):

ermshiperete (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.233-alpha](https://keyman.com/downloads/releases/alpha/17.0.233)

PR #10232, opened 2023-12-12
chore(web): Better match directory structure for headless engine tests

ermshiperete:
This change adds an additional `engine` directory and moves the headless tests to subdirectories of that. It also renames some test directories. Together this makes the layout of the tests better match the layout under `src` and so makes it easier to see what tests what.

@keymanapp-test-bot skip


jahorton:
Something seems wrong here.

From #10238, which is based on `master`:

image

From this PR's most recent build:

image

ermshiperete:
Now on the other side:

image

jahorton (approved):

keyman-server:

PR #10250, opened 2023-12-13
chore(web): Add unit tests for `pageContextAttachment`

ermshiperete:
@keymanapp-test-bot skip

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -73,6 +73,7 @@
     "@sentry/cli": "2.2.0",
     "c8": "^7.12.0",
     "chai": "^4.3.4",
+    "jsdom": "^23.0.1",
     "karma": "^6.4.1",
     "karma-browserstack-launcher": "^1.6.0",
     "karma-chai": "^0.1.0",
@@ -104,7 +105,7 @@
     "@keymanapp/recorder-core": "*",
     "@keymanapp/tslib": "*",
     "@keymanapp/web-utils": "*",
-    "@types/node": "^11.9.4",
+    "@types/node": "^18.19.3",
     "eventemitter3": "^4.0.0",
     "tslib": "^2.5.2"
   },
--- a/web/src/test/auto/headless/engine/attachment/pageContextAttachment.tests.ts
+++ b/web/src/test/auto/headless/engine/attachment/pageContextAttachment.tests.ts
@@ -1,0 +1,108 @@
+import { assert } from 'chai';
+import { JSDOM } from 'jsdom';
+import { PageContextAttachment } from 'keyman/engine/attachment';
+
+declare global {
+  namespace NodeJS {
+    interface Global {
+      document: Document;
+      window: Window;
+      navigator: Navigator;
+    }
+  }
+}
+
+describe('PageContextAttachment', () => {
+  function createDocument(content: string): Document {
+    const jsdom = new JSDOM(content);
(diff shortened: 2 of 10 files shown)


ermshiperete:
@jahorton wrote in https://github.com/keymanapp/keyman/pull/10233#issuecomment-1853269024:

Cause of Web's CI build failure:

```
02:20:09 
  [common/web/es-bundling] ## build completed successfully

02:20:11 
  node_modules/@types/node/globals.d.ts(114,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.

02:20:11 
  [common/web/utils] ## build failed

```

I have vague memories of a minor headache from this in the past, but I don't recall the details at the moment.  Maybe... using JSDOM might have updated `@types/node` (or similar) in a way that leads to this problem?  That's my initial "stab in the dark," at least.

But... I don't see any substitutions or deletions in `package.json` or `package-lock.json`, so that doesn't sound right either.  Huh.

ermshiperete:
I found [this](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/66617) issue: Getting AbortSignal error on compilation for @types/node version 16.6.2 ([#66617](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/66617)).

ermshiperete:
It looks like `@types/jsdom` was the culprit. Looks like we don't need it if we use a newer version of `@types/node`.

jahorton (approved):

web/package.json
```diff
@@ -104,7 +105,7 @@
     "@keymanapp/recorder-core": "*",
     "@keymanapp/tslib": "*",
     "@keymanapp/web-utils": "*",
-    "@types/node": "^11.9.4",
+    "@types/node": "^18.19.3",
```
mcdurdin, line 108: This should be devDependency surely?

web/src/test/auto/headless/engine/attachment/pageContextAttachment.tests.ts
```diff
@@ -0,0 +1,108 @@
...
+    it('empty doc has no elements', () => {
+      const doc = createDocument('<!doctype html><html><body></body></html>');
+      const sut = new PageContextAttachment(doc, null);
+      sut.listInputs();
+
+      assert.isEmpty(sut.sortedInputs);
+    });
+
+    it('supported input types', () => {
+      const doc = createDocument('<!doctype html><html><body> <input type="button"> <input type="checkbox"> <input type="color"> <input type="date"> <input type="datetime-local"> <input type="email"> <input type="file"> <input type="hidden"> <input type="image"> <input type="month"> <input type="number"> <input type="password"> <input type="radio"> <input type="range"> <input type="reset"> <input type="search"> <input type="submit"> <input type="tel"> <input type="text"> <input type="time"> <input type="url"> <input type="week"> <textarea></textarea></body></html>');
```
mcdurdin, line 41: Suggest that we use backtick operators and format this over multiple lines

keyman-server:

srl295:
This PR was where the issue fixed in #10279  first started failing in master

PR #10251, opened 2023-12-13
chore(linux): docker: fix Node.js install, make customizable

srl295:
🐋 
- fix the Node.js install script deprecation 
- fix the user-facing docs to just point people to <https://github.com/nodesource/distributions#ubuntu-versions> for installing.
- also make items configurable such as Ubuntu version and Node version

Fixes: chore(linux): script deprecation warning from nodesource #10024

@keymanapp-test-bot skip

### DIFF
--- a/common/web/input-processor/README.md
+++ b/common/web/input-processor/README.md
@@ -8,13 +8,7 @@ The Keyboard Processor module is an internal component of KeymanWeb, seen within
 * A local installation of [Node.js](https://nodejs.org/) v8.9+.
 	* Builds will call `npm install` to automatically install further necessary build dependencies.
 
-	* Linux users can run the following to update to LTS version of nodejs
-	
-```
-sudo apt-get install python-software-properties
-curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
-sudo apt-get install nodejs
-```
+	* Linux users can update to LTS version of nodejs by following instructions on the [NodeSource Distributions](https://github.com/nodesource/distributions#table-of-contents) page.
 
 **********************************************************************
 
--- a/docs/build/linux-ubuntu.md
+++ b/docs/build/linux-ubuntu.md
@@ -267,3 +267,21 @@ cd linux
 docker pull ubuntu:23.04 # (to make sure you have an up-to-date image)
 docker build . -t keymanapp/keyman-linux-builder:u23.04-node20 --build-arg OS_VERSION=23.04 --build-arg NODE_MAJOR=20
 ````
+
+### Using the builder with VSCode [Dev Containers](https://code.visualstudio.com/docs/devcontainers/tutorial)
+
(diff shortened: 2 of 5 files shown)


docs/build/linux-ubuntu.md
```diff
@@ -262,3 +257,31 @@ Once the image is built, it may be used to build parts of Keyman.
     keymanapp/keyman-linux-builder:latest \
     android/build.sh --debug
   ```
+
+### Customizing the builder
+
+You can use Docker [build args](https://docs.docker.com/build/guide/build-args/) to customize the image build. As an example, the following will build an image explicitly with Ubuntu 23.04 and Node.js 21
```
ermshiperete, line 263: I guess the sample uses that's Node 20 :smile: 

```suggestion
You can use Docker [build args](https://docs.docker.com/build/guide/build-args/) to customize the image build. As an example, the following will build an image explicitly with Ubuntu 23.04 and Node.js 20
```

ermshiperete (approved):

docs/build/linux-ubuntu.md
srl295: 21 on the brain for unrelated reasons.

common/web/input-processor/README.md
```diff
@@ -8,14 +8,8 @@ The Keyboard Processor module is an internal component of KeymanWeb, seen within
 * A local installation of [Node.js](https://nodejs.org/) v8.9+.
```
mcdurdin, line 8: v18 plus!

common/web/input-processor/README.md
```diff
@@ -8,14 +8,8 @@ The Keyboard Processor module is an internal component of KeymanWeb, seen within
...
 	* Builds will call `npm install` to automatically install further necessary build dependencies.
 
-	* Linux users can run the following to update to LTS version of nodejs
-	
-```
-sudo apt-get install python-software-properties
-curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
-sudo apt-get install nodejs
-```
+	* Linux users can update to LTS version of nodejs by following instructions on the [NodeSource Distributions](https://github.com/nodesource/distributions#table-of-contents) page.
```
mcdurdin, line 11: I'd be inclined just to leave this out as the details are in docs/build/linux-ubuntu.md anyway. Less to maintain.

common/web/input-processor/README.md
srl295:
```suggestion
* A local installation of [Node.js](https://nodejs.org/) v18+.
```

?

common/web/input-processor/README.md
srl295: so just a pointer? how much of this doc should be removed? there are a couple like this.

srl295: without this change, users following the existing instructions will get the giant "THIS SCRIPT IS GOING AWAY" message.  maybe we should move forward with this and make another task for reorganizing the docs?

keyman-server:

PR #10271, opened 2023-12-18
fix(mac): unit tests should expect core context clear after invalidate

sgschantz:
Fix two units that are expecting the context to be unchanged after a tab and enter keys are processed by core. These keys invalidate the context and should cause the core to clear the context.

@keymanapp-test-bot skip


rc-swag (approved):
lgtm

PR #10279, opened 2023-12-18
fix(web): unbreak some ts paths

srl295:
@keymanapp-test-bot skip


srl295:
ideally there'd be an eslint to catch this?

rc-swag (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.233-alpha](https://keyman.com/downloads/releases/alpha/17.0.233)

mcdurdin:
> ideally there'd be an eslint to catch this?

It feels like it should be caught at build time. It's kinda weird how it works in some build states but not others. Not sure how best to fix.

mcdurdin:
https://github.com/microsoft/vscode/issues/137446#issuecomment-973563410 is relevant

Issue #10291, opened 2023-12-21
bug(developer): extranous layering message when there's a bad marker


srl295:
to repro:

```xml
  <transform from="\m{lgtm}" to="ANYTHING" />
```

Issue:

```shell
k_008_transform_norm.xml - error KM00021: Markers used for matching but not defined: lgtm
k_008_transform_norm.xml - fatal KM05001: Unexpected exception: Error: Internal error: section bksp depends on uninitialized dependency vars, check ordering

Call stack:
Error: Internal error: section bksp depends on uninitialized dependency vars, check ordering
    at <anonymous> (/Users/srl295/src/keymanapp/keyman/developer/src/kmc-ldml/src/compiler/compiler.ts:185:19)
    at Array.forEach (<anonymous>)
    at LdmlKeyboardCompiler.compile (/Users/srl295/src/keymanapp/keyman/developer/src/kmc-ldml/src/compiler/compiler.ts:179:38)
    at buildLdmlKeyboardToMemory (/Users/srl295/src/keymanapp/keyman/developer/src/kmc/src/commands/buildClasses/BuildLdmlKeyboard.ts:60:13)
    at BuildLdmlKeyboard.build (/Users/srl295/src/keymanapp/keyman/developer/src/kmc/src/commands/buildClasses/BuildLdmlKeyboard.ts:16:25)
    at build (/Users/srl295/src/keymanapp/keyman/developer/src/kmc/src/commands/build.ts:154:18)
    at Command.<anonymous> (/Users/srl295/src/keymanapp/keyman/developer/src/kmc/src/commands/build.ts:83:13)
    at async Command.parseAsync (/Users/srl295/src/keymanapp/keyman/node_modules/commander/lib/command.js:935:5)
    at run (/Users/srl295/src/keymanapp/keyman/developer/src/kmc/src/kmc.ts:58:3)
    at <anonymous> (/Users/srl295/src/keymanapp/keyman/developer/src/kmc/src/kmc.ts:14:3)

    This error has been automatically reported to the Keyman team.
      Identifier:  84dc0e9746634ad1be4e2ec447b62b62
      Application: Keyman Developer
      Reported at: https://sentry.io/organizations/keyman/projects/keyman-developer/events/84dc0e9746634ad1be4e2ec447b62b62/
```

So what happened is that the `vars` section couldn't init due to the error.  But this appeared as a _dependency_ error, showing the internal err `KM05001`

We should track when a section fails to init with an error, and not show the dependency message.

[...]

PR #10300, opened 2023-12-22
fix(developer): quell internal error when  a section fails 🙀

srl295:
- when a required section is missing, only throw an internal error IF we haven't already flagged a problem.  for example, if 'vars' fails to compile because of a problem, and we already have an error on file, then don't throw an internal error that 'bksp' depends on uninitialized 'vars'.

- mimic this flow in the test helpers.

- add a test case for this in the test helpers, in test-tran

- add a strictError flag to say that no additional errs are allowed.

Fixes: #10291

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.238-alpha](https://keyman.com/downloads/releases/alpha/17.0.238)

Issue #10319, opened 2024-01-04
feat(developer): escaping should be \u{…} not \uXXXX 🙀


srl295:
- [X] use proper escaping, not `\uXXXX`
- [x] reject `\uXXXX`
- [X] convert `\u{X}` to `\uXXXX` - even for supplemental
- [X] escape marker matches
- [x] It turns out that we need to use char literals instead of escapes so that normalization works properly.  `e<marker>\\u0300\\u0320` needs to actually be `e<marker>U+0300U+0320` so that the normalization can apply - it won't apply to the escaped chars! So probably make a list of which chars are safe to unescape (skipping parens, brackets etc).
- [x] handle \u{...} in reorders

- split from [#9121](https://github.com/keymanapp/keyman/issues/9121)

srl295:
the "forbid \u...." moved to #10389

Issue #9446, opened 2023-08-09
feat(core,developer): detect bad unicode in kmxplus 🙀 


srl295:
from https://github.com/keymanapp/keyman/pull/9405#discussion_r1288818220

- on the kmx+ loading side, a pre-scan of all key=to and transform=to strings and make sure there's no illegal unicode - this would be a validation error, not an assert. This would be at runtime. 
- an `assert` in `emit_string()` on the C++ side to assert that nothing 'got through'

-----

On the developer side:
- ~on the developer side, kmc should also do the same so we catch it at compile time~ (done, elsewhere)
- ~Messaging strategy (for kmc)~
  - ~hint on PUA~
  - ~warn on unassigned~
  - ~error on illegal~

Related is the following, which is a Unicode-version dependent check for _unassigned_ codepoints, which is a warning not an error.
- #9464

srl295:
reworded to clarify that THIS issue is just checking illegal unicode at load time.

mcdurdin:
Note: using `\u{ffff}\u{0008}` is liable to break things badly because this is the sentinel for markers in strings so this should be prioritised. (Encountered when testing #13469)

PR #10356, opened 2024-01-11
feat(core): unescape u 🙀

srl295:
- for regex, convert all `\u{…}` to `\uXXXX` format _if they are syntax chars_ otherwise leave escaped
- restructure the marker parsing, which will help #10369 
- for analysis (#9446 ), de-escape `\uXXXX` format again

#10319

@keymanapp-test-bot skip


srl295:
issues repro'ed on linux locally and also wasm (butonly under linux, from mac wasm worked)

common/web/types/src/util/util.ts
```diff
@@ -88,6 +100,87 @@ export function unescapeString(s: string): string {
...
+  const code = ch.codePointAt(0);
+  if (code <= 0xFFFF) {
+    return '\\u' + hexQuad(code);
+  } else {
+    return '\\U' + hexOcts(code);
+  }
+}
+
+/** chars that must be escaped: syntax, C0 + C1 controls */
+const REGEX_SYNTAX_CHAR = /^[\u0000-\u001F\u007F-\u009F{}\[\]\\?.^$*-]$/;
```
mcdurdin, line 130: `(`, `)`, and `/` also need escaping, right?

core/src/kmx/kmx_processevent.cpp
```diff
@@ -11,7 +11,7 @@ using namespace kmx;
 
 /* Globals */
 
-KMX_BOOL km::core::kmx::g_debug_ToConsole = FALSE;
+KMX_BOOL km::core::kmx::g_debug_ToConsole = TRUE;
```
mcdurdin, line 14:
```suggestion
KMX_BOOL km::core::kmx::g_debug_ToConsole = FALSE;
```

We don't want this on in a release build. It's debug only.

core/src/ldml/ldml_markers.cpp
```diff
@@ -19,6 +19,10 @@ namespace core {
 namespace ldml {
 
 
+const std::u32string REGEX_PREFIX       = U"\\uffff\\u0008";  // does not include '\\' because it might be a range
```
mcdurdin, line 22: I don't really understand the comment `// does not include '\\' because it might be a range`

core/tests/unit/ldml/test_transforms.cpp
```diff
@@ -645,21 +645,30 @@ int test_strutils() {
...
     assert_equal(map.size(), 0);
   }
+  {
+    marker_map map;
+    std::cout << __FILE__ << ":" << __LINE__ << "   - bad1c" << std::endl;
+    const std::u32string src = U"6\U0000ffffzz"; // missing code
+    const std::u32string dst = remove_markers(src, map);
+    const std::u32string expect = U"6z";
+    zassert_string_equal(dst, expect);
+    assert_equal(map.size(), 0);
```
mcdurdin, line 670: I think given our discussion today, this is something that should never appear in keyboard processor inputs so we shouldn't need to cater to it, right?

mcdurdin (approved):

core/src/kmx/kmx_processevent.cpp
srl295: 🤦 this was supposed to be git skipped.

core/tests/unit/ldml/test_transforms.cpp
srl295: It was for robustness, but I think what we converged on was to leave it unchanged.

core/src/ldml/ldml_markers.cpp
srl295: I'll update the comment.
The whole sequence is (unescaped): `\uffff\u0008\uNNNN` where NNNN is a marker number.  The comment is trying to say that the `\u` part of `\uNNNN` is static, but it's not part of the match because it could also be `\uffff\u0008[...` if it's a "match any marker" `\m{.}` sequence.

mcdurdin: Gotcha I understand now.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.247-alpha](https://keyman.com/downloads/releases/alpha/17.0.247)

PR #10377, opened 2024-01-15
feat(core): add `km_core_state_context_debug` API 🌱

### LINKED ISSUE #10365: refactor(core): Add method to get context for debug purposes 🌱
From [Slack](https://keymanapp.slack.com/archives/C6Q9WS09G/p1704888921104999):

EB: We now have the new `km_core_state_context_set_if_needed` (which takes a UTF-16 string) and `km_core_state_context_clear` API methods which replace the previous `km_core_context_*` methods. It seems to me that we should have a method that returns the context for debug purposes (`km_core_state_context_get_debug`?). Then we wouldn't have to go through `km_core_context_items` by calling `km_core_context_get` and `km_core_context_items_to_utf8` which is a bit confusing compared to setting the context, and not symmetric.

MD: That seems like a good idea -- so long as it is clear it is for debugging use and not integration. The question is, should it return it as a formatted string to enforce that? Can you open an issue for that?

EB: Return a formatted string is a good idea. Currently I use `|string| (len) [Unicode code points]` on Linux, e.g. `|ខ្មែរ| (len:5) [U+1781 U+17d2 U+1798 U+17c2 U+179a ]`

mcdurdin:
Fixes #10365.

Adds a `km_core_state_context_debug` function that returns a km_core_cp string containing the current cached or intermediate context. The function must be used only for debug logging purposes -- the string is not intended to parsed for other purposes and the format may change in the future.

This commit adds the function and corresponding unit tests. Subsequent commits will update existing context debug functions in various engines to use this API endpoint instead.

@keymanapp-test-bot skip


core/src/km_core_state_api.cpp
```diff
@@ -360,3 +362,80 @@ km_core_status km_core_state_context_clear(
...
+    return _new_error_string(u"<could not retrieve context buffer size>");
+  }
+
+  std::vector<char> context_buffer(buf_size);
+  if(km_core_context_items_to_utf8(context_items, &context_buffer[0], &buf_size) != KM_CORE_STATUS_OK) {
+    km_core_context_items_dispose(context_items);
+    return _new_error_string(u"<could not retrieve context buffer>");
+  }
+
+  // construct the error message
```
mcdurdin, line 411:
```suggestion
  // construct the log message
```

core/include/keyman/keyman_core_api.h
```diff
@@ -1140,6 +1140,58 @@ KMN_API
...
+```
+### `km_core_state_context_debug`
+##### Description:
+Returns a debug formatted string of the context from the state.
+##### Return:
+A pointer to a km_core_cp UTF-16 string. Must be disposed of by a call
+to `km_core_cp_dispose`.
+##### Parameters:
+- __state__: A pointer to the opaque state object to be queried.
+- __context_type__: the type of context to retrieve from the state
```
ermshiperete, line 1171: A few more words explaining this parameter might be useful. It's not immediately obvious to me what `KM_CORE_DEBUG_CONTEXT_INTERMEDIATE` will return.

core/src/km_core_state_api.cpp
```diff
@@ -360,3 +362,80 @@ km_core_status km_core_state_context_clear(
...
+
+  int context_item_length = 0;
+  for(auto cp = context_items; cp->type != KM_CORE_CT_END; cp++, context_item_length++);
+
+  buffer << "|" << std::string(&context_buffer[0]) << "| (len: " << context_item_length << ") [";
+  for(auto cp = context_items; cp->type != KM_CORE_CT_END; cp++) {
+    auto flags = buffer.flags();
+    if(cp->type == KM_CORE_CT_CHAR) {
+      // A single Unicode codepoint
+      buffer << " U+" << std::hex << std::setfill('0') << std::setw(4) << std::hex << cp->character;
```
ermshiperete, line 423: Is it necessary to have `std::hex` twice?

ermshiperete (approved):

core/src/km_core_state_api.cpp
srl295: looks like a typo

srl295:
```suggestion
      buffer << " U+" << std::setfill('0') << std::setw(4) << std::hex << cp->character;
```

ermshiperete:
#10386 should fix the packaging failure.

linux/debian/libkeymancore.symbols
```diff
@@ -15,6 +15,7 @@ libkeymancore.so.1 libkeymancore #MINVER#
  km_core_context_length@Base 17.0.195
  km_core_context_set@Base 17.0.195
  km_core_context_shrink@Base 17.0.195
+ km_core_cp_dispose@Base 17.0.241
```
ermshiperete, line 18: You should use the version number that is current when it gets merged otherwise the API check will flag it.

core/include/keyman/keyman_core_api.h
srl295: More context is added in https://github.com/keymanapp/keyman/pull/10382

core/include/keyman/keyman_core_api.h
```diff
@@ -1140,6 +1140,58 @@ KMN_API
 km_core_status
 km_core_state_get_intermediate_context(km_core_state *state, km_core_context_item ** context_items);
 
+/*
+```
+### Context Debug Reporting
+As of version 17, the cached context is an internal property of the
+state, not exposed to the consumer of the API -- apart from the
+Keyman Developer Keyboard Debugger. However, for other debug
+purposes, it is helpful to be able to examine the cached context, so
```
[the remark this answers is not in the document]
mcdurdin: This is for use by consumers of the API -- i.e. Keyman Engine. So not really useful to keyboard authors or users of the Keyman Developer debugger. It's useful for us.

core/src/km_core_state_api.cpp
```diff
@@ -360,3 +362,80 @@ km_core_status km_core_state_context_clear(
...
+    } else {
+      // A marker
+      buffer << " M(" << cp->marker << ")";
```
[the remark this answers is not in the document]
mcdurdin: No, because we used to call markers 'deadkeys', so most other log statements have something like `dk(1)`.

PR #10379, opened 2024-01-15
refactor(linux): use `km_core_state_context_debug` 🌱

### LINKED ISSUE #10378: bug(linux): memory leak using `g_autofree`
When tweaking the debug logging for context in Keyman Engine for Linux (per #10365), I observed that `g_autofree` is probably being incorrectly used, resulting in a memory leak:

https://github.com/keymanapp/keyman/blob/9830f9ea24be608e54084ae189e14cfb363c3547/linux/ibus-keyman/src/engine.c#L937-L940

According to the [documentation](https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html) `g_autofree` (aka `__attribute__((cleanup))`) only operates when the variable goes out of scope, not when it is reassigned. So the first assignment to `debug_context` above will be leaked. There is a 3rd use of the same variable in the function which means the second assignment will also be leaked.

mcdurdin:
Relates to #10365.
Fixes #10378.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -205,27 +205,17 @@ ibus_keyman_engine_class_init (IBusKeymanEngineClass *klass)
 }
 
 static gchar *
-get_context_debug(IBusEngine *engine, km_core_context *context) {
+get_context_debug(IBusEngine *engine) {
   IBusKeymanEngine *keyman       = (IBusKeymanEngine *)engine;
-  km_core_context *state_context = context ? context : km_core_state_context(keyman->state);
 
-  size_t buf_size8 = 512, buf_size16 = 512;
-  km_core_context_item *context_items;
-  gchar *current_context_utf8 = g_new0(gchar, buf_size8);
-  km_core_cp *current_context_utf16 = g_new0(km_core_cp, buf_size16);
-  if (km_core_context_get(state_context, &context_items) == KM_CORE_STATUS_OK) {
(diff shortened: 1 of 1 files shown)


ermshiperete (approved):

linux/ibus-keyman/src/engine.c
```diff
@@ -205,27 +205,17 @@ ibus_keyman_engine_class_init (IBusKeymanEngineClass *klass)
...
-  if (km_core_context_get(state_context, &context_items) == KM_CORE_STATUS_OK) {
-    km_core_context_items_to_utf8(context_items, current_context_utf8, &buf_size8);
-    km_core_context_items_to_utf16(context_items, current_context_utf16, &buf_size16);
-  }
-  GString *output = g_string_new("");
-  g_string_append_printf(output, "|%s| (len:%zu) [", current_context_utf8, km_core_context_length(state_context));
-  for (int i = 0; i <  buf_size16 - 1; i++) {
-    g_string_append_printf(output, "U+%04x ", current_context_utf16[i]);
+
+  km_core_cp *buf = km_core_state_context_debug(keyman->state, KM_CORE_DEBUG_CONTEXT_CACHED);
```
srl295, line 211: Nice and compact!

srl295 (approved):

mcdurdin:
Debian Packaging build failure will be corrected during/after merge (libkeymancore.symbols)

PR #10382, opened 2024-01-15
feat(core): normalize input context 🌱

mcdurdin:
First half of #9999.

Adds the application context member of state object, which starts the process of verifying the handling of the application context.

Adds support for normalization (to NFD) of input app context into the cached context. The keyboard processor will work with the NFD cached context.

Adds unit tests for the normalization as part of the LDML keyboard processor test suite.

For follow-up PRs:
* Comparing modified cached context to app context to determine the transform required to send to the app
* Handling illegal unicode and unpaired surrogates on input context

@keymanapp-test-bot skip


core/include/keyman/keyman_core_api.h
```diff
@@ -1151,11 +1155,26 @@ a debug-formatted version of the context is made available with
...
+       and may be mixed normalization so may require fixups when
+       it is manipulated by keyboard processors that support
+       normalization, such as the LDML keyboard processor.
 ```c
 */
 typedef enum {
   KM_CORE_DEBUG_CONTEXT_CACHED        = 0,
-  KM_CORE_DEBUG_CONTEXT_INTERMEDIATE  = 1
+  KM_CORE_DEBUG_CONTEXT_INTERMEDIATE  = 1,
+  KM_CORE_DEBUG_CONTEXT_APP           = 2
```
srl295, line 1177: good explanation.

core/tests/unit/kmnkbd/debug_api.cpp
```diff
@@ -423,6 +424,7 @@ void test_backspace_markers() {
     {KM_CORE_CT_END}
   };
   try_status(km_core_context_set(km_core_state_context(test_state), marker_context));
+  try_status(km_core_context_set(km_core_state_app_context(test_state), marker_context)); // TODO: this needs to remove markers
```
srl295, line 427: if only there were a function for that.

`remove_markers(std::u32string &str /*... */)` could be used, leaving all other params default.

core/tests/unit/ldml/test_context_normalization.cpp
```diff
@@ -0,0 +1,162 @@
...
+  assert(is_identical_context(application_context, KM_CORE_DEBUG_CONTEXT_APP));
+  assert(is_identical_context(cached_context, KM_CORE_DEBUG_CONTEXT_CACHED));
+  teardown();
+}
+
+void test_context_normalization() {
+  test_context_normalization_already_nfd();
+  test_context_normalization_basic();
+  test_context_normalization_hefty();
+  // TODO: we need to strip illegal chars: test_context_normalization_invalid_unicode(); // -- unpaired surrogate, illegals
```
srl295, line 126: no function for that yet but could use the new-ish macros.

core/src/km_core_state_context_set_if_needed.cpp
```diff
@@ -0,0 +1,220 @@
...
+ * Normalize the input string using ICU
+ */
+bool do_normalize_nfd(km_core_cp const * src, std::u16string &dst) {
+  UErrorCode icu_status = U_ZERO_ERROR;
+  const icu::Normalizer2 *nfd = icu::Normalizer2::getNFDInstance(icu_status);
+  assert(U_SUCCESS(icu_status));
+  if(!U_SUCCESS(icu_status)) {
+    // TODO: log the failure code
+    return false;
+  }
```
srl295, line 197: could be (with including `ldml_utils.hpp` though in the wrong directory):

```suggestion
  if(!UASSERT_SUCCESS(icu_status)) {
    return false;
  }
```

perhaps it should be hoisted?

srl295 (approved):
LGTM

core/src/km_core_state_context_set_if_needed.cpp
mcdurdin: I looked at doing this originally, but would require some refactoring of the ldml_utils.hpp and cleanup of namespaces -- the macro refers implicitly to code in `km::core::ldml` namespace -- so opted to leave it out for now.

core/tests/unit/kmnkbd/debug_api.cpp
mcdurdin: apart from `remove_markers()` working on the internal string representation and not an `km_core_context_item` array.

mcdurdin: TODO: remove markers here and elsewhere where app_context is set directly in unit tests. Consider adding assertions to verify that app_context never contains markers. Or consider making the setter strip markers automatically.

mcdurdin: See #10388.

core/tests/unit/ldml/test_context_normalization.cpp
mcdurdin: See #10392

PR #10386, opened 2024-01-15
chore(core): Ignore C++ symbols

ermshiperete:
We have a C API, but (internal) C++ template instantiations are visible and so get flagged by `dpkg-gensymbols`. This change ignores the one use of C++ by marking it optional. This is listed as a usage scenario in the dpkg-gensymbols man page.

Note this requires a change how we call dpkg-gensymbols: if the `-O` parameter is specified, the file is not treated as being a template file (see man deb-src-symbols) and so the tags are not interpreted. Additionally the file gets overwritten with the non-tag version. So this change removes the `-O` parameter, and changes the archiving of the generated file.

@keymanapp-test-bot skip

### DIFF
--- a/.github/workflows/deb-packaging.yml
+++ b/.github/workflows/deb-packaging.yml
@@ -294,7 +294,7 @@ jobs:
       uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
       with:
         name: libkeymancore.symbols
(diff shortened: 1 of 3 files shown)


mcdurdin (approved):

mcdurdin:
Seems weird that we need to include our internal template instantiations but ok will go with it!

PR #10390, opened 2024-01-15
epic: Keyman Core normalization 🌱

### LINKED ISSUE #9999: feat(core): support normalization in Core 🌱

mcdurdin:
Fixes #9999.

@keymanapp-test-bot skip


rc-swag (approved):
RSLGTM

keyman-server:

PR #10403, opened 2024-01-16
feat(core): infrastructure for normalization of output 🌱

### LINKED ISSUE #9999: feat(core): support normalization in Core 🌱

### LINKED ISSUE #10421: feat(core): ensure that app_context tracks cached_context when normalization is not being used
From #10403.

As of #10403, the app_context is correctly updated to match the new reality after actions are processed by the app, but only when we are using normalization -- the actions_normalize code makes the update. If normalization is not used (i.e. KMN keyboards), then app_context gets out of sync. We need to update the app_context to match the cached_context in this situation -- a relatively simple transform because we just app_context.pop_back actions->code_points_to_delete times, and then append each actions->output USV to app_context (no need to worry about markers).

Plus unit tests of course.

https://github.com/keymanapp/keyman/blob/39f081391e1b57bb40cfc28c54910edc48b54846/core/src/actions_normalize.cpp#L28-L30

### LINKED ISSUE #10422: feat(core): renormalize cached_context across action boundary with markers
From #10403. Blocked by #10369.

https://github.com/keymanapp/keyman/blob/39f081391e1b57bb40cfc28c54910edc48b54846/core/src/actions_normalize.cpp#L81-L84

https://github.com/keymanapp/keyman/blob/39f081391e1b57bb40cfc28c54910edc48b54846/core/tests/unit/kmnkbd/test_actions_normalize.cpp#L259-L260

mcdurdin:
Fixes #9999.
Fixes #10421.
Fixes #10422.

Note TODO items for follow-up PRs: 
- [ ] #10423 
- [ ] #10424

@keymanapp-test-bot skip


core/tests/unit/kmnkbd/test_actions_normalize.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+  // If we don't reach a normalization boundary, we still should continue to work
+  test(
+    "Normalizable letters at start of context",
+    /* app context pre transform: */     u"\u0300",
+    /* cached context post transform: */ u"\u0300\u0323\u0302",
+    /* action del, output: */            0, U"\u0323\u0302",       // NFD input;
+    // ---- results ----
+    /* action del, output: */            1, U"\u0323\u0300\u0302",  // NFC output is still decomposed because there is no base
+    /* app_context: */                   u"\u0323\u0300\u0302"
```
ermshiperete, line 255: I would have expected the output and app_context to be `u"\u0300\u0323\u0302"`. Or did the normalization change the order? But then I would expect the cached context to have the same order.

core/src/actions_normalize.cpp
```diff
@@ -0,0 +1,285 @@
...
+  assert(app_context != nullptr);
+  if(actions == nullptr || cached_context == nullptr || app_context == nullptr) {
+    return false;
+  }
+
+  /*
+    The code_points_to_delete value at this point is in NFD. The cached_context
+    is in NFD and has already been updated by the keyboard processor to the
+    expected result of the action, so we need to remove the output from a copy
+    of the cached_context to start, in order to get it to the same position as
```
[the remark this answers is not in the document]
mcdurdin: Well, it _is_ a copy of the cached context, but not in the same format. That is, cached_context is not being modified. Updated the comment to clarify a little:

> ...so we need to remove the output from a **string** copy of the cached_context to start...

core/tests/unit/kmnkbd/test_actions_normalize.cpp
```diff
@@ -0,0 +1,299 @@
...
+/**
+ * Run a single test on actions_normalize. This is quite nuanced, because the
+ * input state is more than a little complex. We have inputs in NFU and NFD, and
+ * outputs counting NFU and inserting NFC. Be careful!
+ *
+ * @param name                          descriptive name for the test
+ * @param initial_app_context           the app context stored in the state,
+ *                                      _before_ transform is applied -- NFU
+ * @param final_cached_context          cached context _after_ actions have been
+ *                                      applied -- NFD except on the boundary of
```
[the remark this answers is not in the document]
mcdurdin: See #10422 for discussion, now correcting this description.

core/tests/unit/kmnkbd/test_actions_normalize.cpp
mcdurdin: Yes, the normalization should put `\u0323` first. Again, relates to #10422 discussion, my misunderstanding on what we could expect LDML keyboard processor to generate. Updated this test to match input assumption that cached_context is always NFD.

core/src/km_core_action_api.cpp
```diff
@@ -33,7 +33,22 @@ km_core_actions const * km_core_state_get_actions(
...
+
+  if(state->processor().supports_normalization()) {
+    // Normalize to NFC for those keyboard processors that support it
+    if(!actions_normalize(km_core_state_context(state), km_core_state_app_context(state), result)) {
+      km_core_actions_dispose(result);
+      return nullptr;
+    }
+  } else {
+    // For all other keyboard processors, we just copy the cached_context to the app_context
+    if(!actions_update_app_context(km_core_state_context(state), km_core_state_app_context(state))) {
```
mcdurdin, line 46: Perhaps we should add an `_nfu` to this to clarify this is for non-normalized use only?

ermshiperete: I think that would make sense. And the comment on `actions_update_app_context_nfu` should mention that this will be used for processors that don't support normalization.

ermshiperete (approved):

srl295 (approved):

Issue #10410, opened 2024-01-17
feat(core): ldml  to use action struct 🌱 


srl295:
to support #9999 in LDML

also assert(false) for:
- ldml_processor::queue_action
- ldml_processor::process_queued_actions

PR #10425, opened 2024-01-17
chore(core): ldml shuffle some deck chairs 🌱

srl295:
- move the emit_ and context functions into an ldml_event_state object

@keymanapp-test-bot skip

For: #10410 

Note: rebased to make non-changes more clear.

### DIFF
--- a/core/src/ldml/ldml_processor.cpp
+++ b/core/src/ldml/ldml_processor.cpp
@@ -174,58 +174,67 @@ ldml_processor::process_event(
     km_core_virtual_key vk,
     uint16_t modifier_state,
     uint8_t is_key_down,
-    uint16_t _kmn_unused(event_flags)
+    uint16_t event_flags
 ) {
   assert(state);
   if (!state)
     return KM_CORE_STATUS_INVALID_ARGUMENT;
 
-  try {
-    // At the start of every process_event always clear the action_items
-    state->actions().clear();
+  // this will hold the parameters, and our response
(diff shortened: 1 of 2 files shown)


core/src/ldml/ldml_processor.cpp
```diff
@@ -497,5 +508,26 @@ ldml_processor::context_to_string(km_core_state *state, std::u32string &str, boo
...
+    km_core_state *s,
+    km_core_virtual_key v,
+    uint16_t m,
+    uint8_t i,
+    uint16_t e) {
+  this->state          = s;
+  this->vk             = v;
+  this->modifier_state = m;
+  this->is_key_down    = i;
+  this->event_flags    = e;
```
mcdurdin, line 521: Can we use the same parameter names as in the header (and matching member props)?

```suggestion
    km_core_state *state,
    km_core_virtual_key vk,
    uint16_t modifier_state,
    uint8_t is_key_down,
    uint16_t event_flags) {
  this->state          = state;
  this->vk             = vk;
  this->modifier_state = modifier_state;
  this->is_key_down    = is_key_down;
  this->event_flags    = event_flags;
```

mcdurdin (approved):

core/src/ldml/ldml_processor.cpp
srl295: one of the platforms complained about shadowed parameters when I did that.

mcdurdin: ugh ok

Issue #10100, opened 2023-11-30
bug(core): `km_core_state_context_set_if_needed` needs more nuance for when to update cache 🌱


mcdurdin:
Now I am having worries about the algorithm here. There are multiple scenarios, and just want to make sure that I am covering them all.

Ignoring markers..., these are the scenarios I can see:

1. cached context identical to app context
2. cached context same length as app context but content is different
3. cached context is shorter than app context, but content is same as far as it goes
4. cached context is shorter than app context, but content is different
5. cached context is longer than app context, but content is same as far as it goes
6. cached context is longer than app context, but content is different.

I think those are the only 6 scenarios. What do we want to do for each one?

1. unchanged
2. changed, set cache to app context
3. lengthen to app context, but keep markers, consider unchanged?
4. changed, set cache to app context
5. shorten to app context, but keep markers, consider unchanged? or changed + set to app context?
6. changed, set cache to app context

We have to consider the rolling buffer that we'll have as we add new content, and that will impact whether or not the app context is longer than cache (remembering that the cached context may have markers which will shorten the overall non-marker content of it).

So number 3 and number 5 ~~6~~ need careful consideration.

Relates to #10098.

sgschantz:
For cases 3 and 5, I would think that you would return UPDATED simply to reflect exactly what happened. Keyman for Mac is currently ignoring the result, and I don't know of anything to do with the result except log for debug purposes. I don't think the result should drive any platform logic because core is the owner of the context, and the platform needs to accept whatever the core determines.

I thought 6 was straightforward -- why does it need careful consideration?

mcdurdin:
> For cases 3 and 5, I would think that you would return UPDATED simply to reflect exactly what happened. Keyman for Mac is currently ignoring the result, and I don't know of anything to do with the result except log for debug purposes. I don't think the result should drive any platform logic because core is the owner of the context, and the platform needs to accept whatever the core determines.

Agree that returning UPDATED makes sense for 3+5. Does the concept of keeping markers for these two scenarios make sense?

mcdurdin:
> There is a greater chance of that happening if the app context is 1 or 2 characters long, but still seems unlikely in practice.

And I think also the downside to that happening is pretty insignificant -- keeping markers rather than clearing them -- whereas the downside to the alternative -- losing markers when we want them -- would be a real problem.

rc-swag:
[...]

PR #10433, opened 2024-01-18
refactor(core): remove unused context APIs

### LINKED ISSUE #10431: chore(core): remove 'km_core_' prefix from context APIs that are not used externally, and move declarations into context.hpp

mcdurdin:
Fixes #10431.

The `km_core_` prefix has been removed from internal-only functions, and these function declarations moved to context.hpp.

The functions have not been moved from km_core_context_api.cpp at this stage.

Rewrote the function documentation in Javadoc style comments for the internal use functions.

@keymanapp-test-bot skip

### DIFF
--- a/core/include/keyman/keyman_core_api_context.h
+++ b/core/include/keyman/keyman_core_api_context.h
@@ -1,14 +1,7 @@
 /*
  * Keyman is copyright (C) SIL International. MIT License.
  *
- * Keyman Keyboard Processor API - Debugger Interfaces
- *
- * The debugger interfaces are still very dependent on .kmx
- * objects.
- *
- * Note: this file is subject to change; the debugger
- *       interfaces are not stable across versions.
- *
+ * Keyman Keyboard Processor API - Context Interfaces
  */
 
 #pragma once
@@ -17,12 +10,6 @@
 #include <stdlib.h>
 #include <keyman/keyman_core_api_bits.h>
 #include <keyman/keyman_core_api_vkeys.h>
-
-// Currently, the Core unit tests use private context APIs defined in
-// keyman_core_api_context.h, which are unused by other consumers. We are
-// hoping to remove these entirely in the future, so we restrict access
-// by default with this macro. Keyman Core internally uses these functions
-// #define _KM_CORE_ACCESS_PRIVATE_CONTEXT_API
 
 #if defined(__cplusplus)
 extern "C"
@@ -96,7 +83,7 @@ km_core_state_get_intermediate_context(km_core_state *state, km_core_context_ite
 ### `km_core_context_items_dispose`
 ##### Description:
 Free the allocated memory belonging to a `km_core_context_item` array previously
-returned by `km_core_context_items_from_utf16` or `km_core_context_get`
+returned by `km_core_context_items_from_utf16` or `context_get`
 ##### Parameters:
 - __context_items__: A pointer to the start of the `km_core_context_item` array
     to be disposed of.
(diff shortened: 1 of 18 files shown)


core/include/keyman/keyman_core_api_context.h
```diff
@@ -96,7 +83,7 @@ km_core_state_get_intermediate_context(km_core_state *state, km_core_context_ite
 ### `km_core_context_items_dispose`
 ##### Description:
 Free the allocated memory belonging to a `km_core_context_item` array previously
-returned by `km_core_context_items_from_utf16` or `km_core_context_get`
+returned by `km_core_context_items_from_utf16` or `context_get`
```
ermshiperete, line 86:
```suggestion
returned by `context_items_from_utf16` or `context_get`
```

ermshiperete: Hmm, if both methods that will return a `km_core_context_item` array are internal now, I wonder if `km_core_context_items_dispose` still has to  be public?

core/src/context.hpp
```diff
@@ -54,3 +54,196 @@ json & operator << (json &, km_core_context_item const &);
...
+ *                      `km_core_context_item`.
+ *
+ * @param out           A pointer to the result variable: A pointer to the start
+ *                      of the `km_core_context_item` array containing a copy of
+ *                      the context. Terminated with a type of `KM_CORE_CT_END`.
+ *                      Must be disposed of with
+ *                      `km_core_context_items_dispose`.
+ */
+km_core_status
+context_get(km_core_context const *context_items,
```
ermshiperete, line 194: Shouldn't the parameter be named `context`?

```suggestion
context_get(km_core_context const *context,
```

core/src/context.hpp
```diff
@@ -54,3 +54,196 @@ json & operator << (json &, km_core_context_item const &);
...
+ * This must be disposed of by caller using `km_core_context_items_dispose`.
+ *
+ * @return km_core_status
+ *         * `KM_CORE_STATUS_OK`: On success.
+ *         * `KM_CORE_STATUS_INVALID_ARGUMENT`: If non-optional parameters are
+ *           null.
+ *         * `KM_CORE_STATUS_NO_MEM`: In the event not enough memory can be
+ *           allocated for the output buffer.
+ *
+ * @param context_items A pointer to the start of an array
```
ermshiperete, line 184:
```suggestion
 * @param context A pointer to an opaque context object
```

core/tests/unit/kmnkbd/context_api.cpp
```diff
@@ -92,57 +79,57 @@ int main(int, char * [])
...
-  try_status(km_core_context_items_to_utf16(tmp_ctxt, ctxt_buffer, &ctxt_size));
+  try_status(context_get(&mock_ctxt1, &tmp_ctxt));
+  try_status(context_items_to_utf16(tmp_ctxt, ctxt_buffer, &ctxt_size));
   if (std::u16string(u"Hello World!") != ctxt_buffer) return __LINE__;
   km_core_context_items_dispose(tmp_ctxt);
 
   // Test shrink and prepend, delete more than we provide to prepend.
-  try_status(km_core_context_items_from_utf16(u"Bye, ", &ctxt1));
+  try_status(context_items_from_utf16(u"Bye, ", &ctxt1));
   // We delete 7 characters plus 1 marker hence 8 and not 7 as expected if you
```
ermshiperete, line 127:
```suggestion
  // We delete 7 characters (" World!") plus 1 marker hence 8 and not 7 as expected if you
```

core/src/context.hpp
```diff
@@ -54,3 +54,196 @@ json & operator << (json &, km_core_context_item const &);
...
+ * @param context       A pointer to an opaque context object.
+ * @param num           The number of items to remove from the end of context.
+ * @param context_items Pointer to the start of the `KM_CORE_CT_END` terminated
+ *                      array of `km_core_context_item` to add to the front. Up
+ *                      to `num` items will be prepended. This may be null if
+ *                      not required.
+ */
+KMN_API
+km_core_status
+context_shrink(km_core_context *context,
```
ermshiperete, line 247: I have some problems with this method. AFAICT it's only used in test code and for that it is ok, but it seems it should be in the test code then. For use as a general method I think the name doesn't match what it does (it also prepends besides removing from the end),  and it's not possible to prepend additional characters so it's less useful for general use.

I think it would be better to split it in two methods, `context_shrink` (with an additional parameter to be able to remove from beginning) and `context_prepend`. In fact, I can use such methods for #10100, so I'll base my changes on this PR and implement the split.

ermshiperete (approved):

core/src/context.hpp
```diff
@@ -54,3 +54,196 @@ json & operator << (json &, km_core_context_item const &);
...
+ *           memory to grow the context internally.
+ *
+ * @param context       A pointer to an opaque context object.
+ * @param num           The number of items to remove from the end of context.
+ * @param context_items Pointer to the start of the `KM_CORE_CT_END` terminated
+ *                      array of `km_core_context_item` to add to the front. Up
+ *                      to `num` items will be prepended. This may be null if
+ *                      not required.
+ */
+KMN_API
```
ermshiperete, line 245:
```suggestion
```

core/src/context.hpp
mcdurdin: OK. I only left it because it was tangled into the unit tests -- but happy to let you iterate improvement on it too.

core/include/keyman/keyman_core_api_context.h
mcdurdin: `km_core_state_get_intermediate_context` also uses `km_core_context_item`.

mcdurdin:
Debian packaging failed with:

```
[linux/scripts] ## verify starting...
[linux/scripts] OK: libkeymancore API didn't change
[linux/scripts] ERROR: libkeymancore.symbols file got changed without changing the package version number of the symbol
[linux/scripts] Major API change: 9 lines deleted and 0 lines modified
[linux/scripts] ERROR: Major API change without updating API version number in libkeymancore.symbols file
[linux/scripts] OK: API version in .symbols file and in CORE_API_VERSION.md is the same
[linux/scripts] ## verify failed
Error: Process completed with exit code 2.
```

core/src/context.hpp
```diff
@@ -54,3 +54,196 @@ json & operator << (json &, km_core_context_item const &);
...
+ *                passed a null `context` pointer.
+ * @param context A pointer to an opaque context object
+*/
+size_t
+context_length(km_core_context *);
+
+/**
+ * Add more items to the end (insertion point) of the context. If these exceed
+ * the maximum context length the same number of items will be dropped from the
+ * beginning of the context.
```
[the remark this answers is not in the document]
mcdurdin: True, there is a maximum context length (other than memory/int32 limits) in kmn but not in Keyman Core itself. LDML also seems to honour longer contexts, but haven't verified in detail.

mcdurdin:
KeymanWeb build is flaky (#10494). Not related to anything here.

PR #10441, opened 2024-01-18
feat(core): ldml to use action struct 🌱

srl295:
For: #10410

@keymanapp-test-bot skip


core/src/ldml/ldml_processor.cpp
```diff
@@ -519,14 +524,40 @@ ldml_event_state::ldml_event_state(
...
 void ldml_event_state::commit() {
-  state->actions().commit();
+  // TODO-LDML: is this necessary? what's the lifecycle of actions.output?
+  actions.output = new km_core_usv[text.length()+1];
+  memcpy(actions.output, text.c_str(), text.length()+1);
+  // current implementation copies actions.output and doesn't retain it
+  state->set_actions(actions);
+  // clean up
+  delete [] actions.output;
+  actions.output = nullptr;
```
mcdurdin, line 543: I'd move that into `clear()`

core/src/ldml/ldml_processor.cpp
```diff
@@ -519,14 +524,40 @@ ldml_event_state::ldml_event_state(
...
+
+  // initialize actions
+  clear();
 }
 
 void ldml_event_state::commit() {
-  state->actions().commit();
+  // TODO-LDML: is this necessary? what's the lifecycle of actions.output?
+  actions.output = new km_core_usv[text.length()+1];
+  memcpy(actions.output, text.c_str(), text.length()+1);
```
mcdurdin, line 538: memcpy isn't right here -- text.length() is in chars but we are looking at USVs. Something like:

```suggestion
  memcpy(actions.output, text.c_str(), (text.length()+1) * sizeof(text[0]));
```

srl295: Ooh thanks.  I still wonder what the lifecycle is. If it's guaranteed to not be used past the return, i could just pass in `text.c_str()`.

core/src/ldml/ldml_processor.cpp
srl295: Hmm, then I'm puzzled about the lifecycle

core/src/ldml/ldml_processor.cpp
mcdurdin: Yeah, we would always copy it internally so that ownership of the struct is clear. At present `set_actions` just pulls the data out of the struct and sticks it in the queue.

The owner should be able to pass a valid action struct to `km_core_actions_dispose()`. And `set_actions` does not transfer ownership.

core/src/ldml/ldml_processor.cpp
mcdurdin: Technically it's fine here, but it just seemed like `clear()` was a sensible place to flush out the struct entirely (same goes for the persist_options and where they are destroyed). But you own your actions_struct. So you can do what you want with it. set_actions will make a copy of what it needs -- ownership is not transferred.

core/src/ldml/ldml_processor.cpp
srl295: can that string be `const km_core_usv* output;` then?

core/src/ldml/ldml_processor.cpp
srl295: clear clears text() which is what i'm managing to hold the text string

core/src/ldml/ldml_processor.cpp
```diff
@@ -519,14 +524,40 @@ ldml_event_state::ldml_event_state(
...
+  // current implementation copies actions.output and doesn't retain it
+  state->set_actions(actions);
```
srl295, line 540: `km::core::state::set_actions()` currently does NOT call `_actions.commit()` internally, but it kind of seems like it ought to. otherwise the action queue becomes unterminated.

srl295: no, that's not right either, per the action_set_api.cpp test. It calls `set_actions()` and THEN pushes an `KM_CORE_IT_END` item (effectively `commit()`).. 

```c
void test_two_backspaces() {
  puts("test_two_backspaces");
  const km_core_action_item action_items[] = {
    bksp_action_item(KM_CORE_BT_CHAR, 0),
    bksp_action_item(KM_CORE_BT_CHAR, 0),
    end_action_item()
  };

  const km_core_actions actions = {
    2, // unsigned int code_points_to_delete;
    test_empty_output, // km_core_usv* output;
    test_env_opts, // km_core_option_item* persist_options;
    KM_CORE_FALSE, // km_core_bool do_alert;
    KM_CORE_FALSE, // km_core_bool emit_keystroke;
    KM_CORE_CAPS_UNCHANGED // new_caps_lock_state;
  };

  run_test(action_items, actions);
}
```

core/src/ldml/ldml_processor.cpp
```diff
@@ -259,7 +260,8 @@ void ldml_event_state::emit_backspace() {
     dumped in somewhere unknown, so we will have to depend on the app to
     be sensible about backspacing because we know nothing.
   */
-  state->actions().push_backspace(KM_CORE_BT_UNKNOWN);
+  // TODO-LDML: What here?
+  // state->actions().push_backspace(KM_CORE_BT_UNKNOWN);
```
mcdurdin, line 264: This should be `emit_keystroke = 1` I think

core/src/ldml/ldml_processor.cpp
```diff
@@ -242,13 +242,14 @@ void ldml_event_state::emit_backspace() {
...
-      state->actions().push_backspace(KM_CORE_BT_CHAR, (*end).character);
+      actions.code_points_to_delete++;
       state->context().pop_back();
       return;
     } else if ((*end).type == KM_CORE_BT_MARKER) {
-      state->actions().push_backspace(KM_CORE_BT_MARKER, (*end).marker);
+      // nothing in actions
       state->context().pop_back();
       // TODO-LDML: fall through here?
+      // no action?
```
mcdurdin, line 252: Correct, no action

core/src/ldml/ldml_processor.cpp
```diff
@@ -467,23 +468,26 @@ void
...
-  state->actions().push_marker(marker_no);
+  // No markers in the action struct!
   state->context().push_marker(marker_no);
 }
 
 void ldml_event_state::emit_invalidate_passthrough_keystroke() {
   if ((vk < 0x100) && km::core::kmx::vkey_to_contextreset[vk]) {
-    state->actions().push_invalidate_context();
+    // TODO-LDML: how to invalidate context?
+    context_clear();
```
mcdurdin, line 484: Engine decides whether a passed-through keystroke is a context invalidation event. So we don't have to make that decision in each keyboard processor, esp. as the set of invalidating keys may vary from platform to platform.

core/src/ldml/ldml_processor.cpp
mcdurdin: My understanding is that the terminator is only applied on a retrieved list of actions, context items, etc. Internally, the terminator is not used when it is a std::list.

core/src/ldml/ldml_processor.cpp
mcdurdin: hmm, I _think_ so?

core/src/ldml/ldml_processor.cpp
srl295: @mcdurdin but.. core throws an assertion failure if the list isn't terminated.

core/src/ldml/ldml_processor.cpp
srl295: OK. Then I'm back to not being able to test that.

core/src/ldml/ldml_processor.cpp
srl295: 1-ish

core/src/ldml/ldml_processor.cpp
srl295: i'm calling actions().commit() at the end of the processing for now. PTAL if it needs to change

mcdurdin: Yep, for now that seems sensible. One day the need for it may disappear when the action queue goes away entirely.

PR #10493, opened 2024-01-24
fix(developer): sentry-cli no longer injects into .d.ts 🌱

### LINKED ISSUE #10490: bug(common): sentry is injecting boilerplate into .d.ts files?
```
!function(){try{var e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof self?self:{},n=(new Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="efd73433-f1ea-5b91-9e57-c90d1baad20c")}catch(e){}}();
```

See e.g: common/web/types/build/src/util/compiler-interfaces.d.ts

This seems to be happening during kmc's `build.sh bundle` step.

mcdurdin:
Fixes #10490.

Note that the default is to inject into .js, .mjs, .cjs, so we don't need to override that.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/build.sh
+++ b/developer/src/kmc/build.sh
@@ -119,7 +119,6 @@ if builder_start_action bundle; then
     --org keyman \
     --project keyman-developer \
     --release "$VERSION_GIT_TAG"  \
-    --ext js --ext mjs --ext ts --ext map \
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

Issue #10317, opened 2024-01-04
feat(developer): normalization on the developer side 🙀 


srl295:
Yes, core will normalize, but this gives us more consistent files and earlier warnings.

- [x] This requires marker normalization in .ts !
- [x] Warn on use of precomposed ranges, i.e. [é-É] could exceed limits and, might not even be intended?
- Related is #10183 
- Supports #9468

srl295:
- blocked by #10369 in that we need a final marker algo before porting to .ts
- Can use `String.normalize("nfd")` for the implementation, which will be double checked by #10183

srl295:
- I very briefly went down the road of exposing the nfd data via wasm. However, common/web/types doesn't load wasm.
- `Intl.Segmenter()` might work, as in:

```js
> Array.from(new Intl.Segmenter(['und'], {granularity: 'grapheme'}).segment('e\u0302\u0300E\u0300\u0302'))
[
  { segment: 'ề', index: 0, input: 'ềÈ̂' },
  { segment: 'È̂', index: 3, input: 'ềÈ̂' }
]
```

.. however, what I need is the property info, not the segmentation.  It will get the wrong answer with markers:

```js
> Array.from(new Intl.Segmenter(['und'], {granularity: 'grapheme'}).segment('e\u0302\uffff\u0008\u0002\u0300E\u0300\u0302'))
[
  { segment: 'ê', index: 0, input: 'ê￿\b\x02̀È̂' },
  { segment: '￿', index: 2, input: 'ê￿\b\x02̀È̂' },
  { segment: '\b', index: 3, input: 'ê￿\b\x02̀È̂' },
  { segment: '\x02', index: 4, input: 'ê￿\b\x02̀È̂' },
  { segment: '̀', index: 5, input: 'ê￿\b\x02̀È̂' },
  { segment: 'È̂', index: 6, input: 'ê￿\b\x02̀È̂' }
]
```

Hm, maybe I could extract the data:

```js
> new Set(Array.from(new Intl.Segmenter(['und'], {granularity: 'grapheme'}).segment('e\u0302\u0300E\u0300\u0302u\u0300\u0308')).map(({segment})=>segment).map((c)=>c.codePointAt(0)));
Set(3) { 101, 69, 117 }
```

yea, that might work.

PR #10515, opened 2024-01-25
feat(developer): ldml load ccc from icu 🙀 

srl295:
#10317 

- segmenter isn't right, use a table from ICU4C
- also verify that `nfd.hasBoundaryBefore(ch)` is exactly equivalent to `isNFD(ch) && u_getCombiningClass(ch)==0`

Currently this is a .ts file in common/web/types which is updated by a special call to the `test_transforms` executable.  Alternatives:
- ICU-aware regex was considered, but is not exposed in any current implementation
- could be loaded through wasm - but, common doesn't link against wasm!
- this file could be generated at build time, but this then makes common/ depend on core, or on kmcmplib (the two places ICU code is done)

@keymanapp-test-bot skip


common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -197,13 +188,14 @@ export class MarkerParser {
       const rest = a.slice(i).join('');
       const p = MarkerParser.parse_next_marker(rest, forMatch);
       const have_marker = !!(p?.match);
+      const has_nfd_boundary_before = (nfdNoBoundaryBefore.indexOf(a[i]?.codePointAt(0)) === -1);
```
mcdurdin, line 191: When can `a[i]` be undefined and not a string? If `a[i]` is guaranteed to be a string then we should not be using optional chaining.

```suggestion
      const has_nfd_boundary_before = (nfdNoBoundaryBefore.indexOf(a[i].codePointAt(0)) === -1);
```

core/tests/unit/ldml/test_transforms.cpp
```diff
@@ -1033,6 +1033,44 @@ test_normalize() {
...
+  for (km_core_usv ch = 0; ch < 0x10FFFF; ch++) {
+    bool bb = nfd->hasBoundaryBefore(ch);
+    icu::UnicodeString s;
+    s.append((UChar32)ch);
+    bool lccc = nfd->isNormalized(s, status) && u_getCombiningClass(ch) == 0;
+    assert(U_SUCCESS(status));
+    if (bb != lccc) {
+      printf("0x%04x - bb=%s but lccc=%s\n", (unsigned int)ch, bb ? "y" : "n", lccc ? "y" : "n");
+    }
+    assert(bb == lccc);
```
mcdurdin, line 1064: If this is an error condition then it really should fail in all cases, not just print a warning. asserts only fail in debug builds.

core/tests/unit/ldml/test_transforms.cpp
```diff
@@ -1033,6 +1033,44 @@ test_normalize() {
...
+  const icu::Normalizer2 *nfd = icu::Normalizer2::getNFDInstance(status);
+  assert(U_SUCCESS(status));
+  fprintf(f, "// NFD hasBoundaryBefore\n");
+  fprintf(f, "// Generated by test_transforms.cpp\n");
+  fprintf(f, "// regenerate with: core/build/SOMETHING/SOMETHING/tests/unit/ldml/test_transforms --write-nfd\n");
+  fprintf(f, "export const ICU_VERSION='%s';\n", U_ICU_VERSION);
+  fprintf(f, "export const UNICODE_VERSION='%s';\n", U_UNICODE_VERSION);
+
+  fprintf(f, "export const nfdNoBoundaryBefore = [\n");
+  for (km_core_usv ch = 0; ch < 0x10FFFF; ch++) {
```
mcdurdin, line 1055: Might as well start at 0x20 ... less weird than starting at 0, and looks more like a normal "Unicode Character Range".

```suggestion
  for (km_core_usv ch = 0x20; ch < 0x10FFFF; ch++) {
```

core/tests/unit/ldml/test_transforms.cpp
```diff
@@ -1033,6 +1033,44 @@ test_normalize() {
   return EXIT_SUCCESS;
 }
 
+#include <unicode/uchar.h>
+
+#define NFD_FILE "common/web/types/src/ldml-keyboard/nfd-table.ts"
+int
+write_nfd_table() {
```
mcdurdin, line 1040: This function does not appear to be called from `main()`?

mcdurdin (changes requested):
This change looks solid, but I think I'd like to reduce the tech debt hole on file generation, by moving the tooling into a separate module, even though that's a little more work.

Various other feedback too :grin:

common/web/types/src/ldml-keyboard/nfd-table.ts
```diff
@@ -0,0 +1,923 @@
+// NFD hasBoundaryBefore
+// Generated by test_transforms.cpp
+// regenerate with: core/build/SOMETHING/SOMETHING/tests/unit/ldml/test_transforms --write-nfd
```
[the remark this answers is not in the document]
srl295: Ok, I can work on that.

common/web/types/src/ldml-keyboard/nfd-table.ts
```diff
@@ -0,0 +1,923 @@
...
+export const ICU_VERSION='70.1';
+export const UNICODE_VERSION='14.0';
+export const nfdNoBoundaryBefore = [
```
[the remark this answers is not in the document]
srl295: we will need it in Keyman Web.

common/web/types/src/ldml-keyboard/nfd-table.ts
```diff
@@ -0,0 +1,923 @@
+// NFD hasBoundaryBefore
+// Generated by test_transforms.cpp
+// regenerate with: core/build/SOMETHING/SOMETHING/tests/unit/ldml/test_transforms --write-nfd
+export const ICU_VERSION='70.1';
+export const UNICODE_VERSION='14.0';
```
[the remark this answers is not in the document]
srl295: I'm at a loss here.  Where did this ICU come from ?! 

maybe it was picked up on the path?

common/web/types/src/ldml-keyboard/nfd-table.ts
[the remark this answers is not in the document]
srl295: well, yeah, it's a complex folded trie structure in ICU.

core/tests/unit/ldml/test_transforms.cpp
srl295: this is a test, so don't asserts always fail in tests?

core/tests/unit/ldml/test_transforms.cpp
srl295: why 0x20?  Nulls are characters too!

core/tests/unit/ldml/test_transforms.cpp
srl295: bad merge, ok i will fix this.  (I had shelved this in favor of the segment)

common/web/types/src/ldml-keyboard/pattern-parser.ts
srl295: when `i === str_end` (see line 203) - this loop is called once at the end of the string. 

so `a[i]` can be undefined.

common/web/types/src/ldml-keyboard/nfd-table.ts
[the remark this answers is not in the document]
mcdurdin: Oof, if we are picking up ICU from path then we're going to get random versions. Not ideal!

common/web/types/src/ldml-keyboard/pattern-parser.ts
mcdurdin: can you add a comment accordingly somewhere? Because that's not necessarily intuitive to run loop past the end :grin:

core/tests/unit/ldml/test_transforms.cpp
mcdurdin: This is not really a test, it's a build process, and while it is currently hosted within a unit test executable, that will hopefully change, at which point we want to be aborting here even on a release build.

core/tests/unit/ldml/test_transforms.cpp
mcdurdin: ok, no worries. doesn't change output anyway.

srl295:
Obviated, see note on #10565

PR #10532, opened 2024-01-29
fix(developer): support surrogate pairs in ldml debugger 🌱

### LINKED ISSUE #10501: chore(developer): verify that surrogate pairs are handled correctly in the LDML debugger
Long story short: they aren't. Repro:

1. Paste the following into debug window: 😂😆🤣
2. Type a key that generates a marker.
3. See marker index is not maintained (as shown by '???' in the character preview for the marker)

mcdurdin:
Fixes #10501.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tike/child/Keyman.Developer.UI.Debug.UfrmLdmlKeyboardDebug.pas
+++ b/developer/src/tike/child/Keyman.Developer.UI.Debug.UfrmLdmlKeyboardDebug.pas
@@ -474,39 +474,50 @@ procedure TfrmLdmlKeyboardDebug.Run(vkey: Word);
     end;
   end;
 
+
+  // Returns the length of the context items in UTF16, treating markers as a
+  // single UTF16 code unit (U+FFFC)
+  function ContextItemsLengthUTF16(pc: pkm_core_context_item): Integer;
+  begin
+    Result := 0;
+    while pc._type <> KM_CORE_CT_END do
+    begin
+      if pc._type = KM_CORE_CT_CHAR then
+      begin
+        if Uni_IsSurrogate(pc.character) then
+          Inc(Result);
+      end;
+      Inc(Result);
+      Inc(pc);
+    end;
+  end;
+
   function ContextToDebugString(context_items: pkm_core_context_item; offset, length: Integer): string;
   var
-    outText: array of UInt32;
-    ox: Integer;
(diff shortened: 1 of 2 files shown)


developer/src/tike/child/Keyman.Developer.UI.Debug.UfrmLdmlKeyboardDebug.pas
```diff
@@ -474,39 +474,50 @@     TMemoSelectionState = record
     end;
   end;
 
+
+  // Returns the length of the context items in UTF16, treating markers as a
+  // single UTF16 code unit (U+FFFC)
+  function ContextItemsLengthUTF16(pc: pkm_core_context_item): Integer;
+  begin
+    Result := 0;
+    while pc._type <> KM_CORE_CT_END do
```
darcywong00, line 483: Is it possible for context items to never contain KM_CORE_CT_END?

mcdurdin: No, that's contractual on the part of the Core API:

https://github.com/keymanapp/keyman/blob/41cc2f4d1e4e828c791ce8e071eb04afe5968d4d/core/include/keyman/keyman_core_api_context.h#L198-L200

developer/src/tike/child/Keyman.Developer.UI.Debug.UfrmLdmlKeyboardDebug.pas
```diff
@@ -474,39 +474,50 @@     TMemoSelectionState = record
...
   var
-    outText: array of UInt32;
-    ox: Integer;
     pc: pkm_core_context_item;
     dk: TDeadKeyInfo;
   begin
-    SetLength(outText, length + 1);
-
-    ox := 0;
+    Result := '';
```
darcywong00, line 500: Should this initialize to something with `Result.Length = 0`? Could a loop enter the `else` block on the first pass and hit l.514?

mcdurdin: `Result := ''` means that `Result.Length = 0`; `.Length` is a calculated property

darcywong00 (approved):
lgtm

PR #10533, opened 2024-01-29
feat(core): add `deleted_context` to `km_core_actions` struct 🌱

### LINKED ISSUE #10530: feat(core): actions struct should report on deleted_context
```cpp
  result->deleted_context = tail_to_core_usv(km_core_state_app_context(state), result->code_points_to_delete);
```

should be added in `km_core_state_get_actions` immediately before updating app context. 

```diff
   km_core_actions * result = action_item_list_to_actions_object(action_items);
+
+  result->deleted_context = tail_to_core_usv(km_core_state_app_context(state), result->code_points_to_delete);

   if(state->processor().supports_normalization()) {
```

This must also be disposed correctly.

This makes it easier for consumers to figure out surrogate pairs etc (or utf-8 if necessary), and is a nice string to work with for unit tests.

mcdurdin:
Fixes #10530.

Adds the `deleted_context` member to the `km_core_actions` struct, and associated unit tests. Simplifies integration by providing the consumer with all the data they need in order to execute the transform, specifically around number of delete operations required, without needing to query the target application context again. The number of delete operations will vary according to application compliance and selected encoding; for example a UTF-16 string may require 2 delete-backs for surrogate pairs in the text buffer for a compliant app, whereas there will be a single delete-back key event for a non-compliant app.

The `deleted_context` member should also be used for debug assertions.

@keymanapp-test-bot skip

### DIFF
--- a/core/include/keyman/keyman_core_api.h
+++ b/core/include/keyman/keyman_core_api.h
@@ -242,6 +242,12 @@ typedef struct {
 
   // -1=unchanged, 0=off, 1=on
   km_core_caps_state new_caps_lock_state;
+
+  // reference copy of actual UTF32 codepoints deleted from rhs of context
+  // exactly code_points_to_delete in length (plus null terminator). Used to
+  // determine encoding conversion differences when deleting; only set when
+  // using km_core_state_get_actions, otherwise nullptr.
+  const km_core_usv* deleted_context;
 } km_core_actions;
 
 /*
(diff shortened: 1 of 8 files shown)


core/include/keyman/keyman_core_api.h
```diff
@@ -242,6 +242,12 @@ typedef struct {
 
   // -1=unchanged, 0=off, 1=on
   km_core_caps_state new_caps_lock_state;
+
+  // reference copy of actual UTF32 codepoints deleted from rhs of context
```
rc-swag, line 246: I think in the past we avoided "rhs" in the past due to right-to-left orientation. I believe kmx_context refers to the caret.
```reference copy of actual UTF32 codepoints deleted from the end of the context closest to the caret```

rc-swag (approved):
lgtm 

Just a suggestion about how the `end` of the context is described.

PR #10537, opened 2024-01-29
fix(linux): packaging on Ubuntu and Debian servers should give message instead of warning because we don't have patched ibus there

ermshiperete:
When building packages warnings are treated as errors. When building on Ubuntu and Debian servers we don't have a patched ibus version but still want to build the packages. So this change outputs a message instead of a warning when building packages.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.256-alpha](https://keyman.com/downloads/releases/alpha/17.0.256)

Issue #10516, opened 2024-01-25
feat(core,developer): reorders need to be normalization-safe 🙀 


srl295:
interacts with #10317 

- also, fix an issue with double markers found by @jahorton

PR #10539, opened 2024-01-29
feat(core): ldml reorder marker processing 🙀 

srl295:
- reorders don't interact with markers, but must transit them
- add updated tests
- simplify reorder_group::apply(): the 'common prefix' discussion
was out-of-date as std::mismatch is handled by the ldml_processor
and the core context.
- add_back_markers() promoted to SPI so it can be called from reordering.

Fixes: #10516

@keymanapp-test-bot skip

### DIFF
--- a/core/src/ldml/ldml_transforms.cpp
+++ b/core/src/ldml/ldml_transforms.cpp
@@ -331,15 +331,19 @@ reorder_group::apply(std::u32string &str) const {
   /** did we match anything */
   bool some_match = false;
 
+  // markers need to 'pass thorugh' reorders. remove and re-add if needed
+  marker_map markers;
+  std::u32string out = remove_markers(str, markers, plain_sentinel);
+
   // get a baseline sort key
-  auto sort_keys = reorder_sort_key::from(str);
+  auto sort_keys = reorder_sort_key::from(out);
 
   // apply ALL reorders in the group.
   for (const auto &r : list) {
     // work backward from end of string forward
     // That is, see if "abc" matches "abc" or "ab" or "a"
-    for (size_t s = str.size(); s > 0; s--) {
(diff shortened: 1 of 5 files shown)


core/src/ldml/ldml_transforms.cpp
```diff
@@ -331,15 +331,19 @@ reorder_group::apply(std::u32string &str) const {
   /** did we match anything */
   bool some_match = false;
 
+  // markers need to 'pass thorugh' reorders. remove and re-add if needed
```
mcdurdin, line 334:
```suggestion
  // markers need to 'pass through' reorders. remove and re-add if needed
```

core/src/ldml/ldml_transforms.cpp
```diff
@@ -420,29 +412,35 @@ reorder_group::apply(std::u32string &str) const {
...
   }
 #endif
-  return applied;
+  if (markers.empty()) {
+    // no markers, just copy over to str
+    str = out;
+  } else {
+    // copy and re-add markers
+    add_back_markers(str, out, markers, plain_sentinel);
+  }
```
mcdurdin, line 442: I think it'd be better to see this logic inside `add_back_markers()` -- simplifies usage patterns for it. It should have no effective performance difference.

mcdurdin (approved):
LGTM

core/src/ldml/ldml_transforms.cpp
srl295: yes. need to make an update downstream for this too, but added

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.258-alpha](https://keyman.com/downloads/releases/alpha/17.0.258)

PR #10565, opened 2024-01-30
feat(core,developer): simplify markers 🙀

srl295:
- with algorithm changes in #10541 #10563, we actually don't need the boundaryBefore code here. So #10515 is obsolete.
- the NFD algorithm won't reorder text over a normalization-safe boundary.  so the segments are still 'there' but they are ICU's problem.
- We may need the boundary detection in core for minimizing context changes, but it's a first class function in ICU so doesn't need anything in Keyman and definitely not in kmc.

#10317

@keymanapp-test-bot skip


keyman-server:

PR #10569, opened 2024-01-31
chore(core): move action apis into keyman_core_api_actions.h

### LINKED ISSUE #10520: refactor(core): hide old actions APIs from keyman_core_api.h
With the move to consuming action struct, we need to hide the old actions APIs from keyman_core_api.h, probably moving the required APIs to keyman_core_api_actions.h.

This correlates to the recent work done on context APIs in #10384.

To remove from keyman_core_api.h:

* `struct km_core_action_item`
* `struct km_core_backspace_item`
* `enum km_core_backspace_type`
* `enum km_core_action_type`
* `km_core_state_action_items()`
* `km_core_state_queue_action_items()`

# API Usage

## `km_core_state_action_items()`
* Keyman Developer - Debugger - Keyman.System.Debug.DebugEvent.pas
* Keyman for Windows - Engine - kmprocessactions.cpp (blocked on #10415)
* Keyman for Linux - fcitx?

## `km_core_state_queue_action_items()`
* Keyman for Windows - IMX - calldll.cpp

mcdurdin:
Fixes #10520.

The definitions for `struct km_core_action_item`, `struct km_core_backspace_item`, `enum km_core_backspace_type`, `enum km_core_action_type`, `km_core_state_action_items()`, `km_core_state_queue_action_items()`, `km_core_process_queued_actions()` are now in keyman_core_api_actions.h rather than in the primary keyman_core_api.h, as they should not be used by engines in general.

The current users of those APIs are the interactive kmx debugger, and the IMX integration in Keyman Engine for Windows. Both of these depend on the old action queue model rather than the actions struct.

In a future version, we may refactor these further to make them apply directly to the KMX processor, and provide access via an interface to the KMX processor rather than as a general Core API. Once this change is made, then Core itself will no longer support action queues at all, and will expect keyboard processors to fill in an action struct (this is already done for LDML).

@keymanapp-test-bot skip


core/include/keyman/keyman_core_api_actions.h
```diff
@@ -0,0 +1,151 @@
...
+#include <keyman/keyman_core_api_vkeys.h>
+
+#if defined(__cplusplus)
+extern "C"
+{
+#endif
+
+// ----------------------------------------------------------------------------------
+// Actions APIs are now available only to the keyboard debugger, IMX, and Core unit
+// tests (17.0)
```
ermshiperete, line 21: Should we enforce this by putting the definitions inside of an `#ifdef` block and add that define to debugger, IMX and core? Otherwise someone might discover this header file and include it without noticing the disclaimer. Of course it doesn't prevent someone using this but it requires an additional change which hopefully causes the disclaimer to be noticed.

ermshiperete (approved):
LGTM. I'm surprised it doesn't have to be included in debugger somewhere.

mcdurdin:
> I'm surprised it doesn't have to be included in debugger somewhere.

The debugger is still written in Delphi, so uses our internal .pas translation of the keyman_core_api headers, and I opted not to modify that, given this is still just a header change, and that code is effectively deprecated.

core/include/keyman/keyman_core_api_actions.h
mcdurdin: No, I think we should just put it into the v17 documentation for Core (TBD!)

srl295 (approved):

keyman-server:

PR #10576, opened 2024-01-31
fix(linux): Fix version comparison

### LINKED ISSUE #9593: bug(linux): Fix version comparison in `keymanutil.c`
`keymanutil.c` contains at least one comparison of versions. This currently uses string comparison which doesn't work very well for version numbers, e.g. `1.9` and `1.10`.

One scenario where a user would notice this, is if he has version 1.9 of a keyboard installed in the shared area. He then installs version 1.10 (which adds language X) into the user area. Until this bug is fixed the keyboard won't show up as keyboard for X.

ermshiperete:
Implement proper comparison of versions so that 1.10 is newer than 1.9.

Fixes #9593.

@keymanapp-test-bot skip

### DIFF
--- a/linux/ibus-keyman/src/keymanutil.c
+++ b/linux/ibus-keyman/src/keymanutil.c
@@ -212,6 +212,30 @@ get_engine_for_language(
   return engine_desc;
 }
 
+int
+keyman_compare_version(const gchar *version1str, const gchar *version2str) {
+  size_t len1 = strlen(version1str);
+  size_t len2 = strlen(version2str);
+  for (size_t i = 0, j = 0; i < len1 || j < len2; i++, j++) {
+      int version1 = 0, version2 = 0;
+
+      while (i < len1 && version1str[i] != '.') {
+        version1 = version1 * 10 + (version1str[i] - '0');
+        i++;
+      }
+      while (j < len2 && version2str[j] != '.') {
+        version2 = version2 * 10 + (version2str[j] - '0');
+        j++;
+      }
+
+      if (version1 < version2)
+        return -1;
+      if (version1 > version2)
+        return +1;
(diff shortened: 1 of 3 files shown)


linux/ibus-keyman/src/keymanutil.c
```diff
@@ -212,6 +212,30 @@ get_engine_for_language(
...
+        j++;
+      }
+
+      if (version1 < version2)
+        return -1;
+      if (version1 > version2)
+        return +1;
+    }
+  return 0;
+}
```
mcdurdin, line 237: This looks okay. But it might be cleaner to do it with pointers:

```suggestion
int
keyman_compare_version(const gchar *ver1, const gchar *ver2) {
  for (; *ver1 || *ver2; ) {
    int version1 = 0, version2 = 0;

    while (*ver1 && *ver1 != '.') {
      assert(*ver1 >= '0' && *ver1 <= '9');
      version1 = version1 * 10 + (*ver1 - '0');
      ver1++;
    }
    while (*ver2 && *ver2 != '.') {
      assert(*ver2 >= '0' && *ver2 <= '9');
      version2 = version2 * 10 + (*ver2 - '0');
      ver2++;
    }

    if (version1 < version2)
      return -1;
    if (version1 > version2)
      return +1;
    if(*ver1) 
      ver1++;
    if(*ver2)
      ver2++;
  }
  return 0;
}
```

mcdurdin (approved):

linux/ibus-keyman/src/keymanutil.c
ermshiperete: Done. Also added a few more tests.

keyman-server:

PR #10585, opened 2024-02-01
fix(core): make `km_core_state_get_actions()` idempotent

### LINKED ISSUE #10582: bug(core): `km_core_state_get_actions()` is not idempotent
Per discussion with @rc-swag and arising from #10557, we realised that `km_core_state_get_actions()` is mutating the app context that it holds internally, so a subsequent call breaks both the app context and the returned `deleted_context` member.

Bringing the app context into sync with the cached context should be done in `km_core_process_event()`, so then `km_core_state_get_actions()` can be read-only.

Ideally, `km_core_process_event()` should return the actions struct and no separate function call should be required. But this is better suited to a future refactor.

### LINKED ISSUE #10583: chore(windows): Update process actions to km_core_state_get_actions for external events once #10582 is complete
This issue will be a follow-up to  #10582.
For external events we need to be able to call km_core_state_get_actions, however, it needs to be an idempotent event before we can do that. 
Search the code for TODO: #10583 to see find where the code can be updated.

mcdurdin:
Fixes #10582.
Fixes #10583.

Returned struct from the `km_core_state_get_actions()` API is now owned by the state object. This is a change in API contract. The corresponding `km_core_actions_dispose()` API has been made private, because there is never any need for API consumers to call it. As this change is happening in alpha, we won't bump the ABI version.

`state->app_context` is now guaranteed to be in sync with `state->context` after `km_core_process_event()`, with the actions struct populated at that point.

The next and final step in this refactor is to remove the actions queue altogether from the Core and make that a kmn-only concept, but that will wait until 18.0. The only consumers of the actions queue are the kmn-specific interactive debugger in Keyman Developer, and kmn-specific Input Method eXtensions in Keyman Engine for Windows.

@keymanapp-test-bot skip

Note: moved user tests to #10618 because a number of other bugs addressed there which mean running the tests here would be significantly less useful.


core/src/km_core_action_api.cpp
```diff
@@ -30,71 +28,32 @@ km_core_actions const * km_core_state_get_actions(
     return nullptr;
   }
 
-  auto action_items = km_core_state_action_items(state, nullptr);
```
mcdurdin, line 33: This code has moved essentially without change into prepare_actions() in state.cpp

core/src/km_core_processevent_api.cpp
```diff
@@ -50,7 +50,11 @@ km_core_process_event(km_core_state *state,
   if(state == nullptr) {
     return KM_CORE_STATUS_INVALID_ARGUMENT;
   }
-  return state->processor().process_event(state, vk, modifier_state, is_key_down, event_flags);
+  km_core_status status = state->processor().process_event(state, vk, modifier_state, is_key_down, event_flags);
+
+  state->prepare_actions();
```
mcdurdin, line 55: This both updates the actions struct and brings app_context back into sync with cached context. Perhaps could benefit from a rename?

core/src/actions_normalize.cpp
```diff
@@ -201,7 +200,7 @@ bool km::core::actions_normalize(
     return false;
   }
 
-  if((status = km_core_context_set(app_context, app_context_items)) != KM_CORE_STATUS_OK) {
+  if((status = km_core_context_set(static_cast<km_core_context*>(app_context), app_context_items)) != KM_CORE_STATUS_OK) {
```
rc-swag, line 203: Isn't app_context already of type km_core_context*?

core/src/actions_normalize.cpp
```diff
@@ -212,23 +211,23 @@ bool km::core::actions_normalize(
...
  */
-icu::UnicodeString context_items_to_unicode_string(km_core_context const *context) {
+icu::UnicodeString context_items_to_unicode_string(km::core::context const *context) {
   icu::UnicodeString nullString;
   nullString.setToBogus();
 
   km_core_context_item *items = nullptr;
   km_core_status status;
-  if((status = km_core_context_get(context, &items)) != KM_CORE_STATUS_OK) {
+  if((status = km_core_context_get(static_cast<km_core_context const *>(context), &items)) != KM_CORE_STATUS_OK) {
```
rc-swag, line 230: same question as above isn't it already of this type?

core/src/actions_normalize.cpp
mcdurdin: It is `km::core::context` which is a class that subclasses the `km_core_context` struct. It may make more sense to bypass the public api call here, in some ways, and set the context directly, but I was trying to avoid more complex code.

core/src/km_core_processevent_api.cpp
rc-swag: I suggest `apply_actions`.

mcdurdin:
Debian Packaging failure has been addressed by #10573.

rc-swag (approved):
LGTM I suggest `prepare_actions` be renamed to `apply_actions`. 
I don't follow the static casting of the `contexts` to a type they already were.  Or this 
`return static_cast<km_core_actions *>(&(const_cast<km_core_state *>(state)->action_struct())); ` 
However, I think when I look again in the morning I should be able to work it out.

core/src/state.cpp
```diff
@@ -73,3 +76,41 @@ void state::imx_callback(uint32_t imx_id) {
   }
   _imx_callback(static_cast<km_core_state *>(this), imx_id, _imx_object);
 }
+
+state::~state() {
+  km::core::actions_dispose(this->_action_struct);
+}
+
+void state::prepare_actions() {
```
mcdurdin, line 84:
```suggestion
void state::apply_actions_and_merge_app_context() {
```

core/src/state.hpp
```diff
@@ -167,6 +173,7 @@ class state
     bool set_actions(
       km_core_actions const &actions
     );
+    void prepare_actions();
```
mcdurdin, line 176:
```suggestion
    void apply_actions_and_merge_app_context();
```

core/src/km_core_processevent_api.cpp
mcdurdin:
```suggestion
  state->apply_actions_and_merge_app_context();
```

ermshiperete:
Packaging fails because it lacks the fix from #10573

ermshiperete (approved):

srl295 (approved):

keyman-server:

Issue #10554, opened 2024-01-30
feat(developer): if normalization=false, do no normalization in kmc ldml compiler


mcdurdin:
Relates to #10317.
Extracted from #10468.

PR #10586, opened 2024-02-01
feat(core,developer): support normalization=disabled 🙀

srl295:
Fixes: #10554

- updates on the developer and core side
- 'meta' section (both sides) has a getter to return status
- .. plus tests of course
- also, dropped unneeded assert() calls and generally restructured ldml_processor::process_output

@keymanapp-test-bot skip


common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -75,6 +75,11 @@ export class Meta extends Section {
   indicator: StrsItem;
   version: StrsItem; // semver version string, defaults to "0"
   settings: KeyboardSettings;
+
+  /** convenience for checking settings */
+  get normalizionDisabled() {
```
mcdurdin, line 80:
```suggestion
  get normalizationDisabled() {
```

common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -205,11 +210,15 @@ export class Strs extends Section {
...
     // nfd
     if (opts?.nfd) {
-      if (opts?.markers) {
-        s = MarkerParser.nfd_markers(s, false);
-      } else {
-        s = s.normalize("NFD");
-      }
+      if (!sections.meta) {
+        throw Error(`Internal Error: need 'meta' section to check normalization mode.`);
+      } else if (!sections.meta.normalizionDisabled) {
```
mcdurdin, line 215:
```suggestion
      } else if (!sections.meta.normalatizionDisabled) {
```

common/web/types/src/kmx/kmx-plus.ts
```diff
@@ -205,11 +210,15 @@ export class Strs extends Section {
     }
     // nfd
     if (opts?.nfd) {
-      if (opts?.markers) {
-        s = MarkerParser.nfd_markers(s, false);
-      } else {
-        s = s.normalize("NFD");
-      }
+      if (!sections.meta) {
+        throw Error(`Internal Error: need 'meta' section to check normalization mode.`);
```
mcdurdin, line 214: If no `<meta>` section is included, does it default to normalizationDisabled = false?

core/src/ldml/ldml_processor.cpp
```diff
@@ -298,19 +298,26 @@ size_t ldml_processor::process_output(ldml_event_state &ldml_state, const std::u
   //  The normalize functions have assert and Debuglog at the bottom.
   //  so we do not need to assert the status here unless we're going to do something
   //  different with control flow.
-  (void)ldml::normalize_nfd_markers(nfd_str);
-
+  if (!normalization_disabled) {
+    auto norm_ok = ldml::normalize_nfd_markers(nfd_str);
+    assert(norm_ok);
```
mcdurdin, line 303:
```suggestion
    assert(ldml::normalize_nfd_markers(nfd_str));
```

Can we do this and reduce the noise?

core/src/ldml/ldml_processor.cpp
```diff
@@ -298,19 +298,26 @@ size_t ldml_processor::process_output(ldml_event_state &ldml_state, const std::u
   //  The normalize functions have assert and Debuglog at the bottom.
   //  so we do not need to assert the status here unless we're going to do something
   //  different with control flow.
-  (void)ldml::normalize_nfd_markers(nfd_str);
-
+  if (!normalization_disabled) {
```
mcdurdin, line 301: This is a repeated pattern, should we be extracting it out?

core/src/ldml/ldml_processor.cpp
```diff
@@ -330,14 +337,18 @@ size_t ldml_processor::process_output(ldml_event_state &ldml_state, const std::u
 
   // drop last 'matchedContext':
   ctxtstr.resize(ctxtstr.length() - matchedContext);
-  ctxtstr.append(outputString); // TODO-LDML: should be able to do a normalization-safe append here.
-  (void)ldml::normalize_nfd_markers(ctxtstr);
-
+  ctxtstr.append(outputString); // TODO-LDML: should be able to do a normalization-safe append here.'
+  if (!normalization_disabled) {
+    (void)ldml::normalize_nfd_markers(ctxtstr);
```
mcdurdin, line 342: Why no assertion on this instance? Seems inconsistent (See extracting-it-out comment above)

core/src/ldml/ldml_transforms.cpp
```diff
@@ -512,9 +514,11 @@ transform_entry::init() {
   }
   // TODO-LDML: if we have mapFrom, may need to do other processing.
   std::u32string from2 = fFrom;
-  normalize_nfd_markers(from2, regex_sentinel);
+  if (!normalization_disabled) {
+    // normalize, including markers, for regex
+    normalize_nfd_markers(from2, regex_sentinel);
```
mcdurdin, line 519: no assertion?

core/src/ldml/ldml_transforms.cpp
```diff
@@ -637,7 +641,7 @@ transform_entry::apply(const std::u32string &input, std::u32string &output) cons
     }
     output.assign(s.get(), out32len);
     // NOW do a marker-safe normalize
-    if (!normalize_nfd_markers(output)) {
+    if (!normalization_disabled && !normalize_nfd_markers(output)) {
       DebugLog("normalize_nfd_markers(output) failed");
       return 0; // TODO-LDML: normalization failed.
     }
```
mcdurdin, line 647: No assertion?

mcdurdin (approved):
LGTM

There are still many TODO-LDML comments, some of which are a concern, relating to normalization. Can we review and make sure there are no gaps before beta?

core/src/ldml/ldml_processor.cpp
srl295: @mcdurdin except that the `assert()` is not called in production, right? maybe we need a new construct.  

`#define assert(x)` is the usual idiom for production mode.

core/src/ldml/ldml_processor.cpp
srl295: @mcdurdin perhaps an inline member would make sense.

core/src/ldml/ldml_processor.cpp
mcdurdin: Ah of course. My bad ... that could be ugly.

`assert(x)` is dangerous like that... That does mean that `norm_ok` is declared and unused in production code which is possibly a separate painful issue.

srl295: if the code was extracted out, then the pain would be only in one place.

But actually.  the inner level `normalize()` function *already* asserts.  So one thing I could do is just take out all of these other asserts.

srl295: hah, and i had added this comment three weeks ago:

```
size_t ldml_processor::process_output(ldml_event_state &ldml_state, const std::u32string &str, ldml::transforms *with_transforms) const {
  std::u32string nfd_str = str;
  // Note:
  //  The normalize functions have assert and Debuglog at the bottom.
  //  so we do not need to assert the status here unless we're going to do something
  //  different with control flow.
```

mcdurdin (approved):

srl295:
> LGTM
> 
> 
> 
> There are still many TODO-LDML comments, some of which are a concern, relating to normalization. Can we review and make sure there are no gaps before beta?

I think i took out lots of todos- in any event yes will remove as discussed

keyman-server:

PR #10599, opened 2024-02-01
chore(linux): Use correct format specifier for `g_utf8_strlen`

ermshiperete:
The package build on armhf failed because of the wrong format specifier which expects a `size_t` instead of `glong` that `g_utf8_strlen` returns.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #10612, opened 2024-02-02
chore(linux): Fix API check if lines got removed in .symbols file

### LINKED ISSUE #10453: chore(linux): Fix API check if lines got removed in .symbols file
If a PR removes lines from the .symbols file and updates the API version number, the API check will still fail with " ERROR: libkeymancore.symbols file got changed without changing the package version number of the symbol" because it can't find the current version number.

~~Additionally we only check the last commit, but should check all commits in the PR.~~

**Edit:** We already check against the base of the PR.

ermshiperete:
If lines got removed in the .symbols file we won't find the current version number in the .symbols file. This change fixes the expectations.

Also, if a major API change happens during the Alpha cycle we don't expect another API version increment if one already happened since we released the stable version.

This change also adds some unit tests for the new functionality. These can be run manually and won't run during CI builds.

Fixes #10453.

@keymanapp-test-bot skip

### DIFF
--- a/linux/.gitignore
+++ b/linux/.gitignore
@@ -6,7 +6,7 @@ ltmain.sh
 __pycache__
 *.egg-info
 build-*
-deb-*
+./deb-*
 builddebs
 keyman-config/make_deb
 tmp/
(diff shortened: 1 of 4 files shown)


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.260-alpha](https://keyman.com/downloads/releases/alpha/17.0.260)

Issue #10316, opened 2024-01-04
feat(developer): regex: support ranges 🙀


srl295:
> … If I have a transform from [\u{00E8}-\u{00EB}] (èéêë) to [\u{00EC}-\u{00EF}] (ìíîï), how will that work with decomposition? Do we need to expand ranges (beware 0020-10FFFF)?

Note that actually we don't transform TO a range.  That would be a Set which is well-defined here.

but the matching is still an issue to solve in this ticket.

This is under _core_ because it has to be handled on the core side.

Split from [#9468](https://github.com/keymanapp/keyman/issues/9468)
Supports [#9121](https://github.com/keymanapp/keyman/issues/9121)

srl295:
- [ ] Warn on use of precomposed ranges, i.e. [é-É] could exceed limits and, might not even be intended?

srl295:
Icu can support `\Uhhhhhhhh` escapes, so this can be a pass-through if ther'es no decomposition:   `[\U00000-\UFFFFFF]`

mcdurdin:
Idea: we decide to not decompose any characters represented within ranges. This means that some characters will never be matched, but it's not going to break things. There will be some benefit, e.g. allowing large ranges, allowing out-of-alphabet matches (negative matches) etc.

So if we decide not to decompose characters within ranges, then we will end up with something like:
`/èéêë[^èéêë]/` ==> `/e\U00301...[^èéêë]/` or `/e\U00301...[^\U00eb...]/`

srl295:
- HINT if the range crosses non-NFD
- WARN if the range starts or ends with non-NFD

mcdurdin:
This should make the Core changes minimal, correct? Because we no longer need to convert to a Set.

mcdurdin:
Changing this to be a Developer issue. We can push this one into beta if it's only hints and warnings. If there is other work to be done, can you detail that here @srl295?

PR #10614, opened 2024-02-02
feat(developer): range warning for non-NFD chars 🙀 

srl295:
Fixes: #10316 

@keymanapp-test-bot skip


common/web/types/src/util/util.ts
```diff
@@ -350,5 +353,36 @@ export class BadStringAnalyzer {
...
+/** Analyzer that checks if something isn't NFD */
+export class NFDAnalyzer extends StringAnalyzer {
+  protected analyzeCodePoint(c: string, ch: number): BadStringType {
+    const nfd = c.normalize("NFD");
+    if (c !== nfd) {
+      return BadStringType.denormalized;
+    } else {
+      return null;
+    }
+  }
```
mcdurdin, line 387: I wonder about the performance of this -- seems like it could be quite expensive?

mcdurdin (approved):

mcdurdin:
Our messages (errors, hints, and warnings) don't have line numbers at present, which is going to be quite painful for users!

common/web/types/src/util/util.ts
srl295: right. maybe some kind of cache would help.

srl295: yes, could perhaps cache?

srl295: it's only on actual ranges in transforms.

srl295:
> Our messages (errors, hints, and warnings) don't have line numbers at present, which is going to be quite painful for users!

~what about...~ << moving it to a new ticket! #10636

common/web/types/src/util/util.ts
mcdurdin: OK, limited use so we can probably ignore for now. If it's always a single codepoint we could build a map and ship it -- trivial. Danger with that is that if we use `normalize()` elsewhere then the map could be inconsistent wrt Unicode version.

keyman-server:

PR #10650, opened 2024-02-06
fix(windows): support unicode strings properly in logs

### LINKED ISSUE #10604: bug(windows): debug logs that contain surrogate pairs end up with blank strings
`SendDebugMessageFormat("%ws", wide_string)` ends up with an empty string if `wide_string` contains any surrogate pairs. This can make debugging painful!

Looks like a limitation in the underlying format functions.

Note also `wstrtostr` crashes if you pass it surrogate pairs because they won't convert ... ouch.

Suggest replacing with `WideCharToMultiByte`

mcdurdin:
Fixes #10604.

Introduces the following wide-string macros corresponding to existing macros:

* `SendDebugMessageFormatW`
* `SendDebugMessageW`

External (private use) API added, corresponding to existing private API:

* `Keyman_WriteDebugEventW`

Also includes minimal patch to etl2log to support emitting Unicode characters, in UTF-8.

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/engine/keyman32/DebugEventTrace.cpp
+++ b/windows/src/engine/keyman32/DebugEventTrace.cpp
@@ -30,6 +30,16 @@ int GetActualShiftState() {   // I4843
 }
 
 extern "C" void _declspec(dllexport) WINAPI Keyman_WriteDebugEvent(char *file, int line, PWCHAR msg) {
+  // This function exists solely to provide a stable transition between versions
+  // of Keyman, so that things don't crash if an older version of kmtip.dll
+  // happens to call this function in a newer installation. It is not used in
+  // current versions of Keyman. This change came in around 17.0.260-alpha
+  PWCHAR fileW = strtowstr(file);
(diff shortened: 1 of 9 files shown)


ermshiperete (approved):
LGTM, although it's not obvious from the PR description why it's necessary to use `wchar_t*` for the `file` parameter. Doesn't seem to be immediately related to the linked issue. But I guess we want to be able to deal with filenames with emojis :smile:

mcdurdin:
> although it's not obvious from the PR description why it's necessary to use `wchar_t*` for the `file` parameter.

Good catch. It's not important but it does save a memory allocation + conversion + free for every log statement.

rc-swag:
> Doesn't seem to be immediately related to the linked issue.

Yeah I was confused also in the sense I was like wait we already had wide string and it was just the file name? How does that make it not work? All good.

rc-swag (approved):
LGTM

keyman-server:

PR #10651, opened 2024-02-06
fix(developer): ensure project populate files actually adds files

markcsinclair:
also adds unit test.

@keymanapp-test-bot skip


keyman-server:

mcdurdin (approved):
LGTM

PR #10652, opened 2024-02-06
fix(linux): Use raw string for regex

ermshiperete:
This change fixes a syntax warning with Python 3.12. See https://stackoverflow.com/a/50504635/487503 and
https://docs.python.org/3/whatsnew/3.12.html#other-language-changes.

@keymanapp-test-bot skip


darcywong00 (approved):
lgtm

mcdurdin (approved):

keyman-server:

PR #10660, opened 2024-02-07
fix(windows): `wstrtostr` should use `WideCharToMultiByte`

mcdurdin:
Relates to #10604: "Note also wstrtostr crashes if you pass it surrogate pairs because they won't convert ... ouch."

@keymanapp-test-bot skip

### DIFF
--- a/windows/src/engine/keyman32/kmprocess.cpp
+++ b/windows/src/engine/keyman32/kmprocess.cpp
@@ -221,11 +221,11 @@ PWSTR strtowstr(PSTR in)
 PSTR wstrtostr(PCWSTR in)
 {
     PSTR result;
-    size_t len;
+    int len;
 
-    wcstombs_s(&len, NULL, 0, in, wcslen(in));
+    len = WideCharToMultiByte(CP_ACP, 0, in, -1, NULL, 0, NULL, NULL);
     result = new CHAR[len+1];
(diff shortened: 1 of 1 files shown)


windows/src/engine/keyman32/kmprocess.cpp
```diff
@@ -221,11 +221,11 @@ PWSTR strtowstr(PSTR in)
...
     PSTR result;
-    size_t len;
+    int len;
 
-    wcstombs_s(&len, NULL, 0, in, wcslen(in));
+    len = WideCharToMultiByte(CP_ACP, 0, in, -1, NULL, 0, NULL, NULL);
     result = new CHAR[len+1];
-    wcstombs_s(&len, result, len, in, wcslen(in));
+    WideCharToMultiByte(CP_ACP, 0, in, -1, result, len, NULL, NULL);
     result[len] = 0;
```
rc-swag, line 229: They way read if cchWideChar = `-1` then we don't need this line, but I guess it makes it certain.

rc-swag (approved):
lgtm

windows/src/engine/keyman32/kmprocess.cpp
mcdurdin: True. What about failure cases? This is an ugly function anyway and it should be replaced but it is what we have for now...

keyman-server:

PR #10668, opened 2024-02-07
fix(linux): Ignore keys with IBUS_MOD4_MASK set

### LINKED ISSUE #10476: bug(linux): Switching keyboards doesn't work with Keyman keyboard on Ubuntu 23.10 Mantic in Wayland mode
Ubuntu allows you to switch between keyboards with the <kbd>Super</kbd>-<kbd>Space</kbd> shortcut key.

However, on Ubuntu 23.10 Mantic this no longer works if a Keyman keyboard is active and the desktop runs Wayland.

ermshiperete:
If we process keys that have the (Super/Meta/Windows) IBUS_MOD4_MASK modifier set we ignore the key. Core doesn't handle the Meta modifier state and so would treat it as regular key. This fixes switching keyboards in Ubuntu 23.10/24.04 on Wayland.

Fixes #10476.

# User Testing

This should be tested on Ubuntu 23.10 or 24.04 with Wayland with at least one Keyman keyboard installed.

**TEST_SWITCH_KBD**:
- Open a terminal window.
- Press <kbd>Super</kbd><kbd>Space</kbd> to switch to the Keyman keyboard. 
- Press <kbd>Super</kbd><kbd>Space</kbd> again to switch back to the first keyboard
- Verify that the keyboard switch actually happens.

### DIFF
--- a/linux/ibus-keyman/src/engine.c
+++ b/linux/ibus-keyman/src/engine.c
@@ -779,7 +779,11 @@ ibus_keyman_engine_process_key_event(
     return TRUE;
   }
 
-  // REVIEW: why don't we handle these keys?
+  // Depending on the OS/base keyboard layout we might get the same modifier
+  // key whether the user pressed right or left Alt or Ctrl key. We work
+  // around that by setting the `l/r*_pressed` flag when the actual modifier
+  // key is pressed so that when we get the actual key with the modifier
+  // flag set we know which Ctrl/Alt key is still pressed.
   switch (keycode) {
   case KEYMAN_LCTRL:
     keyman->lctrl_pressed = isKeyDown;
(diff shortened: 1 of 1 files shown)


ermshiperete:
@mcdurdin I'm not sure if it would be better to ignore the Meta modifier in Keyman Core? Then the engines don't need to know what Core supports.

mcdurdin:
> I'm not sure if it would be better to ignore the Meta modifier in Keyman Core?

We don't currently have a Meta modifier in the modifier bitmap. So that would take API changes (and probably a big mess because the bitmap is already full). And Meta is only available in Linux, not Windows or macOS. So I think it is appropriate to be done in the Linux Engine.

linux/ibus-keyman/src/engine.c
```diff
@@ -804,16 +808,35 @@ ibus_keyman_engine_process_key_event(
...
+    g_message("%s: Not handling keys with IBUS_MOD4_MASK (Super) modifier", __FUNCTION__);
+    return FALSE;
+  }
+
+  // REVIEW: Do we need to do something about IBUS_MOD2_MASK (NumLock)? I guess
+  // if a keyboard would have rules for the numeric keypad we might need
+  // it. At the moment however we let Core process the keypress and since
+  // it doesn't have rules for the numeric keypad keys we eventually
+  // forward the key to ibus (in process_emit_keystroke_action) and let
+  // Gnome deal with it.
```
mcdurdin, line 828: We do have a flag for numlock - but it is currently commented out. In theory Keyman .kmn keyboards support rules with NUMLOCK modifiers, but this is not recommended and has never really been tested.

mcdurdin (approved):

bharanidharanj:
# Test Results

* **TEST_SWITCH_KBD (PASSED):** Tested with the attached PR build (ver 17.0.262-alpha-local (package version 17.0.262-1~PR-10668-2282.1+mantic1) on Ubuntu 23.10 Mantic Minotaur Linux OS (VM) and here is my observation: 1. Installed some Keyman Keyboards. 2. Verified that the default system keyboard (en) appears on the task bar. 3. Pressed <Windows key + Spacebar> to switch to Keyman (eg., khmer_angkor) keyboard and verified the keyboard was changed to Khmer Angkor. 4. Again, pressed <Windows key + Spacebar> to switch back to the previous keyboard. (ie., en) and verified that the keyboard was changed to the system keyboard (English). Seems to be working fine.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.263-alpha](https://keyman.com/downloads/releases/alpha/17.0.263)

PR #10674, opened 2024-02-08
chore(linux): Ignore missing .symbols file in stable branch

ermshiperete:
This fixes a bug in the API verification if the stable branch doesn't have the expected .symbols file.

@keymanapp-test-bot skip

### DIFF
--- a/linux/scripts/deb-packaging.sh
+++ b/linux/scripts/deb-packaging.sh
@@ -150,6 +150,14 @@ get_api_version_in_symbols_file() {
   echo "${firstline}"
 }
 
+# Check if the API version got updated
+# Returns:
+#   0 - if the API version got updated
+#   1 - the .symbols file got changed but the API version didn't get updated
+#   2 - if we're in the alpha tier and the API version got updated since
+#       the last stable version
+# NOTE: it is up to the caller to check if this is a major version
+# change that requires an API version update.
 is_api_version_updated() {
   local NEW_VERSION OLD_VERSION TIER
   git checkout "${GIT_BASE}" -- "debian/${PKG_NAME}.symbols"
@@ -168,7 +176,12 @@ is_api_version_updated() {
     alpha)
       local STABLE_VERSION
       STABLE_VERSION=$((${VERSION%%.*} - 1))
-      git checkout "stable-${STABLE_VERSION}.0" -- "debian/${PKG_NAME}.symbols"
+      if ! git cat-file -e "origin/stable-${STABLE_VERSION}.0" "debian/${PKG_NAME}.symbols" 2>/dev/null; then
+        # .symbols file doesn't exist in stable branch; that's ok
+        echo "2"
+        return
+      fi
+      git checkout "origin/stable-${STABLE_VERSION}.0" -- "debian/${PKG_NAME}.symbols"
       STABLE_API_VERSION=$(get_api_version_in_symbols_file)
       git checkout "${GIT_SHA}" -- "debian/${PKG_NAME}.symbols"
       if (( NEW_VERSION > STABLE_API_VERSION )); then


linux/scripts/deb-packaging.sh
```diff
@@ -168,7 +176,12 @@ is_api_version_updated() {
     alpha)
       local STABLE_VERSION
       STABLE_VERSION=$((${VERSION%%.*} - 1))
-      git checkout "stable-${STABLE_VERSION}.0" -- "debian/${PKG_NAME}.symbols"
+      if ! git cat-file -e "origin/stable-${STABLE_VERSION}.0" "debian/${PKG_NAME}.symbols" 2>/dev/null; then
```
mcdurdin, line 179: This command looks malformed?

```suggestion
      if ! git cat-file -e "origin/stable-${STABLE_VERSION}.0:debian/${PKG_NAME}.symbols" 2>/dev/null; then
```

mcdurdin (changes requested):
I am a little uncomfortable with the check code messing up the source tree temporarily -- seems to me we could use `git cat-file` to write to a temp file instead of `git checkout` and avoid that? And probably do just a single call then as well to test for existence and write it to the temp file, e.g.

```bash
if ! git cat-file blob "origin/stable-${STABLE_VERSION}.0:debian/${PKG_NAME}.symbols" >"$tmpfilename" 2>/dev/null; then
```

linux/scripts/deb-packaging.sh
ermshiperete: Good catch. Thanks!

ermshiperete:
Refactored the change to not use `git checkout`

linux/scripts/deb-packaging.sh
```diff
@@ -144,19 +144,39 @@ check_updated_version_number() {
...
-  local NEW_VERSION OLD_VERSION TIER
-  git checkout "${GIT_BASE}" -- "debian/${PKG_NAME}.symbols"
-  OLD_VERSION=$(get_api_version_in_symbols_file)
-  git checkout "${GIT_SHA}" -- "debian/${PKG_NAME}.symbols"
-  NEW_VERSION=$(get_api_version_in_symbols_file)
-  if (( NEW_VERSION > OLD_VERSION )); then
+  local OLD_API_VERSION NEW_API_VERSION  TIER TMPFILE
+  TMPFILE=$(mktemp)
+  git cat-file blob "${GIT_BASE}:linux/debian/${PKG_NAME}.symbols" > "${TMPFILE}"
+  OLD_API_VERSION=$(get_api_version_in_symbols_file "${TMPFILE}")
```
mcdurdin, line 175: This is looking better but there are a lot of repeated file and git references.

I think it would be good if we could make another wrapper function that does the full cat-file and parse here, and then that function would also manage the lifecycle of `$TMPFILE`.

```bash
  local SYMBOL_FILE="linux/debian/${PKG_NAME}.symbols"
  local GIT_STABLE_BRANCH="origin/stable-${STABLE_VERSION}.0"
  local OLD_API_VERSION=$(get_api_version_from_git_blob_symbols_file "${GIT_BASE}:${SYMBOL_FILE}")
  if (( OLD_API_VERSION == -1 )); then
    # error occurred, ...
  fi

  ...
  
  local STABLE_API_VERSION=$(get_api_version_from_git_blob_symbols_file "${GIT_STABLE_BRANCH}:${SYMBOL_FILE}"
  if (( STABLE_API_VERSION == -1 )); then
    STABLE_API_VERSION=$(get_api_version_from_git_blob_core "${GIT_STABLE_BRANCH}:core/CORE_API_VERSION.md")
    if (( STABLE_API_VERSION == -1 )); then
      ...
    fi
  fi
```

mcdurdin (commented):
I like the move to cat-file. For maintenance reasons, can we do one more refactor? (I haven't yet reviewed the test script)

mcdurdin (approved):
LGTM. Feels like a lot of work went into this! It's a shame it's so tricky to keep the .symbols file in sync.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.272-beta](https://keyman.com/downloads/releases/beta/17.0.272)

PR #10681, opened 2024-02-09
fix(linux): Use temp dir if we can't create cache dir

### LINKED ISSUE #10033: bug(linux): FileNotFoundError displayed in terminal if cache directory doesn't exist and can't be created
If the cache directory (`~/.cache` or `$XDG_CACHE_HOME`) doesn't exist, various libraries we call will create it and so by the time we try to create `~/.cache/keyman` it will exist. However, if `~/.cache` can't be created, e.g. because of missing permissions, we get a FileNotFoundError. This won't affect the functionality of Keyman (except that it has to download some data again next time).

We should try to create `~/.cache` if it doesn't exist, and if that fails use a temporary directory instead.

Similarly if `~/.cache` exists, but we don't have permission to create `~/.cache/keyman`.

ermshiperete:
If we can't create the cache dir, e.g. if `~/.cache` doesn't exist or if we don't have enough permissions, we now return a temporary directory.

Fixes #10033.

@keymanapp-test-bot skip

### DIFF
--- a/linux/keyman-config/keyman_config/get_kmp.py
+++ b/linux/keyman-config/keyman_config/get_kmp.py
@@ -3,6 +3,7 @@
 import datetime
 import logging
 import os
+import tempfile
 import time
 
 import requests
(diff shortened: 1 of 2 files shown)


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.267-alpha](https://keyman.com/downloads/releases/alpha/17.0.267)

Issue #10471, opened 2024-01-23
bug(core): LDML keyboardprocessor needs to delete U+000D U+000A as a unit on Windows


mcdurdin:
I am anticipating that this will become a problem soon (encountered while working on the keyboard debugger).

I think this belongs under LDML keyboardprocessor. On Windows at least, `\r\n` should always be deleted as a block for K_BKSP.

We may need to special-case this, testing for the presence of this pair at the end of the context and requesting 2 back-deletions rather than 1.

(Note: I think this will become visible with the move from action queue to action struct in #10441, and become more obvious after #10415 is implemented.)

mcdurdin:
@rc-swag assigning to you but happy to discuss on who owns.

srl295:
Q: how does kmx handle this? 

Is this going to be an issue for authors? that is, will keyboards see `A\nB` on some platforms and `A\r\nB` on others?  dare I say, almost a normalization issue

rc-swag:
I don't follow when will there be a need to backspace over a `\r\n`?  Internally in the `core` the context will be invalidated on a "carriage return". On the platform side for Context-aware/Compliant apps the `set_if_needed` will only have the context up to the start of line.

mcdurdin:
Note: see edit in OP for change in perspective post discussion with Ross. Keeping the same issue for now.

mcdurdin:
> Q: how does kmx handle this?
> 
> Is this going to be an issue for authors? that is, will keyboards see `A\nB` on some platforms and `A\r\nB` on others? dare I say, almost a normalization issue

Yes we had the same discussion. Resolution is to normalize by Engine before passing into Core, given it applies mainly to Windows. Then keyboard authors will only ever see `\n`.

mcdurdin:
> Internally in the `core` the context will be invalidated on a "carriage return". On the platform side for Context-aware/Compliant apps the `set_if_needed` will only have the context up to the start of line.

The context may include a `\n` -- some apps start a new context on a new para, others treat the entire text buffer as a single unit.

mcdurdin:
Note: we could add a hint to the keyboard compiler in the future to note that 0x0D will never be seen in context (noting that more work needs to be done on KMW for this to be the case).

PR #10697, opened 2024-02-13
feat(windows): handle CRLF specific to windows

rc-swag:
Fixes: #10471

Added a function to replace `\r\n` with `\n` that is U+000D U+000A with U+000A. And another to do the reverse, that is insert `\r` in front of any `\n` characters to support the CRLF for Windows. Supporting unit tests were also added. 

This PR took a change after my first commit. Initially, I had nice symmetrical pre_process_context post_process_context pair. Then I realised I needed to deal with the UTF-32 encoding. It was more efficient just to process that and insert the CR at the same time, otherwise we a just adding another parse and making an extra buffer we don't need. I have written comments in the code that when if we remove the action queue then we can then make it more 

# User Testing

* **TEST_HIEROGLYPHIC_WINDOWS**
  1. Install Hieroglyphic Keyboard from keyman.com into Keyman for Windows.
  2. Open Wordpad.
  3. Select the Hieroglyphic keyboard
  4. In wordpad, type: <kbd>w</kbd><kbd>d</kbd><kbd>i</kbd><kbd>spacebar</kbd><kbd>spacebar</kbd>
  5. Expected Output : w𓂞
  6. Type:<kbd>backspace</kbd>
  7. Expected Output: w

* **TEST_NEWLINE_LF_ONLY**
Install the PR build of Keyman
  1. Install the attached newline keyboard [newline.zip](https://github.com/keymanapp/keyman/files/14504560/newline.zip)
  2. Open Wordpad.
  3. In wordpad, type: <kbd>a</kbd><kbd>Enter</kbd><kbd>c</kbd>
  4. Expected Output : `Row lf`

* **TEST_NEWLINE_CR_ONLY**
Install the PR build of Keyman
  1. Install the attached newline keyboard [newline.zip](https://github.com/keymanapp/keyman/files/14504560/newline.zip) if not installed already 
  2. Open Firefox Browser (It has to be Firefox as it supports TSF)
  3. In the bowser type url bar type `https://www.editpad.org/`
  4. In new edit document type: <kbd>a</kbd><kbd>Enter</kbd><kbd>c</kbd>
  6. Expected Output : `Row lf`


bharanidharanj:
# Test Results

* **TEST_HIEROGLYPHIC_WINDOWS (PASSED):** Tested with the attached PR build (17.0.266-alpha-test-10697) on Windows 11 OS and here is my observation: 1. Opened Wordpad app. 2. Typed 'wdi spacebar spacebar' and verified that it showed the expected output.3. Pressed backspace key and showed the expected output. Seems to be working fine. 

..Result 1

   


..Result 2

mcdurdin:
The user tests do not appear to test anything relating to CRLF. Is that something we can do?

windows/src/engine/keyman32/appcontext.h
```diff
@@ -102,6 +102,30 @@ class AppContext {
  */
 BOOL ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD len);
 
+/**
+*  frERROR:    If an error occurs, such as null pointers or buffer to small.
```
mcdurdin, line 106:
```suggestion
*  frERROR:    If an error occurs, such as null pointers or buffer too small.
```

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,36 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
   delete[] buf;
   return TRUE;
 }
+
+FormatContextResult
+format_context_for_core(LPCWSTR windows_context, LPWSTR core_context, uint32_t output_size) {
```
mcdurdin, line 169: Can we rename `output_size` to `core_context_size_in_chars` or something like that?

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,36 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+    return frERROR;
+  }
+  // Return early if windowsContext does not contain '\r\n'
+  if (wcsstr(windows_context, L"\r\n") == nullptr) {
+    return frNO_CHANGE;
+  }
+
+  auto windows_context_length = wcsnlen_s(windows_context, MAXCONTEXT);
+
+  if (output_size < windows_context_length) {
```
mcdurdin, line 180:
```suggestion
  if (output_size <= windows_context_length) {
```

Need space for terminating nul

windows/src/engine/keyman32/kmprocessactions.cpp
```diff
@@ -7,21 +7,33 @@
...
-  while (*outputString) {
-    if (Uni_IsSMP(*outputString)) {
-      app->QueueAction(QIT_CHAR, (Uni_UTF32ToSurrogate1(*outputString)));
-      app->QueueAction(QIT_CHAR, (Uni_UTF32ToSurrogate2(*outputString)));
+process_output_string(AITIP* app, const km_core_usv* core_output) {
+  while (*core_output) {
+    if (*core_output == L'\n') {
+      // Insert '\r' for Windows platform applications
+      app->QueueAction(QIT_CHAR, L'\r');
+    }
```
mcdurdin, line 29: It makes no practical difference, I know, but for clarity can we put this test inside the `else` clause below?

mcdurdin (approved):
This LGTM. Would be good to have a user test for the CRLF scenarios. Minor nits otherwise.

rc-swag:
> The user tests do not appear to test anything relating to CRLF. Is that something we can do?
Correct the current test was good regression to make sure nothing was broken. I was working on using a ldml keyboard with rules for the CRLF test. However, it seemed to fail but the debugger and logging was giving some "interesting" results, and proving a bit tricky to test, as it appears something else can strip the CR. Still investigating.

rc-swag:
@mcdurdin  I have just been doing some logging around the contexts and finding some surprising results.
1. AITIP::ReadContext is returning the string with just `\r`  and now `\n' . I have followed the call stack to see if this what it gets back from the TSF API call.
2. Not using AITIP::ReadContext. Then reading the cores version of the app context it is not keeping track of new lines at all.

Using Keyman Developer I as able to create a keyboard with `\r\n` rules and when testing the the keyboard in developer in matched those rules. However developer is clearly not the same as Windows application.

rc-swag:
And with that I guess this has to move back to draft

mcdurdin:
Which compliant apps are you testing with? I am guessing some apps will treat each paragraph as a separate context, but others may inline the `\r\n`.

New thought: ideally, we track what the app gives us in the input context, and give the same pattern back. So we have a function (serious pseudocode ahead warning):

```
const LBType = { none, '\n', '\r\n', '\r' };
function normalizeLineBreaks(&str): LBType {
  // find first line break
  m = /(\r\n|\r|\n)/.match(str);
  if(!m) {
   return none;
  }
  str = str.replaceAll(/(\r\n|\r|\n)/g, '\n');
  return m.firstMatch;
}

function restoreLineBreaks(str, tp:LBType, defaulttp:LBType): string {
  if(tp == none) tp = defaulttp; // use platform default for line breaks if context had no line breaks
  if(tp == none) return str; // if platform has no default and context had no line breaks, do nothing
  return str.replaceAll(/\n/g, tp); // replace line breaks
}
```

And, this could even be buried in Core, ideally, so that the Engine doesn't need to do this legwork. We could pass `defaultLineBreakType` in as a parameter to Core...

Seems like too much for 17.0 though!

rc-swag:
> And, this could even be buried in Core, ideally, so that the Engine doesn't need to do this legwork. We could pass `defaultLineBreakType` in as a parameter to Core...
> 
> Seems like too much for 17.0 though!

This is putting a more of platform knowledge into the core, but it is ok as it we keep it away from the core logic of the processors. I assume when we use the `core context` to load the updated `app context` (stripping out the markers) we also run `restoreLineBreaks`

 This would be more robust to handling any nuances from different platforms and apps. for example would also handle any MacOS apps that also used just `\r` I would still like to know why the Windows ReadContext has only `\r'. 

The platforms would need to make sure to set the platform default line break when loading the ~~keyboard~~ state. This would be straightforward with the core environment array already part of `km_core_state_create`.

mcdurdin:
Let's see if things are breaking with CRLF in visible ways in Windows. If not, we can push this whole project to 18.0 and do it properly there.

rc-swag:
I have implemented the normalize/ restore algorithm but not in core at this late stage as it would effect all platforms.
I have created a issue #10954  to move it to core in version 18.0

bharanidharanj:
# Test Results

* **TEST_NEWLINE_LF_ONLY (FAILED):** Tested with the attached PR build (Keyman ) on Windows 10 OS and here is my observation: 1. Installed NewLine keyboard. 2. Opened WordPad document. 3. Switched to NewLine keyboard. 4. Typed as per the instructions.  Verified it is showing the expected result. 5. Opened Microsoft office 365 Word online document. 6. Switched to NewLine Keyboard. 7. Typed as per instructions. Noticed that it is showing only the typed letters. Seems to be an issue. 

..in WordPad document

   


..in Microsoft Office 365 word online document

rc-swag:
@bharanidharanj  The testing of Word needs to be for an installed desktop version. It can be 365 but still needs to be an installed version. It failed the test because you used the online version in a web browser that doesn't support the Text Services Framework(TSF). Which is needed for this test. Do you have a test machine with Word installed? If not let me know and I will modify the test to suit.

@keymanapp-bot re-test TEST_NEWLINE_LF_ONLY

bharanidharanj:
> @bharanidharanj The testing of Word needs to be for an installed desktop version. It can be 365 but still needs to be an installed version. It failed the test because you used the online version in a web browser that doesn't support the Text Services Framework(TSF). Which is needed for this test. Do you have a test machine with Word installed? If not let me know and I will modify the test to suit.
> 
> @keymanapp-test-bot retest TEST_NEWLINE_LF_ONLY

@rc-swag I think it would be better to modify the test to suite. Thanks.

rc-swag:
> @rc-swag I think it would be better to modify the test to suit. Thanks.

@bharanidharanj  I have updated the `TEST_NEWLINE_LF_ONLY` test to only have Wordpad and added another for `TEST_NEWLINE_CR_ONLY` with Firefox so please do this tests

bharanidharanj:
> > @rc-swag I think it would be better to modify the test to suit. Thanks.
> 
> @bharanidharanj I have updated the `TEST_NEWLINE_LF_ONLY` test to only have Wordpad and added another for `TEST_NEWLINE_CR_ONLY` with Firefox so please do this tests

@rc-swag Okay, Thanks. I will do it.

bharanidharanj:
# Test Results

* **TEST_NEWLINE_LF_ONLY (PASSED):** Retested as per Ross's suggestion in a WordPad document and here is my observation: 1. Followed the steps as mentioned in the test description. 2. Verified that 'Row If' text appears on the document after pressing the Enter key. 

   



* **TEST_NEWLINE_CR_ONLY (PASSED):** Tested with the attached PR build (keyman version 17.0.285-beta-test-10697) and here is my observation: 1. Opened the 'editpad' website. 2. Created a new document. 3. Followed step (iv) and verified that the text 'ROW If' appeared in the document.  Seems to be working as expected.

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+    line_break = default_lb;
+  }
+
+  // return early if doesn't contain any '\n';
+  if (wcsstr(win_out_str, L"\n") == nullptr) {
+    return rsNO_LB;
+  }
+
+  size_t buf_length = wcslen(win_out_str) * 2;
+  LPWSTR buf_string = new WCHAR[buf_length];
```
mcdurdin, line 222: This doesn't allow space for final terminating nul if every char is doubled

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+
+  LPCWSTR in_ptr = win_out_str;
+  LPWSTR buf_ptr   = buf_string;
+
+  while (*in_ptr != L'\0') {
+    if (*in_ptr == '\n') {
```
mcdurdin, line 228: Not consistently using `L` prefix, here and elsewhere in this PR

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+        case lbLF:
+          *buf_ptr++ = *in_ptr++;
+          break;
+        case lbCRLF:
+          *buf_ptr++ = '\r';
+          *buf_ptr++ = *in_ptr++;
+          break;
+        case lbCR:
+          *buf_ptr++ = '\r';
+          *in_ptr++;
```
mcdurdin, line 239:
```suggestion
          in_ptr++;
```

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+  }
+  *buf_ptr = '\0';  // Null terminate the modified string
+
+  // may now need to truncate the string preserving the end closest the caret.
+  auto final_length = wcsnlen_s(buf_string, buf_length);
+  if (final_length < win_out_size_in_chars) {
+    wcscpy_s(win_out_str, win_out_size_in_chars, buf_string);
+  } else {
+    auto diff = final_length + 1 - win_out_size_in_chars;  // +1 for null termination
+    wcscpy_s(win_out_str, win_out_size_in_chars, buf_string + diff);
```
mcdurdin, line 254: This could break a surrogate pair

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+          *in_ptr++;
+          break;
+      }
+    } else {
+      *buf_ptr++ = *in_ptr++;
+    }
+  }
+  *buf_ptr = '\0';  // Null terminate the modified string
+
+  // may now need to truncate the string preserving the end closest the caret.
```
mcdurdin, line 248: Instead of doing this, couldn't we just return the longer buffer?

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+    auto diff = final_length + 1 - win_out_size_in_chars;  // +1 for null termination
+    wcscpy_s(win_out_str, win_out_size_in_chars, buf_string + diff);
+  }
+  delete[] buf_string;
+  return rsSUCCESS;
+}
+
+BOOL
+context_char32_char16(const km_core_usv *core_output, LPWSTR win_out_str, uint32_t win_output_size_in_char) {
+  if (core_output == nullptr || win_out_str == nullptr || win_output_size_in_char <= 0) {
```
mcdurdin, line 262:
```suggestion
  if (core_output == nullptr || win_out_str == nullptr || win_output_size_in_char == 0) {
```

uint32_t cannot be <0

windows/src/engine/keyman32/appcontext.cpp
```diff
@@ -164,3 +164,125 @@ ContextItemToAppContext(km_core_context_item *contextItems, PWSTR outBuf, DWORD
...
+  }
+  buf[idx] = 0;  // Null terminate character array
+
+  auto final_length = wcsnlen_s(buf, buf_length);
+  if (final_length < win_output_size_in_char) {
+    wcscpy_s(win_out_str, win_output_size_in_char, buf);
+  } else {
+    auto diff = final_length + 1 - win_output_size_in_char;  // +1 for null termination
+    wcscpy_s(win_out_str, win_output_size_in_char, buf + diff);
+  }
```
mcdurdin, line 285: Same risk of splitting a surrogate pair

windows/src/engine/keyman32/tests/keyman-engine-tests/appinttests.cpp
```diff
@@ -78,3 +78,58 @@ TEST(AppContext, AppContext_Delete) {
   testContext.Delete();
 
 }
+
+// Test normalize and restore line break functions
```
mcdurdin, line 82: I think we need some tests for surrogate pair edge cases, especially boundary conditions

mcdurdin (changes requested):
I am somewhat uncomfortable about this PR because it makes significant string handling changes very late in the beta cycle. I think we may need some more exercising of the boundary conditions, and I think we should be able to avoid truncation of buffers with a small rework per my notes?

rc-swag:
> I am somewhat uncomfortable about this PR because it makes significant string handling changes very late in the beta cycle.
Based on this I am going to close this and update the issue #10471 for 18.0 but for the algorithm to be in the core (and to check developer debugger). That is where I ultimately wanted it to live and it would then cover linux and macos also. This was just to get a reference solution of the line for the LDML and version 17.  With the buffer in the core truncation I would have done it differently however when I looked at it was just kicking the problem down the road because it was had truncation in there anyway. 
Finally, I am disappointed I missed the surrogates as that is one of the first edge cases I discovered in the `Context::Buf` method. when I started.

rc-swag:
This will be moved to version 18.0 as it is too late in the beta cycle. The logic will also be contained in the core.

mcdurdin:
Agree that this can wait until 18.0; if necessary we can back-port to 17.0 stable.

Issue #10389, opened 2024-01-15
feat(developer): forbid bad regex 🙀 


srl295:
- [x] Forbid `\uXXXX` form escapes: #10319 
- [x] error if JS regex can't compile a `from=…` clause

PR #10701, opened 2024-02-13
feat(developer): errs on \uXXXX escapes 🙀

srl295:
- note, some older test cases had to be fixed also.

Fixes: #10389

@keymanapp-test-bot skip


developer/src/kmc-ldml/src/compiler/tran.ts
```diff
@@ -216,17 +225,35 @@ export abstract class TransformCompiler<T extends TransformCompilerType, TranBas
     return defaults;
   }
 
-  /** true if ok */
-  private checkRanges(cookedFrom: string): string {
+  /** returns ===null on error */
```
mcdurdin, line 228:
```suggestion
  /** @returns null if invalid escapes are found in `cookedFrom` string */
```

But what is an invalid escape? I can parse the function but still not clear exactly what it is we are checking

mcdurdin (approved):

developer/src/kmc-ldml/src/compiler/tran.ts
srl295: it's invalid if it is in `\uXXXX` form without curly braces. I added a function comment.

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.267-alpha](https://keyman.com/downloads/releases/alpha/17.0.267)

Issue #10734, opened 2024-02-15
bug(core): crasher in ldml_transforms 🙀 


srl295:
crasher pressing '6' on:

https://github.com/keyman-keyboards/ldml_test/commit/8334816d0f0599b7bc6024a2bf8db5a81beb8798

srl295:
exception seemed to be https://github.com/keymanapp/keyman/blob/4ec888fd72609bda8f6b8c8ff3e48019e0fd2f26/core/src/ldml/ldml_transforms.cpp#L592-L593 which should be a memory problem, but could be (a) corruption or (b) a bad intermediate string

srl295:
repro'ed

PR #10735, opened 2024-02-15
fix(core): more tests and fix for ldml_transforms.cpp

srl295:
For #10734

- fix for bad generation of from/to maps
- 2 fixes for crasher in ldml_transforms.cpp
- There's still an issue in the transform matcher - test case reflects a TODO

@keymanapp-test-bot skip


core/src/ldml/ldml_transforms.cpp
```diff
@@ -588,7 +588,7 @@ transform_entry::apply(const std::u32string &input, std::u32string &output) cons
     char32_t *s                = new char32_t[group1Len + 1];
     assert(s != nullptr); // TODO-LDML: OOM
     // convert
-    substr.toUTF32((UChar32 *)s, group1Len + 1, status);
+    group1.toUTF32((UChar32 *)s, group1Len + 1, status);
```
srl295, line 591: 😬

srl295: i think this assumed the from= only had the matched group and no extra text.

developer/src/kmc-ldml/src/compiler/tran.ts
```diff
@@ -126,7 +126,6 @@ export abstract class TransformCompiler<T extends TransformCompilerType, TranBas
     let cookedFrom = transform.from;
 
     cookedFrom = sections.vars.substituteStrings(cookedFrom, sections);
-    // TODO: handle 'map' case
```
srl295, line 129: out of date TODO

core/src/ldml/ldml_transforms.cpp
```diff
@@ -749,12 +749,9 @@ transforms::apply(const std::u32string &input, std::u32string &output) {
   for (auto group = transform_groups.begin(); group < transform_groups.end(); group++) {
     // for each transform group
     // break out once there's a match
-    // TODO-LDML: reorders
-    // Assume it's a non reorder group
     /** Length of match within this group*/
 
     // find the first match in this group (if present)
-    // TODO-LDML: check if reorder
```
srl295, line 757: out of date todos'!

common/web/types/src/ldml-keyboard/pattern-parser.ts
```diff
@@ -400,7 +400,7 @@ export class VariableParser {
   /**
    * Pattern for matching a capture set reference `($[set])`
    */
-  public static readonly CAPTURE_SET_REFERENCE = /\(\$\[([0-9A-Za-z_]{1,32})\]\)/g;
+  public static readonly CAPTURE_SET_REFERENCE = /\(\$\[([0-9A-Za-z_]{1,32})\]\)/;
```
mcdurdin, line 403: Why the removal of `g`?

mcdurdin (approved):
LGTM, as far as I can tell, but I don't understand the change to `CAPTURE_SET_REFERENCE`

srl295:
It makes the regex mutable! So it behaves differently on the second use.

common/web/types/src/ldml-keyboard/pattern-parser.ts
srl295: So that using it is idempotent

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.270-beta](https://keyman.com/downloads/releases/beta/17.0.270)

mcdurdin:
> It makes the regex mutable! So it behaves differently on the second use.

Whoa, that's ugly! TIL.

PR #10772, opened 2024-02-19
chore(linux): Fix `debian.sh` by removing quotes

ermshiperete:
We want `EXTRA_ARGS` to be split so that we end up with different arguments.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.272-beta](https://keyman.com/downloads/releases/beta/17.0.272)

PR #10782, opened 2024-02-20
fix(android/engine): Remove logs for uninitialized default keyboard

### LINKED ISSUE #7712: bug(android): Failed getCurrentKeyboardIndex check for keyboard: en_sil_euro_latin
Sentry Issue: [KEYMAN-ANDROID-1E1](https://sentry.io/organizations/keyman/issues/3408522655/?referrer=github_integration)

```
Failed getCurrentKeyboardIndex check for keyboard: en_sil_euro_latin
```

darcywong00:
Fixes #7712

The 4 sets of Sentry reports stem from the following scenario (obscure)

1. User installs Keyman (but doesn't launch the app). Hence, sil_euro_latin never gets "added" (done in MainActivity)
2. From the Android settings --> Languages menus, the user enables Keyman as a system keyboard
3. User then attempts to use Keyman as a keyboard
4. SystemKeyboard initializes (--> KMManger.initialize() which extracts sil_euro_latin.kmp so it can be "used). Since the keyboard sees no keyboards installed (index is -1), we get a bunch of Sentry reports from attempting to fallback.

This PR quiets the Sentry noise so we only log if the key is **not** for the default sil_euro_latin keyboard.

## User Testing
**Setup** - This involves modifying Android Studio configuration so the IDE will install the PR build of Keyman for Android, without launching the app.
1. From Android Studio --> Run --> Edit Configurations
2. In the Run/Debug Configurations dialog, change "Launch Options --> Launch" dropdown from "Default Activity" to "Nothing". 
edit configuration

3. On an Android device/emulator, install the PR build of Keyman for Android

* **TEST_SYSTEM_KEYBOARD** - Verifies system keyboard runs
1. On the Android device, go to the Android Settings --> Languages menu and enable Keyman as a system keyboard
2. Launch Chrome and click on the textbar
3. Select Keyman as the system keyboard
4. Once Keyman loads, verify the OSK appears
5. Verify you can type with the Keyman keyboard
6. Notify @darcywong00 to check Sentry to verify no new Sentry logs appear from the test build.
7. Revert the Android Studio settings so future PR testing with launch the app.

### DIFF
--- a/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
@@ -1633,10 +1633,10 @@ public final class KMManager {
   }
 
   public static boolean isDefaultKey(String key) {
-    if (key != null && key.equals(KMString.format("%s_%s", KMDefault_LanguageID, KMDefault_KeyboardID))) {
-      return true;
-    }
-    return false;
+    return (
+      key != null && 
+      key.equals(KMString.format("%s_%s", KMDefault_LanguageID, KMDefault_KeyboardID)
+    );
   }
 
   /**
--- a/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
@@ -1635,8 +1635,7 @@ public final class KMManager {
   public static boolean isDefaultKey(String key) {
     return (
       key != null && 
-      key.equals(KMString.format("%s_%s", KMDefault_LanguageID, KMDefault_KeyboardID)
-    );
+      key.equals(KMString.format("%s_%s", KMDefault_LanguageID, KMDefault_KeyboardID)));
   }
 
   /**
--- a/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
+++ b/android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
@@ -1632,6 +1632,13 @@ public final class KMManager {
     return KeyboardPickerActivity.removeKeyboard(context, position);
   }
 
+  public static boolean isDefaultKey(String key) {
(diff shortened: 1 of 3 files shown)


bharanidharanj:
# Test Results

* **TEST_SYSTEM_KEYBOARD (OPEN):** Tested with the attached PR build (17.0.272-beta-test-10782) in the Android 12 / API 31 emulator and here is my observation: 1. In the Android Studio, I changed the Launch Options to "Nothing". 2. Installed the attached PR build. Set Keyman as the default system keyboard.  3. Opened Chrome browser then clicked the text bar. 4. Verified that the OSK appeared on the screen. 5. Verified that the typed word appeared on the search bar.

darcywong00:
> Notify to check Sentry to verify no new Sentry logs appear from the test build.

@bharanidharanj - I've checked the Sentry logs for beta test builds and there's no new reports about these issues.

bharanidharanj:
> > Notify to check Sentry to verify no new Sentry logs appear from the test build.
> 
> @bharanidharanj - I've checked the Sentry logs for beta test builds and there's no new reports about these issues.

@darcywong00 As per your instructions, I have changed the Launch options to "Nothing" in Run/Debug Configurations dialog.

android/KMEA/app/src/main/java/com/keyman/engine/KMManager.java
```diff
@@ -1632,6 +1632,13 @@ public static boolean removeKeyboard(Context context, int position) {
     return KeyboardPickerActivity.removeKeyboard(context, position);
   }
 
+  public static boolean isDefaultKey(String key) {
+    if (key != null && key.equals(KMString.format("%s_%s", KMDefault_LanguageID, KMDefault_KeyboardID))) {
+      return true;
+    }
+    return false;
```
mcdurdin, line 1639:
```suggestion
    return (
      key != null && 
      key.equals(KMString.format("%s_%s", KMDefault_LanguageID, KMDefault_KeyboardID)
    );
```

mcdurdin (approved):
LGTM. Good spelunking!

keyman-server:

PR #10784, opened 2024-02-20
chore(linux): Fix `upload-to-debian.sh` script for beta releases

ermshiperete:
@keymanapp-test-bot skip

### DIFF
--- a/linux/scripts/upload-to-debian.sh
+++ b/linux/scripts/upload-to-debian.sh
@@ -99,7 +99,7 @@ cd debianpackage/
 builder_heading "Signing source package"
 debsign -k"$DEBKEYID" --re-sign ./*.changes
 builder_heading "Uploading packages to mentors.debian.net"
-$NOOP dput mentors ./*.changes
+${NOOP} dput mentors ./*.changes
 cd ..
 
 builder_heading "Updating changelog"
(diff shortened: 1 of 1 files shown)


linux/scripts/upload-to-debian.sh
```diff
@@ -73,56 +73,73 @@ if ! git diff --quiet; then
...
-builder_heading "Fetching latest changes"
-git fetch -p origin
-stable_branch=$(git branch -r | grep origin/stable- | sort | tail -1)
-stable_branch=${stable_branch##* }
-# Checkout stable branch so that `scripts/debian.sh` picks up correct version
-git checkout "${stable_branch#origin/}"
-git pull origin "${stable_branch#origin/}"
-
-if git branch -r | grep origin/beta; then
+if [[ "$(cat TIER.md)" == "beta" ]]; then
```
mcdurdin, line 76: Does this script run in master?

linux/scripts/upload-to-debian.sh
```diff
@@ -73,56 +73,73 @@ if ! git diff --quiet; then
...
 fi
 
+builder_heading "Fetching latest changes"
+git fetch -p origin
+if ${ISBETA}; then
+  git checkout beta
+  git pull origin beta
+else
+  stable_branch=$(git branch -r | grep origin/stable- | sort | tail -1)
+  stable_branch=${stable_branch##* }
```
mcdurdin, line 89: We should probably fail the script here if we don't match a stable-x.y name. I'm not entirely clear how you are deriving the stable version here

linux/scripts/upload-to-debian.sh
```diff
@@ -73,56 +73,73 @@ if ! git diff --quiet; then
...
+  local BRANCH=$1
+  local BASE=$2
+  local COMMIT_MSG=$3
+  local PR_MSG=$4
+
+  if [[ -n "${PUSH}" ]]; then
+      ${NOOP} git push --force-with-lease origin "${BRANCH}"
+      # shellcheck disable=2312
+      if (( $(gh pr list --draft --search "${COMMIT_MSG}" --base "${BASE}" --json title --jq '.[].title' | wc -l) > 0)); then
+        builder_echo "PR already exists"
```
mcdurdin, line 122: Should give the PR number here

linux/scripts/upload-to-debian.sh
```diff
@@ -73,56 +73,73 @@ if ! git diff --quiet; then
...
-if git branch -r | grep origin/beta; then
+if [[ "$(cat TIER.md)" == "beta" ]]; then
     ISBETA=true
 else
     ISBETA=false
 fi
 
+builder_heading "Fetching latest changes"
+git fetch -p origin
+if ${ISBETA}; then
```
mcdurdin, line 84: Here you are switching between beta and stable, it seems, but on line 142 you seem to be switching between beta and alpha?

mcdurdin (changes requested):
I'm not clear on the reasons for the changes here. Also there are a bunch of process things embedded here which need to be documented, so it is clear to future maintainers what is happening and why.

ermshiperete:
> I'm not clear on the reasons for the changes here.

When I originally wrote the script I had some misconceptions about the beta branch. We never tried it for beta before. While it was intended to work with beta releases it only did work for stable releases. These changes make it possible to use for beta as well.

linux/scripts/upload-to-debian.sh
ermshiperete: done

linux/scripts/upload-to-debian.sh
ermshiperete: done

mcdurdin: Note: will stable-100.0 sort after stable-90.0? Not a problem today but one day it may be!

linux/scripts/upload-to-debian.sh
```diff
@@ -73,57 +79,75 @@ if ! git diff --quiet; then
...
 fi
 
+function get_latest_stable_branch_name() {
+  # Get the list of stable branches and take the one with the highest number.
+  # Extract `origin/stable-16.0` from `  origin/stable-16.0`
+  local branches stable_branch
+  branches=$(git branch -r | grep origin/stable-)
+  stable_branch=$(echo "${branches}" | sort | tail -1)
+  if [[ -z ${stable_branch} ]]; then
+    builder_die "Can't find stable-* branch" > /dev/stderr
```
mcdurdin, line 89: Will this fail the script, given it is called in a subshell on line 120?

mcdurdin (approved):

linux/scripts/upload-to-debian.sh
darcywong00: alas, I ran a quick test and I got stable-90 instead of stable-100

mcdurdin: This works, if we really care:

```
sort -n -k 1.8
```

* `-n` = numeric
* `-k 1.8` = sort on field 1, char position 8

linux/scripts/upload-to-debian.sh
ermshiperete: Yes, it will work as intended: If one of the commands in the subshell fails we don't end the subshell (because `set -e` is not set in the subshell), but we check that we found a stable branch and return that. If we didn't find a stable branch (either because there is none, or because one of the commands to retrieve it failed) we exit the subshell with an errorcode and that exits the script (because the script itself runs with `set -e`).

linux/scripts/upload-to-debian.sh
```diff
@@ -40,21 +42,25 @@ NOOP=
...
 
 while (( $# )); do
     case $1 in
         -k) shift
-            if [ ! $# -eq 0 ]; then
+            if [[ ! $# -eq 0 ]]; then
                 DEBKEYID=$1
             else
                 builder_die "Error: The -k argument is missing a value. Exiting."
             fi;;
```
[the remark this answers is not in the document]
ermshiperete: Good point. I'll try to remember that.

keyman-server:

PR #10800, opened 2024-02-21
chore(linux): Address comments from Debian mailing list

ermshiperete:
The upload of the beta version caused some comments on the [Debian Input Method](https://lists.debian.org/debian-input-method/2024/02/msg00062.html) mailing list. This change addresses these:

- rename libkeymancore package to include SONAME: libkeymancore1
- remove unnecessary dependency on quilt

Additionally this change sorts the build dependencies alphabetically and explicitly installs `quilt` to be able to build the
source package.

@keymanapp-test-bot skip


keyman-server:

Issue #10752, opened 2024-02-16
bug(core): meson build not failing when -test.json can't build


srl295:
Windows 11.

```shell
core/build.sh -d test:x64
```

... build logs have:

```
Configuring ldml_test-test.json with command
Running command: C:\Program Files\nodejs\node.EXE --enable-source-maps C:/projects/keyman/keyman/core/../developer/src/kmc build ldml-test-data C:\projects\keyman\keyman\core\tests\unit\ldml\keyboards\ldml_test-test.xml --out-file C:\projects\keyman\keyman\core\build\x64\debug\tests\unit\ldml\keyboards\ldml_test-test.json
--- stdout ---
error KM00007: The source file has an invalid structure: Unable to load XML file

--- stderr ---
```

but the 'console' build log showed:

```
Configuring ldml_test-test.json with command
```

and then the test runner failed because there was no .json test file.

The build should have failed when the 'command' failed.

srl295:
looks like the _test_ path doesn't have an exitProcess(1) call in it.

But, couldn't the parent building code exit if any error messages show up in the callbacks?

mcdurdin:
> But, couldn't the parent building code exit if any error messages show up in the callbacks?

There isn't any parent building code for `buildTestData()` -- it's essentially a top-level call in kmc:

https://github.com/keymanapp/keyman/blob/898310e6df05d07c6ddb068fdc96e5489e46b5dd/developer/src/kmc/src/commands/build.ts#L53-L56

srl295:
fixed

PR #10805, opened 2024-02-22
fix(developer): exit kmc if any failure messages

srl295:
- check for failing messages in the callback, not just failing compiles (no output)

Fixes: #10752

@keymanapp-test-bot skip


developer/src/kmc/src/commands/build.ts
```diff
@@ -102,7 +102,7 @@ async function buildFile(filenames: string[], _options: any, commander: any)  {
   }
 
   for(let filename of filenames) {
-    if(!await build(filename, commanderOptions.outFile, callbacks, options)) {
+    if(!await build(filename, commanderOptions.outFile, callbacks, options) || callbacks.hasFailureMessage()) {
```
mcdurdin, line 105: I'm not sure about this. It should probably be an assertion. We should never return `true` from `build()` if we have an error that fails the build. Feels like if we do, we have a bug in the compiler.

developer/src/kmc/src/commands/buildTestData/index.ts
```diff
@@ -21,10 +21,10 @@ export async function buildTestData(infile: string, _options: any, commander: an
     }
   };
 
-  const callbacks: CompilerCallbacks = new NodeCompilerCallbacks(options);
+  const callbacks: NodeCompilerCallbacks = new NodeCompilerCallbacks(options);
```
mcdurdin, line 24:
```suggestion
  const callbacks = new NodeCompilerCallbacks(options);
```

Type should be inferred I think?

srl295: Yes… Needs the subclass to make this call.

developer/src/kmc/src/commands/build.ts
srl295: I could make it an assertion.

developer/src/kmc/src/commands/buildTestData/index.ts
mcdurdin: But shouldn't need to include the type, because `const x = new foo()` gives `x` the type `foo` doesn't it?

developer/src/kmc/src/commands/build.ts
mcdurdin: debug build assertion sounds appropriate (or log to sentry)

developer/src/kmc/src/commands/buildTestData/index.ts
srl295: fixed

developer/src/kmc/src/commands/build.ts
srl295: added a `kmcAssert()` that throws an internal error if falsy.

developer/src/kmc/src/commands/build.ts
```diff
@@ -102,10 +102,11 @@ async function buildFile(filenames: string[], _options: any, commander: any)  {
...
       // Once a file fails to build, we bail on subsequent builds
       await exitProcess(1);
     }
+    kmcAssert(!callbacks.hasFailureMessage(), () => `errors: ${CompilerError.formatEvent(callbacks.messages)}`);
```
mcdurdin, line 109: This can never fire, because of the test on line 105.

developer/src/kmc/src/commands/build.ts
mcdurdin: Given we have the following in `build()`:

https://github.com/keymanapp/keyman/blob/898310e6df05d07c6ddb068fdc96e5489e46b5dd/developer/src/kmc/src/commands/build.ts#L158-L159

This is already being handled by the build() function, and we should keep that logic there.

```suggestion
    if(!await build(filename, commanderOptions.outFile, callbacks, options)) {
```

developer/src/kmc/src/commands/build.ts
mcdurdin: I don't think we need this because the `build()` function already deals with this.

```suggestion
```

developer/src/kmc/src/commands/buildTestData/index.ts
```diff
@@ -21,12 +21,13 @@ export async function buildTestData(infile: string, _options: any, commander: an
...
   };
 
-  const callbacks: CompilerCallbacks = new NodeCompilerCallbacks(options);
+  const callbacks = new NodeCompilerCallbacks(options);
 
   let testData = await loadTestData(infile, callbacks, compilerOptions);
   if (!testData) {
     await exitProcess(1);
   }
+  kmcAssert(!callbacks.hasFailureMessage(), () => `errors: ${CompilerError.formatEvent(callbacks.messages)}`);
```
mcdurdin, line 30:
```suggestion
  if (!testData || callbacks.hasFailureMessage()) {
    await exitProcess(1);
  }
```

mcdurdin (changes requested):
Hmm, I think I missed something in my last review, sorry. The `build()` function already checks `callbacks.hasFailureMessage()` in build.ts. So we don't need to change any logic in that file. So we should match that logic in buildTestData/index.ts. Apologies.

This probably means we don't need the kmcAssert function at this point, so go ahead and remove that again sorry!

---

Side note: we need to be a little careful here because of the way that warnings can be upgraded to fail the build with -w, causing hasFailureMessage to return true even if the build itself had no warnings or errors:

https://github.com/keymanapp/keyman/blob/21a768027831b390a6e4d6d836279351b82ca16a/common/web/types/src/util/compiler-interfaces.ts#L427-L432

https://github.com/keymanapp/keyman/blob/21a768027831b390a6e4d6d836279351b82ca16a/common/web/types/src/util/compiler-interfaces.ts#L384-L394

srl295:
Oh no problem. I missed that too or maybe I was only looking at test data part. . Like I said, my concern was robustness.

srl295:
seems that generally we might just revert build.ts from this PR completely, but i'll take a look

srl295:
@mcdurdin would you be OK if this one were rebased and squashed? so that it's a 2-liner given it's now very different in scope?

mcdurdin (approved):
Yes rebase squash and force push then normal merge is fine

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.277-beta](https://keyman.com/downloads/releases/beta/17.0.277)

PR #10813, opened 2024-02-22
chore(linux): Fix API verification after adding SONAME

ermshiperete:
Due to Debian policy the `libkeymancore` package will now include the `SONAME` (API version number) in the package name (PR #10800). This change adjusts the API verification checks to be able to cope with the modified .symbols file.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:

PR #10818, opened 2024-02-23
fix(developer): make sure scroll bar appears when needed in global welcome

### LINKED ISSUE #10817: bug(developer): project welcome (no project open) does not get a scrollbar when it needs it

mcdurdin:
Fixes #10817.

This fix is a patch rather than a rework to use grid or flex css, which would be better, but should not be done in beta.

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/tike/xml/project/project.css
+++ b/developer/src/tike/xml/project/project.css
@@ -121,7 +121,16 @@ table.io th {
   font: bold 8pt "Segoe UI"
 }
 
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

keyman-server:
Changes in this pull request will be available for download in [Keyman version 17.0.280-beta](https://keyman.com/downloads/releases/beta/17.0.280)

PR #10821, opened 2024-02-23
fix(developer): repeated options in kmc must now be fully specified

### LINKED ISSUE #10396: chore(developer): handle repeated options in kmc better
The variadic options feature in Commander is a bit icky because it accepts multiple space-separated values with one `-m` flag, and it would be better to require `-m` for every instance.

From #10258.

mcdurdin:
Fixes #10396.

Instead of `--message 20ae 5006`, which is ambiguous, because 5006 could be a filename, you must now specify the parameter option flag each time: `--message 20ae --message 5006` (or `-m 20ae -m 5006`).

Turns out this is documented as an example in Commander, just not very visible.

Should we also support `-m 20ae,5006`? This would be a fairly straightforward tweak.

image

@keymanapp-test-bot skip

### DIFF
--- a/developer/src/kmc/src/messages/infrastructureMessages.ts
+++ b/developer/src/kmc/src/messages/infrastructureMessages.ts
@@ -1,4 +1,4 @@
-import { CompilerErrorNamespace, CompilerErrorSeverity, CompilerMessageSpec as m } from "@keymanapp/common-types";
+import { CompilerError, CompilerErrorNamespace, CompilerErrorSeverity, CompilerMessageSpec as m } from "@keymanapp/common-types";
 
 const Namespace = CompilerErrorNamespace.Infrastructure;
 const SevInfo = CompilerErrorSeverity.Info | Namespace;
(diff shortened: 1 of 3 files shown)


developer/src/kmc/src/messages/infrastructureMessages.ts
```diff
@@ -101,15 +101,15 @@ export class InfrastructureMessages {
   static ERROR_InvalidMessageFormat = SevError | 0x0016;
 
   static Error_MessageNamespaceNotFound = (o:{code: number}) => m(this.ERROR_MessageNamespaceNotFound,
-    `Invalid parameter: --message KM${o.code?.toString(16)} does not have a recognized namespace`);
+    `Invalid parameter: --message ${CompilerError.formatCode(o.code)} does not have a recognized namespace`);
```
mcdurdin, line 104: Note: also cleaned up error reports to have a properly formatted code!

darcywong00 (approved):
lgtm

mcdurdin:
@srl295 you'll probably remember this issue. Fixed now!

srl295:
Commas make sense also

PR #10823, opened 2024-02-23
chore(linux): Add running autopkgtests on GHA

ermshiperete:
The autopkgtests run on Debian servers after building a package on Debian. Until now we didn't notice test failures until we uploaded a source package to Debian. This change adds the autopkg tests to the packaging GHA so they run regularly.

@keymanapp-test-bot skip


mcdurdin (approved):

keyman-server:
Changes in this pull request will be available for download in [Keyman version 18.0.5-alpha](https://keyman.com/downloads/releases/alpha/18.0.5)

